PR #5, opened 2016-06-24
this might work, but all the safeties are turned off.

cdeszaq:


cdeszaq:
No idea what this was for. Since builds seem to be working, I'm closing it.

PR #15, opened 2016-08-30
Fix Dimension Serialization Problem with Nested Queries

garyluoex:
* Modified `DimensionToDefaultDimensionSpec` serializer to serialize dimension to apiName if it is not the inner most query
* Added helper `hasInnerQuery` to `Util` in serializer package to determine if current query is the inner most query or not
* Added tests for `DimensionToDefaultDimensionSpec`

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/serializers/Util.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/serializers/Util.java
@@ -9,6 +9,7 @@ import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonStreamContext;
 
 import java.util.Optional;
+import java.util.function.Function;
 
 /**
  * Utility functions for druid serializers.
@@ -24,24 +25,48 @@ public class Util {
      * @return  an Optional String of physical name
      */
     public static Optional<String> findPhysicalName(Dimension value, JsonGenerator gen) {
-        JsonStreamContext context = gen.getOutputContext();
         String apiName = value.getApiName();
         // Search for physical name
+        return mapNearestDruidQuery(
+                gen,
+                druidQuery -> druidQuery.getDataSource()
+                        .getPhysicalTables()
+                        .iterator()
+                        .next()
+                        .getPhysicalColumnName(apiName)
+        );
+    }
+
+    /**
+     * JSON tree walk to determine if there is a nested query below the current json node or not.
+     *
+     * @param gen  the Json Generator to retrieve the tree to walk on.
+     *
+     * @return  a Boolean where true indicates there are more nested query below this node, false otherwise
+     */
+    public static Boolean hasInnerQuery(JsonGenerator gen) {
+        return mapNearestDruidQuery(gen, druidQuery -> druidQuery.getInnerQuery() != null).orElse(false);
+    }
+
+    /**
+     * JSON tree walk to find the druid query context of the current context and apply handler to the DruidQuery,
+     * finds the current context if current context is a druid query.
+     *
+     * @param gen  the Json Generator to retrieve the tree to walk on.
+     * @param mapper  a function that takes an DruidQuery as an argument and return the final desired returned result.
+     *
+     * @return an Optional of the desired result T if DruidQuery is found, Optional.empty otherwise
+     */
+    public static <T> Optional<T> mapNearestDruidQuery(JsonGenerator gen, Function<DruidQuery, T> mapper) {
+        JsonStreamContext context = gen.getOutputContext();
+
         while (context != null) {
             Object parent = context.getCurrentValue();
             if (parent instanceof DruidQuery) {
-                return Optional.of(
-                        ((DruidQuery) parent)
-                                .getDataSource()
-                                .getPhysicalTables()
-                                .iterator()
-                                .next()
-                                .getPhysicalColumnName(apiName)
-                );
+                return Optional.of(mapper.apply((DruidQuery) parent));
             }
             context = context.getParent();
         }
-        // If we cannot find the physical name, then return empty optional.
         return Optional.empty();
     }
 }
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/serializers/DimensionToDefaultDimensionSpecSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/serializers/DimensionToDefaultDimensionSpecSpec.groovy
@@ -1,0 +1,84 @@
+// Copyright 2016 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE file distributed with this work for terms.
+package com.yahoo.bard.webservice.druid.serializers
+
+import static com.yahoo.bard.webservice.data.time.DefaultTimeGrain.HOUR
+import static org.joda.time.DateTimeZone.UTC
+
+import com.yahoo.bard.webservice.data.DruidQueryBuilder
+import com.yahoo.bard.webservice.data.PartialDataHandler
+import com.yahoo.bard.webservice.data.QueryBuildingTestingResources
+import com.yahoo.bard.webservice.data.dimension.Dimension
+import com.yahoo.bard.webservice.data.metric.LogicalMetric
+import com.yahoo.bard.webservice.data.metric.mappers.NoOpResultSetMapper
+import com.yahoo.bard.webservice.data.volatility.DefaultingVolatileIntervalsService
+import com.yahoo.bard.webservice.druid.model.query.DruidAggregationQuery
+import com.yahoo.bard.webservice.table.resolver.DefaultPhysicalTableResolver
+import com.yahoo.bard.webservice.web.DataApiRequest
+
+import com.fasterxml.jackson.databind.ObjectMapper
+
+import org.joda.time.DateTime
+import org.joda.time.Hours
+import org.joda.time.Interval
+
+import spock.lang.Specification
+
+/**
+ * Testing dimension serialization to default dimension spec.
+ */
+class DimensionToDefaultDimensionSpecSpec extends Specification {
+
+    ObjectMapper objectMapper
+    QueryBuildingTestingResources resources
+    DruidQueryBuilder builder
+    DataApiRequest apiRequest
+    DruidAggregationQuery<?> druidQuery
+
+    def setup() {
+        objectMapper = new ObjectMapper()
+        resources = new QueryBuildingTestingResources()
+        DefaultPhysicalTableResolver resolver = new DefaultPhysicalTableResolver(new PartialDataHandler(), new DefaultingVolatileIntervalsService())
+        builder = new DruidQueryBuilder(resources.logicalDictionary, resolver)
+        apiRequest = Mock(DataApiRequest)
+        LogicalMetric lm1 = new LogicalMetric(resources.simpleTemplateQuery, new NoOpResultSetMapper(), "lm1", null)
+
+        apiRequest.getTable() >> resources.lt12
+        apiRequest.getGranularity() >> HOUR.buildZonedTimeGrain(UTC)
+        apiRequest.getTimeZone() >> UTC
+        apiRequest.getLogicalMetrics() >> ([lm1])
+        apiRequest.getIntervals() >> [new Interval(new DateTime("2015"), Hours.ONE)]
+        apiRequest.getFilterDimensions() >> []
+        apiRequest.getTopN() >> OptionalInt.empty()
+        apiRequest.getSorts() >> ([])
+        apiRequest.getCount() >> OptionalInt.empty()
+    }
+
+    def "Serialize to apiName when apiName and physicalName of a dimension is the same"() {
+        given:
+        apiRequest.getDimensions() >> ([resources.d1])
+        druidQuery = builder.buildQuery(apiRequest, resources.simpleTemplateQuery)
+
+        expect:
+        objectMapper.writeValueAsString(druidQuery).contains('"dimensions":["dim1"]')
+    }
+
+    def "Serialize to dimension spec when the apiName and physicalName of a dimension is different"() {
+        given:
+        apiRequest.getDimensions() >> ([resources.d3])
+        druidQuery = builder.buildQuery(apiRequest, resources.simpleTemplateQuery)
+
+        expect:
+        objectMapper.writeValueAsString(druidQuery).contains('"dimensions":[{"type":"default","dimension":"age_bracket","outputName":"ageBracket"}]')
+    }
+
+    def "Serialize dimension of a nested query should only serialize the innermost query's dimension to a dimension spec"() {
+        given:
+        apiRequest.getDimensions() >> ([resources.d3])
+        druidQuery = builder.buildQuery(apiRequest, resources.simpleNestedTemplateQuery)
+        String serializedQuery = objectMapper.writeValueAsString(druidQuery)
+
+        expect:
+        serializedQuery.contains('"dimensions":[{"type":"default","dimension":"age_bracket","outputName":"ageBracket"}]') && serializedQuery.contains('"dimensions":["ageBracket"]')
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ Current
 
 #### Changed:
 
+-  [Fix Dimension Serialization Problem with Nested Queries](https://github.com/yahoo/fili/pull/15)
+    * Modified `DimensionToDefaultDimensionSpec` serializer to serialize dimension to apiName if it is not the inner most query
+    * Added helper `hasInnerQuery` to `Util` in serializer package to determine if current query is the inner most query or not
+    * Added tests for `DimensionToDefaultDimensionSpec`
 
 ### Fixed:
 
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/serializers/DimensionToDefaultDimensionSpec.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/serializers/DimensionToDefaultDimensionSpec.java
@@ -31,7 +31,8 @@ public class DimensionToDefaultDimensionSpec extends JsonSerializer<Dimension> {
                 }
         );
 
-        if (physicalName.equals(apiName)) {
+        // serialize to only apiName if api and physical name is same or there are nested queries
+        if (physicalName.equals(apiName) || Util.hasInnerQuery(gen)) {
             gen.writeString(apiName);
         } else {
             gen.writeObject(new DefaultDimensionSpec(physicalName, apiName));


fili-core/src/main/java/com/yahoo/bard/webservice/druid/serializers/Util.java
```diff
@@ -44,4 +45,25 @@
...
+     * @param gen  the Json Generator to retrieve the tree to walk on
+     *
+     * @return  a Boolean where true indicates there are more nested query below this node, false otherwise
+     */
+    public static Boolean moreNestedQuery(JsonGenerator gen) {
+        JsonStreamContext context = gen.getOutputContext();
+
+        while (context != null) {
+            Object parent = context.getCurrentValue();
+            if (parent instanceof DruidQuery) {
```
cdeszaq: Can we extract the "tree walk to find the next level up" into it's own method? Both of the helper methods in this class are currently doing that work. `Optional<DruidQuery>` may be a useful return type to use, since it may let us keep things clean.

Here's one (untested) idea:
```java
public <T> Optional<T> withContainingQuery(JsonGenerator gen, Function<DruidQuery, T> success) {
    JsonStreamContext context = gen.getOutputContext();
    while (context != null) {
        Object parent = context.getCurrentValue();
        if (parent instanceof DruidQuery) {
            return Optional.of(success.apply((DruidQuery) parent));
        }
        context = context.getParent();
    }
    return Optional.empty();
}
```

fili-core/src/main/java/com/yahoo/bard/webservice/druid/serializers/Util.java
```diff
@@ -44,4 +45,25 @@
...
     }
+
+    /**
+     * JSON tree walk to determine if there is a nested query below the current json node or not.
+     *
+     * @param gen  the Json Generator to retrieve the tree to walk on
+     *
+     * @return  a Boolean where true indicates there are more nested query below this node, false otherwise
+     */
+    public static Boolean moreNestedQuery(JsonGenerator gen) {
```
cdeszaq: Rather than "more", for methods that are essentially "property getters", common convention is to use "has" or "is" as the method prefix. So, in this case, it would be "hasNestedQuery". (or `hasInnerQuery`)

fili-core/src/main/java/com/yahoo/bard/webservice/druid/serializers/Util.java
```diff
@@ -44,4 +45,25 @@
...
+        JsonStreamContext context = gen.getOutputContext();
+
+        while (context != null) {
+            Object parent = context.getCurrentValue();
+            if (parent instanceof DruidQuery) {
+                return ((DruidQuery) parent).getDataSource().getType().equals(DefaultDataSourceType.QUERY);
```
cdeszaq: Rather than look at the type of `DataSource`, I think we might be able to use `DruidQuery::getInnerQuery` to indicate if there is a nested query or not.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/serializers/DimensionToDefaultDimensionSpecSpec.groovy
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Testing dimension serialization to default dimension spec.
+ */
+class DimensionToDefaultDimensionSpecSpec extends Specification {
+
+    static ObjectMapper objectMapper
+    static QueryBuildingTestingResources resources
+    static Set<Interval> intervals
+    static DruidQueryBuilder builder
+    static DefaultPhysicalTableResolver resolver
```
cdeszaq: These shouldn't be `static` unless they _really_ need to be. And even then, they should be `@Shared` instead of `static`, since the behavior of `static` is undefined in some circumstances.

Times when `@Shared` makes sense are when setup of a variable is expensive, or when we _explicitly want to share state_ between tests, which should be rare. In general, we should strive to have as little state leak between tests as possible.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/serializers/DimensionToDefaultDimensionSpecSpec.groovy
```diff
@@ -0,0 +1,96 @@
...
+        objectMapper = new ObjectMapper()
+        resources = new QueryBuildingTestingResources()
+        resolver = new DefaultPhysicalTableResolver(new PartialDataHandler(), new DefaultingVolatileIntervalsService())
+        builder = new DruidQueryBuilder(resources.logicalDictionary, resolver)
+    }
+
+    def "Testing serialization with equal apiName and physicalName, serialize to apiName string"() {
+        setup:
+        DataApiRequest apiRequest = Mock(DataApiRequest)
+        initDefault(apiRequest, resources.d1)
```
cdeszaq: Rather than create a `Mock(DataApiRequest)` in a setup block in every test and then initialize it by calling `initDefault`, it would be better to just have the mock be an instance variable on this Spec, and then have it initialized in `setup()` (which spock gives for free). Then each test only needs to have it's specific setup, and doesn't need to worry about the "general" setup.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/serializers/DimensionToDefaultDimensionSpecSpec.groovy
```diff
@@ -0,0 +1,96 @@
...
+class DimensionToDefaultDimensionSpecSpec extends Specification {
+
+    static ObjectMapper objectMapper
+    static QueryBuildingTestingResources resources
+    static Set<Interval> intervals
+    static DruidQueryBuilder builder
+    static DefaultPhysicalTableResolver resolver
+
+    def initDefault(DataApiRequest apiRequest, Dimension dim) {
+        intervals = [new Interval(new DateTime("2015"), Hours.ONE)]
```
cdeszaq: If `intervals` isn't used anywhere but in the stub setup for `apiRequest.getIntervals`, there's no need for the extra variable, just create the interval list when you're setting the stub.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/serializers/DimensionToDefaultDimensionSpecSpec.groovy
cdeszaq: Any test-specific setup should be handled in the `setup` (or `given`) block for the specific tests, so the dimension stubs, for example.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/serializers/DimensionToDefaultDimensionSpecSpec.groovy
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    def setupSpec() {
+        objectMapper = new ObjectMapper()
+        resources = new QueryBuildingTestingResources()
+        resolver = new DefaultPhysicalTableResolver(new PartialDataHandler(), new DefaultingVolatileIntervalsService())
+        builder = new DruidQueryBuilder(resources.logicalDictionary, resolver)
+    }
+
+    def "Testing serialization with equal apiName and physicalName, serialize to apiName string"() {
```
cdeszaq: In general, we try to name tests in a way that describe the behavior we're testing for as clearly and succinctly as possible, often in a "given, when, then" style that mimics the contents of the blocks in the test method.

Specifically for this test name, here are some improvements:
- Since spock tests are _all_ testing something, having "Testing" in the name doesn't add anything
- It would be good to indicate _what_ is supposed to have the same apiName and physicalName (Dimension, in this case)
- Can drop "string" from the end. Since an apiName _is_ a string, there's nothing really gained by having it in the name

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/serializers/DimensionToDefaultDimensionSpecSpec.groovy
```diff
@@ -0,0 +1,96 @@
...
+        setup:
+        DataApiRequest apiRequest = Mock(DataApiRequest)
+        initDefault(apiRequest, resources.d1)
+
+        when:
+        DruidAggregationQuery<?> dq = builder.buildQuery(apiRequest, resources.simpleTemplateQuery)
```
cdeszaq: This is really not part of the stimulus, but is actually part of the setup (ie. you're building a query to serialize that meets the preconditions for the test), so it should go up in the `setup` block, not the `when` block.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/serializers/DimensionToDefaultDimensionSpecSpec.groovy
```diff
@@ -0,0 +1,96 @@
...
+
+        then:
+        objectMapper.writeValueAsString(dq).contains('"dimensions":["dim1"]')
```
cdeszaq: Since this line is doing both "stimulus" (`objectMapper.writeValueAsString(dq)`) and "result" (`contains('"dimensions":["dim1"]')`), this line is essentially combining the `when` and `then` blocks. When we see that sort of thing in tests, it's usually a good indication that a test should be an `expect` test instead of a` when`/`then` test.

cdeszaq: Note: This applies to the other tests in this file too, it looks like.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/serializers/DimensionToDefaultDimensionSpecSpec.groovy
cdeszaq: Note: This applies to the other tests in this file too, it looks like.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/serializers/DimensionToDefaultDimensionSpecSpec.groovy
cdeszaq: Note: This applies to the other tests in this file too, it looks like.

pom.xml
```diff
@@ -354,6 +354,11 @@
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <version>${maven-jar-plugin-version}</version>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
cdeszaq: This looks like it was mistakenly pulled in. I think this is already on `master`,  right?

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/serializers/DimensionToDefaultDimensionSpecSpec.groovy
```diff
@@ -0,0 +1,96 @@
...
+    def "Testing serialization with nested query as datasource, only serializing the inner-most query dimension"() {
+        setup:
+        DataApiRequest apiRequest = Mock(DataApiRequest)
+        initDefault(apiRequest, resources.d3)
+
+        when:
+        DruidAggregationQuery<?> dq = builder.buildQuery(apiRequest, resources.simpleNestedTemplateQuery)
+
+        then:
+        objectMapper.writeValueAsString(dq).matches("(.*)\"dimensions\":\\[\\{\"type\":\"default\",\"dimension\":\"age_bracket\",\"outputName\":\"ageBracket\"\\}\\](.*)\"dimensions\":\\[\"ageBracket\"\\](.*)")
```
cdeszaq: Rather than using a complex regex, we should just use 2 different `contains` calls, one for each of the things we're looking for.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/serializers/Util.java
garyluoex: I am amazed by how much line reduction this implementation provides, however I am a bit worried about the readability reduction associated with it.

archolewa: I don't see how this reduces readability any. Personally, I generally find it _more_ readable when things are broken out into helper methods, especially if there is duplication. 

First, the helper methods abstract away some of the "how" and replace it with "what." I find this good for two reasons: One, I generally find the "what" more interesting than the "how" when I'm starting to dig into some code. Two, I generally find it easier to understand the "how" when I already know the "what," then I do to derive the "what" from the "how."

Second, this makes it _immediately_ clear that this code is identical both here and in the other helper method. Without this method, I would have to read the two code blocks very closely in order to confirm that they are doing the same thing (as opposed to very subtly different things). That in and of itself is a _huge_ readability bonus.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/serializers/Util.java
```diff
@@ -24,24 +25,47 @@
...
-            if (parent instanceof DruidQuery) {
-                return Optional.of(
-                        ((DruidQuery) parent)
+        return getDruidQuery(gen, druidQuery ->
+                        druidQuery
                                 .getDataSource()
                                 .getPhysicalTables()
                                 .iterator()
                                 .next()
                                 .getPhysicalColumnName(apiName)
```
cdeszaq: This can actually all be unwrapped since it can fit on 1 line now

fili-core/src/main/java/com/yahoo/bard/webservice/druid/serializers/Util.java
```diff
@@ -24,24 +25,47 @@
...
+    public static Boolean hasInnerQuery(JsonGenerator gen) {
+        return getDruidQuery(gen, druidQuery -> druidQuery.getInnerQuery() != null).get();
+    }
+
+    /**
+     * JSON tree walk to find the druid query context of the current context,
+     * returns the current context if current context is a druid query.
+     *
+     * @param gen  the Json Generator to retrieve the tree to walk on.
+     *
```
cdeszaq: Javadoc needed for `success` param

fili-core/src/main/java/com/yahoo/bard/webservice/druid/serializers/Util.java
```diff
@@ -24,24 +25,47 @@
...
+     * @return an Optional of DruidQuery if found, Optional.empty otherwise
+     */
+    public static <T> Optional<T> getDruidQuery(JsonGenerator gen, Function<DruidQuery, T> success) {
```
cdeszaq: Since this method doesn't return the `DruidQuery`, calling it `getDruidQuery` is a bit confusing. I think `withDruidQuery` is perhaps a bit better, since what it's doing is to call the `success` function _in the context of_ (or _with_) a `DruidQuery`.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/serializers/Util.java
```diff
@@ -24,24 +25,47 @@
...
+
+    /**
+     * JSON tree walk to determine if there is a nested query below the current json node or not.
+     *
+     * @param gen  the Json Generator to retrieve the tree to walk on.
+     *
+     * @return  a Boolean where true indicates there are more nested query below this node, false otherwise
+     */
+    public static Boolean hasInnerQuery(JsonGenerator gen) {
+        return getDruidQuery(gen, druidQuery -> druidQuery.getInnerQuery() != null).get();
```
cdeszaq: While it's _unlikely_ that a `DruidQuery` will not be found, it's still _possible_ that this method will return an empty `Optional`. If that were to happen, `get` will throw an exception. Instead, we should handle it by either throwing our own more descriptive exception, or by using a default value instead.

In this case, I'm not sure which would be the better choice. Defaulting to `false` (ideally while logging a warning) probably wouldn't be bad, since not having an enclosing Query also means that you _can't_ have a `DataSource`, and therefore no nested query

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/serializers/DimensionToDefaultDimensionSpecSpec.groovy
```diff
@@ -0,0 +1,88 @@
...
+        objectMapper = new ObjectMapper()
+        resources = new QueryBuildingTestingResources()
+        DefaultPhysicalTableResolver resolver = new DefaultPhysicalTableResolver(new PartialDataHandler(), new DefaultingVolatileIntervalsService())
+        builder = new DruidQueryBuilder(resources.logicalDictionary, resolver)
+        apiRequest = Mock(DataApiRequest)
+    }
+
+    def "Serialize to apiName when apiName and physicalName of a dimension is the same"() {
+        given:
+        initDefault(apiRequest, resources.d1)
```
cdeszaq: Rather than call this method, it would be better to move everything that this method is doing into the `setup` method, and then replace this line with

```groovy
apiRequest.getDimensions() >> ([resources.d1)
```

fili-core/src/main/java/com/yahoo/bard/webservice/druid/serializers/Util.java
garyluoex: hmmmm, it seems like it still wont fit I think

fili-core/src/main/java/com/yahoo/bard/webservice/druid/serializers/Util.java
```diff
@@ -24,24 +25,42 @@
...
+     * @return  a Boolean where true indicates there are more nested query below this node, false otherwise
+     */
+    public static Boolean hasInnerQuery(JsonGenerator gen) {
+        return withDruidQuery(gen, druidQuery -> druidQuery.getInnerQuery() != null).orElse(false);
+    }
+
+    /**
+     * JSON tree walk to find the druid query context of the current context,
+     * returns the current context if current context is a druid query.
+     *
```
michael-mclawhorn: Javadoc is incorrect.  This query returns the mapped value of the druid query under the success function if the current context if the current context is a druid query.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/serializers/Util.java
```diff
@@ -24,24 +25,42 @@
...
+     * @param gen  the Json Generator to retrieve the tree to walk on.
+     * @param success  a function that takes an DruidQuery as an argument and return the final desired returned result.
+     *
```
michael-mclawhorn: success was a good name when this function was a predicate returning true or false, it is not a good name for a function that returns an arbitrary mapped value.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/serializers/Util.java
```diff
@@ -24,24 +25,42 @@
...
+     * @return an Optional of DruidQuery if found, Optional.empty otherwise
+     */
+    public static <T> Optional<T> withDruidQuery(JsonGenerator gen, Function<DruidQuery, T> success) {
+        JsonStreamContext context = gen.getOutputContext();
```
michael-mclawhorn: withDruidQuery is a good name if you were extracting or building a druid query, but in this case you're mapping a value from a druid query.

cdeszaq: the idea behind `with` in this case is that it's executing the function _in the context of_ (or _with_) the `DruidQuery`. Groovy has a very similar construct, which is what this method is (very roughly) mimicking:

```groovy
// If we have an object with methods:
class Foo {
    def doBar() { ... }
    def doBaz() { ... }
}
def myFoo = new Foo()

// This:
myFoo.with {
    doBar()
    doBaz()
}

// is the same as:
myFoo.doBar()
myFoo.doBaz()
```

The previous name for this method in this PR was `getDruidQuery`. Thus far, we don't really have this pattern in the Fili codebase, so better suggestions are extremely welcome.

cdeszaq:
👍

michael-mclawhorn:
👍

michael-mclawhorn:
Build is failing on one of the tests.

cdeszaq:
I'm pretty sure the build is failing due to load during tests caused by a too-tight wait period. Since nothing else relevant to this PR is failing, marking this as mergeable.

PR #20, opened 2016-08-30
[17] Added pagination parameters to ResultSet

dayamr:
- Jobs end point result doesn't contain pagination. Fixed that issue in this PR


cdeszaq:
Tests needed showing that we're now putting the pagination info on the responses, and that we're properly paginating

CHANGELOG.md
```diff
@@ -18,7 +18,8 @@ Current
 
 
 #### Changed:
-
+- [ISSUE-17](https://github.com/yahoo/fili/issues/17) [Added pagination parameters to PreResponse] (https://github.com/yahoo/fili/pull/19)
```
cdeszaq: In the future, if there isn't already an issue for something you're opening a PR for, don't bother with opening an issue as well. In GitHub, PRs _are_ issues, so there's no need to create an additional artifact.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
```diff
@@ -415,11 +423,12 @@ protected void handleResultsResponse(
      * return an Observable wrapping the PreResponse as is.
      *
      * @param preResponse  The PreResponse to be inspected
+     * @param jobsApiRequest  JobsApiRequest object with all the associated info with it
      *
      * @return An Observable wrapping the PreResponse or an Observable wrapping a ResponseException
      *
      */
-    protected Observable<PreResponse> handlePreResponseWithError(PreResponse preResponse) {
+    protected Observable<PreResponse> handlePreResponseWithError(PreResponse preResponse, ApiRequest jobsApiRequest) {
```
archolewa: Rather than taking in the entire `ApiRequest`, this method should only take the pieces that it needs (the `PaginationParameters` and the `UriInfo` I believe). We've been burned before by taking the entire `ApiRequest` when we only needed a couple pieces of it.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
```diff
@@ -431,10 +440,85 @@ protected void handleResultsResponse(
             );
             return Observable.error(responseException);
         }
+
+        if (jobsApiRequest.getPaginationParameters().isPresent()) {
+            Pagination<Result> pages = new AllPagesPagination<>(
+                    preResponse.getResultSet(),
+                    jobsApiRequest.getPaginationParameters().get()
+            );
+            responseContext = paginateResponseContext(responseContext, jobsApiRequest, pages);
```
archolewa: `responseContext` is a method paramater. We should not be reassigning method parameters. You can inline this easily enough.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
```diff
@@ -431,10 +440,85 @@ protected void handleResultsResponse(
...
+            PreResponse paginatedPreResponse = new PreResponse(
+                    new ResultSet(pages.getPageOfData(), preResponse.getResultSet().getSchema()),
+                    responseContext
+            );
+            return Observable.just(paginatedPreResponse);
+        }
```
archolewa: You can make this a bit cleaner (well, I think it's cleaner, YMMV) if you take advantage of the `map` function on `Optional`:

```java
return jobsApiRequest.getPaginationParameters()
        .map(paginationParameters -> new AllPagesPagination<>(preResponse.getResultSet(), paginationParameters))
        .map(pages -> new PreResponse(
                new ResultSet(pages.getPageOfData(), preResponse.getResultSet().getSchema()),
                paginateResponseContext(responseContext, jobsApiRequest, pages)
        ))
        .map(Observable::just)
        .orElse(Observable.just(preResponse));
```
Given an `Optional<T>`, `map` takes a `Function<T, U>`, and returns an `Optional<U>`. If the original `Optional` is empty, then the `map` returns the empty `Optional`. If the original `Optional` is not empty, however, then it applies the `Function<T, U>` to the value contained in the `Optional` and returns a new `Optional` wrapping the result.

Meanwhile, `orElse` takes a default value. If the `Optional` it is applied to is empty, then it returns the default value. Otherwise, it returns the value wrapped in the `Optional`.

This allows us to avoid a fat if statement. I don't like if-statements because as you're looking through the if-block, you have to keep in your head what the check was that triggered entering the if-statement. So instead of thinking "OK, suppose the value is in the `Optional`. Then we do X and return Y. Ok, we're leaving the if-block, so if we the `Optional` is empty then we return Z." This instead allows us to think "OK do X to transform Y' into Y, and return it, defaulting to Z."

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
```diff
@@ -431,10 +440,85 @@ protected void handleResultsResponse(
...
     /**
+     * Add pagination details to ResponseContext.
+     *
+     * @param responseContext  ResponseContext object contains all the meta info of the resultSet
+     * @param jobsApiRequest  JobsApiRequest object with all the associated info with it
+     * @param pages  Paginated resultSet
+     *
+     * @return Updated ResponseContext contains pagination info
+     */
+    protected ResponseContext paginateResponseContext(
```
archolewa: This should take just the `UirInfo`, rather than the entire `ApiRequest`.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
```diff
@@ -431,10 +440,85 @@ protected void handleResultsResponse(
...
+            Pagination<Result> pages
+    ) {
+        LinkedHashMap<String, URI> bodyLinks = new LinkedHashMap<>();
+        Arrays.stream(PaginationLink.values())
+                .forEachOrdered(link -> addLink(
+                                link,
+                                pages,
+                                jobsApiRequest.getUriInfo().getRequestUriBuilder(),
+                                bodyLinks)
+                );
```
archolewa: I'm pretty sure you copy-pasted this from the `ApiRequest`, which does the same thing, which is fine.

Except that this code is ugly, as far as I'm concerned, and doesn't really use streams. I also find it hard to read because we have an overloaded method `addLink` that do rather different things. Instead, I would propose that we take more full advantage of streams, and do the following:

```java
LinkedHashMap<String, URI> bodyLinks = Arrays.stream(PaginationLink.values())
                .map(link -> new ImmutablePair<>(link.getBodyName(),link.getPage(pages)))
                .filter(pair -> pair.getRight().isPresent())
                .map(pair -> Utils.withRight(pair, pair.getRight().getAsInt()))
                .map(pair -> Utils.withRight(pair, uriBuilder.replaceQueryParam("page", pair.getRight())))
                .map(pair -> Utils.withRight(pair, pair.getRight().build()))
                .collect(StreamUtils.toLinkedMap(Pair::getLeft, Pair::getRight));
```

A few comments: The `ImmutablePair` class is from the apache commons. I also added a utility method `withRight` with the expected semantics:

```java
public static <T, U, V> ImmutablePair<T, V> withRight(ImmutablePair<T, U> pair, V right) {
        return new ImmutablePair<>(pair.getLeft(), right);
}
```
And a similar one for replacing the left value.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletSpec.groovy
```diff
@@ -155,6 +192,35 @@ class JobsServletSpec extends Specification {
         GroovyTestUtils.compareJson(result, expectedResponse, JsonSortStrategy.SORT_BOTH)
     }
 
+    def "jobs/pagination/results returns the number of results we requested through pagination parameters" () {
```
archolewa: Where is this segment coming from? Near as I can tell, we aren't storing a ticket called `pagination` anywhere in the `ApiJobStore`. Also, we shouldn't be using `pagination` as the ticket name. We should instead be using `ticket1` or `ticket2` or `ticket3` or something. For a second there, I thought we were doing something screwy, where we had to add an extra `pagination` path segment in order to paginate results!

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletSpec.groovy
```diff
@@ -155,6 +192,35 @@ class JobsServletSpec extends Specification {
...
+                                        }
+                                      }
+                                    }
+                                 """
+
+        when: "We send a request to the jobs/pagination/results endpoint"
+        String result = makeRequest("/jobs/pagination/results", [asyncAfter : ["5"], perPage : [1], page: [1] ])
+
+        then: "We get only first row as we requested against total number of rows"
+        GroovyTestUtils.compareJson(result, expectedResponse, JsonSortStrategy.SORT_BOTH)
```
archolewa: We're missing a few pagination cases:

1. We request a page in the middle.

2. We request a page at the end.

We should have cases for those as well.

fili-core/src/test/java/com/yahoo/bard/webservice/web/endpoints/JobsEndpointResources.java
```diff
@@ -110,6 +110,24 @@ public static PreResponseStore getPreResponseStore() {
...
+        results1.add(result2);
+        results1.add(result3);
+        ResultSet resultSet1 = new ResultSet(results1, schema);
+
+        ResponseContext responseContext1 = new ResponseContext();
+        responseContext1.put("headers", new MultivaluedHashMap<>());
+        responseContext1.put("apiMetricColumnNames", apiMetricColumnNames);
+        responseContext1.put("requestedApiDimensionFields", new HashMap<>());
+        PreResponse preResponse1 = new PreResponse(resultSet1, responseContext1);
+        preResponseStore.save("pagination", preResponse1);
```
archolewa: Ok, I see now. We're not storing any tickets in the `ApiJobStore`, but just in the `PreResponse`. I would suggest also storing the ticket in the `ApiJobStore`. Right now, we don't verify that the ticket is in the `ApiJobStore` before getting it from the `PreResponseStore`, but that leads to some inconsistent behavior, because the user may get a 404 if their long poll times out and there is no ticket in the ApiJobStore, but get the results if the results are ready before the async timeout. A 404 really shouldn't be affected by query parameters.

So that will be changing at some point in the future, which will cause tests using the `pagination` ticket to break if we leave the tests as is.

archolewa:
Finished my review. Looks pretty good. Just a few minor suggestions and tweaks.

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ Current
 
 ### Changed:
 
+- [ISSUE-17](https://github.com/yahoo/fili/issues/17) [Added pagination parameters to PreResponse] (https://github.com/yahoo/fili/pull/19)
```
venisa: Do we have a Jira to track this? if we do, we do not need a Git issue right?

fili-core/src/main/java/com/yahoo/bard/webservice/web/PreResponse.java
```diff
@@ -34,6 +34,17 @@ public ResponseContext getResponseContext() {
         return responseContext;
     }
 
+    /**
+     * Method to have new PreResponse for the given ResultSet.
```
venisa: Can we rephrase this to 'Create a new instance of PreResponse using the supplied ResultSet'

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletSpec.groovy
venisa: Double 👍  for renaming pagination to ticket*. I thought we had a new path too

CHANGELOG.md
archolewa: Since this is Fili (i.e. open-source) we can't reference any internal tickets. That being said, per a request by @cdeszaq earlier (that I'm too lazy to find and link to), we shouldn't be adding Github issues after the fact. Apparently, Github issues are meant to track things todo in the future, not everything that has been done, is being done, and will be done, like Jira.

venisa:
I had a few minor comments but mostly I agree with @archolewa about adding the missing test cases and renaming the ticket to 'ticket*' instead of 'pagination'

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
cdeszaq: If there's specific pieces of ApiRequest we keep needing to have, we can extract an interface from it and depend on that instead if it makes more sense. Not sure if it would help here or not, but it's an option.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
cdeszaq: Note that this is no-longer a method parameter, but it can still be in-lined easily enough. Since it's only used in the one place, there's no point in assigning this to a variable, especially if it's re-assigning a variable already with something semantically different (ie. paginated)

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
cdeszaq: 👍 (though it would be probably a good idea to comment a bit on each step as to what's happening, since the `Utils.withRight` nonsense makes it more obscure than it should be (thanks Java...).

Alternatively, another approach would be to break those operations out into their own methods that would encapsulate the operation in a nice, descriptive name.

cdeszaq: And if we're just copy-pasting this code, we shouldn't be... we should abstract it out into something reusable, otherwise we'll update one but forget the other and then feel bug pain down the road.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletSpec.groovy
cdeszaq: Note to self: We should abstract out "paginatable resources" as a TCK.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
dayamr: Actually i kept that outside for debugging purpose and later didn't make it inline. fixed it now.

archolewa:
@dayamr Don't forget to remove the "Needs Changes" tag when your code is ready for rereview. That's how other Fili contributors know that your PR is ready for another look.

fili-core/src/main/java/com/yahoo/bard/webservice/util/Utils.java
```diff
@@ -220,4 +222,28 @@ public static void emitField(JsonNode node, String fieldName, ObjectMapper mappe
...
     }
+
+    /**
+     *
+     *
+     * @param pair  Immutable Pair instance
+     * @param right  The right value, may be null
+     *
+     * @return New instance of Immutable Pair
+     */
```
archolewa: This `JavaDoc` should either be deleted, or given a method. It's lonely all by itself. 😿

fili-core/src/main/java/com/yahoo/bard/webservice/util/Utils.java
```diff
@@ -220,4 +222,28 @@ public static void emitField(JsonNode node, String fieldName, ObjectMapper mappe
...
+
+    /**
+     * Method to get new Immutable Pair instance with the given pair and right element.
```
archolewa: This description could stand to be clarified a little. Perhaps something like:

```java
Given an `ImmutablePair`, returns a new `ImmutablePair` with the same left value, and the specified right value.
```

fili-core/src/main/java/com/yahoo/bard/webservice/util/Utils.java
```diff
@@ -220,4 +222,28 @@ public static void emitField(JsonNode node, String fieldName, ObjectMapper mappe
...
+     *
+     * @param pair  Immutable Pair instance
+     * @param right  The right value, may be null
+     * @param <T>  Left value of the pair
+     * @param <U>  Right value of the pair
```
archolewa: For these two parameters, `value` should be `type`, since they're type arguments, not regular parameters.

fili-core/src/main/java/com/yahoo/bard/webservice/web/PreResponse.java
```diff
@@ -34,6 +34,17 @@ public ResponseContext getResponseContext() {
...
     }
 
+    /**
+     * Create a new instance of PreResponse using the supplied ResultSet.
+     *
+     * @param resultSet  ResultSet associated with a response
+     *
+     * @return new instance of PreResponse
+     */
+    public PreResponse withResultSet(ResultSet resultSet) {
```
archolewa: No longer being used.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
```diff
@@ -399,7 +412,12 @@ protected void handleResultsResponse(
     ) {
         HttpResponseMaker httpResponseMaker = new HttpResponseMaker(objectMappers, dimensionDictionary);
 
-        preResponseObservable.flatMap(this::handlePreResponseWithError)
+        preResponseObservable
+                .flatMap(preResponse -> handlePreResponseWithError(
+                        preResponse,
+                        apiRequest.getUriInfo(),
+                        apiRequest.getPaginationParameters())
```
archolewa: The last closing paren on this line (for `handlePreResponseWithError`) should be wrapped onto the next line with the closing parent for `flatMap`.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
```diff
@@ -415,11 +433,16 @@ protected void handleResultsResponse(
      * return an Observable wrapping the PreResponse as is.
      *
      * @param preResponse  The PreResponse to be inspected
+     * @param uriInfo  uriInfo object to get uriBuilder
+     * @param paginationParameters  Pagination parameters received through an API
```
archolewa: Missing `request` at the end of this parameter description. Though I would probably rephrase this as something like `the user's requested pagination`

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
```diff
@@ -431,7 +454,44 @@ protected void handleResultsResponse(
             );
             return Observable.error(responseException);
         }
-        return Observable.just(preResponse);
+
+        return paginationParameters
+                .map(pageParams -> new AllPagesPagination<>(preResponse.getResultSet(), pageParams))
+                .map(pages -> new PreResponse(
```
archolewa: Nitpick: Should probably call the lambda argument `page` rather than `pages` since it only contains one page of data.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
archolewa: We might want to rephrase this as `addPaginationMetadata`. I know that the method name doesn't make clear what the metadata is being added to, but seeing as how the only argument this method is taking that's mutable is the `ResponseContext` (and it's the first argument), it _should_ be fairly clear from context.  Regardless, we should probably change the method name to something else, because the current name is misleading. It sounds like we're paginating the `responseContext` object, which doesn't make any sense.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletSpec.groovy
```diff
@@ -171,4 +240,74 @@ class JobsServletSpec extends Specification {
...
+              "currentPage": 1,
+              "rowsPerPage": 1,
+              "numberOfResults": 3
+            }
+          }
+        }
+       """
+    }
+
+    String getPaginatedExpectedResponse2(){
```
archolewa: I would suggest swapping the name of this method, with the name of `getPaginatedExpectedResponse3`. Basically, I think we should have the number in each method name line up with the page being returned.

archolewa:
A few small comments. Other than that, I think this is looking pretty close.

We could probably stand to make the pagination metadata stream a bit clearer (either giving the `pairs` variables better names, or abstracting the pagination stream into its own well-named method), but I'm not going to block on that.

dayamr:
@archolewa @cdeszaq this PR should be ready for your thumps up.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
```diff
@@ -399,27 +412,36 @@ protected void handleResultsResponse(
...
-                                apiRequest.getFormat(),
-                                apiRequest.getUriInfo()
-                        )
-                );
+                new HttpResponseChannel(
+                        asyncResponse,
+                        httpResponseMaker,
+                        apiRequest.getFormat(),
+                        apiRequest.getUriInfo()
+                ));
```
cdeszaq: The wrapping on this isn't quite right. It should be:

```java
preResponseObservable
        .flatMap(preResponse -> handlePreResponseWithError(
                preResponse,
                apiRequest.getUriInfo(),
                apiRequest.getPaginationParameters()
        ))
        .subscribe(
                new HttpResponseChannel(
                        asyncResponse,
                        httpResponseMaker,
                        apiRequest.getFormat(),
                        apiRequest.getUriInfo()
                )
        );
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletSpec.groovy
```diff
@@ -46,6 +46,41 @@ class JobsServletSpec extends Specification {
...
+                }
+              }
+            }"""
+
+        when: "We send a request for the first page of jobs"
+        //The buildApiJobStore method in TestBinderFactory sets up the ApiJobStore and stores the metadata for ticket1 in it.
+        String result = makeRequest("/jobs", [perPage : [1], page: [1] ])
+
+        then: "what we expect is one job row with pagination meta data"
+        GroovyTestUtils.compareJson(result, expectedResponse, JsonSortStrategy.SORT_BOTH)
```
cdeszaq: Do we need to sort both, or is just sorting maps enough?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletSpec.groovy
```diff
@@ -155,6 +201,33 @@ class JobsServletSpec extends Specification {
...
+        then: "We get only first row as we requested against total number of rows"
+        GroovyTestUtils.compareJson(result1, getExpectedFirstPage(), JsonSortStrategy.SORT_BOTH)
+
+        when: "We send a request to the jobs/ticket3p/results endpoint"
+        String result2 = makeRequest("/jobs/ticket3p/results", [asyncAfter : ["5"], perPage : [1], page: [3] ])
+
+        then: "We get last row as we requested against total number of rows"
+        GroovyTestUtils.compareJson(result2, getExpectedLastPage(), JsonSortStrategy.SORT_BOTH)
+
+        when: "We send a request to the jobs/ticket3p/results endpoint"
```
cdeszaq: Unless these comments are different in some way to indicate the steps happening, they are not providing any benefit. We should either remove these noisy comments, or we should update them to reflect what's going on in the `when` blocks

cdeszaq:
👍 after the remaining small things are addressed.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletSpec.groovy
dayamr: I feel it would be safe to sort both.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
archolewa: This still needs to be addressed.

archolewa:
`JobsServlet::paginateResponseContext` still needs to be renamed, but other than that I don't see any problems. 👍 Once `JobsServlet::paginateResponseContext` has been renamed.

PR #21, opened 2016-08-30
Update changelog description around issue links

cdeszaq:
We don't need (nor want) an issue for any changes that are coming in as a PR directly, so I updated the changelog description text to reflect that.


michael-mclawhorn:
👍  Sure.

archolewa:
👍

PR #22, opened 2016-08-31
Checkstyle back on

michael-mclawhorn:


fili-wikipedia-example/src/test/java/com/yahoo/wiki/webservice/application/WikiJerseyTestBinder.java
```diff
@@ -13,14 +13,25 @@
...
 /**
- * TestBinder with Wiki configuration specialization
+ * TestBinder with Wiki configuration specialization.
  */
 public class WikiJerseyTestBinder extends JerseyTestBinder {
 
+    /**
+     * Constructor.
+     *
+     * @param resourceClasses  Resource classes to load into the application..
```
archolewa: There's an extra period here.

pom.xml
```diff
@@ -108,7 +108,7 @@
         <maven-source-plugin-version>3.0.1</maven-source-plugin-version>
         <maven-surefire-plugin.version>2.17</maven-surefire-plugin.version>
         <maven-site-plugin-version>3.5</maven-site-plugin-version>
-        <maven-checkstyle-plugin-version>2.17</maven-checkstyle-plugin-version>
+        <maven-checkstyle-plugin-version>2.16</maven-checkstyle-plugin-version>
```
archolewa: Why the downgrade?

pom.xml
```diff
@@ -231,6 +231,19 @@
...
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-checkstyle-plugin</artifactId>
+                    <executions>
+                        <execution>
+                            <id>validate</id>
+                            <phase>validate</phase>
+                            <configuration>
+                                <suppressionsLocation>${checkstyle.suppressions.location}</suppressionsLocation>
+                                <suppressionsFileExpression>checkstyle.suppressions.file</suppressionsFileExpression>
+                            </configuration>
```
cdeszaq: We don't seem to be retaining some of the previous properties. I'm not sure, but I _think_ we want to keep (most) of them?

```xml
<includes>**\/*.java</includes> <!-- Makes it look at only .java code files, since it chokes on groovy files -->
<resourceIncludes>**/*.properties,**/*.sample</resourceIncludes> <!-- Look at some of the resource files -->
<includeTestSourceDirectory>true</includeTestSourceDirectory> <!-- Include test code -->
<consoleOutput>true</consoleOutput> <!-- Dump output to the console -->
```

pom.xml
cdeszaq: I would be in favor of bumping to current if we can, but if it causes other issues to get picked up, don't worry about it right now, we can bump it later.

pom.xml
michael-mclawhorn: All that is in the parent pom plugin management section.

pom.xml
michael-mclawhorn: I downgraded because it did cause other issues to get picked up.

pom.xml
cdeszaq: This _is_ the parent pom's plugin management section, isn't it?

cdeszaq: got it, it's below

cdeszaq:
👍

archolewa:
👍

PR #23, opened 2016-08-31
Enrich ApiJobStore to support filtering

venisa:
ApiJobStore now supports filtering JobRows
Added filtering logic to HashJobStore
Added `ApiJobStoreFilter` which stores filter information like the JobField to be inspected, the values to be compared and the filter operation to be performed.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/ApiJobStore.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/ApiJobStore.java
@@ -6,6 +6,8 @@ import com.yahoo.bard.webservice.async.jobs.jobrows.JobRow;
 
 import rx.Observable;
 
+import java.util.Set;
+
 /**
  * An ApiJobStore is responsible for storing the metadata about Bard jobs. Conceptually, the ApiJobStore is a table
  * where each row is the metadata of a particular job, and the columns are the metadata stored with each job
@@ -51,4 +53,21 @@ public interface ApiJobStore {
      * @return An Observable that emits a stream of all the JobRows in the store
      */
     Observable<JobRow> getAllRows();
+
+    /**
+     * This method takes a Set of JobRowFilters, ANDS them by default, and returns a cold observable that emits a
+     * stream of JobRows which satisfy the given filter.
+     * <p>
+     * Any of the fields may be not filterable for every implementation of the {@code ApiJobStore} as the efficiency of
+     * filtering is dependent on the backing store. An IllegalArgumentException is thrown if filtering on any given
+     * field is not supported.
+     *
+     * @param jobRowFilters  A Set of JobRowFilters where each JobRowFilter contains the JobField to be
+     * filtered on, the filter operation and the values to be compared to.
+     *
+     * @return An Observable that emits a stream of JobRows that satisfy the given filters
+     *
+     * @throws IllegalArgumentException if filtering on any field is not supported
+     */
+    Observable<JobRow> getFilteredRows(Set<JobRowFilter> jobRowFilters) throws IllegalArgumentException;
 }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
@@ -2,12 +2,21 @@
 // Licensed under the terms of the Apache license. Please see LICENSE file distributed with this work for terms.
 package com.yahoo.bard.webservice.async.jobs.stores;
 
+import static com.yahoo.bard.webservice.web.ErrorMessageFormat.FILTER_OPERATOR_INVALID;
+import static com.yahoo.bard.webservice.web.ErrorMessageFormat.FILTER_JOBFIELD_UNDEFINED;
+
+import com.yahoo.bard.webservice.async.jobs.jobrows.JobField;
 import com.yahoo.bard.webservice.async.jobs.jobrows.JobRow;
+import com.yahoo.bard.webservice.web.FilterOperation;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import rx.Observable;
 
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * An ApiJobStore backed by an in-memory map. This is meant as a stub implementation for
@@ -16,6 +25,8 @@ import java.util.Map;
  * so its memory footprint will grow until the system is rebooted.
  */
 public class HashJobStore implements ApiJobStore {
+
+    private static final Logger LOG = LoggerFactory.getLogger(HashJobStore.class);
 
     private final Map<String, JobRow> store;
 
@@ -50,5 +61,67 @@ public class HashJobStore implements ApiJobStore {
     @Override
     public Observable<JobRow> getAllRows() {
         return Observable.from(store.values());
+    }
+
+    @Override
+    public Observable<JobRow> getFilteredRows(Set<JobRowFilter> jobRowFilters) throws IllegalArgumentException {
+        return getAllRows().filter(jobRow -> satisfiesFilters(jobRowFilters, jobRow));
+    }
+
+    /**
+     * This method checks if the given JobRow satisfies all the JobRowFilters and returns true if it does.
+     * If a JobField in any of the filters is not a part of the JobRow, this method throws an IllegalArgumentException.
+     *
+     * @param jobRowFilters  A Set of JobRowFilters specifying the different conditions to be satisfied
+     * @param jobRow  The JobRow which needs to be inspected
+     *
+     * @return true if the JobRow satisfies all the filters, false otherwise
+     *
+     * @throws IllegalArgumentException if a JobField in any of the filters is not a part the JobRow
+     */
+    private boolean satisfiesFilters(Set<JobRowFilter> jobRowFilters, JobRow jobRow) throws IllegalArgumentException {
+        return jobRowFilters.stream().allMatch(filter -> satisfiesFilter(jobRow, filter));
+    }
+
+    /**
+     * This method checks if the given JobRow satisfies the given JobRowFilter and returns true if it does.
+     * If a JobField in the filter is not a part the JobRow, this method throws an IllegalArgumentException.
+     *
+     * @param jobRow  The JobRow which needs to be inspected
+     * @param jobRowFilter  A JobRowFilter specifying the filter condition
+     *
+     * @return true if the JobRow satisfies the filter, false otherwise
+     *
+     * @throws IllegalArgumentException if a JobField in the filter is not a part the JobRow
+     */
+    private boolean satisfiesFilter(JobRow jobRow, JobRowFilter jobRowFilter) throws IllegalArgumentException {
+        JobField filterJobField = jobRowFilter.getJobField();
+        FilterOperation filterOperation = jobRowFilter.getOperation();
+        Set<String> filterValues = jobRowFilter.getValues();
+
+        if (!jobRow.containsKey(filterJobField)) {
+            Set<JobField> actualJobFields = jobRow.keySet();
+            LOG.debug(FILTER_JOBFIELD_UNDEFINED.logFormat(filterJobField, actualJobFields));
+            throw new IllegalArgumentException(
+                    FILTER_JOBFIELD_UNDEFINED.format(filterJobField, actualJobFields)
+            );
+        }
+
+        String actualValue = jobRow.get(filterJobField);
+
+        switch (filterOperation) {
+            case notin:
+                return !filterValues.contains(actualValue);
+            case startswith:
+                return filterValues.stream().anyMatch(actualValue::startsWith);
+            case contains :
+                return filterValues.stream().anyMatch(actualValue::contains);
+            case in: // the fall-through is intentional because in is a synonym for eq
+            case eq:
+                return filterValues.contains(actualValue);
+            default:
+                LOG.debug(FILTER_OPERATOR_INVALID.logFormat(filterOperation));
+                throw new IllegalArgumentException(FILTER_OPERATOR_INVALID.format(filterOperation));
+        }
     }
 }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/JobRowFilter.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/JobRowFilter.java
@@ -1,0 +1,218 @@
+// Copyright 2016 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE file distributed with this work for terms.
+package com.yahoo.bard.webservice.async.jobs.stores;
+
+import static com.yahoo.bard.webservice.web.ErrorMessageFormat.FILTER_ERROR;
+import static com.yahoo.bard.webservice.web.ErrorMessageFormat.FILTER_INVALID;
+import static com.yahoo.bard.webservice.web.ErrorMessageFormat.FILTER_JOBFIELD_UNDEFINED;
+import static com.yahoo.bard.webservice.web.ErrorMessageFormat.FILTER_OPERATOR_INVALID;
+
+import com.yahoo.bard.webservice.async.jobs.jobrows.DefaultJobField;
+import com.yahoo.bard.webservice.async.jobs.jobrows.JobField;
+import com.yahoo.bard.webservice.util.FilterTokenizer;
+import com.yahoo.bard.webservice.web.BadFilterException;
+import com.yahoo.bard.webservice.web.FilterOperation;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Arrays;
+import java.util.LinkedHashSet;
+import java.util.Objects;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * Class containing filter information to filter JobRows in ApiJobStore.
+ */
+public class JobRowFilter {
+    private static final Logger LOG = LoggerFactory.getLogger(JobRowFilter.class);
+
+    private final JobField jobField;
+    private final FilterOperation operation;
+    private final Set<String> values;
+
+    /*  url filter query pattern:  (JobField name)-(operation)[(value or comma separated numeric values)]?
+    *
+    *  e.g.    userId-eq[Foo]
+    *
+    *          JobField name:    userId
+    *          operation:        eq
+    *          values:           Foo
+    */
+    private static final Pattern QUERY_PATTERN = Pattern.compile("([^\\|]+)-([^\\[]+)\\[([^\\]]+)\\]?");
+
+    /**
+     * Parses the URL ApiJobStore filter query and generates the JobRowFilter object.
+     *
+     * @param filterQuery  Expects a URL ApiJobStore filter query String in the format:
+     * <p>
+     * <code>(JobField name)-(operation)[?(value or comma separated values)]?</code>
+     *
+     * @throws BadFilterException when filter pattern is not matched or when any of its properties are not
+     * valid.
+     */
+    public JobRowFilter(@NotNull String filterQuery) throws BadFilterException {
+        LOG.trace("filterQuery: {}", filterQuery);
+
+        Matcher tokenizedQuery = QUERY_PATTERN.matcher(filterQuery);
+
+        // if pattern match found, extract values else throw exception
+        if (!tokenizedQuery.matches()) {
+            LOG.debug(FILTER_INVALID.logFormat(filterQuery));
+            throw new BadFilterException(FILTER_INVALID.format(filterQuery));
+        }
+
+        jobField = extractJobField(tokenizedQuery);
+        operation = extractOperation(tokenizedQuery);
+        values = extractValues(tokenizedQuery, filterQuery);
+    }
+
+    /**
+     * Constructor for an JobRowFilter object whose data has already been parsed.
+     *
+     * @param jobField  The JobField to perform the filtering on
+     * @param operation  The operation to perform (eg: eq)
+     * @param values  A Set of Strings to compare the JobField's value to.
+     */
+    private JobRowFilter(JobField jobField, FilterOperation operation, Set<String> values) {
+        this.jobField = jobField;
+        this.operation = operation;
+        this.values = values;
+    }
+
+    public JobField getJobField() {
+        return jobField;
+    }
+
+    public FilterOperation getOperation() {
+        return operation;
+    }
+
+    public Set<String> getValues() {
+        return values;
+    }
+
+    /**
+     * Construct an JobRowFilter object using the same FilterOperation and values as the object on
+     * which this method is called and using the supplied JobField.
+     *
+     * @param jobField  The JobField to perform the filtering on
+     *
+     * @return An instance of JobRowFilter created using the supplied JobField
+     */
+    public JobRowFilter withJobField(JobField jobField) {
+        return new JobRowFilter(jobField, operation, values);
+    }
+
+    /**
+     * Construct an JobRowFilter object using the same JobField and values as the object on
+     * which this method is called and using the supplied FilterOperation.
+     *
+     * @param operation  The operation to perform (eg: eq)
+     *
+     * @return An instance of JobRowFilter created using the supplied FilterOperation
+     */
+    public JobRowFilter withOperation(FilterOperation operation) {
+        return new JobRowFilter(jobField, operation, values);
+    }
+
+    /**
+     * Construct an JobRowFilter object using the same JobField and FilterOperation as the object on
+     * which this method is called and using the supplied values.
+     *
+     * @param values  A Set of Strings to compare the JobField's value to
+     *
+     * @return  An instance of JobRowFilter created using the supplied values
+     */
+    public JobRowFilter withValues(Set<String> values) {
+        return new JobRowFilter(jobField, operation, values);
+    }
+
+    /**
+     * Extracts the JobField to be examined from the tokenizedQuery.
+     *
+     * @param tokenizedQuery  The tokenized filter expression.
+     *
+     * @return  The JobField to be examined
+     * @throws BadFilterException is the JobField does not exist
+     */
+    private JobField extractJobField(Matcher tokenizedQuery) throws BadFilterException {
+        String fieldName = tokenizedQuery.group(1);
+
+        return Arrays.stream(DefaultJobField.values())
+                .filter(field -> field.getName().equals(fieldName))
+                .findFirst()
+                .orElseThrow(() -> {
+                            LOG.debug(FILTER_JOBFIELD_UNDEFINED.logFormat(fieldName, DefaultJobField.values()));
+                            return new BadFilterException(FILTER_JOBFIELD_UNDEFINED.format(
+                                    fieldName, DefaultJobField.values()
+                            ));
+                });
+    }
+
+    /**
+     * Extracts the operation to be performed by the ApiJobStore filter query.
+     *
+     * @param tokenizedQuery  The tokenized filter expression.
+     *
+     * @return The operation to be performed by the ApiJobStore filter query.
+     * @throws BadFilterException if the operation does not exist
+     */
+    private FilterOperation extractOperation(Matcher tokenizedQuery) throws BadFilterException {
+        String operationName = tokenizedQuery.group(2);
+        try {
+            return FilterOperation.valueOf(operationName);
+        } catch (IllegalArgumentException ignored) {
+            LOG.debug(FILTER_OPERATOR_INVALID.logFormat(operationName));
+            throw new BadFilterException(FILTER_OPERATOR_INVALID.format(operationName));
+        }
+    }
+
+    /**
+     * Extracts the values to be used in the JobRowFilter query from the query.
+     *
+     * @param tokenizedQuery  The tokenized filter expression..
+     * @param filterQuery  The raw query. Used for logging.
+     *
+     * @return The set of values to be used in the JobRowFilter query.
+     * @throws BadFilterException If the fragment of the query that specifies the values is malformed.
+     */
+    private Set<String> extractValues(Matcher tokenizedQuery, String filterQuery) throws BadFilterException {
+        try {
+            // replaceAll takes care of any leading ['s or trailing ]'s which might mess up the values set.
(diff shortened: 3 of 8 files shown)


venisa:
@archolewa @cdeszaq @dayamr Please review

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/ApiJobStore.java
```diff
@@ -51,4 +53,21 @@
      * @return An Observable that emits a stream of all the JobRows in the store
      */
     Observable<JobRow> getAllRows();
+
+    /**
+     * This method takes a Set of ApiJobStoreFilters, ANDS them by default, and returns a cold observable that emits a
+     * stream of JobRows which satisfy the given filter.
+     * <p>
+     * Every field may not be filterable for every implementation of the `ApiJobStore` as the efficiency of filtering is
```
archolewa: Should replace the markdown style backticks with the JavaDoc `{@code ...}`. I too wish that JavaDocs supported Markdown, but alas this is not Elixir (also I think Java predates Markdown).

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/ApiJobStoreFilter.java
```diff
@@ -0,0 +1,213 @@
...
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * Class containing filter information to filter JobRows in ApiJobStore.
+ */
+public class ApiJobStoreFilter {
```
archolewa: I would probably call this a `JobRowFilter` since that's what it's doing.

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/ApiJobStoreFilter.java
```diff
@@ -0,0 +1,213 @@
...
+     */
+    private JobField extractJobField(Matcher tokenizedQuery) throws BadFilterException {
+        String fieldName = tokenizedQuery.group(1);
+        for (JobField field : DefaultJobField.values()) {
+            if (field.getName().equals(fieldName)) {
+                return field;
+            }
+        }
+        LOG.debug(FILTER_JOBFIELD_UNDEFINED.logFormat(fieldName));
+        throw new BadFilterException(FILTER_JOBFIELD_UNDEFINED.format(fieldName));
```
archolewa: This can be changed into a pretty little stream:

```java
return Arrays.stream(DefaultJobField.values())
                .filter(field -> field.getName().equals(fieldName))
                .findFirst()
                .orElseThrow(() -> {
                    LOG.debug(FILTER_JOBFIELD_UNDEFINED.logFormat(fieldName));
                    return new BadFilterException(FILTER_JOBFIELD_UNDEFINED.format(fieldName));
                });
```

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/ApiJobStoreFilter.java
```diff
@@ -0,0 +1,213 @@
...
+     * @return  An instance of ApiJobStoreFilter created using the supplied values
+     */
+    public ApiJobStoreFilter withValues(Set<String> values) {
+        return new ApiJobStoreFilter(jobField, operation, values);
+    }
+
+    /**
+     * Extracts the JobField to be examined from the tokenizedQuery.
+     *
+     * @param tokenizedQuery  The parsed ApiJobStore filter tokenizedQuery.
```
archolewa: This reads a little bit garbled (also the filter hasn't been parsed yet, only tokenized). I'd suggest something more like:

```java
@param tokenizedQuery  The tokenized ApiJobStoreFilter
```

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/ApiJobStoreFilter.java
```diff
@@ -0,0 +1,213 @@
...
+            }
+        }
+        LOG.debug(FILTER_JOBFIELD_UNDEFINED.logFormat(fieldName));
+        throw new BadFilterException(FILTER_JOBFIELD_UNDEFINED.format(fieldName));
+    }
+
+    /**
+     * Extracts the operation to be performed by the ApiJobStore filter query.
+     *
+     * @param tokenizedQuery  The parsed ApiJobStore filter tokenizedQuery.
```
archolewa: [ditto](https://github.com/yahoo/fili/pull/23/files#r77255575)

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/ApiJobStoreFilter.java
```diff
@@ -0,0 +1,213 @@
...
+        } catch (IllegalArgumentException ignored) {
+            LOG.debug(FILTER_OPERATOR_INVALID.logFormat(operationName));
+            throw new BadFilterException(FILTER_OPERATOR_INVALID.format(operationName));
+        }
+    }
+
+    /**
+     * Extracts the values to be used in the ApiJobStoreFilter query from the query.
+     *
+     * @param tokenizedQuery  The parsed ApiJobStore filter tokenizedQuery.
```
archolewa: [ditto](https://github.com/yahoo/fili/pull/23/files#r77255575)

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/ApiJobStoreFilterOperation.java
```diff
@@ -0,0 +1,10 @@
+// Copyright 2016 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE file distributed with this work for terms.
+package com.yahoo.bard.webservice.async.jobs.stores;
+
+/**
+ * Types of legal ApiJobStoreFilter operations.
+ */
+public enum ApiJobStoreFilterOperation {
+    eq
```
archolewa: Why are we using our own special filter operations instead of the `FilterOperation` enum? We might as well support the full filter syntax, even if the full filter semantics isn't necessary supported by every `ApiJobStore`.

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
```diff
@@ -51,4 +61,43 @@ public HashJobStore() {
     public Observable<JobRow> getAllRows() {
         return Observable.from(store.values());
     }
+
+    @Override
+    public Observable<JobRow> getFilteredRows(Set<ApiJobStoreFilter> apiJobStoreFilters)
+            throws IllegalArgumentException {
+        return Observable.from(
+                store.entrySet().stream()
```
archolewa: Since we don't care about the keys, we can just stream over the `store.values()`. Then we don't have to worry about extracting the value from the entries.

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
```diff
@@ -51,4 +61,43 @@ public HashJobStore() {
...
+                        .filter(entry -> satisfiesFilters(apiJobStoreFilters, entry.getValue()))
+                        .map(entry -> entry.getValue())
+                        .collect(Collectors.toSet())
```
archolewa: We can make this _much_ simpler if we just build a reactive chain:

```java
return getAllRows().filter(jobRow -> satisfiesFilters(apiJobStoreFilters, jobRow));
```

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
```diff
@@ -51,4 +61,43 @@ public HashJobStore() {
...
+     * @return true if the JobRow satisfies all the filters, false otherwise
+     *
+     * @throws IllegalArgumentException if a JobField in any of the filters is not a part the JobRow
+     */
+    private boolean satisfiesFilters(Set<ApiJobStoreFilter> apiJobStoreFilters, JobRow jobRow)
+            throws IllegalArgumentException {
+        for (ApiJobStoreFilter filter : apiJobStoreFilters) {
+            JobField jobField = filter.getJobField();
+            if (!jobRow.containsKey(jobField)) {
+                Set<JobField> jobFields = jobRow.keySet();
```
archolewa: I would suggest renaming this to something like `actualJobFields` rather than `jobFields`. It ensures that two variables that represent very different things (the `JobField` to filter on, vs. the `JobFields` actually in the `JobRow`) have names that differ by more than one letter.

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
```diff
@@ -51,4 +61,43 @@ public HashJobStore() {
...
+                LOG.debug(JOBFIELD_NOT_PRESENT_IN_JOB_META_DATA.logFormat(jobField, jobFields));
+                throw new IllegalArgumentException(JOBFIELD_NOT_PRESENT_IN_JOB_META_DATA.format(jobField, jobFields));
+            }
+
+            if (!filter.getValues().contains(jobRow.get(jobField))) {
+                return false;
+            }
+        }
+        return true;
```
archolewa: You know, if you really wanted to, you could pull the body of the for loop into another method `satisfiesFilter` that only looks at one JobRow and one Filter. Then, you can write this function as a very simple stream:

```java
return apiJobStoreFilters.stream().allMatch(filter -> satisfiesFilter(jobRow, filter);
```

And the method `satisfiesFilter` can be simplified to:

```java
 if(!jobRow.containsKey(jobField)) {
   // exception stuff
}
return filter.getValues().contains(jobRow.get(jobField));
```

In fact, we'll probably want to do this anyway, because we should also support the other filter operations, and this method will become too complex for my tastes if it handles the checks for every possible filter operation.

archolewa:
Finished my first pass.

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/ApiJobStoreFilterOperation.java
venisa: I initially used FilterOperation but since we are currently only supporting eq, I created a special one. Yea yea I know we can throw IllegalArgumentException. I will correct this

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/ApiJobStoreFilter.java
venisa: 👍

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
venisa: 👍

venisa:
@archolewa Addressed all comments

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
```diff
@@ -51,4 +62,58 @@ public HashJobStore() {
...
+     *
+     * @throws IllegalArgumentException if a JobField in any of the filters is not a part the JobRow
+     */
+    private boolean satisfiesFilters(Set<JobRowFilter> jobRowFilters, JobRow jobRow)
+            throws IllegalArgumentException {
+        return jobRowFilters.stream().allMatch(filter -> satisfiesFilter(jobRow, filter));
+    }
+
+    /**
+     * This method checks if the given JobRow satisfies the given JobRowFilter and returnd true if it does.
```
archolewa: Typo. `returnd` should be `returns`.

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
```diff
@@ -51,4 +62,58 @@ public HashJobStore() {
...
+     * If a JobField in the filter is not a part the JobRow, this method throws a IllegalArgumentException.
```
archolewa: Typo. `a IllegalArgumentException` should be `an IllegalArgumentException`.

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
```diff
@@ -51,4 +62,58 @@ public HashJobStore() {
...
+        if (!filterOperation.equals(FilterOperation.eq)) {
+            LOG.debug(FILTER_OPERATOR_INVALID.logFormat(filterOperation));
+            throw new IllegalArgumentException(FILTER_OPERATOR_INVALID.format(filterOperation));
+        }
+        if (!jobRow.containsKey(filterJobField)) {
+            Set<JobField> actualJobFields = jobRow.keySet();
+            LOG.debug(JOBFIELD_NOT_PRESENT_IN_JOB_META_DATA.logFormat(filterJobField, actualJobFields));
+            throw new IllegalArgumentException(
+                    JOBFIELD_NOT_PRESENT_IN_JOB_META_DATA.format(filterJobField, actualJobFields)
+            );
```
archolewa: The `HashJobStore` should probably be supporting every filter operation. It shouldn't be too difficult, because the values of the `JobRow` map are all Strings, so the filter operations all translate directly into String operations.

archolewa:
Finished my latest round. We really should be supporting every filter operation for the `HashJobStore`. Especially since they are can be translated in a very straightforward way into simple String operations.

venisa:
@archolewa Addressed second round of review comments

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
```diff
@@ -99,10 +99,10 @@ private boolean satisfiesFilters(Set<JobRowFilter> jobRowFilters, JobRow jobRow)
     private boolean satisfiesFilter(JobRow jobRow, JobRowFilter jobRowFilter) throws IllegalArgumentException {
         JobField filterJobField = jobRowFilter.getJobField();
         FilterOperation filterOperation = jobRowFilter.getOperation();
-        if (!filterOperation.equals(FilterOperation.eq)) {
-            LOG.debug(FILTER_OPERATOR_INVALID.logFormat(filterOperation));
-            throw new IllegalArgumentException(FILTER_OPERATOR_INVALID.format(filterOperation));
-        }
+        Set<String> filterValues = jobRowFilter.getValues();
+
+        String actualValue = jobRow.get(filterJobField);
```
archolewa: Either this should happen after the if-statement (extracting the value stored at a given key before checking if the key exists is a little bit weird), or the if block should check if `actualValue` is `null`. Personally I would recommend the first.

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
```diff
@@ -111,9 +111,20 @@ private boolean satisfiesFilter(JobRow jobRow, JobRowFilter jobRowFilter) throws
...
-        if (!jobRowFilter.getValues().contains(jobRow.get(filterJobField))) {
-            return false;
+        switch (filterOperation) {
+            case notin:
+                return !filterValues.contains(actualValue);
+            case startswith:
+                return filterValues.stream().anyMatch(filterValue -> actualValue.startsWith(filterValue));
+            case contains :
+                return filterValues.stream().anyMatch(filterValue -> actualValue.contains(filterValue));
+            case in:
```
archolewa: We should probably include a comment here indicating that the fall-through is intentional because `in` is a synonym for `eq`. I believe we do that in other places as well.

archolewa:
A few very minor comments. 👍 once they've been addressed.

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/ApiJobStore.java
```diff
@@ -51,4 +53,21 @@
      * @return An Observable that emits a stream of all the JobRows in the store
      */
     Observable<JobRow> getAllRows();
+
+    /**
+     * This method takes a Set of JobRowFilters, ANDS them by default, and returns a cold observable that emits a
+     * stream of JobRows which satisfy the given filter.
+     * <p>
+     * Every field may not be filterable for every implementation of the {@code ApiJobStore} as the efficiency of
```
michael-mclawhorn: 'Any of the fields may be not' would be slightly clearer than 'Every field may not be'.  As written it seems to mean that sometimes all the fields are not filterable and sometimes they are all filterable.

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/JobRowFilter.java
```diff
@@ -0,0 +1,216 @@
...
+                            LOG.debug(FILTER_JOBFIELD_UNDEFINED.logFormat(fieldName));
+                            return new BadFilterException(FILTER_JOBFIELD_UNDEFINED.format(fieldName));
+                });
+    }
+
+    /**
+     * Extracts the operation to be performed by the ApiJobStore filter query.
+     *
+     * @param tokenizedQuery  The tokenized filter expression..
+     *
```
michael-mclawhorn: Too many '.'s

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/JobRowFilter.java
```diff
@@ -0,0 +1,216 @@
...
+     * @param filterQuery  The raw query. Used for logging.
+     *
+     * @return The set of values to be used in the JobRowFilter query.
+     * @throws BadFilterException If the fragment of the query that specifies the values is malformed.
+     */
+    private Set<String> extractValues(Matcher tokenizedQuery, String filterQuery) throws BadFilterException {
+        try {
+            // replaceAll takes care of any leading ['s or trailing ]'s which might mess up the values set.
+            return new LinkedHashSet<>(
+                    FilterTokenizer.split(
```
michael-mclawhorn: Future work note, this algorithm needs to be improved to handles quoted strings containing ',' or '[' at some point.  No change needed, I just wanted to note the issue.

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
```diff
@@ -51,4 +62,69 @@ public HashJobStore() {
...
+            case contains :
+                return filterValues.stream().anyMatch(filterValue -> actualValue.contains(filterValue));
+            case in: // the fall-through is intentional because in is a synonym for eq
+            case eq:
+                return filterValues.contains(actualValue);
+            default:
+                LOG.debug(FILTER_OPERATOR_INVALID.logFormat(filterOperation));
+                throw new IllegalArgumentException(FILTER_OPERATOR_INVALID.format(filterOperation));
+
+        }
```
michael-mclawhorn: Extra whitespace.

fili-core/src/main/java/com/yahoo/bard/webservice/web/ErrorMessageFormat.java
```diff
@@ -131,6 +131,7 @@
     RESOURCE_RETRIEVAL_FAILURE("Unable to retrieve the resource for given resource name: %s. %s"),
     RESOURCE_STORAGE_FAILURE("Unable to store the resource for resource name %s. %s"),
     RESOURCE_DELETION_FAILURE("Unable to delete the resource for resource name %s. %s"),
+    UNKNOWN_EXCEPTION_MOBSTOR("Unknown error during transaction with mobstor . Additional info %s."),
```
michael-mclawhorn: We cannot reference mobstor in the public api.  This needs to move to the private yahoo packages.

michael-mclawhorn: Also, I don't see this error used in this PR.

michael-mclawhorn:
I'm happy with this. 👍

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/ApiJobStore.java
```diff
@@ -51,4 +53,21 @@
...
+
+    /**
+     * This method takes a Set of JobRowFilters, ANDS them by default, and returns a cold observable that emits a
+     * stream of JobRows which satisfy the given filter.
+     * <p>
+     * Any of the fields may be not filterable for every implementation of the {@code ApiJobStore} as the efficiency of
+     * filtering is dependent on the backing store. An IllegalArgumentException is thrown if filtering on any given
+     * field is not supported.
+     *
+     * @param jobRowFilters  A List of JobRowFilters where each JobRowFilter contains the JobField to be
```
cdeszaq: Note: The type is `Set` but this calls it a `List`

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
```diff
@@ -51,4 +62,69 @@ public HashJobStore() {
     public Observable<JobRow> getAllRows() {
         return Observable.from(store.values());
     }
+
+    @Override
+    public Observable<JobRow> getFilteredRows(Set<JobRowFilter> jobRowFilters)
+            throws IllegalArgumentException {
```
cdeszaq: Can unwrap

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/ApiJobStore.java
```diff
@@ -51,4 +53,21 @@
...
+     * filtered on, the filter operation and the values to be compared to.
+     *
+     * @return An Observable that emits a stream of JobRows that satisfy the given filters
+     *
+     * @throws IllegalArgumentException if filtering on any field is not supported
+     */
+    Observable<JobRow> getFilteredRows(Set<JobRowFilter> jobRowFilters) throws IllegalArgumentException;
```
cdeszaq: In the same way that we recently bolted Pagination on to the `SearchProvider` interface, we'll want to do that here as well.

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
```diff
@@ -51,4 +62,69 @@ public HashJobStore() {
...
+        return getAllRows().filter(jobRow -> satisfiesFilters(jobRowFilters, jobRow));
+    }
+
+    /**
+     * This method checks if the given JobRow satisfies all the JobRowFilters and returns true if it does.
+     * If a JobField in any of the filters is not a part the JobRow, this method throws a IllegalArgumentException.
```
cdeszaq: s/a part the/a part of the

cdeszaq: s/a Illegal/an Illegal

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
```diff
@@ -51,4 +62,69 @@ public HashJobStore() {
...
+     *
+     * @param jobRowFilters  A Set of JobRowFilters specifying the different conditions to be satisfied
+     * @param jobRow  The JobRow which needs to be inspected
+     *
+     * @return true if the JobRow satisfies all the filters, false otherwise
+     *
+     * @throws IllegalArgumentException if a JobField in any of the filters is not a part the JobRow
+     */
+    private boolean satisfiesFilters(Set<JobRowFilter> jobRowFilters, JobRow jobRow)
+            throws IllegalArgumentException {
```
cdeszaq: Can unwrap

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
cdeszaq: Is this comment outdated? It looks like it _does_ support all operations?

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
```diff
@@ -51,4 +62,69 @@ public HashJobStore() {
...
+
+        String actualValue = jobRow.get(filterJobField);
+
+        switch (filterOperation) {
+            case notin:
+                return !filterValues.contains(actualValue);
+            case startswith:
+                return filterValues.stream().anyMatch(filterValue -> actualValue.startsWith(filterValue));
+            case contains :
+                return filterValues.stream().anyMatch(filterValue -> actualValue.contains(filterValue));
```
cdeszaq: Both of these lambdas can be method references

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/JobRowFilter.java
cdeszaq: We already have code to handle that case (the ApiFilters handles that already, at least somewhat). There's no reason to _not_ have the filter processing here be at least as good...

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/NoOpApiJobStore.java
```diff
@@ -30,4 +32,10 @@
     public Observable<JobRow> getAllRows() {
         return Observable.empty();
     }
+
+    @Override
+    public Observable<JobRow> getFilteredRows(Set<JobRowFilter> jobRowFilter)
+            throws IllegalArgumentException {
```
cdeszaq: Can unwrap.

Also, we should probably document what this will return, since the NoOp version is a little odd, in general, and it may not be clear what filtering on the NoOp store would do.

fili-core/src/main/java/com/yahoo/bard/webservice/web/ErrorMessageFormat.java
```diff
@@ -162,7 +163,11 @@
 
     MISSING_JOB_ID("Bard experienced an internal error. Sorry.", "Missing id for job row %s"),
 
-    INVALID_ASYNC_AFTER( "Invalid 'asyncAfter' parameter: '%s'. 'asyncAfter' must be either 'never' or an integer number of milliseconds.");
+    INVALID_ASYNC_AFTER( "Invalid 'asyncAfter' parameter: '%s'. 'asyncAfter' must be either 'never' or an integer number of milliseconds."),
+
+    FILTER_JOBFIELD_UNDEFINED("Filter JobField '%s' does not exist."),
+
+    JOBFIELD_NOT_PRESENT_IN_JOB_META_DATA("JobField '%s' is not a part of job meta data. The possible fields to filter on are '%s'");
```
cdeszaq: If we move the semicolon to the next line, then we can avoid the (admittedly trivial) line removal in this file in the future 😄 

/NotABlocker

cdeszaq: How are these two errors different? Can we just use the more expressive one?

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/HashJobStore.java
venisa: Yes the comment is outdated

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/JobRowFilter.java
venisa: I copied this piece from ApiFilters...is there some piece I missed?

https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/web/ApiFilter.java#L173

fili-core/src/main/java/com/yahoo/bard/webservice/web/ErrorMessageFormat.java
venisa: I have no clue how that got there. I must have been sleep typing

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/ApiJobStore.java
venisa: Set

fili-core/src/main/java/com/yahoo/bard/webservice/web/ErrorMessageFormat.java
venisa: FILTER_JOBFIELD_UNDEFINED means if the user tries to filter on some weird JobField that does not exist in the system. Let's say the user says filter jobs by userTeam or something which is not a valid field.

JOBFIELD_NOT_PRESENT_IN_JOB_META_DATA means that the Field is a valid field in the system like dateUpdated(it is a part of `DefaultJobField`)but it is not a part of the JobRow. I am not sure if we will ever have this scenario.

May be we should treat both these scenarios as the same error. 

Thoughts @archolewa @cdeszaq ?

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/JobRowFilter.java
cdeszaq: If it's a copy, then we should be OK. We'll fix this problem in the future when we move to a grammar-based parser for all filters

fili-core/src/main/java/com/yahoo/bard/webservice/web/ErrorMessageFormat.java
archolewa: I think treating the two as the same is fine, especially since what "valid" means isn't particularly well defined. We have a default set of fields, but the `JobRow` itself has no notion of valid or invalid fields.

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/stores/ApiJobStore.java
venisa: Created a ticket for this since we will also need to add Pagination information to `getAllRows` and the focus of this PR was adding filtering functionality.

https://jira.corp.yahoo.com/browse/ADBARD-978

PR #25, opened 2016-08-31
Preserve collection order of dimensions, dimension fields and metrics

archolewa:


michael-mclawhorn:
👍 
I think I can live without a CHANGELOG entry on this one.

cdeszaq:
Changelog needed, since this changes some method signatures.

fili-core/src/main/java/com/yahoo/bard/webservice/jobs/AsyncUtils.java
```diff
@@ -46,7 +47,7 @@ public static Response makeJobMetadataResponse(String jobMetadata) {
      * @return A PreResponse containing the error information that should be sent to the user
      */
     public static PreResponse buildErrorPreResponse(Throwable throwable) {
-        ResponseContext responseContext = new ResponseContext(Collections.emptyMap());
+        ResponseContext responseContext = new ResponseContext(new LinkedHashMap<>());
```
cdeszaq: Rather than make a new object each time, it would be good if we could only create 1 object and just re-use that over and over (like what `Collections.emptyMap()` did).

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -553,7 +559,7 @@ private static void validateTimeAlignment(
             return dimension.getDimensionFields();
         } else if (showFields.size() == 1 && showFields.contains(DimensionFieldSpecifierKeywords.NONE.toString())) {
             // Show no fields
-            return Collections.emptySet();
+            return new LinkedHashSet<>();
```
cdeszaq: Should use singleton of this if we can

archolewa: What do you mean `singleton of this`? There are no elements, it's an empty set. Do you mean `Collections.emptySet`? Can't do that because, the `Collections.emptySet()` is of type `Set` not `LinkedHashSet`, which is the new return type for this method.

fili-core/src/main/java/com/yahoo/bard/webservice/jobs/AsyncUtils.java
archolewa: That sounds dangerous to me, not unless we take steps to make sure the empty map is immutable. The reason `Collections.emptyMap` is safe is because it's immutable, so you don't have to worry about someone accidentally modifying the map and spreading the changes all over the system.

Frankly, the overhead of building an empty collection is _so tiny_ that I doubt creating our own custom empty, immutable `LinkedHashMap`is not worth the effort.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
archolewa: I understand your comment now. You mean a singleton instance of an empty `LinkedHashMap`, not a map with one key-value pair, for which I have [the same concerns as with a singleton `LinkedHashSet`](https://github.com/yahoo/fili/pull/25#discussion_r77239307)

venisa:
👍

PR #26, opened 2016-08-31
Enrich jobs endpoint with filtering functionality

venisa:
This PR adds filtering capability to Job endpoint.
The default job payload also contains userId information

Please note: This PR is based on https://github.com/yahoo/fili/pull/23
Please ignore the first commit


fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
```diff
@@ -164,8 +165,15 @@ public void getJobs(
...
 
-            apiRequest.getJobViews()
-                    .toList()
+            Observable<Map<String, String>> jobViews;
+
+            if (filters == null || "".equals(filters)) {
+                jobViews = apiRequest.getJobViews();
+            } else {
+                jobViews = apiRequest.getFilteredJobViews(apiRequest.buildJobStoreFilter(filters));
+            }
```
venisa: Now that I look at this...I am not sure if it's better to do this here or to call a generic getJobViews(filters) and have JobsApiRequest perform this check and call the correct function accordingly

venisa:
@dayamr @archolewa @cdeszaq Please review

fili-core/src/main/java/com/yahoo/bard/webservice/web/JobsApiRequest.java
```diff
@@ -124,6 +129,46 @@ public JobsApiRequest(
...
+     * JobRequestFailedException is returned. If the ApiJobStore is empty, we return an empty Observable.
+     *
+     * @param apiJobStoreFilters  A Set containing apiJobStoreFilters which contain all the filter information needed
+     * to filter JobRows
+     *
+     * @return An Observable containing a stream of Maps representing the jobs to be returned to the user
+     */
+    public Observable<Map<String, String>> getFilteredJobViews(Set<ApiJobStoreFilter> apiJobStoreFilters) {
+        return apiJobStore.getFilteredRows(apiJobStoreFilters)
+                .map(this::mapJobRowsToJobViews);
```
archolewa: Can be unwrapped.

fili-core/src/main/java/com/yahoo/bard/webservice/web/JobsApiRequest.java
```diff
@@ -124,6 +129,46 @@ public JobsApiRequest(
...
+     *
+     * @param filterQuery  Expects a URL filterQuery String that may contain multiple filters separated by
+     * comma.  The format of a filter String is :
+     * (JobField name)-(operation)[(value or comma separated values)]?
+     *
+     * @return  A Set of ApiJobStoreFilters
+     */
+    public LinkedHashSet<ApiJobStoreFilter> buildJobStoreFilter(@NotNull String filterQuery) {
+        // split on '],' to get list of filters
+        return Arrays.asList(filterQuery.split(COMMA_AFTER_BRACKET_PATTERN)).stream()
```
archolewa: Can be replaced by `Arrays.stream(filterQuery.split(COMMA_AFTER_BRACKET_PATTERN))`

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
archolewa: I think it would be better if we did what the `DataServlet` does: pass in the filters to the `JobsApiRequest` and let it parse the filters internally. Then, when `getJobViews()` is invoked, the `JobsApiRequest` applies the filters if present. So, in the `DataServlet`, the code would look like:

```java
            JobsApiRequest apiRequest = new JobsApiRequest (
                    format,
                    null,  //asyncAfter is null so it behaves like a synchronous request
                    perPage,
                    page,
                    filters, //Note the addition of the filters to the constructor
                    uriInfo,
                    jobPayloadBuilder,
                    apiJobStore,
                    preResponseStore,
                    broadcastChannel
            );
... //Other stuff
jobViews = apiRequest.getJobViews(); //Inside JobsApiRequest::getJobViews, JobsApiRequest::getFilteredJobViews is invoked if there are filters
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/jobs/ApiJobStoreFilterSpec.groovy
```diff
@@ -0,0 +1,67 @@
...
+
+        then:
+        apiJobStoreFilter.jobField?.name == jobField
+        apiJobStoreFilter.operation == ApiJobStoreFilterOperation.valueOf(op)
+        apiJobStoreFilter.values == expected as Set
+
+        where:
+        jobField  | op   | values           | expected
+        'userId'  | 'eq' | '[foo,bar,baz]'  | ['foo', 'bar', 'baz']
+        'userId'  | 'eq' | '[foo]'          | ['foo']
```
archolewa: We should have at least one test for each operation.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/JobsApiRequestSpec.groovy
```diff
@@ -232,6 +195,78 @@ class JobsApiRequestSpec extends Specification {
...
+        filters.size() == 1
+        ApiJobStoreFilter filter = filters[0]
+        filter.jobField?.name == "userId"
+        filter.operation == ApiJobStoreFilterOperation.valueOf("eq")
+        filter.values == ["foo"] as Set
+    }
+
+    def "buildJobStoreFilter correctly parses multiple filter query"() {
+        setup:
+        String filterQuery = "userId-eq[foo],status-eq[successful]"
```
archolewa: We should probably use `success` rather than `successful` since `success` is a `DefaultJobStatus`, while `successful` is not, so that our test data is actually valid.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletSpec.groovy
```diff
@@ -155,6 +159,28 @@ class JobsServletSpec extends Specification {
...
+                                            "userId": "momo"
+                                        }
+                                  ]}"""
+
+        when: "We send a request to the /jobs endpoint with a valid filters parameter"
+        String result = makeRequest("/jobs", [filters : ["userId-eq[momo]"]])
+
+        then: "We only get the job payload that satisfies the filter"
+        GroovyTestUtils.compareJson(result, expectedResponse, JsonSortStrategy.SORT_BOTH)
+        }
```
archolewa: Should be unindented one level.

It would be nice if we could also at a minimum have a test verifying that we get back the empty list if none of the tickets satisfy the given filter.

archolewa:
Finished my first round. Overall it looks pretty good, though the testing feels a bit sparse. It would be nice if we could have some more tests involving different operations than just `eq`, and combinations of operations. The `SearchProviderSpec` has some good examples of very thorough filter tests. I don't feel a strong need to be quite that thorough here, but we may want to lift a few of the more complex filter tests from that specification.

venisa:
@archolewa I will add the tests for the filters in https://github.com/yahoo/fili/pull/23 as that is where the code was added

venisa:
@archolewa 
Added tests for each operator: https://github.com/yahoo/fili/pull/23/commits/89ca91d9dfa8c4d4e39397dc132dea2c57ff2804

Added test for multiple filter query with different operators: https://github.com/yahoo/fili/pull/23/commits/a93cb6cc2169661fe8a701b6aa638e98334e83a1

Added test for ensuring we return empty list when no filters are satisfied: https://github.com/yahoo/fili/pull/26/commits/d01a771db3538a8e8a251e770c79561c014a7d2d

We already have tests that the different filter operators select the correct JobRows:
https://github.com/yahoo/fili/blob/a93cb6cc2169661fe8a701b6aa638e98334e83a1/fili-core/src/test/groovy/com/yahoo/bard/webservice/async/jobs/stores/HashJobStoreSpec.groovy

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletSpec.groovy
```diff
@@ -155,6 +159,36 @@ class JobsServletSpec extends Specification {
...
+        then: "We only get the job payload that satisfies the filter"
+        GroovyTestUtils.compareJson(result, expectedResponse, JsonSortStrategy.SORT_BOTH)
+    }
+
+    def "jobs endpoint returns an empty list when none of the JobRows satisfy the filter"() {
+        when: "We send a request to the /jobs endpoint with a valid filters parameter"
+        String result = makeRequest("/jobs", [filters : ["userId-eq[pikachu]"]])
+
+        then: "We only get the job payload that satisfies the filter"
+        GroovyTestUtils.compareJson(result, "{\"jobs\":[]}", JsonSortStrategy.SORT_BOTH)
```
archolewa: If you use single quotes to delimitate the string, you don't have to escape the double quotes around `jobs`.

archolewa:
One small comment. 👍 once it's been addressed.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
```diff
@@ -164,8 +165,15 @@ public void getJobs(
                                     )
                     );
 
-            apiRequest.getJobViews()
-                    .toList()
+            Observable<Map<String, String>> jobViews;
+
+            if (filters == null || "".equals(filters)) {
+                jobViews = apiRequest.getJobViews();
```
michael-mclawhorn: If you give filters a default value in the parameter list for this method, then you wouldn't have to test the null case.

michael-mclawhorn: Sorry to be late bringing this up, but in the other ApiRequest objects we've avoided using the 'get' verb when the method is essentially a factory.  Since the ApiRequest continues to be both a data object and a factory, it would be completely obscure whether 'getFoo()' was retrieving a built foo or an accessor for an already built foo.  

This applies to getJobViews, but not by any means to just getJobViews in this code now.

michael-mclawhorn: I'll drop this issue for this PR, but I want someone to pick this up sooner rather than later:

https://jira.corp.yahoo.com/browse/ADBARD-977

fili-core/src/main/java/com/yahoo/bard/webservice/jobs/ApiJobStoreFilterOperation.java
```diff
@@ -0,0 +1,10 @@
+// Copyright 2016 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE file distributed with this work for terms.
+package com.yahoo.bard.webservice.jobs;
+
+/**
+ * Types of legal ApiJobStoreFilter operations.
+ */
+public enum ApiJobStoreFilterOperation {
+    eq
```
cdeszaq: There should be many more in here, right?

archolewa: This should be deleted (it's cruft from an earlier version of #23). Basically, I think this PR needs to be rebased on top of #23.

venisa: That's right. I am working on some final changes for #23 after which I will rebase

venisa:
@michael-mclawhorn This has been rebased on top of master after merging #23

fili-core/src/main/java/com/yahoo/bard/webservice/web/JobsApiRequest.java
```diff
@@ -112,15 +123,45 @@ public JobsApiRequest(
     }
 
     /**
-     * Returns an Observable containing a stream of job payloads for all the jobs in the ApiJobStore. If, for any
-     * JobRow, the mapping from JobRow to job view fails, an Observable over JobRequestFailedException is returned. If
-     * the ApiJobStore is empty, we return an empty Observable.
+
```
cdeszaq: Space not needed

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/JobsApiRequestSpec.groovy
```diff
@@ -230,23 +192,116 @@ class JobsApiRequestSpec extends Specification {
...
+                syncResults : "https://localhost:9998/v1/jobs/2/results?asyncAfter=never",
+                self : "https://localhost:9998/v1/jobs/2",
+                status : "pending",
+                dateCreated : "0002-04-29T00:00:00.000Z",
+                userId : "Number 2"
+        ]
+
         when:
-        apiRequest.getJobViews().subscribe(errorSubscriber)
+        apiRequest.getJobViews().subscribe(testSubscriber)
```
cdeszaq: Should this be calling with a filter?

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
```diff
@@ -138,9 +139,10 @@ public void getJobs(
 
             JobsApiRequest apiRequest = new JobsApiRequest (
                     format,
-                    null, //asynAfter is null so it behaves like a synchronous request
+                    null, //asyncAfter is null so it behaves like a synchronous request
```
michael-mclawhorn: Async after shouldn't be rolling up to the parent object.  This requires a larger design discussion.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/JobsApiRequestSpec.groovy
venisa: Nope. The filter is now baked into apiRequest

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
michael-mclawhorn: To be clear, this isn't a ding against this PR, but rather prior ApiRequest changes around asyncAfter that might need to be revisited.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletSpec.groovy
```diff
@@ -155,6 +159,36 @@ class JobsServletSpec extends Specification {
...
+        then: "We only get the job payload that satisfies the filter"
+        GroovyTestUtils.compareJson(result, expectedResponse, JsonSortStrategy.SORT_BOTH)
+    }
+
+    def "jobs endpoint returns an empty list when none of the JobRows satisfy the filter"() {
+        when: "We send a request to the /jobs endpoint with a valid filters parameter"
+        String result = makeRequest("/jobs", [filters : ["userId-eq[pikachu]"]])
+
+        then: "We only get the job payload that satisfies the filter"
+        GroovyTestUtils.compareJson(result, '{"jobs":[]}', JsonSortStrategy.SORT_BOTH)
```
cdeszaq: Sort not needed here, right?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletSpec.groovy
```diff
@@ -155,6 +159,36 @@ class JobsServletSpec extends Specification {
...
+                                            "syncResults":"http://localhost:9998/jobs/ticket1/results?asyncAfter=never",
+                                            "userId": "momo"
+                                        }
+                                  ]}"""
+
+        when: "We send a request to the /jobs endpoint with a valid filters parameter"
+        String result = makeRequest("/jobs", [filters : ["userId-eq[momo]"]])
+
+        then: "We only get the job payload that satisfies the filter"
+        GroovyTestUtils.compareJson(result, expectedResponse, JsonSortStrategy.SORT_BOTH)
```
cdeszaq: Sorting on Map (the default I think) should be sufficient, right?

cdeszaq:
👍 after those last couple of small tweaks

PR #27, opened 2016-08-31
Remove timing from JobsApiRequestSpec

archolewa:
-- The spec has test that verifies that when the timeout expires, an
empty Observable is emitted. However, the timeout was 2 milliseconds,
and the main test thread was only sleeping 5 milliseconds before sending
the message. These two times were too close together, so that the test
may result in a false failure in a highly contested environment (i.e.
Travis). So the test was rewritten so that it doesn't rely on any timing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@ Current
 
 #### Changed:
 
+-  [Removed timing component in JobsApiRequestSpec](https://github.com/yahoo/fili/pull/27)
+    * Rather than setting an async timeout, and then sleeping, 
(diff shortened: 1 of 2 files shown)


cdeszaq:
👍

michael-mclawhorn:
You say you increased the sleep time, but I don't see that change.

archolewa:
@michael-mclawhorn You were seeing a slightly out of date title for the PR. Originally I was just increasing the buffer for the timing, but then I realized that we could test what we wanted to test without worrying about any timing at all (besides the standard timeout timing we see on many tests). I made that change, but forgot at first to update the name of the PR.

michael-mclawhorn:
Makes sense.
👍

PR #29, opened 2016-08-31
Add jobs endpoint documentation

venisa:
Add documentation for Jobs endpoint.

Please note: 
I have provided a very brief description for async. However, the focus of this PR is to document the Jobs endpoint.

@cdeszaq @archolewa you guys had reviewed this when this was opened in bard.
I addressed all outstanding comments and added a new commit for filters on jobs endpoint

### DIFF
--- a/api-docs.md
+++ b/api-docs.md
@@ -27,6 +27,8 @@ Table of Contents
     - [Filtering](#filtering)
     - [Having](#having)
     - [Sorting](#sorting)
+- [Asynchronous Queries](#asynchronous-queries)
+    - [Jobs endpoint](#jobs-endpoint)
 - [Misc](#misc)
     - [Dates and Times](#dates-and-times)
     - [Case Sensitivity](#case-sensitivity)
@@ -561,6 +563,102 @@ There are, however, a few catches to this:
 - Sorting is only applied at the Druid level. Therefore, the results of a sort are not guaranteed to be accurate if Fili 
 performs any post-Druid calculations on one of the metrics that you are sorting on.
 
+Asynchronous Queries
+--------------------
+Fili supports asynchronous data queries. A new parameter `asyncAfter` is added on data queries. The `asyncAfter`
+parameter will control whether a data query should always be synchronous, or transition from synchronous to asynchronous
+ on the fly. If `asyncAfter=never` then Fili will wait indefinitely for the data, and hold the connection with the
+ client open as long as allowed by the network. This will be the default. However, the default behavior of `asyncAfter`
+ may be modified by setting the `default_asyncAfter` configuration parameter. If `asyncAfter=0`, the query is
+ asynchronous immediately.
+
+If it takes less time than the timeout for the data to be sent back, then the results are sent to the client.
+If `asyncAfter=t` for `t` an integer in milliseconds then the query will start synchronously. If the query takes longer
+than `t` milliseconds, then the query becomes asynchronous. If the timeout passes, and the data has not come back, then
+the user receives a `202 Accepted` response and the [job meta-data](#job-meta-data).
+
+### Jobs Endpoint
+The jobs endpoint is the one stop shop for queries about asynchronous jobs. This endpoint is responsible for:
+
+1. Providing a list of [all jobs](#get-ting-summary-of-all-jobs) in the system.
+2. Providing the status of a [particular job](#get-ting-job-status) queried via the `jobs/TICKET` resource.
+3. Providing access to the [results](#get-ting-results) via the `jobs/TICKET/results` resource.
+
+#### GET-ting summary of all jobs
+A user may get the status of all jobs by sending a `GET` to `jobs` endpoint.
+
+```
+https://HOST:PORT/v1/jobs
+```
+If no jobs are available in the system, an empty collection is returned.
+
+The `jobs` endpoint supports filtering on job fields (i.e. `userId`, `status`), using the same syntax as the
+[data endpoint filters](#filtering). For example:
+
+`userId-eq[greg, joan], status-eq[success]`
+
+resolves into the following boolean operation:
+
+`(userId = greg OR userId = joan) AND status = success`
+
+which will return only those Jobs created by `greg` and `joan` that have completed successfully.
+
+#### GET-ting Job Status
+When the user sends a `GET` request to `jobs/TICKET`, Fili will look up the specified ticket and return the job's
+meta-data as follows:
+
+###### Job meta-data
+```json
+{
+    "query": "https://HOST:PORT/v1/data/QUERY",
+    "results": "https://HOST:PORT/v1/jobs/TICKET/results",
+    "syncResults": "https://HOST:PORT/v1/jobs/TICKET/results?asyncAfter=never",
+    "self": "https://HOST:PORT/v1/jobs/TICKET",
+    "status": ONE OF ["pending", "success", "error"],
+    "jobTicket": "TICKET",
+    "dateCreated": "ISO 8601 DATETIME",
+    "dateUpdated": "ISO 8601 DATETIME",
+    "userId": "Foo"
+}
+```
+* `query` is the original query
+* `results` provides a link to the data, whether it is fully synchronous or switches from
+   synchronous to asynchronous after a timeout depends on the default setting of `asyncAfter`.
+* `syncResults` provides a synchronous link to the data (note the `asyncAfter=never` parameter)
+* `self` provides a link that returns the most up-to-date version of this job
+* `status` indicates the status of the results pointed to by this ticket
+    - `pending` - The job is being worked on
+    - `success` - The job has been completed successfully
+    - `error` - The job failed with an error
+    - `canceled` - The job was canceled by the user (coming soon)
+* `jobTicket` is a unique identifier for the job
+* `dateCreated` is the date on which the job was created
+* `dateUpdated` when the job's status was last updated
+* `userId` is an identifier for the user who submitted this job
+
+If the ticket is not available in the system, we get a 404 error with the message `No job found with job ticket TICKET`
+
+#### GET-ting Results
+The user may access the results of a query by sending a `GET` request to `jobs/TICKET/results`. This
+resource takes the following parameters:
+
+ 1. **`format`** - Allows the user to specify a response format, i.e. csv, or JSON. This behaves
+just like the [`format`](#response-format) parameter on queries sent to the `data` endpoint.
+
+2. **`page`, `perPage`** - The [pagination](#pagination--limit) parameters. Their behavior is the same as when sending
+a query to the `data` endpoint, and allow the user to get pages of the results.
+
+3. **`asyncAfter`** - Allows the user to specify how long they are willing to wait for results from the
+result store. Behaves like the [`asyncAfter`](async) parameter on the `data` endpoint.
+
+If the results for the given ticket are ready, we get the results in the format specified. Otherwise, we get the
+[job's metadata](#job-meta-data).
+
+##### Long Polling
+If clients wish to long poll for the results, they may send a `GET` request to
+`https://HOST:PORT/v1/jobs/TICKET/results?asyncAfter=never` (the query linked to under the
+`syncResults` field in the async response). This request will perform like a synchronous query: Fili
+will not send a response until all of the data is ready.
 
 Misc
 ----


api-docs.md
```diff
@@ -561,6 +563,98 @@ There are, however, a few catches to this:
...
+```json
+{
+    "query": "https://HOST:PORT/v1/data/QUERY",
+    "results": "https://HOST:PORT/v1/jobs/TICKET/results",
+    "syncResults": "https://HOST:PORT/v1/jobs/TICKET/results?asyncAfter=never",
+    "self": "https://HOST:PORT/v1/jobs/TICKET",
+    "status": ONE OF ["pending", "success", "error"],
+    "jobTicket": "TICKET",
+    "dateCreated": "ISO 8601 DATETIME",
+    "userId": "Foo"
```
venisa: @archolewa This does not include dateUpdated as our job payload does not currently give dateUpdated. (It's not a part of the job payload in the design docs)

venisa:
@cdeszaq @archolewa Please review

venisa:
@archolewa Added dateUpdated

api-docs.md
```diff
@@ -561,6 +563,100 @@ There are, however, a few catches to this:
 - Sorting is only applied at the Druid level. Therefore, the results of a sort are not guaranteed to be accurate if Fili 
 performs any post-Druid calculations on one of the metrics that you are sorting on.
 
+Asynchronous Queries
+----
```
archolewa: The `---` should be the same length as the title.

api-docs.md
```diff
@@ -561,6 +563,100 @@ There are, however, a few catches to this:
...
+Fili supports asynchronous data queries. A new parameter `asyncAfter` is added on data queries. The `asyncAfter`
+parameter will control whether a data query should always be synchronous, or transition from synchronous to asynchronous
+ on the fly. If `asyncAfter=never` then Fili will wait indefinitely for the data, and hold the connection with the
+ client open as long as allowed by the network. This will be the default. However, the default behavior of `asyncAfter`
+ may be modified by setting the `default_asyncAfter` configuration parameter. If `asyncAfter=0`, the query is
+ asynchronous immediately.
+
+If it takes less time than the timeout for the data to be sent back, then the results are sent to the client.
+If `asyncAfter=t` for t an integer in milliseconds then the query will start synchronously. If the query takes a longer
```
archolewa: I would wrap the `t` in back-quotes in `for t an integer`.

archolewa: Should drop `a` in `If the query takes a longer`

api-docs.md
```diff
@@ -561,6 +563,100 @@ There are, however, a few catches to this:
...
+than t milliseconds, then the query becomes asynchronous. If the timeout passes, and the data has not come back, then
```
archolewa: [ditto](https://github.com/yahoo/fili/pull/29/files#r77696623)

api-docs.md
```diff
@@ -561,6 +563,100 @@ There are, however, a few catches to this:
...
+only returns the JobRows that satisfy the filter condition.
+
+The format of a filter string is:
+```
+jobFieldName-filterOperation[values]
+```
+Multiple comma separated filters are supported and `And`ed by default
+
+```
+https://HOST:PORT/v1/jobs
```
archolewa: I think this is supposed to be between lines 588 and 590.

api-docs.md
```diff
@@ -561,6 +563,100 @@ There are, however, a few catches to this:
...
+#### GET-ting status of all tickets
+A user may get the status of all tickets by sending a `GET` to `jobs` endpoint.
+
+The `jobs` endpoint supports the `filters` query param that filters the JobRows according to the filter condition and
+only returns the JobRows that satisfy the filter condition.
+
+The format of a filter string is:
+```
+jobFieldName-filterOperation[values]
+```
```
archolewa: We should have a bit more information about filters. Fortunately, the data section should provide that information, since the syntax is the same. So we should say that users may filter on job fields (i.e. `userId`), show the same syntax, and then link to the data filters for more information about the filter syntax.

archolewa:
Finished my latest pass. A bit more about the filters (a link to the data filter section will go a long way) would be nice.

api-docs.md
```diff
@@ -561,6 +563,101 @@ There are, however, a few catches to this:
...
+#### GET-ting status of all tickets
+A user may get the status of all tickets by sending a `GET` to `jobs` endpoint.
+
+```
+https://HOST:PORT/v1/jobs
+```
+If no jobs are available in the system, we get an empty collection.
+
+The `jobs` endpoint supports the `filters` query param that filters the JobRows according to the filter condition and
+only returns the JobRows that satisfy the filter condition.
```
archolewa: I think I would rephrase this to make it clearer _what_ we can filter on:

```
The `jobs` endpoint supports filtering on job fields (i.e. `userId`, `status`), using the same syntax as the [data endpoint filters](#filtering). For example:

```
userId-eq[greg, joan], status-eq[success]
```
resolves into the following boolean operation:

(userId = greg OR userId = joan) AND status = success

which will return only those Jobs created by `greg` and `joan` that have completed successfully.
```

archolewa:
Made a small suggestion to hopefully make the filter documentation a bit clearer. 👍 Once that's been addressed.

api-docs.md
```diff
@@ -561,6 +563,101 @@ There are, however, a few catches to this:
...
+If `asyncAfter=t` for `t` an integer in milliseconds then the query will start synchronously. If the query takes longer
+than `t` milliseconds, then the query becomes asynchronous. If the timeout passes, and the data has not come back, then
+the user receives a `202 Accepted` response and the [job meta-data](#job-meta-data).
+
+### Jobs Endpoint
+The jobs endpoint is the one stop shop for queries about asynchronous jobs. This endpoint is responsible for:
+
+1. Providing the status of tickets queried via the `jobs/TICKET` resource.
+2. Provide a list of all jobs in the system.
+3. Providing access to the results via the `jobs/TICKET/results` resource.
```
cdeszaq: Would be good to have these in order of coarsest to finest resource (ie. `/jobs`, `/jobs/TICKET`, and `/jobs/TICKET/results`)

/NotABlocker

api-docs.md
```diff
@@ -561,6 +563,101 @@ There are, however, a few catches to this:
...
+
+#### GET-ting status of all tickets
+A user may get the status of all tickets by sending a `GET` to `jobs` endpoint.
```
cdeszaq: This isn't really getting the _status_ of all tickets, it's just getting _all tickets_ (specifically, a "summary" version of each of them).

api-docs.md
```diff
@@ -561,6 +563,101 @@ There are, however, a few catches to this:
...
+
+```
+https://HOST:PORT/v1/jobs
+```
+If no jobs are available in the system, we get an empty collection.
```
cdeszaq: s/we get an empty collection/an empty collection is returned

api-docs.md
cdeszaq: To add to @archolewa comments a bit: the core idea here is to tie it back to the filters that users are already familiar with, and leave the terminology centered around "jobs" since that's what they are interacting with. JobRow is an internal concept, really, so we don't want to expose that in the API if we don't have to

cdeszaq: Also, add a link in there back up to where we talk about filters in general

api-docs.md
```diff
@@ -561,6 +563,101 @@ There are, however, a few catches to this:
...
+only returns the JobRows that satisfy the filter condition.
+
+The format of a filter string is:
+```
+jobFieldName-filterOperation[values]
+```
+Multiple comma separated filters are supported and `And`ed by default. Filters on jobs endpoint are similar to
+[filters](#filtering) on data endpoint where jobField name is used to specify the field to filter on.
+
+#### GET-ting Ticket Status
```
cdeszaq: s/Ticket/Job

api-docs.md
```diff
@@ -561,6 +563,101 @@ There are, however, a few catches to this:
...
+When the user sends a `GET` request to `jobs/TICKET`, Fili will look up the specified ticket and return the ticket's
+meta-data as follows:
```
cdeszaq: s/ticket/job

api-docs.md
```diff
@@ -561,6 +563,101 @@ There are, however, a few catches to this:
...
+```json
+{
+    "query": "https://HOST:PORT/v1/data/QUERY",
+    "results": "https://HOST:PORT/v1/jobs/TICKET/results",
+    "syncResults": "https://HOST:PORT/v1/jobs/TICKET/results?asyncAfter=never",
+    "self": "https://HOST:PORT/v1/jobs/TICKET",
+    "status": ONE OF ["pending", "success", "error"],
+    "jobTicket": "TICKET",
+    "dateCreated": "ISO 8601 DATETIME",
+    "dateUpdated": "ISO 8601 DATETIME",
```
cdeszaq: Are we including all of these fields now? I know there was some confusion previously... I'm in favor of including all of these

api-docs.md
```diff
@@ -561,6 +563,101 @@ There are, however, a few catches to this:
...
+    "userId": "Foo"
+}
+```
+* `query` is the original query
+* `results` provides a link to the data, whether it is fully synchronous or switches from
+   synchronous to asynchronous after a timeout depends on the default setting of `asyncAfter`.
+* `syncResults` provides a synchronous link to the data (note the `asyncAfter=never` parameter)
+* `self` provides a link that returns the most up-to-date version of this metadata
```
cdeszaq: s/this metadata/this job metadata

api-docs.md
cdeszaq: Could turn each of these into links into the section that talks about it below.

/NotABlocker

cdeszaq:
Just a couple final small things and this looks good

api-docs.md
archolewa: I think we should. Especially since this is JSON (so fields can be easily ignored), I would rather err on the side of including more rather than less, so that we don't have to go back and add them later (thereby changing the API...), especially since it's so easy to ignore unwanted fields in JSON.

cdeszaq: @archolewa, additions to the API are _safe_ (clients should be OK with stuff they are not looking for being there), but removals are  unsafe. Your thoughts about changing the API are, therefor, backwards 😉

api-docs.md
venisa: Added link to filters in the latest commit

api-docs.md
venisa: The only missing field is dateUpdated. @archolewa will be adding all the fields as a part of his change to Fili

venisa:
@archolewa @cdeszaq all comments have been addressed

api-docs.md
archolewa: @cdeszaq Safety, shmafety. I just don't want to interrupt whatever I'm doing in six months to come back and add the `dateUpdated` field because some client decided they wanted it.

api-docs.md
```diff
@@ -561,6 +563,102 @@ There are, however, a few catches to this:
...
+than `t` milliseconds, then the query becomes asynchronous. If the timeout passes, and the data has not come back, then
+the user receives a `202 Accepted` response and the [job meta-data](#job-meta-data).
+
+### Jobs Endpoint
+The jobs endpoint is the one stop shop for queries about asynchronous jobs. This endpoint is responsible for:
+
+1. Provide a list of [all jobs](#get-ting-summary-of-all-jobs) in the system.
+2. Providing the status of a [particular job](#get-ting-job-status) queried via the `jobs/TICKET` resource.
+3. Providing access to the [results](#get-ting-results) via the `jobs/TICKET/results` resource.
+
```
michael-mclawhorn: #1 should probably also be Providing.

api-docs.md
```diff
@@ -561,6 +563,102 @@ There are, however, a few catches to this:
...
+#### GET-ting Results
+The user may access the results of a query by sending a `GET` request to `jobs/TICKET/results`. This
+resource takes the following parameters:
+
+ 1. **`format`** - Allows the user to specify a response format, i.e. csv, or JSON. This behaves
+just like the [`format`](#response-format) parameter on queries sent to the `data` endpoint.
+
+2. **`page`, `perPage`** - The [pagination](#pagination--limit) parameters. Their behavior is the same as when sending
+a query to the `data` endpoint, and allow the user to get pages of the results.
```
michael-mclawhorn: Should we flag this as also 'coming soon'?  It's blocked on: 
https://github.com/yahoo/fili/issues/36

api-docs.md
```diff
@@ -580,30 +580,31 @@ the user receives a `202 Accepted` response and the [job meta-data](#job-meta-da
...
+The `jobs` endpoint supports filtering on job fields (i.e. `userId`, `status`), using the same syntax as the
+[data endpoint filters](#filtering). For example:
 
-The format of a filter string is:
-```
-jobFieldName-filterOperation[values]
-```
-Multiple comma separated filters are supported and `And`ed by default. Filters on jobs endpoint are similar to
-[filters](#filtering) on data endpoint where jobField name is used to specify the field to filter on.
+userId-eq[greg, joan], status-eq[success]
```
cdeszaq: Should be "code" (ie. back-tick-escaped)

api-docs.md
```diff
@@ -580,30 +580,31 @@ the user receives a `202 Accepted` response and the [job meta-data](#job-meta-da
...
+
+resolves into the following boolean operation:
+
+(userId = greg OR userId = joan) AND status = success
```
cdeszaq: Same here (should be code)

api-docs.md
cdeszaq: I wouldn't worry about it too much. We'll have it shortly, and this is a patch version, not a stable version, so it's OK for the docs to be a bit ahead of the implementation.

api-docs.md
cdeszaq: @archolewa the flip-side: all the various clients (ie. many more engineers) don't want to come back in 6 months to handle a field getting removed if they were depending on it.

cdeszaq:
👍 after the few final small tweaks

michael-mclawhorn:
👍 Seems fine.

PR #32, opened 2016-08-31
increase sleep time for flaky tests

venisa:
JobsServletReactiveChainforResultsEndpointSpec had a number of tests with sleep to mimic several situations in which we get results from via BroadcastChannel notification or through PreResponseStore directly.

The Thread sleep duration was too small causing atleast 1 test to fail in Travis:
https://travis-ci.org/yahoo/fili/builds/156637523

This PR increased the sleep duration to ensure that the timeouts are not too close together


venisa:
@archolewa @dayamr Please review

archolewa:
Increasing test latency by a second per test makes me 😢. I think the key problem here is that we don't have the abstractions right for the `JobsApiRequest`. My current thoughts are that the `JobsApiRequest` shouldn't have any notion of timeouts. Instead, it has a notion of `asynchronous vs. synchronous` queries, and then builders of the `JobsApiRequest` can inject the method of determining whether a query is synchronous or asynchronous (i.e. via a timeout). That being said, it isn't a high enough priority to dig into now, and this is good enough, so

👍

dayamr:
👍

PR #34, opened 2016-09-02
QTL Integration Style Tests

garyluoex:
* Added two tests `LookupDimensionFilteringDataServeletSpec` and `LookupDimensionGroupingDataServletSpec` to test QTL functionality
* Modified some testing resources (PETS table and corresponding dimensions) to allow better testing on `LookupDimension`s

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
@@ -35,7 +35,7 @@ import com.yahoo.bard.webservice.data.config.ConfigurationLoader;
 import com.yahoo.bard.webservice.data.config.ResourceDictionaries;
 import com.yahoo.bard.webservice.data.config.dimension.DimensionConfig;
 import com.yahoo.bard.webservice.data.config.dimension.DimensionLoader;
-import com.yahoo.bard.webservice.data.config.dimension.KeyValueStoreDimensionLoader;
+import com.yahoo.bard.webservice.data.config.dimension.TypeAwareDimensionLoader;
 import com.yahoo.bard.webservice.data.config.metric.MetricLoader;
 import com.yahoo.bard.webservice.data.config.table.TableLoader;
 import com.yahoo.bard.webservice.data.dimension.DimensionDictionary;
@@ -837,7 +837,7 @@ public abstract class AbstractBinderFactory implements BinderFactory {
      * @return a Dimension Loader instance
      */
     protected DimensionLoader getDimensionLoader() {
-        return new KeyValueStoreDimensionLoader(getDimensionConfigurations());
+        return new TypeAwareDimensionLoader(getDimensionConfigurations());
     }
 
     /**
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/web/TableFullViewProcessorSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/web/TableFullViewProcessorSpec.groovy
@@ -68,14 +68,14 @@ class TableFullViewProcessorSpec extends Specification {
     def "Check table meta data info at grain level"() {
 
         setup:
-        String expectedResponse = "[name:all, description:The pets all grain, metrics:[[category:General, name:rowNum, longName:rowNum, uri:http://localhost:9998/v1/rowNum], [category:General, name:limbs, longName:limbs, uri:http://localhost:9998/v1/limbs], [category:General, name:dayAvgLimbs, longName:dayAvgLimbs, uri:http://localhost:9998/v1/dayAvgLimbs]], retention:P1Y, longName:All, dimensions:[[category:General, name:species, longName:species, uri:http://localhost:9998/v1/species, cardinality:0], [category:General, name:sex, longName:sex, uri:http://localhost:9998/v1/sex, cardinality:0], [category:General, name:breed, longName:breed, uri:http://localhost:9998/v1/breed, cardinality:0]]]"
+        String expectedResponse = "[description:The pets all grain, dimensions:[[category:General, name:sex, longName:sex, uri:http://localhost:9998/v1/sex, cardinality:0], [category:General, name:species, longName:species, uri:http://localhost:9998/v1/species, cardinality:0], [category:General, name:breed, longName:breed, uri:http://localhost:9998/v1/breed, cardinality:0]], longName:All, metrics:[[category:General, name:rowNum, longName:rowNum, uri:http://localhost:9998/v1/rowNum], [category:General, name:limbs, longName:limbs, uri:http://localhost:9998/v1/limbs], [category:General, name:dayAvgLimbs, longName:dayAvgLimbs, uri:http://localhost:9998/v1/dayAvgLimbs]], name:all, retention:P1Y]"
         LogicalTable petsTable = petsShapesTables.find {it.getName() == "pets"}
 
         when:
         TableGrainView tableGrainView = fullViewProcessor.formatTableGrain(petsTable, "all", uriInfo)
 
         then:
-        tableGrainView.toString() == expectedResponse
+        tableGrainView.sort().toString() == expectedResponse
     }
 
     def "check all the tables full view at grain level"() {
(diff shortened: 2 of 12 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -837,7 +837,7 @@ protected ConfigurationLoader buildConfigurationLoader(
      * @return a Dimension Loader instance
      */
     protected DimensionLoader getDimensionLoader() {
-        return new KeyValueStoreDimensionLoader(getDimensionConfigurations());
+        return new TypeAwareDimensionLoader(getDimensionConfigurations());
```
cdeszaq: Need to call out this change in Changelog

fili-core/src/test/java/com/yahoo/bard/webservice/data/config/dimension/TestLookupDimensionConfig.java
```diff
@@ -13,37 +18,88 @@
  * Hold all of the Configuration information for the look up dimensions.
  */
 public class TestLookupDimensionConfig extends TestDimensionConfig implements LookupDimensionConfig {
-    private List<String> namespaces;
+    private final List<String> namespaces;
```
cdeszaq: Since this is for testing, it's OK for it to be non-final. Tests often need to reach inside things and mess with them, and `final` makes that hard to do. For non-test code, `final` is good, because most things should be immutable, and `final` helps a lot with that.

fili-core/src/test/java/com/yahoo/bard/webservice/data/config/dimension/TestLookupDimensionConfig.java
```diff
@@ -13,37 +18,88 @@
...
+            TestApiDimensionName apiName,
+            String physicalName,
+            KeyValueStore keyValueStore,
+            SearchProvider searchProvider,
+            LinkedHashSet<DimensionField> fields,
+            LinkedHashSet<DimensionField> defaultFields,
+            List<String> namespaces
+    ) {
+        super(apiName, physicalName, keyValueStore, searchProvider, fields, defaultFields);
+        this.namespaces = Collections.unmodifiableList(namespaces);
```
cdeszaq: Generally, we're OK letting test-code be modifiable. This is a good impulse, but not for test code 😉

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/TableFullViewProcessorSpec.groovy
```diff
@@ -68,14 +68,14 @@ class TableFullViewProcessorSpec extends Specification {
...
+        String expectedResponse = "[description:The pets all grain, dimensions:[[category:General, name:sex, longName:sex, uri:http://localhost:9998/v1/sex, cardinality:0], [category:General, name:species, longName:species, uri:http://localhost:9998/v1/species, cardinality:0], [category:General, name:breed, longName:breed, uri:http://localhost:9998/v1/breed, cardinality:0]], longName:All, metrics:[[category:General, name:rowNum, longName:rowNum, uri:http://localhost:9998/v1/rowNum], [category:General, name:limbs, longName:limbs, uri:http://localhost:9998/v1/limbs], [category:General, name:dayAvgLimbs, longName:dayAvgLimbs, uri:http://localhost:9998/v1/dayAvgLimbs]], name:all, retention:P1Y]"
         LogicalTable petsTable = petsShapesTables.find {it.getName() == "pets"}
 
         when:
         TableGrainView tableGrainView = fullViewProcessor.formatTableGrain(petsTable, "all", uriInfo)
 
         then:
-        tableGrainView.toString() == expectedResponse
+        tableGrainView.sort().toString() == expectedResponse
```
cdeszaq: Why are we sorting here? What changed that we need to sort now but didn't before?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/LookupDimensionFilteringDataServletSpec.groovy
```diff
@@ -0,0 +1,287 @@
...
+            addDimensionRow(BardDimensionField.makeDimensionRow(it, "species2", "species2Desc"))
+            addDimensionRow(BardDimensionField.makeDimensionRow(it, "species3", "species3Desc"))
+        }
+    }
+
+    @Override
+    boolean compareResult(String result, String expectedResult, JsonSortStrategy sortStrategy) {
+        // Using default SORT_MAPS provided by Util
+        GroovyTestUtils.compareJson(result, expectedResult)
+    }
```
cdeszaq: Isn't this just the default? Does this need to be overridden?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/LookupDimensionFilteringDataServletSpec.groovy
```diff
@@ -0,0 +1,287 @@
...
+
+    @Override
+    Class<?>[] getResourceClasses() {
+        [DataServlet.class]
+    }
+
+    @Override
+    String getTarget() {
+        return "data/pets/day/sex/breed/species"
```
cdeszaq: Since we're just focused on the filtering in this test, we shouldn't do grouping as well. As written, this test is actually doing _both_ the grouping _and_ the filtering using the namespaced dimensions. I _like_ the test, but either it's name should change to reflect that it's doing both, or we should remove the grouping.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/LookupDimensionGroupingDataServletSpec.groovy
```diff
@@ -0,0 +1,179 @@
...
+    @Override
+    Map<String, List<String>> getQueryParams() {
+        [
+                "metrics" : ["limbs"],
+                "dateTime": ["2014-05-01%2F2014-05-02"]
+        ]
+    }
+
+    @Override
+    boolean compareResult(String result, String expectedResult, JsonSortStrategy sortStrategy) {
```
cdeszaq: Does this need the override? It's doing what the default does anyways I think

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/TableFullViewProcessorSpec.groovy
garyluoex: Instead of containing `KeyValueStoreDimension`, the table now contains `LookupDimension` which changed the order of the dimensions. I can look into it if you think it is necessary and remove the sort?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/LookupDimensionFilteringDataServletSpec.groovy
garyluoex: The default for "test api response" in `BaseDataServletComponentSpec` is `SORT_BOTH`, so we have to override it to use the default sorting mechanism `SORT_MAP` instead by not passing in a sorting mechanism. (or I can add `SORT_MAP`) explicitly.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/LookupDimensionGroupingDataServletSpec.groovy
garyluoex: Same as above.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/LookupDimensionFilteringDataServletSpec.groovy
cdeszaq: Ahh, I see. We should add a comment in here that mentions that. Otherwise, simple tracing of this doesn't find the `test druid query` call that switches it to `SORT_BOTH`.

(applies to the other file as well)

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/TableFullViewProcessorSpec.groovy
cdeszaq: Nah, it's probably fine.

cdeszaq:
👍 after those 2 small comments

prakashnm:
👍

PR #35, opened 2016-09-08
Async functional tests

archolewa:
Adds the functional tests for asynchronous queries.

In order to support this, I had to slightly modify the TestDruidServlet to have a `Producer<String>` that generates a fake Druid response on demand, rather than having the Druid response directly. This was needed to support delaying the Druid response, so that we could test the path where results aren't yet ready from the backend.

The base spec for the asynchronous queries is rather complex, and I still don't think I explained it well. Suggestions on improving the documentation for that test are especially welcome.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/payloads/DefaultJobPayloadBuilder.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/payloads/DefaultJobPayloadBuilder.java
@@ -2,7 +2,13 @@
 // Licensed under the terms of the Apache license. Please see LICENSE file distributed with this work for terms.
 package com.yahoo.bard.webservice.async.jobs.payloads;
 
-import com.yahoo.bard.webservice.async.jobs.jobrows.DefaultJobField;
+import static com.yahoo.bard.webservice.async.jobs.jobrows.DefaultJobField.DATE_CREATED;
+import static com.yahoo.bard.webservice.async.jobs.jobrows.DefaultJobField.DATE_UPDATED;
+import static com.yahoo.bard.webservice.async.jobs.jobrows.DefaultJobField.JOB_TICKET;
+import static com.yahoo.bard.webservice.async.jobs.jobrows.DefaultJobField.QUERY;
+import static com.yahoo.bard.webservice.async.jobs.jobrows.DefaultJobField.STATUS;
+import static com.yahoo.bard.webservice.async.jobs.jobrows.DefaultJobField.USER_ID;
+
 import com.yahoo.bard.webservice.async.jobs.jobrows.JobRow;
 import com.yahoo.bard.webservice.web.ErrorMessageFormat;
 import com.yahoo.bard.webservice.web.JobRequestFailedException;
@@ -31,16 +37,16 @@ public class DefaultJobPayloadBuilder implements JobPayloadBuilder {
         Map<String, String> job = new LinkedHashMap<>();
 
         String ticket = jobRow.getId();
-        Map<String, String> fieldValueMap = jobRow.getRowMap();
 
-        job.put("query", fieldValueMap.get(DefaultJobField.QUERY.getName()));
+        job.put(QUERY.getName(), jobRow.get(QUERY));
         job.put("results", JobPayloadBuilder.getResultsUrl(ticket, uriInfo));
         job.put("syncResults", JobPayloadBuilder.getSyncResultsUrl(ticket, uriInfo));
         job.put("self", JobPayloadBuilder.getSelfUrl(ticket, uriInfo));
-        job.put("status", fieldValueMap.get(DefaultJobField.STATUS.getName()));
-        job.put("jobTicket", fieldValueMap.get(DefaultJobField.JOB_TICKET.getName()));
-        job.put("dateCreated", fieldValueMap.get(DefaultJobField.DATE_CREATED.getName()));
-        job.put("userId", fieldValueMap.get(DefaultJobField.USER_ID.getName()));
+        job.put(STATUS.getName(), jobRow.get(STATUS));
+        job.put(JOB_TICKET.getName(), jobRow.get(JOB_TICKET));
+        job.put(DATE_CREATED.getName(), jobRow.get(DATE_CREATED));
+        job.put(DATE_UPDATED.getName(), jobRow.get(DATE_UPDATED));
+        job.put(USER_ID.getName(), jobRow.get(USER_ID));
 
         //throw exception if any of the JobFields are missing in the map
         if (job.containsValue(null)) {
@@ -50,7 +56,7 @@ public class DefaultJobPayloadBuilder implements JobPayloadBuilder {
                     .map(Map.Entry::getKey)
                     .collect(Collectors.toSet());
 
-            String msg = ErrorMessageFormat.JOB_MAPPING_FAILED.format(ticket, fieldValueMap, missingFields);
+            String msg = ErrorMessageFormat.JOB_MAPPING_FAILED.format(ticket, jobRow.getRowMap(), missingFields);
             LOG.error(msg);
             throw new JobRequestFailedException(msg);
         }
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncDruidSendsErrorSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncDruidSendsErrorSpec.groovy
@@ -1,0 +1,151 @@
+package com.yahoo.bard.webservice.async
+
+import static com.yahoo.bard.webservice.async.jobs.jobrows.DefaultJobStatus.FAILURE
+import static com.yahoo.bard.webservice.async.jobs.jobrows.DefaultJobStatus.PENDING
+
+import com.yahoo.bard.webservice.util.GroovyTestUtils
+
+import spock.util.concurrent.AsyncConditions
+import spock.util.concurrent.PollingConditions
+
+/**
+ * Verifies that the job status is updated to error, and results links return the error message when Druid experiences
+ * an error.
+ */
+class AsyncDruidSendsErrorSpec extends AsyncFunctionalSpec {
+    /*
+        This executes the following requests, and expects the following responses. All caps are placeholders.
+             Send:
+                 http://localhost:9998/data/shapes/day?dateTime=2016-08-30%2F2016-08-31&metrics=height&asyncAfter=0
+             Receive:
+                 {
+                     "self": "http://localhost:9998/jobs/gregUUID",
+                     "userId": "greg",
+                     "ticket": "gregUUID",
+                     "results": "http://localhost:9998/jobs/gregUUID/results",
+                     "syncResults": "http://localhost:9998/jobs/gregUUID/results?asyncAfter=never",
+                     "dateCreated": "DATETIME",
+                     "dateUpdated": "SAME_DATETIME_AS_ABOVE",
+                     "status": "pending"
+                 }
+             Send:
+                 http://localhost:9998/jobs/gregUUID/results?asyncAfter=never
+             Receive:
+                {
+                    "status" : 500,
+                    "statusName" : "Internal Server Error",
+                    "reason" : "All the things have broken.",
+                    "description" : "All the things have broken.",
+                    "druidQuery" : null
+                }
+            Send:
+                 http://localhost:9998/jobs/gregUUID/results
+            Receive:
+                {
+                    "status" : 500,
+                    "statusName" : "Internal Server Error",
+                    "reason" : "All the things have broken.",
+                    "description" : "All the things have broken.",
+                    "druidQuery" : null
+                }
+
+            Send:
+                 http://localhost:9998/jobs/gregUUID
+            Receive:
+                {
+                     "self": "http://localhost:9998/jobs/gregUUID",
+                     "userId": "greg",
+                     "ticket": "gregUUID",
+                     "results": "http://localhost:9998/jobs/gregUUID/results",
+                     "syncResults": "http://localhost:9998/jobs/gregUUID/results?asyncAfter=never",
+                     "dateCreated": "DATETIME",
+                     "dateUpdated": "SAME_DATETIME_AS_ABOVE",
+                     "status": "failure"
+                }
+    */
+
+    static final String QUERY =
+            "http://localhost:9998/data/shapes/day?dateTime=2016-08-30%2F2016-08-31&metrics=height&asyncAfter=0"
+
+    static final String ERROR_MESSAGE = """{
+                        "status" : 500,
+                        "statusName" : "Internal Server Error",
+                        "reason" : "All the things have broken.",
+                        "description" : "All the things have broken.",
+                        "druidQuery" : null
+                    }"""
+
+    @Override
+    Map<String, Closure<String>> getResultsToTargetFunctions() {
+        [
+                data: { "data/shapes/day" },
+                //By querying the syncResults link first, we wait until the results are ready, thanks to the
+                //BroadcastChannel. So we have no race condition between the data request updating the job ticket, and
+                //the query for the job metadata
+                syncResults: { AsyncTestUtils.extractTargetFromField(it.data.readEntity(String), "syncResults") },
+                results: { AsyncTestUtils.extractTargetFromField(it.data.readEntity(String), "results") },
+                jobs: { AsyncTestUtils.buildTicketLookup(it.data.readEntity(String)) }
+        ]
+    }
+
+    @Override
+    Map<String, Closure<Void>> getResultAssertions() {
+        [
+                data: {
+                    assert it.status == 202
+                    AsyncTestUtils.validateJobPayload(it.readEntity(String), QUERY, PENDING.name)
+                },
+                syncResults: {
+                    // However, there was a problem in the backend, and the job failed. So when we go to get the
+                    // results, we instead get the error status (500) of the error, along with an error message
+                    // describing the problem.
+                    assert it.status == 500
+                    assert GroovyTestUtils.compareJson(it.readEntity(String), ERROR_MESSAGE)
+                },
+                results: {
+                    // This returns the same results as syncResults, since the two only differ in how long they
+                    // wait for a response.
+                    assert it.status == 500
+                    assert GroovyTestUtils.compareJson(it.readEntity(String), ERROR_MESSAGE)
+                },
+                jobs: { response ->
+                    // The job metadata is the same as the metadata returned by the data endpoint, except the status is
+                    // failure rather than pending.
+                    //There is a small window between when the results are stored and the job row's status is updated.
+                    PollingConditions pollingConditions = new PollingConditions()
+                    pollingConditions.eventually {
+                        assert response.status == 200
+                        AsyncTestUtils.validateJobPayload(response.readEntity(String), QUERY, FAILURE.name)
+                    }
+                }
+        ]
+    }
+
+    @Override
+    Map<String, Closure<Map<String, List<String>>>> getQueryParameters() {
+        [
+                data: {[
+                        metrics: ["height"],
+                        asyncAfter: ["0"],
+                        dateTime: ["2016-08-30/2016-08-31"]
+                ]},
+                syncResults: {
+                    AsyncTestUtils.extractQueryParameters(
+                            new URI(AsyncTestUtils.getJobFieldValue(it.data.readEntity(String), "syncResults"))
+                    )
+                },
+                results: {[:]},
+                jobs: {[:]},
+        ]
+    }
+
+    @Override
+    Closure<String> getFakeDruidResponse() {
+        return {"All the things have broken."}
+    }
+
+    @Override
+    int getDruidStatusCode() {
+        return 500
+    }
+}
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncFunctionalSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncFunctionalSpec.groovy
@@ -1,0 +1,215 @@
+package com.yahoo.bard.webservice.async
+
+import com.yahoo.bard.webservice.application.JerseyTestBinder
+import com.yahoo.bard.webservice.models.druid.client.impl.TestDruidWebService
+import com.yahoo.bard.webservice.table.availability.AvailabilityTestingUtils
+import com.yahoo.bard.webservice.web.endpoints.BaseDataServletComponentSpec
+import com.yahoo.bard.webservice.web.endpoints.DataServlet
+import com.yahoo.bard.webservice.web.endpoints.JobsServlet
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.datatype.jdk8.Jdk8Module
+
+import org.joda.time.Interval
+
+import spock.lang.Specification
+import spock.lang.Timeout
+
+import javax.validation.constraints.NotNull
+import javax.ws.rs.core.Response
+
+/**
+ * Serves as a base class for testing asynchronous processes.
+ *
+ * This test allows us to execute a series of queries one after the other, and use the results of previous queries to
+ * build subsequent queries.
+ * <p>
+ * As such, this spec expects a map of closures rather than a single target. The test will do something similar to, but
+ * not quite a reduction over this map. First, it will execute the first closure in the map (note that we are using the
+ * fact that Groovy's maps have a well-defined ordering) to get a target. We will make a request to the test server
+ * using that target (plus query parameters). Then we will take the result of that query (a Response), verify it, and
+ * pass it into the next closure, which will build the next target to make a request against. We will get the result,
+ * verify it, and then pass it and the result from the first request to the third closure to get the third target to
+ * execute, and so on.
+ * <p>
+ * The keys of the map are arbitrary, but must be consistent across the maps of target builders, the map of
+ * result verifiers, and the map of query parameters within a single specification. Essentially, the keys
+ * serve as names for each request.
+ * <p>
+ * Note that this test is expected to be used for testing asynchronous interactions, _not_ that Bard constructs the
+ * correct Druid query for a given request. For tests of that nature, see the {@link BaseDataServletComponentSpec}.
+ */
+@Timeout(30)
+abstract class AsyncFunctionalSpec extends Specification {
+
+    private static final ObjectMapper MAPPER = new ObjectMapper()
+            .registerModule(new Jdk8Module().configureAbsentsAsNulls(false))
+
+    JerseyTestBinder jtb
+
+    /**
+     * Returns a map of closures, each of which transforms a map of named Responses into request targets.
+     * <p>
+     * The map keys (i.e. names) are arbitrary, but used to associate targets with query parameters and result
+     * validation routines.
+     * <p>
+     * Each closure is a function Map&ltString, Response&gt -> String that takes a map of named responses, and returns a
+     * target to make a request with.
+     *
+     * @return a map of closures that derives request targets from maps of named query Responses
+     */
+    abstract Map<String, Closure<String>> getResultsToTargetFunctions()
+
+    /**
+     * Returns a map of closures, each of which is an operation Response -> Void that takes a query response, and
+     * performs whatever validation on the Response that needs to be performed.
+     * <p>
+     * The key for each validation must correspond to the key of the target of the associated request.
+     *
+     * @return a map of closures that take a query result, and performs whatever validation needs to be performed.
+     */
+    abstract Map<String, Closure<Void>> getResultAssertions()
+
+    /**
+     * Get the map of closures generating query parameters for each target.
+     * <p>
+     * Each closure is a function Map&ltString, Response&gt -> Map&ltString, List&ltString&gt&gt that takes
+     * named Responses and returns the query parameters for the associated query.
+     * <p>
+     * The key for each query parameter generator must correspond to the key of the associated target
+     *
+     * @return The map of query parameters for each target.
+     */
+    abstract Map<String, Closure<Map<String, List<String>>>> getQueryParameters()
+
+    /**
+     * Returns the classes for the Jetty resources (i.e. servlets) that need to be set up in the test environment.
+     *
+     * @return An array of classes representing Jetty resources
+     */
+    Class<?>[] getResourceClasses() {
+        return [DataServlet.class, JobsServlet.class]
+    }
+
+    /**
+     * Returns the generator for the Druid response that should be returned by the backend
+     *
+     * @return The closure that generates the Druid response that should be returned by the test framework's Druid stub
+     */
+    abstract Closure<String> getFakeDruidResponse()
+
+    /**
+     * Returns the status code that the fake Druid should send when sending results back.
+     *
+     * @return The status code that the fake Druid should use when sending results back, defaults to 200 (OK)
+     */
+    int getDruidStatusCode() {
+        return 200
+    }
+
+    def setup() {
+        // Create the test web container to test the resources
+        jtb = buildTestBinder()
+
+        populatePhysicalTableAvailability()
+    }
+
+    /**
+     * Populates the interval availability of the physical tables.
+     * <p>
+     * By default, every Physical table believes that it has complete data from January 1st 2010 to December 31 2500.
+     */
+    void populatePhysicalTableAvailability() {
+        AvailabilityTestingUtils.populatePhysicalTableCacheIntervals(jtb, new Interval("2010-01-01/2500-12-31"))
+    }
+
+    def cleanup() {
+        // Release the test web container
+        jtb.tearDown()
+    }
+
+    JerseyTestBinder buildTestBinder() {
+        new JerseyTestBinder(getResourceClasses())
+    }
+
+    void validateJson(String json) {
+        MAPPER.readTree(json)
+    }
+
+    @Timeout(10)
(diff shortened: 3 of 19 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/payloads/DefaultJobPayloadBuilder.java
```diff
@@ -31,15 +37,16 @@
...
         job.put("syncResults", JobPayloadBuilder.getSyncResultsUrl(ticket, uriInfo));
         job.put("self", JobPayloadBuilder.getSelfUrl(ticket, uriInfo));
-        job.put("status", fieldValueMap.get(DefaultJobField.STATUS.getName()));
-        job.put("jobTicket", fieldValueMap.get(DefaultJobField.JOB_TICKET.getName()));
-        job.put("dateCreated", fieldValueMap.get(DefaultJobField.DATE_CREATED.getName()));
+        job.put(STATUS.getName(), jobRow.get(STATUS));
+        job.put(JOB_TICKET.getName(), jobRow.get(JOB_TICKET));
+        job.put(DATE_CREATED.getName(), jobRow.get(DATE_CREATED));
+        job.put(DATE_UPDATED.getName(), jobRow.get(DATE_UPDATED));
+        job.put(USER_ID.getName(), jobRow.get(USER_ID));
```
venisa: I am having a 'Why didn't I do this moment'

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncDruidSendsErrorSpec.groovy
```diff
@@ -0,0 +1,90 @@
...
+    @Override
+    Closure<String> getFakeDruidResponse() {
+        return {"All the things have broken."}
+    }
+
+    @Override
+    int getDruidStatusCode() {
+        return 500
+    }
+}
```
venisa: New line needed

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncFunctionalSpec.groovy
```diff
@@ -0,0 +1,213 @@
...
+    /**
+     * Returns the status code that the fake Druid should send when sending results back.
+     *
+     * @return The status code that the fake Druid should use when sending results back, defaults to 200 (OK)
+     */
+    int getDruidStatusCode() {
+        return 200
+    }
+
+
```
venisa: Extra blank line

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncFunctionalSpec.groovy
```diff
@@ -0,0 +1,213 @@
...
+            jtb.nonUiDruidWebService.statusCode = statusCode
+            if (statusCode == 200) {
+                jtb.nonUiDruidWebService.jsonResponse = druidResponse
+            } else {
+                jtb.nonUiDruidWebService.reasonPhrase = druidResponse.call()
+            }
+        }
+    }
+
+
```
venisa: Extra blank line

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncFunctionalSpec.groovy
```diff
@@ -0,0 +1,213 @@
...
+        // Add query params to call
+        queryParameters.each { String key, List<String> values ->
+            httpCall = httpCall.queryParam(key, values.join(","))
+        }
+
+        // Make the call
+        return httpCall.request().get()
+    }
+}
```
venisa: New line needed

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncInvalidApiRequest.groovy
```diff
@@ -0,0 +1,44 @@
...
+                        dateTime: ["2016-08-30/2016-08-31"]
+                ] }
+        ]
+    }
+
+    @Override
+    Closure<String> getFakeDruidResponse() {
+        return {"I will never leave this curly prison."}
+    }
+}
```
venisa: New line needed

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncResultsNotReadySpec.groovy
```diff
@@ -0,0 +1,72 @@
...
+                    AsyncTestUtils.validateJobPayload(response.readEntity(String), QUERY, DefaultJobStatus.PENDING.name)
+                },
+                results: {Response response ->
+                    assert response.status == 200
+                    AsyncTestUtils.validateJobPayload(response.readEntity(String), QUERY, DefaultJobStatus.PENDING.name)
+                }
+        ]
+    }
+
+
```
venisa: Extra whitespace

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncResultsNotReadySpec.groovy
```diff
@@ -0,0 +1,72 @@
...
+    @Override
+    Closure<String> getFakeDruidResponse() {
+        return {
+            //30 seconds, the same length as the test timeout. In this test, we don't want Druid to respond at all.
+            Thread.sleep(30000)
+            return "[]"
+        }
+    }
+}
```
venisa: new line needed

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncResultsReadySpec.groovy
```diff
@@ -0,0 +1,105 @@
...
+import com.yahoo.bard.webservice.async.jobs.jobrows.DefaultJobStatus
+import com.yahoo.bard.webservice.util.GroovyTestUtils
+
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+/**
+ * Verifies that when the results of an asynchronous request are ready, that the job status is updated to ready and
+ * the results are accessible through both the {@code results} and {@code syncResults} links.
+ */
```
venisa: So once the results are ready, we set the result status to 'success' right? I thought the only 3 status messages we had were 'success, pending and failure'

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncResultsReadySpec.groovy
```diff
@@ -0,0 +1,105 @@
...
+class AsyncResultsReadySpec extends AsyncFunctionalSpec {
+    static final Logger LOG = LoggerFactory.getLogger(AsyncResultsReadySpec.class)
+
+    static final String QUERY =
+            "http://localhost:9998/data/shapes/day?dateTime=2016-08-30%2F2016-08-31&metrics=height&asyncAfter=0"
+
+
```
venisa: Extra whitespace

fili-core/src/main/java/com/yahoo/bard/webservice/async/jobs/payloads/DefaultJobPayloadBuilder.java
michael-mclawhorn: There's almost always a better way to do something.  That's why I love this review process, I get smarter by watching other people's work.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncResultsReadySpec.groovy
```diff
@@ -0,0 +1,105 @@
...
+        """{
+              "rows" : [
+                  {
+                    "dateTime" : "2016-08-30 00:00:00.000",
+                    "height" : 100
+                  }
+              ]
+        }"""
+    }
+}
```
venisa: new line needed

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncResultsReadySpec.groovy
```diff
@@ -0,0 +1,105 @@
...
+                results: {
+                    assert it.status == 200
+                    assert GroovyTestUtils.compareJson(it.readEntity(String), getExpectedApiResponse())
+                },
+                jobs: { response ->
+                    assert response.status == 200
+                    AsyncTestUtils.validateJobPayload(
+                            response.readEntity(String),
+                            QUERY,
+                            DefaultJobStatus.SUCCESS.name
```
venisa: So we are using the status SUCCESS... we should juts update the Javadoc above

venisa:
I just had a bunch of silly comments about new lines and whitespace. We should probably update the Javadoc for `AsyncResultsReadySpec` to say result 'SUCCESS' instead of result 'READY'

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/BaseDataServletComponentSpec.groovy
```diff
@@ -139,15 +143,28 @@ abstract class BaseDataServletComponentSpec extends Specification {
         GroovyTestUtils.compareJson(result, expectedResult)
     }
 
+
```
dayamr: extra line

dayamr:
Apart from new line and extra line comments, it looks good to me. 👍

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncResultsReadySpec.groovy
archolewa: Yes, that should say "success" not "ready." Good catch.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncDruidSendsErrorSpec.groovy
```diff
@@ -0,0 +1,90 @@
...
+                        "status" : 500,
+                        "statusName" : "Internal Server Error",
+                        "reason" : "All the things have broken.",
+                        "description" : "All the things have broken.",
+                        "druidQuery" : null
+                    }"""
+
+    @Override
+    Map<String, Closure<String>> getResultsToTargetFunctions() {
+        return [
```
cdeszaq: `return` not needed

/NotABlocker

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncDruidSendsErrorSpec.groovy
```diff
@@ -0,0 +1,90 @@
...
+                //the query for the job metadata
+                syncResults: { AsyncTestUtils.extractTargetFromField(it.data.readEntity(String), "syncResults") },
+                results: { AsyncTestUtils.extractTargetFromField(it.data.readEntity(String), "results") },
+                jobs: { AsyncTestUtils.buildTicketLookup(it.data.readEntity(String)) }
+        ]
+    }
+
+    @Override
+    Map<String, Closure<Void>> getResultAssertions() {
+        return [
```
cdeszaq: `return` not needed

/NotABlocker

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncDruidSendsErrorSpec.groovy
```diff
@@ -0,0 +1,90 @@
...
+                jobs: { response ->
+                    assert response.status == 200
+                    AsyncTestUtils.validateJobPayload(response.readEntity(String), QUERY, FAILURE.name)
+                }
+        ]
+    }
+
+    @Override
+    Map<String, Closure<Map<String, List<String>>>> getQueryParameters() {
+        return [
```
cdeszaq: `return` not needed

/NotABlocker

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncFunctionalSpec.groovy
```diff
@@ -0,0 +1,211 @@
...
+
+import org.joda.time.Interval
+
+import spock.lang.Specification
+import spock.lang.Timeout
+
+import javax.validation.constraints.NotNull
+import javax.ws.rs.core.Response
+
+
```
cdeszaq: Extra space not needed

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncDruidSendsErrorSpec.groovy
```diff
@@ -0,0 +1,90 @@
...
+                data: {[
+                        metrics: ["height"],
+                        asyncAfter: ["0"],
+                        dateTime: ["2016-08-30/2016-08-31"]
+                ]},
+                jobs: {[:]},
+                results: {[:]},
+                syncResults: {
```
cdeszaq: The ordering of this map should (for clarity at least) match the ordering of the other closure maps.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncFunctionalSpec.groovy
```diff
@@ -0,0 +1,211 @@
...
+ * result verifiers, and the map of query parameters within a single specification. Essentially, the keys
+ * serve as names for each request.
+ * <p>
+ * Note that this test is expected to be used for testing asynchronous interactions, _not_ that Bard constructs the
+ * correct Druid query for a given request. For tests of that nature, see the {@link BaseDataServletComponentSpec}.
+ *
+ * @param <T>  The type to use for keys into the maps containing test-specific resources
+ */
+@Timeout(30)
+abstract class AsyncFunctionalSpec extends Specification {
```
cdeszaq: You've not _actually_ made this class generic, though the JavaDoc seems to indicate as such.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncFunctionalSpec.groovy
```diff
@@ -0,0 +1,211 @@
...
+
+    private static final ObjectMapper MAPPER = new ObjectMapper()
+            .registerModule(new Jdk8Module().configureAbsentsAsNulls(false))
+
+    JerseyTestBinder jtb
+
+    /**
+     * Returns a map of closures, each of which transforms a map of named Responses into request targets.
```
cdeszaq: Giving the expected method signature of these Closures (since the Closure type interface doesn't indicate it) would be good, I think. Based on this JavaDoc, it looks like it might be something like `Map<String, Response> -> requestTarget`, but it isn't clear.

cdeszaq: Note that this "give us the signature" sentiment applies to the rest of these Closure maps.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncFunctionalSpec.groovy
```diff
@@ -0,0 +1,211 @@
...
+    /**
+     * Get the map of closures generating query parameters for each target.
+     * <p>
+     * Each closure takes as argument the map of all previous results (like {@code getResultsToTargetFunctions})
+     * and returns the query parameters for the associated query.
+     * <p>
+     * The key for each query parameter generating closure must correspond to the key of the associated target
+     *
+     * @return The map of query parameters for each target.
+     *
```
cdeszaq: Blank comment line not needed

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncFunctionalSpec.groovy
```diff
@@ -0,0 +1,211 @@
...
+     *  applicable.
+     *
+     *  @param  druidResponse The fake response to be injected.
+     */
+    def injectDruidResponse(Closure<String> druidResponse) {
+        if (jtb.nonUiDruidWebService instanceof TestDruidWebService) {
+            int statusCode = getDruidStatusCode()
+            jtb.nonUiDruidWebService.statusCode = statusCode
+            if (statusCode == 200) {
+                jtb.nonUiDruidWebService.jsonResponse = druidResponse
```
cdeszaq: Should this be `druidResponse.call()`?

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncFunctionalSpec.groovy
```diff
@@ -0,0 +1,211 @@
...
+        Map<String, Closure<Map<String, List<String>>>> queryParameters = getQueryParameters()
+
+        expect: "All the asynchronous interactions behave appropriately"
+        getResultsToTargetFunctions().each {interactionName, resultsToTarget ->
+            Response response = makeAbstractRequest(
+                    resultsToTarget(previousResponses),
+                    queryParameters.get(interactionName)(previousResponses)
+            )
+            response.bufferEntity()
+            resultValidations[interactionName](response)
```
cdeszaq: Some comments in here around what's happening would be very helpful, especially for those less familiar with what calling functions retrieved via map access or method calls looks like (ie. `myMap[foo](bar)` or `myMap.get(foo)(bar)`)

cdeszaq: Also, we should use 1 mapp access style: either the index accessor as used on this line, or the getter as used 3 lines up. I like index access, personally, but we should use just one.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncFunctionalSpec.groovy
```diff
@@ -0,0 +1,211 @@
...
+            previousResponses.put(interactionName, response)
```
cdeszaq: Could also use index accessor to "set" it,  if you opt for that style of access

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncFunctionalSpec.groovy
archolewa: Ah, yes. Originally it was generic. Then I decided that allowing the user to provide some arbitrary type for keys just made  a (already very abstract) class more abstract than it needed to be.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncDruidSendsErrorSpec.groovy
```diff
@@ -0,0 +1,90 @@
...
+                syncResults: { AsyncTestUtils.extractTargetFromField(it.data.readEntity(String), "syncResults") },
+                results: { AsyncTestUtils.extractTargetFromField(it.data.readEntity(String), "results") },
+                jobs: { AsyncTestUtils.buildTicketLookup(it.data.readEntity(String)) }
+        ]
+    }
+
+    @Override
+    Map<String, Closure<Void>> getResultAssertions() {
+        return [
+                data: {
```
cdeszaq: I feel like having comments in here indicating the scenario and what the assertion is expecting would help aid in understanding both what behavior is being tested for, as well as what the queries are for which the behavior is expected.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncFunctionalSpec.groovy
archolewa: No. The `jsonResponse` in the DruidWebService is now a `Function`, not a `String` in order to support delaying the response.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncInvalidApiRequest.groovy
```diff
@@ -0,0 +1,44 @@
...
+                data: { [
+                        asyncAfter: ["0"],
+                        dateTime: ["2016-08-30/2016-08-31"]
+                ] }
+        ]
+    }
+
+    @Override
+    Closure<String> getFakeDruidResponse() {
+        return {"I will never leave this curly prison."}
```
cdeszaq: Rather than having to override a method when we don't care about the response, does it make sense to have the method be non-abstract with a default Druid response of an empty object, or list or something?

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncResultsNotReadySpec.groovy
```diff
@@ -0,0 +1,71 @@
...
+
+import com.yahoo.bard.webservice.async.jobs.jobrows.DefaultJobField
+import com.yahoo.bard.webservice.async.jobs.jobrows.DefaultJobStatus
+
+import spock.lang.Timeout
+
+import javax.ws.rs.core.Response
+
+/**
+ * Verifies that if the result of an asynchronous query is not ready yet, then the user get back the exact same payload
```
cdeszaq: s/get/gets

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncResultsNotReadySpec.groovy
```diff
@@ -0,0 +1,71 @@
...
+            "http://localhost:9998/data/shapes/day/color?dateTime=2016-08-30%2F2016-08-31&metrics=height&asyncAfter=0"
+
+    @Override
+    Map<String, Closure<String>> getResultsToTargetFunctions() {
+        return [
+                data: {"data/shapes/day/color"},
+                jobs: {AsyncTestUtils.buildTicketLookup(it.data.readEntity(String))},
+                results: {payloads ->
+                    String dataPayload = payloads.data.readEntity(String)
+                    "jobs/${AsyncTestUtils.getJobFieldValue(dataPayload, DefaultJobField.JOB_TICKET.name)}/results"
```
cdeszaq: A comment here around what's being looked at would be good.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncResultsNotReadySpec.groovy
```diff
@@ -0,0 +1,71 @@
...
+                    assert response.status == 202
+                    AsyncTestUtils.validateJobPayload(response.readEntity(String), QUERY, DefaultJobStatus.PENDING.name)
+                },
+                jobs: {Response response ->
+                    assert response.status == 200
+                    AsyncTestUtils.validateJobPayload(response.readEntity(String), QUERY, DefaultJobStatus.PENDING.name)
+                },
+                results: {Response response ->
+                    assert response.status == 200
+                    AsyncTestUtils.validateJobPayload(response.readEntity(String), QUERY, DefaultJobStatus.PENDING.name)
```
cdeszaq: Note that if you _truely_ want to assert that the Response is the same as the `/data` response, you should save the `Response` object you got in the `data` assertion, and then check that the other assertions have the same thing (or at least that their thing has the same content (entity) and relevant headers.

As it stands, these assertions are checking that each response matches it's own criteria, not that all 3 are the same.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncResultsNotReadySpec.groovy
```diff
@@ -0,0 +1,71 @@
...
+                jobs: {[:]},
+                results: { [ asyncAfter: ["0"] ] }
+        ]
+    }
+
+    @Override
+    Closure<String> getFakeDruidResponse() {
+        return {
+            //30 seconds, the same length as the test timeout. In this test, we don't want Druid to respond at all.
+            Thread.sleep(30000)
```
cdeszaq: A different (and _possibly_ better) way to do this would be to use a Latch (like a countdown latch), which can be shared between the thread the `getFakeDruidResponse` runs on and the thread that the request validation code runs on. Then, when the final assertion (or something) runs, it can "un-latch" and allow the Druid response to flow. That prevents the need to block that thread for 30 seconds, letting it be blocked only as long as needed.

This is the general mechanism for dealing with ordering across threads that is timing / load-safe, and doesn't waste any more time than needed, since it's not time-based.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncResultsReadySpec.groovy
```diff
@@ -0,0 +1,99 @@
...
+
+    @Override
+    Map<String, Closure<Void>> getResultAssertions() {
+        return [
+                data: {
+                    assert it.status == 202
+                    AsyncTestUtils.validateJobPayload(it.readEntity(String), QUERY, DefaultJobStatus.PENDING.name)
+                },
+                syncResults: {
+                    assert it.status == 200
```
cdeszaq: Comments in here would be good (see earlier comment)

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncTestUtils.groovy
```diff
@@ -0,0 +1,100 @@
...
+ * Contains a collection of functions to aid in testing asynchronous
+ * queries. Particularly around parsing asynchronous payloads and extracting
+ * from them the information needed by subsequent requests.
+ */
+class AsyncTestUtils {
+
+   static final JsonSlurper JSON_PARSER = new JsonSlurper()
+
+   /**
+    * Given a String representation of the job metadata sent to the user, and a job payload field, returns the value
```
cdeszaq: A JSON representation, specifically, right? Not just String?

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncTestUtils.groovy
```diff
@@ -0,0 +1,100 @@
...
+    * @param field  The field whose data is of interest
+    *
+    * @return The value of the specified field in the specified job payload
+    */
+   static String getJobFieldValue(String response, String field) {
+      JSON_PARSER.parseText(response)[field]
+   }
+
+   /**
+    * Given a String representation of the job information sent to the user, returns the URI path for a point lookup of
```
cdeszaq: s/String/JSON

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/ErrorDataServletSpec.groovy
```diff
@@ -138,7 +138,7 @@ class ErrorDataServletSpec extends Specification {
         setup:
         DruidServiceConfig oldConfig = nonUiTestWebService.serviceConfig
         nonUiTestWebService.serviceConfig = new DruidServiceConfig("Non-Ui Broker", oldConfig.url, 123, oldConfig.priority)
-        nonUiTestWebService.jsonResponse = standardGoodDruidResponse
+        nonUiTestWebService.jsonResponse ={standardGoodDruidResponse}
```
cdeszaq: whitespace needed

fili-core/src/test/java/com/yahoo/bard/webservice/models/druid/client/impl/TestDruidWebService.java
```diff
@@ -39,7 +40,7 @@
     );
     private static ObjectWriter writer = mapper.writer();
 
-    public String jsonResponse = "[]";
+    public Producer<String> jsonResponse = () -> "[]";
```
cdeszaq: Definitely need Changelog entry for this, since it's a breaking change for anyone trying to use this.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncResultsReadySpec.groovy
```diff
@@ -0,0 +1,99 @@
...
+                        asyncAfter: ["0"],
+                        dateTime: ["2016-08-30/2016-08-31"]
+                ]},
+                syncResults: {
+                    new URI(AsyncTestUtils.getJobFieldValue(it.data.readEntity(String), "syncResults"))
+                            .query.split("&")
+                            .collectEntries { parameter ->
+                                    def (String key, String value) = parameter.tokenize("=")
+                                    return [(key): value.tokenize(",")]
+                            }
```
cdeszaq: Does it make sense to extract this bit as a helper method as well? I think I saw it at least 2 places in this PR and it's not the most easy-to-understand code under the sun.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncFunctionalSpec.groovy
cdeszaq: Ok, makes sense. Need changelog entry for this.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncInvalidApiRequest.groovy
archolewa: Personally, I think that depends on how often we expect to have tests that ignore the results from Druid. I don't like the idea of an exceptional case being the default. Every test that I've written except for this one do interact with Druid, so it's exceptional in this case, at least.

Now that I'm done using fancy speak that makes me sound like a craftsman, the real reason is that I'm lazy, and one nice thing about making it abstract is that with a few keystrokes in IntelliJ I can automatically generate method signatures for all the abstract methods. If I make this non-abstract, I can generate all the methods but one that I need to override, and then need to generate the `getFakeDruidResponse` method signature as a separate step. 

Not a big deal, but I find it mildly annoying when overriding the `BaseDataServletComponentSpec`, and I figured that faking Druid responses would be common enough to justify making it abstract to save myself a few keystrokes.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncFunctionalSpec.groovy
cdeszaq: Handled

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncFunctionalSpec.groovy
cdeszaq: Handled

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncFunctionalSpec.groovy
cdeszaq: Handled

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncResultsNotReadySpec.groovy
archolewa: Yeah, originally I was going to do that, but it ended up being easier just to use the same validation method with the same arguments. Since the validation method is side-effect free, I _am_ verifying that they're the same. ;)

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncInvalidApiRequest.groovy
cdeszaq: Fair enough, I'm OK leaving it as abstract

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncResultsNotReadySpec.groovy
cdeszaq: In that case, toss a comment in there that indicates the caveat of not _really_ checking that they are the same, but approximating it by asserting the same expected values.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncResultsNotReadySpec.groovy
```diff
@@ -0,0 +1,128 @@
...
+                 http://localhost:9998/jobs/gregUUID/results?asyncAfter=0
+            Receive:
+                SAME AS ABOVE
+            Send:
+                 http://localhost:9998/jobs/gregUUID
+            Receive:
+                SAME AS ABOVE
+    */
+
+    final CountDownLatch validationFinished = new CountDownLatch(3)
```
archolewa: https://www.youtube.com/watch?v=9jK-NcRmVcw

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncResultsNotReadySpec.groovy
```diff
@@ -63,8 +119,9 @@ class AsyncResultsNotReadySpec extends AsyncFunctionalSpec {
     @Override
     Closure<String> getFakeDruidResponse() {
         return {
-            //30 seconds, the same length as the test timeout. In this test, we don't want Druid to respond at all.
-            Thread.sleep(30000)
+            //We don't want the backend to return until all the validation has been performed, to simulate the backend
+            //taking a while to respond.
+            validationFinished.await()
```
cdeszaq: Beautiful.

cdeszaq:
Also, not sure why, but GitHub's PRs seem to be rather borked today. Everything I've commented on has either been addressed (though it doesn't show it), or was mitigated via comments.

archolewa:
Yeah, I've noticed that too. It looks like there have been some changes to the UI since last I reviewed, so it's likely they've got some bugs because of their changes.

PR #38, opened 2016-09-13
Possibly remove duplicate artifact push

michael-mclawhorn:
Artifacts are trying to be generated and pushed twice:

javadoc jar
sources jar

Removing extra callouts.


cdeszaq:
👍

PR #39, opened 2016-09-15
Fixes some flaky asynchronous tests.

archolewa:
--A test version of the AsynchronousWorkflowsBuilder is introduced. This
version extends the DefaultAsynchronousWorkflowsBuilder, and builds the
exact same workflow. However, it also provides a mechanism for outside
classes to add additional Observers to the workflows, allowing them to
do such things as add countdown latches to have thread-safe tests.


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ Current
 -------
 
 ### Added:
+-  [A test implementation of the `AsynchronousWorkflowsBuilder`, `TestAsynchronousWorkflowsBuilder`](http://github.com/yahoo/fili/pull/36)
```
cdeszaq: PR 39, yes? (also other change log entry)

dayamr:
Thank you for explaining Andrew. it looks good to me. 👍

fili-core/src/test/java/com/yahoo/bard/webservice/async/workflows/TestAsynchronousWorkflowsBuilder.java
```diff
@@ -0,0 +1,101 @@
...
+                payloadEmitter,
+                jobMetadata,
+                jobMetadataSerializer
+        );
+
+        Map<Workflow, Observable> workflowMap = new HashMap<>();
+        workflowMap.put(Workflow.SYNCHRONOUS, workflows.getSynchronousPayload());
+        workflowMap.put(Workflow.ASYNCHRONOUS, workflows.getAsynchronousPayload());
+        workflowMap.put(Workflow.PRERESPONSE_READY, workflows.getPreResponseReadyNotifications());
+        workflowMap.put(Workflow.JOB_MARKED_COMPLETE, workflows.getJobMarkedCompleteNotifications());
```
cdeszaq: it _might_ be nice to decouple this from the specific workflow steps a bit. As it stands, if a workflow _adds_ steps, there's no way to attach a subscriber to it. I'm not sure we can do it right now, since this is Java code, and to do it cleanly we'd want to use groovy's ability to call methods as strings, but it would be a nice-to-have capability. Likely more for "do later" rather than any time soon.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncDruidSendsErrorSpec.groovy
```diff
@@ -75,6 +77,33 @@ class AsyncDruidSendsErrorSpec extends AsyncFunctionalSpec {
                         "druidQuery" : null
                     }"""
 
+    final CountDownLatch jobMetadataReady = new CountDownLatch(1)
+
+    def setup() {
+        TestAsynchronousWorkflowsBuilder.addSubscriber(
+                TestAsynchronousWorkflowsBuilder.Workflow.JOB_MARKED_COMPLETE,
+                new Observer() {
```
cdeszaq: This is ugly. We can do better (after all, this _is_ Groovy):

```groovy
TestAsynchronousWorkflowsBuilder.addSubscriber(TestAsynchronousWorkflowsBuilder.Workflow.JOB_MARKED_COMPLETE) 
        { jobMetadataReady.countDown() }
        { throw it }
```
Or we can pass them as parameters if you don't like the implicit closure-passing style

```groovy
TestAsynchronousWorkflowsBuilder.addSubscriber(
        TestAsynchronousWorkflowsBuilder.Workflow.JOB_MARKED_COMPLETE,
        {jobMetadataReady.countDown()},
        {throw it}
)
```

To achieve this, we need to add some more helpers to the `TestAsynchronousWorkflowsBuilder`:

```java
/**
 * Adds the specified subscriber to the specified workflow.
 *
 * @param workflow  The workflow to add the countdown latch to
 * @param workflowSubscriber  The subscriber that should be added to the specified workflow
 */
public static void addSubscriber(Workflow workflow, Observer workflowSubscriber) {
    SUBSCRIBERS.put(workflow, workflowSubscriber);
}

/**
 * Adds the subscriber (specified in individual method components) to the specified workflow.
 *
 * @param workflow  The workflow to add the countdown latch to
 * @param onNext  onNext method for the observer
 * @param onCompleted  onCompleted method for the observer
 * @param onError  onError method for the observer
 */
public static void addSubscriber(
        Workflow workflow,
        Consumer<Object> onNext,
        Runnable onCompleted,
        Consumer<Throwable> onError
) {
    Observer workflowSubscriber = new Observer() {
        @Override
        public void onNext(Object next) {
            onNext.accept(next);
        }
        @Override
        public void onCompleted() {
            onCompleted.run();
        }
        @Override
        public void onError(Throwable error) {
            onError.accept(error);
        }
    };
    SUBSCRIBERS.put(workflow, workflowSubscriber);
}

/**
 * Subscribes the partially-specified observer to the specified workflow.
 * <p>
 * Uses a no-op onCompleted method.
 *
 * @param workflow  The workflow to add the countdown latch to
 * @param onNext  onNext method for the observer
 * @param onError  onError method for the observer
 */
public static void addSubscriber(Workflow workflow, Consumer<Object> onNext, Consumer<Throwable> onError) {
    addSubscriber(workflow, onNext::accept, () -> { }, ignored -> { });
}

/**
 * Subscribes the partially-specified observer to the specified workflow.
 * <p>
 * Uses no-op onCompleted and onError methods.
 *
 * @param workflow  The workflow to add the countdown latch to
 * @param onNext  onNext method for the observer
 */
public static void addSubscriber(Workflow workflow, Consumer<Object> onNext) {
    addSubscriber(workflow, onNext::accept, ignored -> { });
}
```

cdeszaq: Note that this suggestion follows into the other spec that gets updated in this PR also.

CHANGELOG.md
archolewa: Heheh. Looks like I copied-pasted-forgot-to-modify.

fili-core/src/test/java/com/yahoo/bard/webservice/async/workflows/TestAsynchronousWorkflowsBuilder.java
archolewa: Yeah, I wasn't a particularly big fan of this, either. In Java, the best approach to handle this sort of thing may be to make the `AsynchronousWorkflows` object less rigid. Give it a map that maps interfaces to `Observables`, with an enum as default implementation of the interface that enumerates the current workflows, rather than having explicit fields and getters. Then, this class iterate over the map.

I consider that to be outside the scope of this PR, though.

cdeszaq: Yeah, that is a good way of doing it in Java land. And yeah, it can happen later. No need to slow down this PR.

fili-core/src/main/java/com/yahoo/bard/webservice/web/JobsApiRequest.java
```diff
@@ -95,34 +81,6 @@ public JobsApiRequest(
...
-                .timeout(asyncAfter, TimeUnit.MILLISECONDS, Observable.empty())
-                .flatMap(preResponseStore::get)
-                .take(1)
-                .replay(1);
-
-        //connect to the broadCastChannelPreResponseObservable so that it begins collecting values
-        broadCastChannelPreResponseObservable.connect();
-
-        return broadCastChannelPreResponseObservable;
-    }
```
archolewa: This has been moved into the method [JobsServlet::getResults](https://github.com/yahoo/fili/pull/39/files#diff-97b8f8f634a71f2ff451f07dbecd29dcR363)

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/JobsApiRequestSpec.groovy
```diff
@@ -121,63 +117,6 @@ class JobsApiRequestSpec extends Specification {
...
-
-        when:
-        Observable<PreResponse> preResponseObservable = defaultJobsApiRequest.handleBroadcastChannelNotification("ticket1")
-
-        broadcastChannel.publish("ticket1")
-        preResponseObservable.subscribe(testSubscriber)
-
-        then:
-        testSubscriber.assertReceivedOnNext([ticket1PreResponse])
-    }
```
archolewa: This test is subsumed by tests in the `JobsServletReactiveChainforResultsEndpointSpec` that verify the results come back synchronously if they are ready within the `asyncAfter` timeout.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/JobsApiRequestSpec.groovy
```diff
@@ -121,63 +117,6 @@ class JobsApiRequestSpec extends Specification {
...
-        Observable<PreResponse> preResponseObservable = defaultJobsApiRequest.handleBroadcastChannelNotification("ticket1")
-        broadcastChannel.publish("ticket_before_connecting_is_dropped")
-        broadcastChannel.publish("ticket1")
-        broadcastChannel.publish("ticket2")
-        broadcastChannel.publish("ticket1")
-        preResponseObservable.subscribe(testSubscriber)
-
-        then:
-        testSubscriber.assertReceivedOnNext([ticket1PreResponse])
-    }
```
archolewa: This is subsumed by a [new test in the JobsServletReactiveChainforResultsEndpointSpec](https://github.com/yahoo/fili/pull/39/files#diff-fd9bfd667dcfa0b4c6d56f006840de39R208)

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/JobsApiRequestSpec.groovy
```diff
@@ -121,63 +117,6 @@ class JobsApiRequestSpec extends Specification {
...
-        when: "We start listening to the broadcastChannel and timeout occurs"
-        Observable<PreResponse> preResponseObservable = apiRequest.handleBroadcastChannelNotification("ticket1")
-
-        and: "we subscribe to the preResponseObservable"
-        preResponseObservable.subscribe(testSubscriber)
-
-        then: "preResponseObservable is empty (the chain is complete, and no values were sent)"
-        ReactiveTestUtils.assertCompletedWithoutError(testSubscriber)
-        testSubscriber.assertNoValues()
-    }
```
archolewa: This is subsumed by tests in the `JobsServletReactiveChainforResultsEndpointSpec` that verify that the `getResults` method returns an empty Observable if the results are not ready within the asynchronous timeout.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletReactiveChainforResultsEndpointSpec.groovy
```diff
@@ -152,73 +181,119 @@ class JobsServletReactiveChainforResultsEndpointSpec extends Specification {
...
+        and: "We receive the notification after async timeout"
         broadcastChannel.publish("ticket4")
 
-        then: "then we go to the PreResponseStore twice to get the ticket"
-        2 * mockPreResponseStore.get(_) >> Observable.just(Mock(PreResponse))
+        then: "then we go to the PreResponseStore exactly once for the initial check to see if the data is already there"
+        1 * mockPreResponseStore.get(_) >> Observable.empty()
     }
 
-    def "If the PreResponse is not available in the PreResponseStore initially and the notification from broadcastChannel is received within the async timeout, we go to the PreResponsestore twice"() {
```
archolewa: [ditto](https://github.com/yahoo/fili/pull/39/files#diff-fd9bfd667dcfa0b4c6d56f006840de39L190)

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletReactiveChainforResultsEndpointSpec.groovy
```diff
@@ -152,73 +181,119 @@ class JobsServletReactiveChainforResultsEndpointSpec extends Specification {
...
+
+        then: "The subscriber only gets the results for ticket1"
+        resultsSubscriber.assertValue(results)
+
+        cleanup:
+        preResponseStore.clearGetLatches()
+        preResponseStore.clearSaveLatches()
     }
 
-    def "If the PreResponse is available in the PreResponseStore and the notification from broadcastChannel is received within the async timeout, we go to the PreResponsestore twice"() {
```
archolewa: Whether or not the results are available actually has zero impact on this test. This test is testing how notifications from the broadcastChannel affects processing.

fili-core/src/main/java/com/yahoo/bard/webservice/web/JobsApiRequest.java
```diff
@@ -34,29 +30,23 @@
...
     private final ApiJobStore apiJobStore;
-    private final PreResponseStore preResponseStore;
-    private final BroadcastChannel<String> broadcastChannel;
     private final String filters;
 
     /**
      * Parses the API request URL and generates the Api Request object.
-     *
-     * @param format  response data format JSON or CSV. Default is JSON.
+     *  @param format  response data format JSON or CSV. Default is JSON.
```
RVRSKumar: Seems like an extra `whitespace` got added here.

fili-core/src/main/java/com/yahoo/bard/webservice/web/JobsApiRequest.java
```diff
@@ -34,29 +30,23 @@
...
      * @param asyncAfter  How long the user is willing to wait for a synchronous request in milliseconds
      * @param perPage  number of rows to display per page of results. If present in the original request,
-     * must be a positive integer. If not present, must be the empty string.
+ * must be a positive integer. If not present, must be the empty string.
```
RVRSKumar: Seems like an indentation issue with the editor. Can we fix this as well?

fili-core/src/main/java/com/yahoo/bard/webservice/web/JobsApiRequest.java
archolewa: Addressed

fili-core/src/main/java/com/yahoo/bard/webservice/web/JobsApiRequest.java
archolewa: Addressed

fili-core/src/main/java/com/yahoo/bard/webservice/async/preresponses/stores/HashPreResponseStore.java
```diff
@@ -8,32 +8,74 @@
...
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.concurrent.CountDownLatch;
 
 /**
  * An in-memory implementation of PreResponseStore mainly for testing purposes. It only provides functionality to save
  * an entry to store and get an entry from the store. It does not have delete functionality nor does it take care of
  * cleaning stale data.
+ * <p>
+ * Since the HashPreResponseStore is intended primarily for testing, it also includes two maps of tickets to
```
cdeszaq: Some thoughts here:

1.) If this is primarily for testing, we should move this class to the `test` source root
2.) If this class _is_ useful for non-testing, then we should leave it here and remove the latches (assuming they are not useful for non-testing workloads)
3.) If we want _both_ (ie. keep this class for real use, and also give it the latch counting behavior for tests to use, then we should have a `LatchedHashPreResponseStore` in the `test` source root that extends this one, adding the latching behavior.

Either way, things that are only for testing don't belong in code that is in the `src` root. That's what the `test` root is for.

fili-core/src/main/java/com/yahoo/bard/webservice/web/JobsApiRequest.java
cdeszaq: Needs note in Changelog

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
```diff
@@ -359,20 +355,50 @@ protected void handlePreResponse(
...
+     *
+     * @return An Observable wrapping a PreResponse or an empty Observable in case a timeout occurs.
+     */
+    protected Observable<PreResponse> getResults(@NotNull String ticket, long asyncAfter) {
+        /*
+         * BroadCastChannel is a hot observable i.e. it emits notification irrespective of whether it has any
+         * subscribers. We use the replay operator so that the preResponseObservable upon connection, will begin
+         * collecting values.
+         * Once a new observer subscribes to the observable, it will have all the collected values replayed to it.
+         */
```
cdeszaq: Comment should get moved inside the `else` where it pertains.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
```diff
@@ -359,20 +355,50 @@ protected void handlePreResponse(
...
+        if (asyncAfter == JobsApiRequest.ASYNCHRONOUS_ASYNC_AFTER_VALUE) {
+            return Observable.empty();
```
cdeszaq: Comment on why this is good to do would be nice. (It's not immediately clear _why_ we're doing this)

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
```diff
@@ -359,20 +355,50 @@ protected void handlePreResponse(
...
+        } else {
+            ConnectableObservable<String> broadcastChannelNotifications = broadcastChannel.getNotifications()
+                    .filter(ticket::equals)
+                    .take(1)
+                    .replay(1);
+            broadcastChannelNotifications.connect();
+            return preResponseStore.get(ticket).switchIfEmpty(
+                    applyTimeoutIfNeeded(broadcastChannelNotifications, asyncAfter).flatMap(preResponseStore::get)
```
cdeszaq: Are you sure this is doing the correct thing? To me, this looks like it will wait until `preResponseStore.get()` completes, and then activate the "timeout" (if needed).

I'm _guessing_ this is now doing the right thing, but it's not immediately obvious, so more comments are likely needed around what's happening, and why it's correct. (talking about the ordering of things might not hurt either)

cdeszaq: Note that before, the timeout step was injected _before_ the `preResponseStore::get` call...

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
```diff
@@ -359,20 +355,50 @@ protected void handlePreResponse(
...
-    protected Observable<PreResponse> getPreResponseObservable(String ticket, JobsApiRequest apiRequest) {
-        //Subscribe to the BroadcastChannel so we do not miss notifications about the ticket being stored in the
-        //PreResponseStore.
-        Observable<PreResponse> broadCastChannelPreResponseObservable =
-                apiRequest.handleBroadcastChannelNotification(ticket);
-
-         //Check the PreResponseStore to see if the PreResponse associated with the given ticket has been stored in
-        //the PreResponseStore. If not, wait for a PreResponse for the amount of time specified in async.
-        return preResponseStore.get(ticket).switchIfEmpty(broadCastChannelPreResponseObservable);
+    private <T> Observable<T> applyTimeoutIfNeeded(Observable<T> observable, long asyncAfter) {
```
cdeszaq: It feels like what we're _actually_ making is a new "operator" that _extends_ the timeout and adds "always" and "never" semantics in the same parameter. With that in mind, does it makes sense for this method to take care of all 3 options? I see it looking something like this, perhaps:

```java
private <T> Observable<T> applyTimeoutIfNeeded(Observable<T> primary, long asyncAfter, Observable<T> alternate) {
    return asyncAfter == JobsApiRequest.ASYNCHRONOUS_ASYNC_AFTER_VALUE ?
        // Always alternate
        alternate :
        asyncAfter == JobsApiRequest.SYNCHRONOUS_ASYNC_AFTER_VALUE ?
                // Always primary
                primary :
                // Timeout-based switch
                primary.timeout(asyncAfter, TimeUnit.MILLISECONDS, alternate);
}
```

Note that I also decoupled the "alternate" stream from the method, so that it's passed in, and gave params better names.

cdeszaq: If we have this method dealing with the "tri-state" behavior, then we can consolidate that logic in one place, and I _think_ it may simplify the switching behavior a bit.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
cdeszaq: Tracing through how this got called a bit more, I see how this is still doing the same thing, but it still wasn't obvious what was going on as far as ordering, so comments would definitely help make that better

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletReactiveChainforResultsEndpointSpec.groovy
```diff
@@ -152,73 +181,119 @@ class JobsServletReactiveChainforResultsEndpointSpec extends Specification {
...
+                apiJobStore
+        )
 
-        when: "PreResponse is available in the PreResponseStore"
-        mockPreResponseStore.save("ticket4", PreResponseTestingUtils.buildPreResponse("2016-04-24")).subscribe()
+        and:
+        PreResponse preResponse = PreResponseTestingUtils.buildPreResponse("2016-04-24")
 
-        and: "We miss the notification that the preResponse is stored in the PreResponseStore"
+        when: "We miss the notification that the preResponse is stored in the PreResponseStore"
```
cdeszaq: This should also be a `given` piece. The only "when" involved here is the `getResults` call.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletReactiveChainforResultsEndpointSpec.groovy
```diff
@@ -152,73 +181,119 @@ class JobsServletReactiveChainforResultsEndpointSpec extends Specification {
...
-        mockJobServlet.getPreResponseObservable("ticket4", apiRequest1)
-        //The delay is to ensure that we get the notification after async timeout
-        Thread.sleep(1000)
+        and:
+        jobsServlet.getResults("ticket1", apiRequest.asyncAfter).subscribe(resultsSubscriber)
 
-        and: "We receive the notification after async timeout"
-        broadcastChannel.publish("ticket4")
+        and: "We wait for the first attempt to get results from the store to come up empty before we add fake results"
+        getTicket1Latch.await(30, TimeUnit.SECONDS)
```
cdeszaq: Why are we waiting for up to 30 seconds before continuing here? (note that if this timeout expires, this just returns false, even if the latch has not released.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletReactiveChainforResultsEndpointSpec.groovy
```diff
@@ -152,73 +181,119 @@ class JobsServletReactiveChainforResultsEndpointSpec extends Specification {
...
+        PreResponse results = Mock(PreResponse)
+        preResponseStore.save("ticket1", results)
 
-        then: "then we go to the PreResponseStore exactly once to get the ticket"
-        1 * mockPreResponseStore.get(_) >> Observable.just(Mock(PreResponse))
+        and: "We wait for the results to be successfully stored before sending a ready notification"
+        saveTicket1Latch.await(30, TimeUnit.SECONDS)
```
cdeszaq: Why are we waiting for up to 30 seconds before continuing here? (note that if this timeout expires, this just returns false, even if the latch has not released.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletReactiveChainforResultsEndpointSpec.groovy
```diff
@@ -152,73 +181,119 @@ class JobsServletReactiveChainforResultsEndpointSpec extends Specification {
...
+                jobPayloadBuilder,
+                apiJobStore
+        )
 
-        and: "We start the async chain"
-        mockJobServlet.getPreResponseObservable("ticket4", apiRequest1)
+        when: "We start the async chain"
+        mockJobServlet.getResults("ticket4", apiRequest1.asyncAfter)
+        //The delay is to ensure that we get the notification after async timeout
+        Thread.sleep(1000)
```
cdeszaq: Can we not sleep here?

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
archolewa: Since `asyncAfter` is best effort for anything but `always` (which doesn't go to the store), it doesn't really matter either way. This basically gives the `asyncAfter=0` the semantics of "If the results are available, give them to me, otherwise quickly send back the async payload."

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
archolewa: If you look closely, you'll see that the result of `applyTimeOutIfNeeded` is actually being applied _inside_ of a `preResponseStore.get().switchIfEmpty` call. In other words, we aren't checking the PreResponseStore if the request is always asynchronous (there's no point), but we are in the other two cases.

So we can't really merge the three cases into a single helper method without more thought than is worth it. If we did, we'd probably just end up with implementing all of `getResults` in `applyTimeoutIfNeeded`.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletReactiveChainforResultsEndpointSpec.groovy
archolewa: So that the test doesn't hang indefinitely in case something goes wrong and we never actually countdown the latches. I've moved that timeout into a timeout annotation on the test.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletReactiveChainforResultsEndpointSpec.groovy
archolewa: This test is testing that things behave appropriately in the case where we have a numeric `asyncAfter`. Unfortunately, we have currently have no way of injecting latches or anything into the workflow. So our choices are to sleep (which the test was already doing before I simplified it), or refactor the JobsServlet so that we have some way of injecting additional components into the workflow.

While the second may or may not be a valid approach (depending on how much control we want to give clients over the JobServlet), it's outside the scope of this PR.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
cdeszaq: ok, makes sense

CHANGELOG.md
```diff
@@ -41,9 +45,18 @@ Current
...
     * This allows the asynchronous functional tests to add countdown latches to the workflows where necessary, allowing
       for thread-safe tests.
+=======
+-  [HashPreResponseStore moved to `test` root directory.](https://github.com/yahoo/fili/pull/39)
+   * The `HashPreResponseStore` is really intended only for testing, and does not have capabilities (i.e. TTL) that are needed for production.
+
+- [Removed `JobsApiRequest::handleBroadcastChannelNotification`](https://github.com/yahoo/fili/pull/39)
+   * That logic does not really belong in the `JobsApiRequest` (which is responsible for modeling a response, not processing it), and has
+        been consolidated into the `JobsServlet`.
+>>>>>>> Stashed changes
```
cdeszaq: Some merge conflict markers in here still...

archolewa: Apparently (vanilla) git says nothing if it has conflicts when applying a stash. This is annoying.

CHANGELOG.md
```diff
@@ -45,18 +43,17 @@ Current
...
+-  [HashPreResponseStore moved to `test` root directory.](https://github.com/yahoo/fili/pull/39)
+   * The `HashPreResponseStore` is really intended only for testing, and does not have capabilities (i.e. TTL) that are needed for production.
+
 -  [The `TestBinderFactory` now uses the `TestAsynchronousWorkflowsBuilder`](http://github.com/yahoo/fili/pull/39)
     * This allows the asynchronous functional tests to add countdown latches to the workflows where necessary, allowing
       for thread-safe tests.
-=======
--  [HashPreResponseStore moved to `test` root directory.](https://github.com/yahoo/fili/pull/39)
-   * The `HashPreResponseStore` is really intended only for testing, and does not have capabilities (i.e. TTL) that are needed for production.
+
```
cdeszaq: Not neeed

CHANGELOG.md
```diff
@@ -37,6 +43,18 @@ Current
 
 
 ### Changed:
+-  [HashPreResponseStore moved to `test` root directory.](https://github.com/yahoo/fili/pull/39)
+   * The `HashPreResponseStore` is really intended only for testing, and does not have capabilities (i.e. TTL) that are needed for production.
+
+-  [The `TestBinderFactory` now uses the `TestAsynchronousWorkflowsBuilder`](http://github.com/yahoo/fili/pull/39)
+    * This allows the asynchronous functional tests to add countdown latches to the workflows where necessary, allowing
+      for thread-safe tests.
+
```
cdeszaq: There's an extra blank line in here that's not needed
/NotABlocker

cdeszaq:
👍

fili-core/src/main/java/com/yahoo/bard/webservice/async/preresponses/stores/HashPreResponseStore.java
dayamr: 👍

dayamr:
It is good to add a test case for `asyncAfter=<non-zero-integer>` to check that behavior. It requires custom timeout to mimic this behavior across the workflow and we can consider it as a separate effort. Is it good candidate to create this as an issue? If not, what is the right way to track this?

Apart from the above comment, it looks good to me. 👍

PR #42, opened 2016-09-20
Cleaned up dependencies in pom files

cdeszaq:
Note: Based on top of the final v0.5.x commit. Once that merges, this PR will be re-pointed to master.

This PR cleans up pom dependencies so that the parent pom manages the specific versions of all of the dependencies the child poms have. It also cleans up the parent pom's `dependencies` section so that child poms are only inheriting dependencies they _all_ should have (ie. Spock, logging, etc.)

This also bumps the version to v0.6.x.

### DIFF
--- a/fili-core/pom.xml
+++ b/fili-core/pom.xml
@@ -16,11 +16,10 @@
     <parent>
         <groupId>com.yahoo.fili</groupId>
         <artifactId>fili-parent-pom</artifactId>
-        <version>0.1-SNAPSHOT</version>
+        <version>0.6-SNAPSHOT</version>
     </parent>
 
     <properties>
-        <version.druid.api>0.3.8</version.druid.api>
         <checkstyle.config.location>../checkstyle-style.xml</checkstyle.config.location>
         <checkstyle.suppressions.location>../checkstyle-suppressions.xml</checkstyle.suppressions.location>
     </properties>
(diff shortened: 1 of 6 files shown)


fili-core/pom.xml
```diff
@@ -29,230 +28,185 @@
         <dependency>
             <groupId>com.yahoo.fili</groupId>
             <artifactId>fili-system-config</artifactId>
-            <version>0.1-SNAPSHOT</version>
```
archolewa: The version information is now being picked up from the root `pom.xml` yes?

pom.xml
```diff
@@ -114,91 +115,378 @@
...
+            </dependency>
+
+            <!-- All things Spring -->
+            <dependency>
+                <groupId>org.springframework</groupId>
+                <artifactId>spring-core</artifactId>
+                <version>4.0.5.RELEASE</version>
+            </dependency>
+
+            <!-- Test -->
```
archolewa: We should probably move all of the Spock (and maybe Groovy?) dependencies under this "section" as well, since those are for testing.

archolewa:
👍

pom.xml
cdeszaq: Spock definitely is (I'll make that change), but Groovy might not be... 😉

cdeszaq: Updated

RVRSKumar:
👍

cdeszaq:
Merged

PR #46, opened 2016-09-22
Fixes #45, strip sorts from weight check queries

### LINKED ISSUE #45: Weight check query improperly retains sorting on removed columns
When the weight check query is build, any LimitSpec that's present on the initial query gets included on the weight check query. Since the post-aggregations are all removed when building the weight check query, sorting on post-agg columns results in a Druid error for the query.

cdeszaq:
Note: Based on other PRs. Once those merge, this PR will be repointed to master.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/orderby/LimitSpec.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/orderby/LimitSpec.java
@@ -45,7 +45,7 @@ public class LimitSpec {
      * @return type
      */
     public String getType() {
-        return this.type;
+        return type;
     }
 
     /**
@@ -55,7 +55,7 @@ public class LimitSpec {
      */
     @JsonInclude(JsonInclude.Include.NON_ABSENT)
     public OptionalInt getLimit() {
-        return this.limit;
+        return limit;
     }
 
     /**
@@ -66,6 +66,16 @@ public class LimitSpec {
     public LinkedHashSet<OrderByColumn> getColumns() {
         return this.columns;
     }
+
+    // CHECKSTYLE:OFF
+    public LimitSpec withColumns(LinkedHashSet<OrderByColumn> sortColumns) {
+        return new LimitSpec(sortColumns, limit);
+    }
+
+    public LimitSpec withLimit(OptionalInt limit) {
+        return new LimitSpec(columns, limit);
+    }
+    // CHECKSTYLE:ON
 
     @Override
     public boolean equals(final Object o) {
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/WeightEvaluationQuery.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/WeightEvaluationQuery.java
@@ -13,6 +13,7 @@ import com.yahoo.bard.webservice.druid.model.datasource.QueryDataSource;
 import com.yahoo.bard.webservice.druid.model.datasource.UnionDataSource;
 import com.yahoo.bard.webservice.druid.model.filter.Filter;
 import com.yahoo.bard.webservice.druid.model.having.Having;
+import com.yahoo.bard.webservice.druid.model.orderby.LimitSpec;
 import com.yahoo.bard.webservice.druid.model.postaggregation.ConstantPostAggregation;
 import com.yahoo.bard.webservice.druid.model.postaggregation.PostAggregation;
 import com.yahoo.bard.webservice.util.IntervalUtils;
@@ -22,6 +23,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Collections;
+import java.util.LinkedHashSet;
 import java.util.List;
 
 /**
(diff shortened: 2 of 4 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/orderby/LimitSpec.java
```diff
@@ -67,6 +67,16 @@ public OptionalInt getLimit() {
         return this.columns;
     }
 
+    // CHECKSTYLE:OFF
+    public LimitSpec withColumns(LinkedHashSet<OrderByColumn> sortColumns) {
+        return new LimitSpec(sortColumns, this.limit);
+    }
+
+    public LimitSpec withLimit(OptionalInt limit) {
+        return new LimitSpec(this.columns, limit);
```
archolewa: `this` is unnecessary in both withers.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/WeightEvaluationQuery.java
```diff
@@ -200,7 +205,9 @@ private static DataSource makeInnerQuery(DruidAggregationQuery<?> query, double
                         aggregations,
                         postAggregations,
                         innerQuery.getIntervals(),
-                        ((GroupByQuery) innerQuery).getLimitSpec()
+                        ((GroupByQuery) innerQuery).getLimitSpec() == null ?
+                                null :
+                                ((GroupByQuery) innerQuery).getLimitSpec().withColumns(new LinkedHashSet<>())
```
archolewa: Seeing as how we're doing this twice, and all the casting makes for a fairly complex statement, I would suggest extracting this into a helper method:

```java
private LimitSpec extractLimitSpec(GroupByQuery query) {
   return query.getLimitSpec() == null ? null : query.getLimitSpec().withColumns(new LinkedHashSet());
} 
```
This will simplify the constructor call to:
```java
   ...
   query.getQueryType() == QueryType.GROUP_BY ? extractLimitSpec((GroupByQuery) query) : null
```
and this code to:

```java
...
extractLimitSpec((GroupByQuery) innerQuery)
```

It also means that we can have a positive condition, which is always nice.

archolewa: You could also have it take a `Query` instead, and do the type checking and casting inside the method. Either approach is fine.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/query/WeightEvaluationQuerySpec.groovy
```diff
@@ -185,6 +185,38 @@ class WeightEvaluationQuerySpec extends Specification {
...
+                "", //perPage
+                "", //page
+                null, //uriInfo
+                dataServlet
+        )
+
+        GroupByQuery groupByQuery = builder.buildQuery(apiRequest, merger.merge(apiRequest))
+
+        expect: "The groupBy query has a sort, as a pre-condition"
+        !groupByQuery.limitSpec.columns.empty
```
archolewa: Since `columns` is a collection, you can simplify this using groovy truthiness:

```groovy
groupByQuery.limitSpec.columns
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/query/WeightEvaluationQuerySpec.groovy
```diff
@@ -185,6 +185,38 @@ class WeightEvaluationQuerySpec extends Specification {
...
+
+        and: "The weight-check query doesn't have a sort at either level"
+        new WeightEvaluationQuery(groupByQuery, 1).limitSpec.columns.empty
+        new WeightEvaluationQuery(groupByQuery, 1).innermostQuery.limitSpec.columns.empty
```
archolewa: You can do something [similar](https://github.com/yahoo/fili/pull/46/files#diff-682dade2e010757acb0dec5327e1c196R213) here.

will-lauer:
:+1:

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/orderby/LimitSpec.java
michael-mclawhorn: It can underline which parameters are pulled from local instance.

michael-mclawhorn:
I'm warm to andrew's requests.  Otherwise 👍  for me.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/orderby/LimitSpec.java
cdeszaq: Done

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/WeightEvaluationQuery.java
cdeszaq: Done

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/query/WeightEvaluationQuerySpec.groovy
cdeszaq: Done

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/query/WeightEvaluationQuerySpec.groovy
cdeszaq: Done

cdeszaq:
Merged

PR #49, opened 2016-09-23
Synchronous queries are no longer stored in ApiJobStore.

archolewa:
-- The `delay` operator considers `Observable::onCompleted` to be an
"emitted item." Therefore, the job row was still being stored in the
ApiJobStore even when the query was synchronous.

--This has been fixed by replacing `delay` with a subscription, and/or
zip when appropriate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,14 @@ Current
 
 ### Fixed:
 
+- [Fixes a bug where job metadata was being stored in the `ApiJobStore` even when the results came back synchronously](https://github.com/yahoo/fili/pull/49)
+  * The workflow that updates the job's metadata with `success` was running even when the query was synchronous. That 
+    update also caused the ticket to be stored in the `ApiJobStore`.
+  * The delay operator didn't stop the "update" workflow from executing because it viewed an `Observable::onCompleted`
+    call as a message for the purpose of the delay. Since the two observables that that the metadata update gated on are
+    empty when the query is synchronous, the "update metadata" workflow was being triggered every time.
+  * The delay operator was replaced by `zipWith` as a gating mechanism.
+    
 - [#45, removing sorting from weight check queries](https://github.com/yahoo/fili/pull/46)
 
 
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/async/workflows/DefaultAsynchronousWorkflowsBuilder.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/async/workflows/DefaultAsynchronousWorkflowsBuilder.java
@@ -113,12 +113,14 @@ public class DefaultAsynchronousWorkflowsBuilder implements AsynchronousWorkflow
         // JobRow has been stored because otherwise we have a race condition where the version of the JobRow with
         // a success/failure status could be written to the store first, and then immediately overwritten by the
         // original JobRow, leaving the JobRow in a state of "pending" for its lifetime.
-        Observable<JobRow> jobRowUpdatedNotification = buildUpdateRowChain(
+        ConnectableObservable<JobRow> jobRowUpdatedNotification = buildUpdateRowChain(
                 jobRowStoredNotification,
                 preResponseStoredNotification,
                 errorHandlingPreResponseEmitter,
                 jobMetadata
-        );
+        )
+                .replay(1);
+        jobRowUpdatedNotification.connect();
 
         return new AsynchronousWorkflows(
                 synchronousPayload,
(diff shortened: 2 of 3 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/async/workflows/DefaultAsynchronousWorkflowsBuilder.java
```diff
@@ -113,12 +113,14 @@ public AsynchronousWorkflows buildAsynchronousWorkflows(
...
-        Observable<JobRow> jobRowUpdatedNotification = buildUpdateRowChain(
+        ConnectableObservable<JobRow> jobRowUpdatedNotification = buildUpdateRowChain(
                 jobRowStoredNotification,
                 preResponseStoredNotification,
                 errorHandlingPreResponseEmitter,
                 jobMetadata
-        );
+        )
+                .replay(1);
+        jobRowUpdatedNotification.connect();
```
archolewa: The other notifications are hot, and since this observable relies on hot observables, I felt it was safest to also make it hot.

archolewa:
Needs a CHANGELOG.

fili-core/src/main/java/com/yahoo/bard/webservice/async/workflows/DefaultAsynchronousWorkflowsBuilder.java
```diff
@@ -204,8 +206,8 @@ public AsynchronousWorkflows buildAsynchronousWorkflows(
         // The job status should not be updated until both the PreResponse has been stored, and the storage of the
         // original Job ticket has been attempted.
         return preResponseEmitter
-                .delay(ignored -> preResponseStoredNotification)
-                .delay(ignored -> jobRowStoredNotification)
+                .zipWith(jobRowStoredNotification, (preResponse, ignored) -> preResponse)
+                .zipWith(preResponseStoredNotification, (preResponse, ignored) -> preResponse)
```
cdeszaq: A comment around why we can ignore the other side of these zips would be good.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/SynchronousQueriesAreNotStoredSpec.groovy
```diff
@@ -0,0 +1,47 @@
...
+        return [
+                data: {[
+                        metrics: ["height"],
+                        asyncAfter: ["never"],
+                        dateTime: ["2016-08-30/2016-08-31"]
+                ]},
+                jobs: {[
+                        //'greg' is the UserId hard-coded in TestBinderFactory::buildJobRowBuilder and used for
+                        //all queries created in the test environment.
+                        filters: ["userId-eq[greg]"]
```
cdeszaq: Any reason we couldn't just get all of them?

dayamr:
After CHANGELOG entry. 👍

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/SynchronousQueriesAreNotStoredSpec.groovy
archolewa: Because the job store is prepopulated with jobs for functional tests against the jobs endpoint, so it's not empty.

fili-core/src/main/java/com/yahoo/bard/webservice/async/workflows/DefaultAsynchronousWorkflowsBuilder.java
archolewa: The comment above the stream is supposed to explain it, by explaining that we need to wait until the job row has been stored, and the pre response has been stored. I can make it more explicit that we're using the zipWith as a gating mechanism if you'd like.

cdeszaq: Yeah, the "zip as gate" is a little odd, I think.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/SynchronousQueriesAreNotStoredSpec.groovy
cdeszaq: ahh, ok. Makes sense

fili-core/src/main/java/com/yahoo/bard/webservice/async/workflows/DefaultAsynchronousWorkflowsBuilder.java
archolewa: It is, but I don't really know of a better way. Not if we want to release it iff the two observables emit items.

CHANGELOG.md
```diff
@@ -26,6 +26,14 @@ Current
 
 ### Fixed:
 
+- [Fixes a bug where job metadata was being stored in the `ApiJobStore` even when the results came back synchronously](https://github.com/yahoo/fili/pull/49)
+    * The workflow that updates the job's metadata with `success` was running even when the query was synchronous. That update had the impact of
+      storing the ticket in the ApiJobStore.
+    * The reason the delay operator didn't stop the update workflow from executing viewed an `Observable::onCompleted` as a message for the purpose of
+      delays. Since the two observables that that the metadata update it gated on are empty when the query is synchronous, the update metadata workflow 
+      was being updated.
```
cdeszaq: This bullet seems... confusing? Could it be cleaned / tightened up a bit?

PR #50, opened 2016-09-26
Corrected bug with Fili sub-module dependency specification

cdeszaq:
Dependency versions are now set via a fixed property at deploy time, rather than
relying on `project.version`


archolewa:
👍

michael-mclawhorn:
👍

PR #51, opened 2016-09-26
Implemented slices servlet surfacing segment info information.

michael-mclawhorn:
This change surfaces the segment information on a physical table to the slices/slice endpoint.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
@@ -272,6 +272,8 @@ public abstract class AbstractBinderFactory implements BinderFactory {
 
                 DataSourceMetadataService dataSourceMetadataService = buildDataSourceMetadataService();
 
+                bind(dataSourceMetadataService).to(DataSourceMetadataService.class);
+
                 QuerySigningService<?> querySigningService = buildQuerySigningService(
                         loader.getPhysicalTableDictionary(),
                         dataSourceMetadataService
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/SlicesApiRequest.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/SlicesApiRequest.java
@@ -7,21 +7,26 @@ import static com.yahoo.bard.webservice.web.ErrorMessageFormat.SLICE_UNDEFINED;
 
 import com.yahoo.bard.webservice.data.dimension.Dimension;
 import com.yahoo.bard.webservice.data.dimension.DimensionColumn;
+import com.yahoo.bard.webservice.metadata.DataSourceMetadataService;
+import com.yahoo.bard.webservice.metadata.SegmentInfo;
 import com.yahoo.bard.webservice.table.Column;
 import com.yahoo.bard.webservice.table.PhysicalTable;
 import com.yahoo.bard.webservice.table.PhysicalTableDictionary;
 import com.yahoo.bard.webservice.web.endpoints.DimensionsServlet;
 import com.yahoo.bard.webservice.web.endpoints.SlicesServlet;
 
+import org.joda.time.DateTime;
 import org.joda.time.Interval;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.SortedMap;
 import java.util.stream.Collectors;
 
 import javax.validation.constraints.NotNull;
@@ -36,6 +41,7 @@ public class SlicesApiRequest extends ApiRequest {
 
     private final Set<Map<String, String>> slices;
     private final Map<String, Object> slice;
+    private final DataSourceMetadataService dataSourceMetadataService;
 
     /**
      * Parses the API request URL and generates the Api Request object.
@@ -50,6 +56,7 @@ public class SlicesApiRequest extends ApiRequest {
      * @param page  desired page of results. If present in the original request, must be a positive
      * integer. If not present, must be the empty string.
      * @param tableDictionary  cache containing all the valid physical table objects.
+     * @param dataSourceMetadataService  a resource holding the available datasource metadata
      * @param uriInfo  The URI of the request object.
      *
      * @throws BadApiRequestException is thrown in the following scenarios:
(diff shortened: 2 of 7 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/web/SlicesApiRequest.java
```diff
@@ -174,18 +190,47 @@ protected SlicesApiRequest() {
...
+     * @param sliceMetadata  The raw set of maps of datetime to maps of segment names to segment details
+     *
+     * @return A simplified Map of dateTime to set of segment identifiers
+     */
+    private static Map<DateTime, Set<String>> generateSegmentMetadataView(
+            Set<SortedMap<DateTime, Map<String, SegmentInfo>>> sliceMetadata
+    ) {
+        Map<DateTime, Set<String>> segmentView;
+        if (!sliceMetadata.isEmpty()) {
+            SortedMap<DateTime, Map<String, SegmentInfo>> segmentInfo = sliceMetadata.iterator().next();
```
archolewa: It looks like we are only ever grabbing one element from the `sliceMetadata` is that intentional? Shouldn't be building the map from _all_ the buckets in the `sliceMetadata`?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/SlicesApiRequestSpec.groovy
```diff
@@ -96,13 +107,17 @@ class SlicesApiRequestSpec extends Specification {
...
+                Collections.singleton(table)
+        );
 
         Map<String, Object> expected = new LinkedHashMap<>()
         expected.put("name", name);
         expected.put("timeGrain", fullDictionary.get(name).getTimeGrain().getName());
         expected.put("timeZone", "UTC");
         expected.put("dimensions", dimensionsResult);
         expected.put("metrics", metricsResult);
+        expected.put("segmentInfo", SlicesApiRequest.generateSegmentMetadataView(sliceMetadata))
```
archolewa: Isn't this test supposed to be testing `SlicesApiRequest::generateSegmentMetadataView` (amongst other things)? As is, we're using the code under test to generate the expected value!

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -268,6 +268,8 @@ protected void configure() {
 
                 DataSourceMetadataService dataSourceMetadataService = buildDataSourceMetadataService();
 
+                bind(dataSourceMetadataService).to(DataSourceMetadataService.class);
```
cdeszaq: What were we doing to have this service injected in other places? Was it only being used by us to inject it, and never actually bound for others to use?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/SlicesApiRequestSpec.groovy
```diff
@@ -4,23 +4,30 @@ package com.yahoo.bard.webservice.web
...
+import com.yahoo.bard.webservice.metadata.BaseDataSourceMetadataSpec
+import com.yahoo.bard.webservice.metadata.DataSourceMetadata
+import com.yahoo.bard.webservice.metadata.DataSourceMetadataService
+import com.yahoo.bard.webservice.metadata.SegmentInfo
 import com.yahoo.bard.webservice.table.Column
 import com.yahoo.bard.webservice.table.PhysicalTableDictionary
+import com.yahoo.bard.webservice.table.Table
 import com.yahoo.bard.webservice.web.endpoints.SlicesServlet
 
+import org.joda.time.DateTime
```
cdeszaq: Unused import

fili-core/src/main/java/com/yahoo/bard/webservice/web/SlicesApiRequest.java
michael-mclawhorn: dataSourceMetadataService.getTableSegments( Set of Tables )
returns the set of segments for a set of tables.

dataSourceMetadataService.getTableSegments(
                Collections.singleton(table)
        );
should return the set of segments for a particular table.  We don't have a way to cut that set of one down to the value of one other than by popping it out of the singleton set.

cdeszaq: If this is handled, I _think_ the conditional around `!sliceMetadata.isEmpty()` can go away too, since a stream over the top-level set (if empty) would produce an empty Map anyways. I'm always a fan of simplifying conditionals and special-cases

michael-mclawhorn:
The reason I haven't done more due dilligence about the segment metadata view is that there is a larger change needed to determine what segment metadata should look like when surfaced to the user, and rather than try to spec that out, I simply tried to cut it down to a _barely_ readable format.

We can block this on design decisions about what that data should look like.  Ultimately, the awkwardness of Set<Map<DateTime,Map<String,InfoThingie>>> is a massive code smell and there should be real java objects around that mess of nested types.  But I didn't want to pull that string if I can help it.

fili-core/src/main/java/com/yahoo/bard/webservice/web/SlicesApiRequest.java
cdeszaq: Even though it returns a set, handling the collection (just flatten it) shouldn't be too bad.

Otherwise, if you must, do the set-unwraping at the parent level, rather than inside the processing method.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
michael-mclawhorn: We constructed it and passed it to objects which we bound separately or loaded directly into tasks
line 275: queryBindingService
line 290, datasource metadata loader

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/SlicesApiRequestSpec.groovy
michael-mclawhorn: Us
ed on line 110

fili-core/src/main/java/com/yahoo/bard/webservice/web/SlicesApiRequest.java
michael-mclawhorn: I was having trouble with nesting the streams of streams of streams.  I can take another stab at it but readability was limited.

cdeszaq: `mySetOfThings.stream().flatmap(Set::stream)`?

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
cdeszaq: Ah, ok

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/SlicesApiRequestSpec.groovy
cdeszaq: Ah, so it is. My bad. (damn gnarly types...)

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/SlicesApiRequestSpec.groovy
archolewa: Ah. Well, we may want a test for `generateSegmentMetadataView` then.

archolewa: We should add a comment though indicating that this test is not intended as a test of `generateSegmentMetadataView` but only that the results get added to the payload.

archolewa:
@michael-mclawhorn I don't really want to block this PR on proper output design. Having some way of displaying the segment metadata that is loaded (even if it's ugly) is better than having none. Especially since the slices endpoint isn't officially supported anyway, so I feel no qualms about massively changing the response format for the segments later.

archolewa:
👍

cdeszaq:
👍

cdeszaq:
Where's this PR at? Are we planning on merging it, or is it dead / going to die?

archolewa:
👍

PR #52, opened 2016-09-27
Adds read locks to LuceneSearchProvider.

archolewa:
The `LuceneSearchProvider` is missing read locks before reading the lucene database. As a result, Bad Things can happen if you attempt to read from Lucene (i.e. resolving dimension filters) while loading dimension indices.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
@@ -101,24 +101,28 @@ public class LuceneSearchProvider implements SearchProvider {
     }
 
     /**
-     * Get the IndexSearcher.
-     *
-     * @return the IndexSearcher
+     * Initializes the `luceneIndexSearcher` if it has not been initialized already.
+     * <p>
+     * Note that the index searcher cannot be built at construction time, because it needs the dimension and
+     * associated key-value store. However, because of a circular dependency between the `SearchProvider` and the
+     * `Dimension` classes, we cannot provide the dimension and key-value store to the search provider at
+     * construction time.
      */
-    private IndexSearcher getIndexSearcher() {
-        // Open the searcher if we don't have one
+    private void initializeIndexSearcher() {
         if (luceneIndexSearcher == null) {
             reopenIndexSearcher(true);
         }
-        return luceneIndexSearcher;
     }
 
     /**
      * Re-open the Index Searcher, opening it for the first time if it's never been opened.
+     * <p>
+     * This method will attempt to acquire and release a write lock.
      *
      * @param firstTimeThrough  If true, will write an empty index and will then re-open the searcher
      */
     private synchronized void reopenIndexSearcher(boolean firstTimeThrough) {
+        lock.writeLock().lock();
         try {
             // Close the current reader if open
             if (luceneIndexSearcher != null) {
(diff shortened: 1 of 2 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -547,17 +554,23 @@ private Query getFilterQuery(Set<ApiFilter> filters, int perPage) {
                 .map(document -> document.get(idKey))
                 .map(dimension::findDimensionRowByKeyValue)
                 .collect(Collectors.toCollection(TreeSet::new));
+
+        int documentCount;
+        lock.readLock().lock();
```
cdeszaq: Don't we need to get this lock much earlier in this method? We're getting `indexSearcher` up near the top of the method...

michael-mclawhorn:
What would be great would be if we had some way of testing access to the indexer low levels so we could validate that all calls to them were only done in a readLocked context.

It looks pretty good.  Needs a CHANGELOG.

cdeszaq:
A few more thoughts:
-Probably also need a write-lock in the `reopenIndexSearcher` method when we close/re-open it
- `clearDimension` needs some help:
  -It currently will deadlock, since it holds a write-lock, and calls `refreshCardinality` from within it, which tries to grab a read-lock. I think moving the method calls to`refreshCardinality` and `reopenIndexSearcher` outside the `try/catch/finally` is all that's needed, but I didn't trace it too carefully

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
archolewa: Yes.

archolewa:
@michael-mclawhorn You _might_ be able to do something by taking a write lock in a separate thread and then inspecting the guts of the read-write lock. I did a little bit of digging, but it's not immediately clear to me how to do that.

Alternatively, you could take a write lock and then verify that the method doesn't finish, but that would significantly increase our testing time.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -513,51 +542,57 @@ private Query getFilterQuery(Set<ApiFilter> filters, int perPage) {
...
-            }
-        }
-        for (int currentPage = 1; currentPage < requestedPageNumber; currentPage++) {
-            ScoreDoc lastEntry = hits[hits.length - 1];
-            hits = getPageOfData(indexSearcher, lastEntry, query, perPage, requestedPageNumber).scoreDocs;
+        TreeSet<DimensionRow> filteredDimRows;
+        int documentCount;
+        initializeIndexSearcher();
+        try {
+            lock.readLock().lock();
```
cdeszaq: Move outside the `try` to match the other calls in this class?

cdeszaq:
👍

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
wcekan: Correct.  any lock must complete before the `try`

wcekan:
👍

PR #54, opened 2016-09-29
Correct the license to be accurate

cdeszaq:
We had the "pointer" piece, but not the _actual_ license, which we need.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,15 @@ Current
 
 ### Changed:
 
+- Clean up build plugins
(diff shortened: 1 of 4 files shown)


archolewa:
I'm assuming you copy-pasted this from some apache license website or other?

cdeszaq:
@archolewa I grabbed it from GitHub's little helper tool that recognizes when you're trying to add a file like "LICENSE" through the web-ui, and it gives a little dropdown with licenses you can add.

archolewa:
👍

michael-mclawhorn:
👍

PR #55, opened 2016-09-29
Eliminated unwanted warning about logical name to physical name mappi…

dayamr:
We don't need to look for logical name to physical name mapping for metrics.  This mapping was causing unwanted warnings in log like.

"No mapping found for logical name: '<metricName>' to physical name (will use logical name as physical name). This is unexpected and should not happen for properly configured dimensions."

Hence removed that lookup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -140,6 +140,10 @@ Jobs resource. Here are the highlights of what's in this release:
 
 ### Changed:
 
+- Removed `physicalName` lookup for metrics in `TableUtils::getColumnNames` to remove spurious warnings
+     * Metrics are not mapped like dimensions are. Dimensions are aliased per physical table and metrics are aliazed per logical table.
+     * Logical metric is mapped with one or many physical metrics. Same look up logic for dimension and metrics doesn't make sense.
+
 #### Jobs:
 
 - [HashPreResponseStore moved to `test` root directory.](https://github.com/yahoo/fili/pull/39)
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/util/TableUtilsSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/util/TableUtilsSpec.groovy
@@ -89,4 +89,17 @@ class TableUtilsSpec extends  Specification {
         expect:
         TableUtils.getColumnNames(request, query, new PhysicalTable("", DefaultTimeGrain.DAY.buildZonedTimeGrain(DateTimeZone.UTC), [:])) == [d1Name, metric1, metric2, metric3] as Set
     }
+
+    def "logicalName to physicalName mapping not required for metrics" () {
+        setup:
+        request.dimensions >> []
+        request.filterDimensions >> []
+        query.metricDimensions >> []
+        query.dependentFieldNames >> ([metric1] as Set)
+        PhysicalTable physicalTable = Mock(PhysicalTable)
+        0 * physicalTable.getPhysicalColumnName(_)
+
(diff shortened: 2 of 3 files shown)


archolewa:
The warning is a side-effect of the refactorings made to support dimensions that have the same api name, but different physical names on each logical table. That lookup is absolutely necessary to support systems that have that property and should not be removed. 

If we want to remove the warnings, then either the configuration of projects that don't have that property need to reworked, or we should revisit the question of whether we should consider systems that don't need that mapping (and thus don't an empty `logicalToPhysicalNames` map in their `LogicalTable`s) are perfectly valid systems, and so the warning should be removed.

cdeszaq:
@archolewa You'll note that this PR isn't about dimensions, but is about _metrics_. The lookup for resolving dimensions isn't going away (I agree, it's super needed), but applying the same lookup for metrics doesn't make sense, since metrics are not handled the same way as dimensions in this regard. (ie. changing metrics happens per-logical table, not per-physical table at the moment)

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/TableUtilsSpec.groovy
```diff
@@ -85,8 +85,10 @@ class TableUtilsSpec extends  Specification {
         request.filterDimensions >> ds1
         query.metricDimensions >> ds1
         query.dependentFieldNames >> ([metric1, metric2, metric3] as Set)
+        PhysicalTable physicalTable = new PhysicalTable("", DefaultTimeGrain.DAY.buildZonedTimeGrain(DateTimeZone.UTC), [:])
+        0 * physicalTable.getPhysicalColumnName(_)
```
cdeszaq: This assertion doesn't make sense to me, for a couple of reasons:

1. Why are we adding the assertion? We didn't have it before, what value does adding it provide? <sub>Note: I'm not saying we _shouldn't_ add it, I just was curious why we might want to</sub>
2. Why does this test pass now? It looks like `TableUtils::getColumnNames` _should_ be calling `getPhysicalColumnName` for _dimensions_, so I think it should be called at least _once_. Something is very fishy here.

CHANGELOG.md
```diff
@@ -95,6 +95,9 @@ Jobs resource. Here are the highlights of what's in this release:
 
 ### Changed:
 
+- [Removed the physicaName lookup for the logical metric name](https://github.com/yahoo/fili/pull/55)
+  * Updated `TableUtils::getColumnNames`
```
cdeszaq: This 2nd line isn't really adding any benefit. Or rather, perhaps the 2 lines can be combined?:

```
   - Removed `physicalName` lookup for metrics in `TableUtils::getColumnNames` to remove spurious warnings
     * Metrics are not mapped like dimensions are
```

archolewa:
@cdeszaq This is what happens when I look at a PR late in the day...

cdeszaq:
What's the status of this PR? Is it being worked on?

archolewa:
👍

CHANGELOG.md
```diff
@@ -122,6 +122,9 @@ Jobs resource. Here are the highlights of what's in this release:
 
 ### Changed:
 
+- Removed `physicalName` lookup for metrics in `TableUtils::getColumnNames` to remove spurious warnings
+     * Metrics are not mapped like dimensions are
+
```
michael-mclawhorn: Maybe just a few more words here for context.

cdeszaq:
👍 pending Mike's suggestion for a bit better change-log entry.

PR #57, opened 2016-10-03
Support custom Druid queries.

archolewa:
--`QueryType` has been turned into an interface backed by an enum.

--The `DruidResponseParser` and `DruidQueryBuilder` are now injectable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,13 @@ Current
 
 ### Changed:
 
+- [Fili now supports custom Druid query types](https://github.com/yahoo/fili/pull/57)
+    * `QueryType` has been turned into an interface, backed by an enum `DefaultQueryType`.
+        - The default implementations of `DruidResponseParser` `DruidQueryBuilder`, `WeightEvaluationQuery` and
+          `TestDruidWebService` only support `DefaultQueryType`.
+    * `DruidResponseParser` is now injectable by overriding `AbstractBinderFactory::buildDruidResponseParser` method.
+    * `DruidQueryBuilder` is now injectable by overriding `AbstractBinderFactory::buildDruidQueryBuilder` method.
+
 - [Updated commons-collections4 dependency from 4.0 to 4.1 to address a security vulnerability in the library.](https://github.com/yahoo/fili/pull/52)
   * For details see: https://commons.apache.org/proper/commons-collections/security-reports.html#Apache_Commons_Collections_Security_Vulnerabilities
   * It should be noted that Fili does not make use of any the serialization/deserialization capabilities of any classes
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidResponseParser.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidResponseParser.java
@@ -8,6 +8,7 @@ import com.yahoo.bard.webservice.data.dimension.DimensionColumn;
 import com.yahoo.bard.webservice.data.dimension.DimensionRow;
 import com.yahoo.bard.webservice.data.metric.MetricColumn;
 import com.yahoo.bard.webservice.druid.model.QueryType;
+import com.yahoo.bard.webservice.druid.model.DefaultQueryType;
 import com.yahoo.bard.webservice.table.ZonedSchema;
 
 import com.fasterxml.jackson.databind.JsonNode;
@@ -38,7 +39,8 @@ public class DruidResponseParser {
      *
      * @param jsonResult  Druid results in json
      * @param schema  Schema for results
-     * @param queryType  the type of query
+     * @param queryType  the type of query, note that this implementation only supports instances of
+     * {@link DefaultQueryType}
      *
      * @return the set of results
      */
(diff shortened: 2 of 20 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidResponseParser.java
```diff
@@ -38,20 +39,28 @@
...
+        DefaultQueryType defaultQueryType = (DefaultQueryType) queryType;
+
         /* Get dimension and metric columns */
         Set<DimensionColumn> dimensionColumns = schema.getColumns(DimensionColumn.class);
         Set<MetricColumn> metricColumns = schema.getColumns(MetricColumn.class);
 
-        List<Result> results;
-        switch (queryType) {
+        List<Result> results = null;
+        switch (defaultQueryType) {
```
archolewa: The "missing" `break` is in the default case.

CHANGELOG.md
```diff
@@ -14,6 +14,11 @@ Current
 
 ### Changed:
 
+- [Fili now supports custom Druid query types](https://github.com/yahoo/fili/pull/57)
+    * `QueryType` has been turned into an interface, backed by an enum `DefaultQueryType`.
+    * `DruidResponseParser` is now injectable by overriding `AbstractBinderFactory::buildDruidResponseParser` method.
+    * `DruidQueryBuilder` is now injectable by overriding `AbstractBinderFactory::buildDruidQueryBuilder` method.
```
cdeszaq: Wasn't this already overridable?

fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidResponseParser.java
```diff
@@ -38,20 +39,28 @@
...
+        }
+        DefaultQueryType defaultQueryType = (DefaultQueryType) queryType;
+
         /* Get dimension and metric columns */
         Set<DimensionColumn> dimensionColumns = schema.getColumns(DimensionColumn.class);
         Set<MetricColumn> metricColumns = schema.getColumns(MetricColumn.class);
 
-        List<Result> results;
-        switch (queryType) {
+        List<Result> results = null;
```
cdeszaq: Is this needed? It defaults to `null`

fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidResponseParser.java
```diff
@@ -38,20 +39,28 @@
...
+     * @param queryType  the type of query, note that this implementation only supports instances of
+     * {@link DefaultQueryType}
      *
      * @return the set of results
      */
     public ResultSet parse(JsonNode jsonResult, ZonedSchema schema, QueryType queryType) {
 
         LOG.trace("Parsing druid query {} by json result: {} using schema: {}", queryType, jsonResult, schema);
 
+
```
cdeszaq: Extra whitespace not needed

fili-core/src/test/java/com/yahoo/bard/webservice/models/druid/client/impl/TestDruidWebService.java
```diff
@@ -114,7 +115,12 @@ public void postDruidQuery(
         }
 
         // Set the response to use based on the type of the query we're processing
-        switch (lastQuery.getQueryType()) {
+        if (!(lastQuery.getQueryType() instanceof DefaultQueryType)) {
+            throw new IllegalArgumentException("Illegal query type : " + lastQuery.getQueryType());
```
cdeszaq: We should add a note (likely to the changelog) that this test extension point doesn't support new types of queries.

fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidResponseParser.java
```diff
@@ -38,20 +39,28 @@
...
+        if (!(queryType instanceof DefaultQueryType)) {
+            // Throw an exception for unsupported query types
+            unsupportedQueryType(queryType);
```
cdeszaq: We should add a note (likely to the changelog) that this test extension point doesn't support new types of queries.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/WeightEvaluationQuery.java
```diff
@@ -185,13 +183,18 @@ private static DataSource makeInnerQuery(DruidAggregationQuery<?> query, double
...
         List<Aggregation> aggregations;
-        aggregations = Collections.<Aggregation>singletonList(new CountAggregation("ignored"));
+        aggregations = Collections.singletonList(new CountAggregation("ignored"));
 
         // Get the inner post aggregation
         List<PostAggregation> postAggregations;
-        postAggregations = Collections.<PostAggregation>singletonList(new ConstantPostAggregation("count", weight));
+        postAggregations = Collections.singletonList(new ConstantPostAggregation("count", weight));
+
+        if (!(innerQuery.getQueryType() instanceof DefaultQueryType)) {
```
cdeszaq: We should add a note (likely to the changelog) that this test extension point doesn't support new types of queries.

michael-mclawhorn:
Do we want to make a test on DruidResponseParserSpec to exercise the unsupported operation exception on a non enum instance of the query type interface.

CHANGELOG.md
archolewa: [Nope.](https://github.com/yahoo/fili/pull/57/files#diff-4645aafefd3d237deea57399558cfb49L211)

fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidResponseParser.java
archolewa: It's needed to make the java compiler happy. Otherwise, it complains that `results` may not have been initialized (likely because it doesn't realize that the default case throws an exception, now that the exception throwing is hidden behind a method.

michael-mclawhorn:
👍

PR #58, opened 2016-10-06
Add ability to sort mixed-type lists to JsonSlurper

cdeszaq:
Before, if a list started with a `Comparable` (ie. string, boolean, or number), it would be sorted using the objects' comparability. This broke if the list contained more than 1 type, like a String and a Map. With this change, that is fixed, and it _should_ retain any sorts that existed before.

Note: Only impacts test code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,8 @@ Current
     
 - [#45, removing sorting from weight check queries](https://github.com/yahoo/fili/pull/46)
 
+- [`JsonSlurper` can now handle sorting lists with mixed-type entries](https://github.com/yahoo/fili/pull/58)
+  * even if the list starts with a string, number, or boolean
 
 ### Known Issues:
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -66,6 +66,8 @@ Current
     
 - [#45, removing sorting from weight check queries](https://github.com/yahoo/fili/pull/46)
 
+- `JsonSlurper` can now handle sorting lists with mixed-type entries, even if the list starts with a string, number, or
+  boolean
```
archolewa: This should have a link to the PR.

fili-core/src/test/java/com/yahoo/bard/webservice/util/JsonSlurper.java
```diff
@@ -5,6 +5,7 @@
 /*
  * Changed HashMap to LinkedHashMap for consistency. Added ability to optionally sort the lists and maps.
  */
+
```
archolewa: This code was borrowed from Groovy right? While we're here, we really should update the copyright notification to reflect that.

michael-mclawhorn:
👍

codingwhatever:
👍

fili-core/src/test/java/com/yahoo/bard/webservice/util/JsonSlurper.java
cdeszaq: Fixed

CHANGELOG.md
cdeszaq: Fixed

PR #61, opened 2016-10-10
Improves error messages around talking to Druid

archolewa:
--The `ResponseException` now includes a message that serializes
its internal state.

--The status code and druid query are logged when building a
`ResponseException` in the async callbacks.


cdeszaq:
👍

codingwhatever:
👍

PR #62, opened 2016-10-10
Ability to add auth headers to Druid requests.

codingwhatever:
We are adding security to our Druid cluster and need Fili to be able to send a token in a header for authenticating our Druid requests.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
@@ -109,6 +109,7 @@ import org.slf4j.LoggerFactory;
 import rx.subjects.PublishSubject;
 
 import java.io.IOException;
+import java.lang.reflect.Constructor;
 import java.time.Clock;
 import java.time.ZoneId;
 import java.util.Arrays;
@@ -119,6 +120,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -143,6 +145,8 @@ public abstract class AbstractBinderFactory implements BinderFactory {
     private static final String METER_CACHE_HIT_RATIO = "queries.meter.cache.hit_ratio";
     private static final String METER_SPLITS_TOTAL_RATIO = "queries.meter.split_queries.total_ratio";
     private static final String METER_SPLITS_RATIO = "queries.meter.split_queries.ratio";
+
+    private static final String DRUID_HEADER_SUPPLIER_CLASS = "druid_header_supplier_class";
 
     // Two minutes in milliseconds
     public static final int HC_LAST_RUN_PERIOD_MILLIS_DEFAULT = 120 * 1000;
@@ -937,7 +941,33 @@ public abstract class AbstractBinderFactory implements BinderFactory {
      * @return A DruidWebService
      */
     protected DruidWebService buildDruidWebService(DruidServiceConfig druidServiceConfig, ObjectMapper mapper) {
-        return new AsyncDruidWebServiceImpl(druidServiceConfig, mapper);
+        Supplier<Map<String, String>> supplier = buildDruidWebServiceHeaderSupplier();
+        return new AsyncDruidWebServiceImpl(druidServiceConfig, mapper, supplier);
+    }
+
+    /**
+     * Build the Supplier for Druid data request headers.
+     *
+     * @return The Druid data request header Supplier.
+     */
+    protected Supplier<Map<String, String>> buildDruidWebServiceHeaderSupplier() {
+        Supplier<Map<String, String>> supplier = HashMap::new;
+        String customSupplierClassString = SYSTEM_CONFIG.getStringProperty(DRUID_HEADER_SUPPLIER_CLASS, null);
+        if (customSupplierClassString != null && customSupplierClassString.equals("")) {
+            try {
+                Class<?> c = Class.forName(customSupplierClassString);
+                Constructor<?> constructor = c.getConstructor();
+                supplier = (Supplier<Map<String, String>>) constructor.newInstance();
+            } catch (Exception e) {
+                LOG.error(
+                        "Unable to load the Druid query header supplier, className: {}, exception: {}",
+                        customSupplierClassString,
+                        e
+                );
+                throw new IllegalStateException(e);
+            }
+        }
+        return supplier;
     }
 
     /**
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
@@ -39,7 +39,10 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Supplier;
 
 import javax.ws.rs.core.Response.Status;
 
@@ -61,7 +64,7 @@ public class AsyncDruidWebServiceImpl implements DruidWebService {
     public static final String DRUID_WEIGHTED_QUERY_TIMER = DRUID_TIMER + "_W_";
     public static final String DRUID_SEGMENT_METADATA_TIMER = DRUID_TIMER + "_S_0";
 
-
+    private final Supplier<Map<String, String>> headersToAppend;
     private final DruidServiceConfig serviceConfig;
 
     /**
@@ -69,9 +72,30 @@ public class AsyncDruidWebServiceImpl implements DruidWebService {
      *
      * @param serviceConfig  Configuration for the Druid Service
      * @param mapper  A shared jackson object mapper resource
+     * @deprecated We now require a header supplier parameter.
+     *      Use {@link #AsyncDruidWebServiceImpl(DruidServiceConfig, ObjectMapper, Supplier)}
      */
-    public AsyncDruidWebServiceImpl(DruidServiceConfig serviceConfig, ObjectMapper mapper) {
-        this(serviceConfig, initializeWebClient(serviceConfig.getTimeout()), mapper);
+    @Deprecated
+    public AsyncDruidWebServiceImpl(
+            DruidServiceConfig serviceConfig,
+            ObjectMapper mapper
+    ) {
+        this(serviceConfig, initializeWebClient(serviceConfig.getTimeout()), mapper, HashMap::new);
+    }
+
+    /**
+     * Friendly non-DI constructor useful for manual tests.
+     *
+     * @param serviceConfig  Configuration for the Druid Service
+     * @param mapper  A shared jackson object mapper resource
+     * @param headersToAppend Supplier for map of headers for Druid requests
+     */
+    public AsyncDruidWebServiceImpl(
+            DruidServiceConfig serviceConfig,
+            ObjectMapper mapper,
+            Supplier<Map<String, String>> headersToAppend
+    ) {
+        this(serviceConfig, initializeWebClient(serviceConfig.getTimeout()), mapper, headersToAppend);
     }
 
     /**
@@ -101,8 +125,32 @@ public class AsyncDruidWebServiceImpl implements DruidWebService {
      * @param config  the configuration for this druid service
      * @param asyncHttpClient  the HTTP client
      * @param mapper  A shared jackson object mapper resource
+     * @deprecated  We now require a header supplier parameter.
+     *      Use {@link #AsyncDruidWebServiceImpl(DruidServiceConfig, AsyncHttpClient, ObjectMapper, Supplier)}
      */
-    public AsyncDruidWebServiceImpl(DruidServiceConfig config, AsyncHttpClient asyncHttpClient, ObjectMapper mapper) {
+    @Deprecated
+    public AsyncDruidWebServiceImpl(
+            DruidServiceConfig config,
+            AsyncHttpClient asyncHttpClient,
+            ObjectMapper mapper
+    ) {
+        this(config, asyncHttpClient, mapper, HashMap::new);
+    }
+
+    /**
+     * IOC constructor.
+     *
+     * @param config  the configuration for this druid service
+     * @param asyncHttpClient  the HTTP client
+     * @param mapper  A shared jackson object mapper resource
+     * @param headersToAppend Supplier for map of headers for Druid requests
+     */
+    public AsyncDruidWebServiceImpl(
+            DruidServiceConfig config,
+            AsyncHttpClient asyncHttpClient,
+            ObjectMapper mapper,
+            Supplier<Map<String, String>> headersToAppend
+    ) {
         this.serviceConfig = config;
 
         if (serviceConfig.getUrl() == null) {
@@ -112,6 +160,7 @@ public class AsyncDruidWebServiceImpl implements DruidWebService {
         }
 
         LOG.info("Configured with druid server config: {}", config.toString());
+        this.headersToAppend = headersToAppend;
         this.webClient = asyncHttpClient;
         this.writer = mapper.writer();
         this.httpErrorMeter = REGISTRY.meter("druid.errors.http");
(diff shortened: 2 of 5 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -100,6 +100,7 @@
 import com.codahale.metrics.health.HealthCheckRegistry;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
+import org.apache.commons.collections.map.HashedMap;
```
cdeszaq: Can we use a built-in (ie. Java Collections) map instead of apache commons?

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -70,8 +72,10 @@
      * @param serviceConfig  Configuration for the Druid Service
      * @param mapper  A shared jackson object mapper resource
      */
-    public AsyncDruidWebServiceImpl(DruidServiceConfig serviceConfig, ObjectMapper mapper) {
-        this(serviceConfig, initializeWebClient(serviceConfig.getTimeout()), mapper);
+    public AsyncDruidWebServiceImpl(DruidServiceConfig serviceConfig, ObjectMapper mapper,
+            Supplier<Map<String, String>> authHeaders
```
cdeszaq: Wrapping param lists should "chop" the list, rather than just wrap it. eg.:

```java
public AsyncDruidWebServiceImpl(
        DruidServiceConfig serviceConfig, 
        ObjectMapper mapper,
        Supplier<Map<String, String>> authHeaders
) {
```

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -70,8 +72,10 @@
      * @param serviceConfig  Configuration for the Druid Service
      * @param mapper  A shared jackson object mapper resource
```
cdeszaq: Javadoc needed for the new parameter

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
cdeszaq: Can we re-name this? Rather than auth-specific, can this be named so that it can be _any_ headers?

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -102,7 +106,9 @@ private static AsyncHttpClient initializeWebClient(int requestTimeout) {
      * @param asyncHttpClient  the HTTP client
      * @param mapper  A shared jackson object mapper resource
      */
-    public AsyncDruidWebServiceImpl(DruidServiceConfig config, AsyncHttpClient asyncHttpClient, ObjectMapper mapper) {
+    public AsyncDruidWebServiceImpl(DruidServiceConfig config, AsyncHttpClient asyncHttpClient, ObjectMapper mapper,
+            Supplier<Map<String, String>> authHeaders
```
cdeszaq: Same comments re: wrapping and name apply here as well

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -61,7 +63,7 @@
     public static final String DRUID_WEIGHTED_QUERY_TIMER = DRUID_TIMER + "_W_";
     public static final String DRUID_SEGMENT_METADATA_TIMER = DRUID_TIMER + "_S_0";
 
-
+    private final Supplier<Map<String, String>> authHeaders;
```
cdeszaq: can we rename this to indicate _any_ headers?

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
cdeszaq: We should also have constructors that default the header supplier for backwards compatibility. We can deprecate them so that people are required to supply a supplier eventually, but we should try to maintain backwards compatibility if we can.

The supplier can just return an empty map in the defaulting case.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
codingwhatever: Addressed this in the constructor below.

codingwhatever:
@cdeszaq Addressed all the comments.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -72,10 +73,27 @@
      * @param serviceConfig  Configuration for the Druid Service
      * @param mapper  A shared jackson object mapper resource
      */
-    public AsyncDruidWebServiceImpl(DruidServiceConfig serviceConfig, ObjectMapper mapper,
-            Supplier<Map<String, String>> authHeaders
+    @Deprecated
```
archolewa: If you could add the `@deprecated` annotation to the JavaDoc as well, with a brief explanation of why this constructor is being deprecated, and which constructor should be used instead, I would be grateful.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -70,8 +73,27 @@
      * @param serviceConfig  Configuration for the Druid Service
      * @param mapper  A shared jackson object mapper resource
      */
-    public AsyncDruidWebServiceImpl(DruidServiceConfig serviceConfig, ObjectMapper mapper) {
-        this(serviceConfig, initializeWebClient(serviceConfig.getTimeout()), mapper);
+    @Deprecated
```
archolewa: The `@deprecated` annotation should also be added to the JavaDoc, with a brief explanation of why this is being deprecated, and a pointer to which constructor should be used instead.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -102,7 +124,29 @@ private static AsyncHttpClient initializeWebClient(int requestTimeout) {
      * @param asyncHttpClient  the HTTP client
      * @param mapper  A shared jackson object mapper resource
      */
-    public AsyncDruidWebServiceImpl(DruidServiceConfig config, AsyncHttpClient asyncHttpClient, ObjectMapper mapper) {
+    @Deprecated
```
archolewa: [ditto](https://github.com/yahoo/fili/pull/62/files#diff-292ba74271eb4c3e7ec871dbae87751cR76)

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
cdeszaq: For these deprecations, we should add an `@deprecated` tag to the javadoc indicating why we're deprecating them (because you should specify your header supplier), and what you should use instead (the matching constructor with that new parameter)

Applies to the other `@Deprecated` instance in this PR as well

archolewa:
This also needs a CHANGELOG entry.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -919,7 +924,24 @@ protected JobPayloadBuilder buildJobPayloadBuilder() {
      * @return A DruidWebService
      */
     protected DruidWebService buildDruidWebService(DruidServiceConfig druidServiceConfig, ObjectMapper mapper) {
-        return new AsyncDruidWebServiceImpl(druidServiceConfig, mapper, HashMap::new);
+        Supplier<Map<String, String>> supplier = HashMap::new;
+        String customSupplierClassString = SYSTEM_CONFIG.getStringProperty(DRUID_HEADER_SUPPLIER_CLASS, null);
```
cdeszaq: @michael-mclawhorn does this usage fit with how we usually use systemConfig?

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -919,7 +924,24 @@ protected JobPayloadBuilder buildJobPayloadBuilder() {
...
+        if (customSupplierClassString != null && customSupplierClassString != "") {
+            try {
+                Class<?> c = Class.forName(customSupplierClassString);
+                Constructor<?> constructor = c.getConstructor();
+                supplier = (Supplier<Map<String, String>>) constructor.newInstance();
+            } catch (ClassNotFoundException | NoSuchMethodException |
+                    IllegalAccessException | InstantiationException | InvocationTargetException e) {
```
cdeszaq: Rather than enumerate all of these, should we just catch `Exception`? If anything goes wrong loading this, we should just fail anyway, no matter what the exception is, right?

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -919,7 +924,24 @@ protected JobPayloadBuilder buildJobPayloadBuilder() {
...
+                LOG.error(
+                        "Unable to load the Druid query header supplier, className: {}, exception: {}",
+                        customSupplierClassString,
+                        e
+                );
+                throw new IllegalStateException(e);
+            }
+        }
```
cdeszaq: I think we should pull the `buildDruidWebServiceHeaderSupplier` code into it's own method. That will give another (fairly easy) override-based config mechanism if someone's `Supplier` needs more than a no-param constructor.

codingwhatever: I assumed we could just pull the other params from system_config in the supplier. I guess there could be more complicated use cases?

codingwhatever:
@archolewa @cdeszaq Addressed all review comments.

michael-mclawhorn:
I think Supplier<Map<String, String>> might be inferior as the type here to Iterable<Pair<String, String>>

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -61,17 +64,38 @@
     public static final String DRUID_WEIGHTED_QUERY_TIMER = DRUID_TIMER + "_W_";
     public static final String DRUID_SEGMENT_METADATA_TIMER = DRUID_TIMER + "_S_0";
 
-
+    private final Supplier<Map<String, String>> anyHeaders;
     private final DruidServiceConfig serviceConfig;
```
michael-mclawhorn: I dislike the name of the variable.  If I understand the scope of this, these are mandatory headers appended (or possibly prepended, it looks like the latter based on digging into the netty code) to the client request.  So possible 'requestAppendHeaders' or 'clientAppendHeaders'.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -267,14 +316,18 @@ public void postDruidQuery(
             timerName = DRUID_WEIGHTED_QUERY_TIMER + String.format(format, seqNum);
         }
 
+        BoundRequestBuilder requestBuilder = webClient.preparePost(serviceConfig.getUrl())
+                .setBody(entityBody)
+                .addHeader("Content-Type", "application/json");
+
+        anyHeaders.get().forEach((k, v) -> requestBuilder.addHeader(k, v));
+
```
michael-mclawhorn: It does not appear that headers are maplike in that there can't be more than one of a given key value.  Given this, you might be better served with an iterable here.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -937,7 +941,33 @@ protected JobPayloadBuilder buildJobPayloadBuilder() {
...
+    protected Supplier<Map<String, String>> buildDruidWebServiceHeaderSupplier() {
+        Supplier<Map<String, String>> supplier = HashMap::new;
+        String customSupplierClassString = SYSTEM_CONFIG.getStringProperty(DRUID_HEADER_SUPPLIER_CLASS, null);
+        if (customSupplierClassString != null && customSupplierClassString != "") {
+            try {
+                Class<?> c = Class.forName(customSupplierClassString);
+                Constructor<?> constructor = c.getConstructor();
+                supplier = (Supplier<Map<String, String>>) constructor.newInstance();
+            } catch (Exception e) {
+                LOG.error(
```
michael-mclawhorn: Can we catch something more specific than 'Exception' here?

cdeszaq: We _were_ matching something more specific in an earlier version of this PR and I suggested replacing that list of roughly 5 exceptions with just `Exception`. Since we're re-throwing the exception anyways, catching everything so that we can log it (since it'll be fatal anyways) is (I think) fine, and lets the code be smaller / cleaner.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
cdeszaq: Since we're adding the headers (and it looks like added headers strictly add, don't replace), I think a name like `headersToAdd` or `headersToAppend` helps clarify this.

michael-mclawhorn:
To summarize my responses, since I distributed them across Rick's responses to my comments:
The only change I am currently asking for is the anyHeaders variable to be named slightly differently.  Otherwise 👍  from me.

codingwhatever:
@cdeszaq Addressed remaining comments and added a unit test.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplSpec.groovy
```diff
@@ -0,0 +1,53 @@
+package com.yahoo.bard.webservice.druid.client.impl
```
cdeszaq: Copyright needed

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplSpec.groovy
```diff
@@ -0,0 +1,53 @@
...
+            Map<String, String> get() {
+                return expectedHeaders
+            }
+        }
+
+        AsyncDruidWebServiceImplWrapper webServiceImplWrapper = new AsyncDruidWebServiceImplWrapper(
+                DruidClientConfigHelper.getNonUiServiceConfig(),
+                new ObjectMapper(),
+                supplier
+        )
```
cdeszaq: All of this stuff from here up that's in the `when` block should go up in the `setup` block, since it's still setting up the scenario.

If needed, you can use `and:` blocks (with an optional comment: `and: "a set of expected headers to add"`) to break up any block.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplSpec.groovy
```diff
@@ -0,0 +1,53 @@
...
+                null,
+                null,
+                null,
+                weightEvaluationQuery
+        );
+
+        then:
+        HttpHeaders actualHeaders = webServiceImplWrapper.getHeaders()
+        for (Map.Entry<String, String> header : expectedHeaders) {
+            actualHeaders.get(header.getKey()) == header.getValue()
```
cdeszaq: Only top-level expressions in a `then` block are treated as assertions. To treat non-top-level expressions as assertions, you need to explicitly add the `assert` keyword: `assert actualHeaders.get(header.getKey()) == header.getValue()`

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -937,7 +941,33 @@ protected JobPayloadBuilder buildJobPayloadBuilder() {
...
+
+    /**
+     * Build the Supplier for Druid data request headers.
+     *
+     * @return The Druid data request header Supplier.
+     */
+    protected Supplier<Map<String, String>> buildDruidWebServiceHeaderSupplier() {
+        Supplier<Map<String, String>> supplier = HashMap::new;
+        String customSupplierClassString = SYSTEM_CONFIG.getStringProperty(DRUID_HEADER_SUPPLIER_CLASS, null);
+        if (customSupplierClassString != null && customSupplierClassString != "") {
```
cdeszaq: We should fix this too.

codingwhatever:
@cdeszaq Addressed the latest comments.

cdeszaq:
:+1: just needs to be squashed and rebased onto master and I think this is good to merge.

PR #64, opened 2016-10-13
Add in filter

archolewa:
We don't currently have a Filter for Druid's In Filter. I'd like to have one, so I've added it. 😄 

Note that we are not currently using it, though we should be able to use it to simplify the filter builders. 

I tried, but fixing the tests turned out to be taking more time than I can really afford right now. Especially if we want to make the tests readable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,12 @@ Current
 
 ### Added:
 
-- [Support for X-Request-ID](https://github.com/yahoo/fili/pull/68)
-- [Documentation for topN](https://github.com/yahoo/fili/pull/43)
+- [Support for Druid's In Filter](https://github.com/yahoo/fili/pull/64)
+    * The in-filter only works with Druid versions 0.9.0 and up.
 
+- [Support for X-Request-ID](https://github.com/yahoo/fili/pull/68)
+
+- [Documentation for topN](https://github.com/yahoo/fili/pull/43)
 
 ### Changed:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -140,6 +140,10 @@ Jobs resource. Here are the highlights of what's in this release:
 
 ### Changed:
 
+- Removed `physicalName` lookup for metrics in `TableUtils::getColumnNames` to remove spurious warnings
+     * Metrics are not mapped like dimensions are. Dimensions are aliased per physical table and metrics are aliazed per logical table.
+     * Logical metric is mapped with one or many physical metrics. Same look up logic for dimension and metrics doesn't make sense.
+
 #### Jobs:
 
 - [HashPreResponseStore moved to `test` root directory.](https://github.com/yahoo/fili/pull/39)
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -10,10 +10,16 @@ Current
 
 ### Added:
 
+- [Support for Druid's In Filter](https://github.com/yahoo/fili/pull/64)
```
cdeszaq: Should add a note that it only works with v0.9.1 (I think... I don't think it was there in earlier versions)

CHANGELOG.md
```diff
@@ -10,10 +10,16 @@ Current
...
+
 - [Documentation for topN](https://github.com/yahoo/fili/pull/43)
 
 ### Changed:
 
+- [Cleaned up the type for `DimensionalFilter::withDimension`](https://github.com/yahoo/fili/pull/64)
+    -- Previously, `DimensionalFilter::withDimension` had the signature `public abstract <T> T withDimension(Dimension dimension)`,
+    but `T` is already defined as part of the class type of `DimensionalFilter`. Therefore, `<T>` has been removed.
```
cdeszaq: Does that actually change anything? If not (ie. if it's the same as changing `List<Fool> f = new ArrayList<Foo>()` into `List<Fool> f = new ArrayList<>()`), then I wouldn't bother noting it in the change log.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/InFilter.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Objects;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * Fili model of the Druid In Filter: http://druid.io/docs/0.9.1.1/querying/filters.html.
+ * <p>
+ * An In Filter is a generalization of the {@link SelectorFilter}. Rather than filtering on a specific value for a
+ * specific diimension, the In Filter takes a list of values. A dimension satisfies the In Filter iff its value is
+ * contained in the specified list.
```
cdeszaq: May be worth mentioning that an `in` filter is the same as an `or` filter around a collection of `selector` filters. (I think that's right, right?)

cdeszaq: Also should mention here that this is only supported v0.9.1 and later.

CHANGELOG.md
archolewa: Looks like it is also in v0.9.0 (at least it's included in the docs...)

CHANGELOG.md
archolewa: No, I don't think it changes anything.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/InFilter.java
archolewa: Yes.

michael-mclawhorn:
👍 
I don't disagree with Rick's input, but this is techincally sound.

CHANGELOG.md
cdeszaq: If it's included, sweet. But that means what we should just indicate that it's available in v0.9.0 instead o v0.9.1 :-P

PR #68, opened 2016-10-14
Support passing X-Request-ID to Druid

DeathByTape:
A note from offline discussion, HTTP headers are case **in**sensitive. See [RFC 7230](https://tools.ietf.org/html/rfc7230) for more information.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
@@ -401,6 +401,19 @@ public class RequestLog {
     }
 
     /**
+     * Prepend an id prefix to generated druid query id.
+     *
+     * @param idPrefix  Prefix for queryId sent to druid
+     */
+    public static void addIdPrefix(String idPrefix) {
+        RequestLog current = RLOG.get();
+        String newId = idPrefix + getId();
+        current.info = current.info.withUuid(newId);
+        current.logId = newId;
+        MDC.put(ID_KEY, newId);
+    }
+
+    /**
      * Overwrite current thread's request log context.
      * It first clears the request log of the current thread and then fills it with the contents found in ctx
      *
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/BardLoggingFilter.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/BardLoggingFilter.java
@@ -26,6 +26,7 @@ import java.net.URI;
 import java.util.List;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 import javax.annotation.Priority;
 import javax.inject.Singleton;
@@ -68,10 +69,12 @@ public class BardLoggingFilter implements ContainerRequestFilter, ContainerRespo
     private static final String PROPERTY_NANOS = LoggingFilter.class.getName() + ".nanos";
     private static final String PROPERTY_REQ_LEN = LoggingFilter.class.getName() + ".reqlen";
     private static final String PROPERTY_OUTPUT_STREAM = LoggingFilter.class.getName() + ".ostream";
+    private static final Pattern VALID_REQUEST_ID = Pattern.compile("[a-zA-Z0-9+/=-]+");
 
     public static final double MILLISECONDS_PER_NANOSECOND = 1000000.0;
     public static final String TOTAL_TIMER = "TotalTime";
     public static final String CLIENT_TOTAL_TIMER = "ClientRequestTotalTime";
+    public static final String X_REQUEST_ID_HEADER = "x-request-id";
 
     /**
      * Intercept the Container request to add length of request and a start timestamp.
@@ -83,6 +86,7 @@ public class BardLoggingFilter implements ContainerRequestFilter, ContainerRespo
     @Override
     public void filter(ContainerRequestContext request) throws IOException {
 
+        appendRequestId(request.getHeaders().getFirst(X_REQUEST_ID_HEADER));
         RequestLog.startTiming(TOTAL_TIMER);
         RequestLog.startTiming(this);
         RequestLog.record(new Preface(request));
@@ -108,6 +112,7 @@ public class BardLoggingFilter implements ContainerRequestFilter, ContainerRespo
     @Override
     public void filter(ContainerRequestContext request, ContainerResponseContext response)
         throws IOException {
+        appendRequestId(request.getHeaders().getFirst(X_REQUEST_ID_HEADER));
         RequestLog.startTiming(this);
         StringBuilder debugMsgBuilder = new StringBuilder();
 
@@ -168,6 +173,7 @@ public class BardLoggingFilter implements ContainerRequestFilter, ContainerRespo
      */
     @Override
     public void filter(ClientRequestContext request) throws IOException {
+        appendRequestId(request.getStringHeaders().getFirst(X_REQUEST_ID_HEADER));
         RequestLog.startTiming(CLIENT_TOTAL_TIMER);
         request.setProperty(PROPERTY_NANOS, System.nanoTime());
     }
@@ -196,6 +202,30 @@ public class BardLoggingFilter implements ContainerRequestFilter, ContainerRespo
 
         LOG.debug(debugMsgBuilder.toString());
         RequestLog.stopTiming(CLIENT_TOTAL_TIMER);
+    }
+
+    /**
+     * Add a request id to pass to druid. Only added if x-request-id follows the format specified at:
+     * https://devcenter.heroku.com/articles/http-request-id
+     *
+     * @param requestId  Request id to add as queryId prefix to druid
+     */
+    private void appendRequestId(String requestId) {
+        if (isValidRequestId(requestId)) {
+            return; // Ignore according to https://devcenter.heroku.com/articles/http-request-id
+        }
+        RequestLog.addIdPrefix(requestId);
+    }
+
+    /**
+     * Validate whether or not a string is acceptable as an x-request-id header argument.
+     *
+     * @param requestId  Request id to validate
+     * @return True if valid, false otherwise
+     */
+    private boolean isValidRequestId(String requestId) {
+        return requestId == null || requestId.isEmpty() || requestId.length() > 200
+               || !VALID_REQUEST_ID.matcher(requestId).matches();
     }
 
     /**
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/BaseDataServletComponentSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/BaseDataServletComponentSpec.groovy
@@ -21,6 +21,7 @@ import org.slf4j.LoggerFactory
 import spock.lang.Specification
 import spock.lang.Timeout
 
+import javax.ws.rs.core.MultivaluedHashMap
 import javax.ws.rs.core.MultivaluedMap
 import javax.ws.rs.core.Response
 
@@ -69,6 +70,13 @@ abstract class BaseDataServletComponentSpec extends Specification {
         jtb = buildTestBinder()
 
         populatePhysicalTableAvailability()
+    }
+
+    /**
+     * Used to append headers to API request made by this spec.
+     */
+    MultivaluedHashMap<String, String> getAdditionalApiRequestHeaders() {
+        return [:]
     }
 
     /**
@@ -206,7 +214,7 @@ abstract class BaseDataServletComponentSpec extends Specification {
         }
 
         // Make the call
-        Response response = httpCall.request().get()
+        Response response = httpCall.request().headers(getAdditionalApiRequestHeaders()).get()
         if (response.status != 200) {
             LOG.error( "***  *** Response status: ${response.status}: ${response.readEntity(String)}")
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Current
 
 ### Added:
 
+- [Support for X-Request-ID](https://github.com/yahoo/fili/pull/68)
 - [Documentation for topN](https://github.com/yahoo/fili/pull/43)
 
 
@@ -57,6 +58,8 @@ Current
 
 
 ### Fixed:
+
+- [Fixed typo emit -> omit in Utils.java omitField()](https://github.com/yahoo/fili/pull/68)
 
 - [Adds read locking to all attempts to read the Lucene index](https://github.com/yahoo/fili/pull/52)
   * Before, if Fili attempted to read from the Lucene indices (i.e. processing a query with filters)
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogBlock.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogBlock.java
@@ -37,6 +37,18 @@ public class LogBlock {
     }
 
     /**
+     * Create a new LogBlock with the same body but updated uuid.
+     *
+     * @param uuid  New uuid.
+     * @return New LogBlock with updated uuid.
+     */
+    public LogBlock withUuid(String uuid) {
+        LogBlock newLogBlock = new LogBlock(uuid);
+        body.entrySet().forEach(entry -> newLogBlock.body.put(entry.getKey(), entry.getValue()));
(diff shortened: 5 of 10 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
```diff
@@ -92,6 +94,7 @@ private RequestLog(RequestLog rl) {
         times = new LinkedHashMap<>(rl.times);
         threadIds = new LinkedHashSet<>(rl.threadIds);
         MDC.put(ID_KEY, logId);
+        idPrefix = "";
```
archolewa: This should be set to the `idPrefix` on the passed-in RequestLog.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
```diff
@@ -397,7 +402,20 @@ public static String getId() {
...
+     * Set id prefix.
+     *
+     * @param idPrefix  Prefix for queryId sent to druid
+     */
+    public static void setIdPrefix(String idPrefix) {
+        RequestLog current = RLOG.get();
+        if (current.info == null) {
+            current.init();
+        }
+        current.idPrefix = (idPrefix == null) ? "" : idPrefix;
```
archolewa: This isn't enough. If you look at line [157](https://github.com/yahoo/fili/pull/68/files#diff-79f5168b37854c3172bc2994b14d3168R157), you'll see that we are putting the id into `MDC`. That `MDC` is what allows us to put the id into every log line generated by a request.

As a result, if we want the prefix to _also_ show up in every log line (we do), then we need to make sure the `logId` and the `idPrefix` are added to `MDC`. 

So at the minimum, in the last line in this method, we should add the following:

```java
MDPC.put(ID_KEY, getId())
```

Or manually do the string concatentation if calling the getter in the setter gives you heebie-jeebies.

Although in all honesty, I feel like that opens the door to subtle bugs, if `init` gets called later and overwrites `MDC` with the prefix. 

Instead, I think that `setIdPrefix` should prepend the id to `logId`:

```java
current.logId = idPrefix + current.logId;
MDC.put(ID_KEY, current.logId);
```

That way, if `init` gets called again at some other random point, we don't have to worry about the prefix being dropped.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/DruidQueryIdSpec.groovy
```diff
@@ -0,0 +1,95 @@
...
+            addDimensionRow(BardDimensionField.makeDimensionRow(it, "Foo", "FooDesc"))
+            addDimensionRow(BardDimensionField.makeDimensionRow(it, "Bar", "BarDesc"))
+            addDimensionRow(BardDimensionField.makeDimensionRow(it, "Baz", "BazDesc"))
+        }
+    }
+
+    @Override
+    Class<?>[] getResourceClasses() {
+        [ DataServlet.class
+        , BardLoggingFilter.class ]
```
archolewa: I would put this on one line. We don't generally wrap unless we have to.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/DruidQueryIdSpec.groovy
```diff
@@ -0,0 +1,95 @@
...
+    @Override
+    String getTarget() {
+        return "data/shapes/week/color"
+    }
+
+    @Override
+    Map<String, List<String>> getQueryParams() {
+        [
+                "metrics": ["width","depth"],
+                "dateTime": ["2014-06-02%2F2014-06-30"],
```
archolewa: The date doesn't have to be URL-encoded.

/NotABlocker

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/DruidQueryIdSpec.groovy
```diff
@@ -0,0 +1,95 @@
...
+                "sort": ["width|desc","depth|asc"]
+        ]
+    }
+
+    @Override
+    boolean compareResult(String result, String expectedResult, JsonSortStrategy sortStrategy = JsonSortStrategy.SORT_MAPS) {
+        if (!testedDruidQuery) {
```
archolewa: What is this and why is it necessary?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/DruidQueryIdSpec.groovy
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @Override
+    String getExpectedApiResponse() {
+        // Noop
+        """{}"""
+    }
+
+    @Override
+    Response makeAbstractRequest(Closure queryParams=this.&getQueryParams) {
```
archolewa: I would appreciate a comment indicating that the only difference between this version of `makeAbstractRequest` and the parent version is that we are adding the `X-Request-ID` to the header. Otherwise, the two implementations are so similar that people may think this implementation is unnecessary, delete it, and then wonder why the test started failing.

Though a better approach may be to add a hook to `makeAbstractRequest`. Basically, add a method `getAdditionalHeaders` that returns headers to add to the request before making the call. We can then have `BaseDataServletComponentSpec::makeAbstractRequest` add the headers from `getAdditionalHeaders` to the request just before making the call.

archolewa:
One other note: If you could remember to add PR labels when you open the PR (`Reviewable` and `Need 2 Reviews` in particular), it would be much appreciated. The labels make it a lot easier for us to determine the sttaus of the various PR's at a glance.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/DruidQueryIdSpec.groovy
DeathByTape: copy/pasta'd :) can unencode if we like

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/DruidQueryIdSpec.groovy
DeathByTape: I like the latter solution!

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/DruidQueryIdSpec.groovy
archolewa: I see. We use it compare both the API and the Druid response. How about, instead of using a boolean, we only do the comparision of the expected result contains the context field?

The API response doesn't contain any context mapping.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/DruidQueryIdSpec.groovy
```diff
@@ -0,0 +1,81 @@
...
+import com.yahoo.bard.webservice.util.JsonSlurper
+import com.yahoo.bard.webservice.util.JsonSortStrategy
+import com.yahoo.bard.webservice.web.filters.BardLoggingFilter
+
+import javax.ws.rs.core.MultivaluedHashMap
+
+
+class DruidQueryIdSpec extends BaseDataServletComponentSpec {
+    def prefixId = "abcdef"
+    static def testedDruidQuery = false
```
archolewa: This field is no longer necessary, and can be removed.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
```diff
@@ -401,6 +401,18 @@ public static String getId() {
...
     /**
+     * Set id prefix.
+     *
+     * @param idPrefix  Prefix for queryId sent to druid
+     */
+    public static void addIdPrefix(String idPrefix) {
+        RequestLog current = RLOG.get();
+        String newId = idPrefix + getId();
+        current.logId = newId;
+        MDC.put(ID_KEY, newId);
```
cdeszaq: While this will update the `current.logId` field (and the MDC value), I _don't_ think it will update the UUID that is part of the `RequestLog` payload. I _think_ that the `LogInfo.uuid` field (on the object stored in `current.info`) needs to be updated as well in order for that to happen.

I think the reason this was missed is because we're holding that UUID in 2 places `current.logId` (we read from that all over), and in the `LogInfo.uuid` field (`current.info`). It would be _great_ if we could get rid of the `current.logId` field and only use the value from the `LogInfo` object. This refactor isn't required for this PR, but it would be great if we could pull it in sooner, rather than later.

As part of the change around where the logId is accessed from, we should update `getId` to pull from the single source of truth as well.

Also, since this is the closest I can get to it, it would be cool (if we do this refactoring) to move the `MDC.put` call in the `init` method up much higher, so that there's less of a gap between getting a UUID and setting it into MCD.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
```diff
@@ -401,6 +401,18 @@ public static String getId() {
     }
 
     /**
+     * Set id prefix.
```
cdeszaq: Since it's no longer setting, but is instead appending (so subsequent calls will keep appending), this JavaDoc should make that more clear.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/BaseDataServletComponentSpec.groovy
```diff
@@ -71,6 +72,10 @@ abstract class BaseDataServletComponentSpec extends Specification {
         populatePhysicalTableAvailability()
     }
 
+    MultivaluedHashMap<String, String> getAdditionalHeaders() {
```
cdeszaq: It's not clear what these headers are for without tracing how this is used. Perhaps we should expand the method name to something like `getAdditionalApiRequestHeaders`? A JavaDoc would also help, but not critical if the method name is clearer.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/DruidQueryIdSpec.groovy
```diff
@@ -0,0 +1,80 @@
...
+import com.yahoo.bard.webservice.data.dimension.BardDimensionField
+import com.yahoo.bard.webservice.data.dimension.DimensionDictionary
+import com.yahoo.bard.webservice.util.JsonSlurper
+import com.yahoo.bard.webservice.util.JsonSortStrategy
+import com.yahoo.bard.webservice.web.filters.BardLoggingFilter
+
+import javax.ws.rs.core.MultivaluedHashMap
+
+
+class DruidQueryIdSpec extends BaseDataServletComponentSpec {
```
cdeszaq: It would be good to have this name reflect something about looking for the request ID header to be connected to the druid query id. Perhaps something like `RequestIdPrefixesDruidQueryIdSpec`?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/DruidQueryIdSpec.groovy
```diff
@@ -0,0 +1,80 @@
...
+                "sort": ["width|desc","depth|asc"]
+        ]
+    }
+
+    @Override
+    boolean compareResult(String result, String expectedResult, JsonSortStrategy sortStrategy = JsonSortStrategy.SORT_MAPS) {
+        JsonSlurper slurper = new JsonSlurper(sortStrategy)
+        def parsedJson = slurper.parseText(result)
+        if (parsedJson.context != null) {
+            testedDruidQuery = true
```
cdeszaq: This variable isn't a thing I don't think...

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/DruidQueryIdSpec.groovy
```diff
@@ -0,0 +1,80 @@
...
+            return parsedJson.context.queryId.startsWith(prefixId)
+        }
```
cdeszaq: I think this all can be simplified using Groovy's [safe navigation operator](http://groovy-lang.org/operators.html#_safe_navigation_operator):

```groovy
def parsedQuery = new JsonSlurper(sortStrategy).parseText(result)
assert parsedQuery?.context?.queryId?.startsWith(prefixId)
```

fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/BardLoggingFilter.java
```diff
@@ -83,6 +84,7 @@
     @Override
     public void filter(ContainerRequestContext request) throws IOException {
 
+        RequestLog.addIdPrefix(request.getHeaders().getFirst(X_REQUEST_ID_HEADER));
```
DeathByTape: What are allowed chars for this header? Are we going to use something like heroku's definition? https://devcenter.heroku.com/articles/http-request-id

I'll go with that for now.

cdeszaq: Yeah, that'll work, though it would be good to add periods and underscores (`.` and `_`)

cdeszaq: And I would say that min-length doesn't matter, and 200 chars for max length seems sane. Stripping any unallowed character (and then truncating) is probably the most permissive approach too

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/DruidQueryIdSpec.groovy
DeathByTape: This didn't actually work. This fails in the case where we want this function to act as a noop.

fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/BardLoggingFilter.java
```diff
@@ -199,6 +205,20 @@ public void filter(ClientRequestContext request, ClientResponseContext response)
     }
 
     /**
+     * Add a request id to pass to druid. Only added if x-request-id follows the format specified at:
+     * https://devcenter.heroku.com/articles/http-request-id
+     *
+     * @param requestId  Request id to add as queryId prefix to druid
+     */
+    private void appendRequestId(String requestId) {
```
cdeszaq: It would be good to have a test for this, validating the rules.

cdeszaq:
👍

fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/BardLoggingFilter.java
```diff
@@ -199,6 +205,30 @@ public void filter(ClientRequestContext request, ClientResponseContext response)
...
+
+    /**
+     * Validate whether or not a string is acceptable as an x-request-id header argument.
+     *
+     * @param requestId  Request id to validate
+     * @return True if valid, false otherwise
+     */
+    private boolean isValidRequestId(String requestId) {
+        return requestId == null || requestId.isEmpty() || requestId.length() > 200
+               || !VALID_REQUEST_ID.matcher(requestId).matches();
```
archolewa: It looks like this method is actually returning the opposite of what it claims. It's returning `true` if the requestId is _not_ valid, and `false` if it is valid.

fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/BardLoggingFilter.java
```diff
@@ -199,6 +205,30 @@ public void filter(ClientRequestContext request, ClientResponseContext response)
     }
 
     /**
+     * Add a request id to pass to druid. Only added if x-request-id follows the format specified at:
+     * https://devcenter.heroku.com/articles/http-request-id
+     *
+     * @param requestId  Request id to add as queryId prefix to druid
+     */
+    private void appendRequestId(String requestId) {
+        if (isValidRequestId(requestId)) {
```
cdeszaq: Good catch. I think the _behavior_ is correct (yay testing) the name is just wrong. @DeathByTape, could you open a quick PR to correct this?

PR #69, opened 2016-10-17
Fixes BardLoggingFilter::isValidRequestId

archolewa:
-- It actually returns the opposite of what it claims. So the name and
JavaDoc have been modified to more accurately reflect that.


cdeszaq:
👍

PR #70, opened 2016-10-19
ResponseException stores the JSON sent to Druid.

archolewa:
-- `DruidAggregationQuery` does not have `toString` defined.
Furthermore, it would be non-trivial to define a `toString` that
serializes the query into the actual query sent to Druid (we would need
to inject a `ObjectMapper` from the `ObjectMapperSuite` into the
aggregation query). Therefore, the `ResponseException` now takes an
`ObjectWriter` that can be used to serialize the druid query that is
logged.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/async/ResponseException.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/async/ResponseException.java
@@ -5,6 +5,13 @@ package com.yahoo.bard.webservice.async;
 import com.yahoo.bard.webservice.druid.model.query.DruidAggregationQuery;
 import com.yahoo.bard.webservice.druid.model.query.DruidQuery;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectWriter;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import javax.ws.rs.core.Response;
 
 /**
@@ -12,6 +19,7 @@ import javax.ws.rs.core.Response;
  * type is enforced as Exception.
  */
 public class ResponseException extends Exception {
+    private static final Logger LOG = LoggerFactory.getLogger(ResponseException.class);
 
     private final int statusCode;
     private final String reason;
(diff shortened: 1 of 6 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/async/ResponseException.java
```diff
@@ -37,6 +45,29 @@ public ResponseException(Response.StatusType statusType, DruidAggregationQuery<?
     }
 
     /**
+     * Class constructor with throwable, other parameters and a mapper for serializing the druid query.
+     * @param statusType  Status type of the response
```
cdeszaq: Space needed

fili-core/src/main/java/com/yahoo/bard/webservice/async/ResponseException.java
```diff
@@ -64,7 +95,29 @@ public ResponseException(
             DruidQuery<?> druidQuery,
             Throwable cause
     ) {
-        super(buildMessage(reason, description, statusCode, druidQuery, cause));
+        this(statusCode, reason, description, druidQuery, cause, new ObjectMapper().writer());
+    }
```
cdeszaq: We should deprecate this constructor, since creating a new "naked" mapper w/o the extensions we have on the primary ones will result in a funky serialization.

cdeszaq: And by that token, probably deprecate other constructors to this thing that don't take in the writer they'll need as well.

cdeszaq: Oh, and also not increase the number of deprecation warnings the compiler throws at us if we can 😉

fili-core/src/main/java/com/yahoo/bard/webservice/async/ResponseException.java
```diff
@@ -88,14 +152,22 @@ private static String buildMessage(
...
-            Throwable cause
+            Throwable cause,
+            ObjectWriter objectWriter
     ) {
+        String druidQueryString;
+        try {
+            druidQueryString = objectWriter.writeValueAsString(druidQuery);
+        } catch (JsonProcessingException e) {
+            druidQueryString = druidQuery.toString();
+            LOG.warn(String.format("Failed to properly serialize druid query %s", druidQueryString), e);
```
cdeszaq: Flip these. Otherwise, if the `toString()` blows up, we never see the next log message

archolewa: If I do that, then the best we can say is "Failed to properly serialize a druid query." Which admittedly isn't going to be that different from what will be displayed now, since we don't override `toString`.

cdeszaq: All, makes sense. With that in mind, you can actually just in-line `druidQuery` in the `LOG` call, and let the logging framework call `toString` if it decides it needs to actually log it.

DeathByTape:
👍

PR #75, opened 2016-10-21
Disable query split for "all" grain

mangesh-pardeshi:
Query split breaks for "all" grain when multiple disjoint intervals specified. Disable split in general for "all" grain

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,11 @@ Current
 
 ### Fixed:
 
+- [Disable Query split for "all" grain ](https:://github.com/yahoo/fili/pull/75)
+    - Before, if we requested "all" grain with multiple intervals, the `SplitQueryRequestHandler` would incorrectly split the query 
+      and we would get multiple buckets in the output. Now, the query split is disabled for "all" grain and we correctly
+      get only one bucket in the response.
+
 - [Fixed typo emit -> omit in Utils.java omitField()](https://github.com/yahoo/fili/pull/68)
 
 - [Adds read locking to all attempts to read the Lucene index](https://github.com/yahoo/fili/pull/52)
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/SplitQueryRequestHandler.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/SplitQueryRequestHandler.java
@@ -5,6 +5,7 @@ package com.yahoo.bard.webservice.web.handlers;
 import static com.yahoo.bard.webservice.web.ErrorMessageFormat.EMPTY_INTERVAL_FORMAT;
 
 import com.yahoo.bard.webservice.application.MetricRegistryFactory;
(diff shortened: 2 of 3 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/SplitQueryRequestHandler.java
```diff
@@ -63,6 +64,11 @@ public boolean handleRequest(
         Granularity granularity = druidQuery.getGranularity();
         List<Interval> queryIntervals = druidQuery.getIntervals();
 
+        if (granularity instanceof AllGranularity) {
+           // by-pass splitting for "all" grain
```
cdeszaq: It would be good to include _why_ we can't split for "all" grain. (ie. because there's only 1 response bucket, and so we can't split it)

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/handlers/SplitQueryRequestHandlerSpec.groovy
```diff
@@ -67,6 +69,28 @@ class SplitQueryRequestHandlerSpec extends Specification {
...
+        then:
+        (0) * groupByQuery.withAllIntervals(_) >> groupByQuerySplit
+        (1) * next.handleRequest(rc, apiRequest, groupByQuery, response)
+        
+        where:
+        timeGrain | interval
+        all       | week
+        all       | month
+        all       | month
+        all       | year
```
cdeszaq: Since the error we observed was that it was only incorrectly splitting for a disjoint interval, could we add a case that tests that condition as well?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/handlers/SplitQueryRequestHandlerSpec.groovy
```diff
@@ -67,6 +69,28 @@ class SplitQueryRequestHandlerSpec extends Specification {
         1         | MONTH     | month
         12        | MONTH     | year
     }
+    
+    def "Handler skips splitting for all time grain"() {
```
cdeszaq: Since we're using a `where` block, we should add the `@Unroll` annotation to the test so that each entry in the `where` block ends up being it's own test. The advantage of this is that we'll get info for each of the test failures, rather than just 1 entry for any of the cases failing.

As part of this, it would be good to add the interval attribute to the test name so that it shows up in the test report. You can do that by changing the name of the test to something like `Handler skips splitting for all time grain when the interval is #interval`

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ Current
 
 ### Changed:
 
+- [Disable Query split for "all" grain ](https:://github.com/yahoo/fili/pull/75)
```
cdeszaq: This should really go under the `### Fixed` category, since it's fixing a bug.

michael-mclawhorn:
👍

dayamr:
👍

PR #76, opened 2016-10-24
Change csv attachment name to use '__' instead of ','

mangesh-pardeshi:
Chrome does not accept ',' as part of the filename so changing it to use '__' instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ Current
 
 ### Changed:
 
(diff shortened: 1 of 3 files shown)


cdeszaq:
👍 Looks fine to me. 

@codingwhatever, you guys pushed for having better download file names initially, right? Since this is a fix to otherwise broken functionality, I'm guessing this change is fine, but I wanted to ping you just in case 😉

archolewa:
👍

codingwhatever:
👍 This should be fine

PR #78, opened 2016-10-27
Adds the request id to response processing errors.

archolewa:
--Now, whenever we return an error message from the post-processing
pipeline, we also include the request's UUID. This will make it a lot
easier to trace the query through the logs when users send us the error
message.


fili-core/src/test/groovy/com/yahoo/bard/webservice/util/GroovyTestUtils.groovy
```diff
@@ -190,4 +192,46 @@ abstract class GroovyTestUtils extends Specification {
...
+     * contain the queryId key, if the entry's key is "context."
+     *
+     * @param entry  The entry to be transformed
+     *
+     * @return  Either a shallow copy of the entry, or a version of the entry where its value does not contain the
+     * "queryId" key.
+     */
+    private static Map stripQueryId(def entry) {
+        return [(entry.key): entry.key == "context" ? entry.value.findAll{ it.key != "queryId"} : entry.value]
+    }
```
cdeszaq: There's also the `Utils::omitField` method (which is called from roughly line 43 of this class). Does that do the same thing this one's doing? They feel like they might be similar?

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/MappingResponseProcessor.java
```diff
@@ -107,14 +110,32 @@ public HttpErrorCallback getStandardError(
...
+                try {
+                    druidQueryString = druidQuerySerializer.writeValueAsString(druidQuery);
+                } catch (JsonProcessingException jse) {
+                    try {
+                        druidQueryString = druidQuery.toString();
+                    } catch (Exception e) {
+                        LOG.warn("Error invoking a druid query's toString.", e);
+                        druidQueryString = "QUERY'S `toString` FAILED";
+                    }
+                    LOG.warn(String.format("Failed to serialize druid query %s", druidQueryString), jse);
```
archolewa: This isn't supposed to be here. I think I noticed that we weren't serializing the druid query in this log message, and added code to do so, but then was all "Meh, not worth another PR, thanks to the query id showing up in every log message." I must not have checked out a fresh master before adding the query id to the errror message.

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/GroovyTestUtils.groovy
archolewa: I forgot about that class.

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/GroovyTestUtils.groovy
```diff
@@ -190,4 +192,46 @@ abstract class GroovyTestUtils extends Specification {
...
+     * A special-purpose operation that verifies that an error response is correct.
+     * <p>
+     * This function expects JSON Strings. It filters out both context from the Druid query contained in the error
+     * message (because it contains an untestable UUID for the query) and it only verifies that the queryId is present,
+     * since that is also an untestable UUID.
+     *
+     * @param errorBody  The error generated by the test
+     * @param expectedBody  The expected error body
+     */
+    static boolean compareErrorPayload(String errorBody, String expectedBody) {
```
QubitPi: `boolean` is not documented

cdeszaq: Nice catch!

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/MappingResponseProcessor.java
michael-mclawhorn: Is this going to get cleaned up somehow? Or is it already cleaned up and the comments haven't noticed it?

cdeszaq: @michael-mclawhorn as far as I can tell, @archolewa has not pushed any new commits to this PR since this comment was left.

archolewa: I haven't had a chance to work on it these past few days.

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/GroovyTestUtils.groovy
archolewa: Urg, it looks like `omitField` uses an `ObjectMapper` . I don't want to use an `ObjectMapper`. Maps are so much nicer in a scripting language. And I don't really want to rewrite this entire method to use the (in my opinion rather clunky) `JsonNode` interface.

cdeszaq: Duplicated code is the devil...

cdeszaq: ok, I'm sufficiently convinced

archolewa: 😲  I can't believe an argument that stemmed primarily from laziness actually convinced you. 

Be honest with me. It's because you like how groovy that code there is (well, except for the for-loop maybe), and you know that if I rewrite it to use an `ObjectMapper` it'll look all Javay and gross.

cdeszaq: Mostly b/c I figured a response of "I agree, yours is better, you should re-write everywhere we're using the old one to use the new one" would fall flat :)

archolewa: Yes.

michael-mclawhorn:
👍

PR #79, opened 2016-10-27
PhysicalTable uses `getAvailableIntervals`

archolewa:
--`PhysicalTable::getAvailableIntervals` returns
`availableIntervalsRef.get()`. However, the other methods didn't actually
use `getAvailableIntervals` internally. Instead, they used
`availableIntervalsRef.get()` directly. As a result, overriding
`getAvailableIntervals` would not actually be sufficient to restrict a
`PhysicalTable`'s view of its availability, despite the method name.

--Now, all invocations of `availableIntervalsRef.get()` have been
replaced with `getAvailableIntervals`, making it much easier to enforce
a consistent view of availability simply by overriding
`getAvailableIntervals`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ Current
 
 ### Changed:
 
+- [`PhysicalTable` now uses `getAvailableIntervals` internally rather than directly referencing its intervals](https://github.com/yahoo/fili/pull/79)
+
 - [CSV attachment name for multi-interval request now contain '__' instead of ','](https://github.com/yahoo/fili/pull/76)
     - This change is made to allow running multi-api request with csv format using chrome browser.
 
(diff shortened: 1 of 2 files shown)


cdeszaq:
👍  (though a changelog entry would be nice, since this _could_ break someone if they were relying on internal things _not_ calling that getter

archolewa:
@cdeszaq The only way I can think of that happening would be if someone is overriding and changing `getAvailableIntervals` not to affect Fili's behavior (since we don't actually use it anywhere but tests), but solely for use in their own code.

michael-mclawhorn:
👍 
I would actually regard this as safe not to CL, although this is a perfect candidate for establishing guidance around what is and isn't a public api.

cdeszaq:
My 1st thought was that this was below the level of noting in the CL, but since PhysicalTable is one of the Core Domains within the system, and it seems that the thrust of the PR is to make it _easier_ to extend, it made sense to err on the side of being overly communicative in the CL about it than surprise possible extenders.

PR #80, opened 2016-10-28
Give more time for ClassScannerSpec tests to run

michael-mclawhorn:
The tests run too slowly for this timeout in an underprovisioned VM.  Given that they currently pass this timeout in all our existing pipelines, this should be a no op for well provisioned systems.


cdeszaq:
👍

archolewa:
👍

Issue #74, opened 2016-10-20
Customize Logging Format


archolewa:
Currently, Fili exports the log information accumulated by `RequestLog` as a JSON. However, different clients may wish to use different formats for that information. Therefore, we should add an injection point that allows customers to provide their own output format.

Here's a brief sketch of how we could do that:

a. Add a LogFormat interface with a single method: `format` that takes in the `RequestLog` and returns a formatted `String` to be logged.
b. Add a configuration parameter that takes the fully-qualified classname of the formatter.
c. Provide a default implementation that formats the `RequestLog` into JSON.
c. Update `RequestLog` to pull in the `LogFormat` implementation at initialization time. Modify `RequestLog::export` to  pass the log information through that formatter.

cdeszaq:
This looks good to me. 👍

archolewa:
This has been implemented.

PR #81, opened 2016-10-28
Implement log formatter

QubitPi:
Issue https://github.com/yahoo/fili/issues/74

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Current
 
 ### Added:
 
-- [Customize Logging Format](https://github.com/yahoo/fili/pull/81)
+- [Customize Logging Format in RequestLog](https://github.com/yahoo/fili/pull/81)
 
 - [Support for Druid's In Filter](https://github.com/yahoo/fili/pull/64)
     * The in-filter only works with Druid versions 0.9.0 and up.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ Current
 
 ### Added:
 
+- [Customize Logging Format](https://github.com/yahoo/fili/pull/81)
+
 - [Support for Druid's In Filter](https://github.com/yahoo/fili/pull/64)
     * The in-filter only works with Druid versions 0.9.0 and up.
 
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/logging/JsonLogFormatter.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/logging/JsonLogFormatter.java
@@ -1,0 +1,41 @@
+// Copyright 2016 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.logging;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import com.fasterxml.jackson.databind.ser.std.ToStringSerializer;
+import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
+import com.fasterxml.jackson.datatype.joda.JodaModule;
+import org.joda.time.Interval;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A log formatter that prints log in JSON format.
+ */
+public class JsonLogFormatter implements LogFormatter {
+    private static final Logger LOG = LoggerFactory.getLogger(RequestLog.class);
+    private final ObjectMapper objectMapper;
+
+    /**
+     *  Configure ObjectMapper.
+     */
+    public JsonLogFormatter() {
+        objectMapper = new ObjectMapper();
+        objectMapper.registerModule((new JodaModule()).addSerializer(Interval.class, new ToStringSerializer()));
+        objectMapper.registerModule(new Jdk8Module().configureAbsentsAsNulls(false));
+    }
+
+    @Override
+    public String format(LogBlock logBlock) {
+        try {
+            return objectMapper.writeValueAsString(logBlock);
+        } catch (JsonProcessingException jsonProcessingException) {
+            String message = String.format("Exporting mega log line: '%s' to JSON failed.", logBlock);
+            LOG.warn(message, jsonProcessingException);
+            return logBlock.toString();
+        }
+    }
+}
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatter.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatter.java
@@ -1,0 +1,18 @@
+// Copyright 2016 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.logging;
+
+/**
+ * Interface for log formatter that are meant to facilitate log exploration and data extraction.
+ */
+public interface LogFormatter {
+
+    /**
+     * Takes in a log block and returns a formatted String to be logged.
+     *
+     * @param logBlock  the log block that represents a log and that will be formatted in to the needed format.
+     *
+     * @return a formatted String to be logged
+     */
+    String format(LogBlock logBlock);
+}
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
@@ -34,14 +34,12 @@ public class LogFormatterProvider {
      */
     public static LogFormatter getInstance() {
         if (LOG_FORMATTER == null) {
-            String logFormatterImplementation = System.getenv(LOG_FORMATTER_IMPLEMENTATION_KEY);
             try {
-                if (logFormatterImplementation == null) {
-                    logFormatterImplementation = SystemConfigProvider.getInstance().getStringProperty(
-                            SYSTEM_CONFIG.getPackageVariableName(LOG_FORMATTER_IMPLEMENTATION_KEY),
-                            DEFAULT_LOG_FORMATTER_IMPL
-                    );
-                }
+                String logFormatterImplementation = SystemConfigProvider.getInstance().getStringProperty(
+                        SYSTEM_CONFIG.getPackageVariableName(LOG_FORMATTER_IMPLEMENTATION_KEY),
+                        DEFAULT_LOG_FORMATTER_IMPL
+                );
+
                 return (LogFormatter) Class.forName(logFormatterImplementation).newInstance();
             } catch (Exception exception) {
                 LOG.error("Exception while loading Log formatter: {}", exception);
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
@@ -45,7 +45,7 @@ public class LogFormatterProvider {
                 return (LogFormatter) Class.forName(logFormatterImplementation).newInstance();
             } catch (Exception exception) {
                 LOG.error("Exception while loading Log formatter: {}", exception);
-                throw new IllegalStateException(exception.getMessage());
+                throw new IllegalStateException(exception);
             }
         }
 
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
@@ -1,0 +1,54 @@
+// Copyright 2016 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.logging;
+
+import com.yahoo.bard.webservice.config.SystemConfig;
+import com.yahoo.bard.webservice.config.SystemConfigProvider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A class to provide a log formatter instance.
+ */
+public class LogFormatterProvider {
+
+    private static final Logger LOG = LoggerFactory.getLogger(LogFormatterProvider.class);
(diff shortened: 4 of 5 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/logging/JsonLogFormatter.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A log formatter that prints log in JSON format.
+ */
+public class JsonLogFormatter implements LogFormatter {
+    private static final Logger LOG = LoggerFactory.getLogger(RequestLog.class);
+    private final ObjectMapper mapper = new ObjectMapper();
```
archolewa: This `ObjectMapper` should be configured with the [same modules](https://github.com/yahoo/fili/pull/81/files#diff-79f5168b37854c3172bc2994b14d3168R77) as the original mapper in `RequestLog` (which incidentally also needs to be removed).

fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatter.java
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Interface for log formatter that are meant to facilitate log exploration and data extraction.
+ */
+public interface LogFormatter {
+
+    /**
+     * Takes in a log block and returns a formatted String to be logged.
+     *
+     * @param logBlock the log block that represents a log and that will be formatted in to the needed format.
```
archolewa: Should have two spaces between parameter name and description.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A class to provide a log formatter instance.
+ */
+public class LogFormatterProvider {
+
+    private static final Logger LOG = LoggerFactory.getLogger(SystemConfigProvider.class);
```
archolewa: Incorrect class (should be `LogFormatterProvider.class` not `SystemConfigProvider.class`

fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private static final String LOG_FORMATTER_IMPL_KEY = "fili__log_formatter_impl";
```
archolewa: This shouldn't be prefixed with `fili`. We should be getting the module name from the `SystemConfig`.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
```diff
@@ -0,0 +1,46 @@
...
+     *  @return an instance of LogFormatter
+     */
+    public static LogFormatter getInstance() {
+        if (LOG_FORMATTER != null) {
+            return LOG_FORMATTER;
+        }
+        String logFormatterImplementation = System.getenv(LOG_FORMATTER_IMPL_KEY);
+        try {
+            if (logFormatterImplementation == null) {
+                logFormatterImplementation = System.getProperty(LOG_FORMATTER_IMPL_KEY, DEFAULT_LOG_FORMATTER_IMPL);
```
archolewa: I think here you should actually be using an instance of `SystemConfig`:

```java
logFormatterImplementation = SystemConfigProvider.getInstance().getStringProperty(
        SYSTEM_CONFIG.getPackageVariableName(LOG_FORMATTER_IMPL_KEY),
        DEFAULT_LOG_FORMATTER_IMPL
);
```

Also, I _think_ line `35` can be removed, and we can drop the `null` check.

@michael-mclawhorn Could probably can verify that for me.

The reason the `SystemConfigProvider` accesses the environment directly, is because it can't very well ask itself about its configuration parameter!

Everything else should go through the `SystemConfig` object though.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
```diff
@@ -472,13 +471,7 @@ public static String export() {
...
         record(new Durations(current.aggregateDurations()));
         record(new Threads(current.threadIds));
-        try {
-            return current.mapper.writeValueAsString(current.info);
-        } catch (JsonProcessingException jpe) {
-            String msg = String.format("Exporting mega log line with id: '%s' to JSON failed.", current.logId);
-            LOG.warn(msg, jpe);
-            return msg;
-        }
+        return LogFormatterProvider.getInstance().write(current.info);
```
archolewa: We should make sure to remove the `mapper` field from this class, since it's no longer necessary.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
archolewa: (That's what `SystemConfig::getPackageVariableName` does).

fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
```diff
@@ -0,0 +1,53 @@
...
+ * A class to provide a log formatter instance.
+ */
+public class LogFormatterProvider {
+
+    private static final Logger LOG = LoggerFactory.getLogger(LogFormatterProvider.class);
+    private static final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance();
+
+    private static final String LOG_FORMATTER_IMPL_KEY = SystemConfigProvider.getInstance().getStringProperty(
+            SYSTEM_CONFIG.getPackageVariableName("log_formatter_impl"),
+            "fili__log_formatter_impl"
```
archolewa: The second argument to `getStringProperty` is not a fully qualified config variable, it's a default setting for the config parameter. In other words, the second argument to `getStringProperty` should be `DEFAULT_LOG_FORMATTER_IMPL` not `fili__log_formatter_impl`.

archolewa:
👍

fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public class LogFormatterProvider {
+
+    private static final Logger LOG = LoggerFactory.getLogger(LogFormatterProvider.class);
+    private static final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance();
+
+    private static final String DEFAULT_LOG_FORMATTER_IMPL = JsonLogFormatter.class.getCanonicalName();
+
+    private static final String LOG_FORMATTER_IMPL_KEY = SystemConfigProvider.getInstance().getStringProperty(
+            SYSTEM_CONFIG.getPackageVariableName("log_formatter_impl"),
```
cdeszaq: Can we expand this to `implementation`? In general, we try to avoid abbreviations if we can so that the code is more understandable and clear.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     *  Get an instance of LogFormatter.
+     *
+     *  @return an instance of LogFormatter
+     */
+    public static LogFormatter getInstance() {
+        if (LOG_FORMATTER != null) {
+            return LOG_FORMATTER;
+        }
+        String logFormatterImplementation = System.getenv(LOG_FORMATTER_IMPL_KEY);
```
cdeszaq: This construction allows for the `LOG_FORMATTER` singleton holder to be unset if somehow something calls `getInstance` before this class does. It _shouldn't_ happen, but it would be good if we can insulate ourselves. A better construction is something like this:

```java
public String getInstance() {
    if (MY_INSTANCE == null) {
        // create myInstance
        MY_INSTANCE = myInstance;
    }
    return MY_INSTANCE;
}
```

By having only 1 return path, it's easier to reason about having set the singleton holder.

And if you want to get even more paranoid, you can have the body of that `if` be it's own `synchronized` helper method (which should also check if it's already set).

fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
```diff
@@ -0,0 +1,53 @@
...
+        try {
+            if (logFormatterImplementation == null) {
+                logFormatterImplementation = SystemConfigProvider.getInstance().getStringProperty(
+                        SYSTEM_CONFIG.getPackageVariableName(LOG_FORMATTER_IMPL_KEY),
+                        DEFAULT_LOG_FORMATTER_IMPL
+                );
+            }
+            return (LogFormatter) Class.forName(logFormatterImplementation).newInstance();
+        } catch (Exception e) {
+            LOG.error("Exception while loading Log formatter: {}", e.getMessage(), e);
```
cdeszaq: There's no use in having the exception message as it's own parameter, and in fact it's actively hiding information in this call because you only have 1 placeholder in the log message template. (ie. `e` will not get logged)

Instead, since SLF4J log calls can natively handle exceptions, just have a call like this: `LOG.error("message", e)`, passing the exception as the only parameter.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
```diff
@@ -0,0 +1,53 @@
...
+            throw new IllegalStateException(e);
```
cdeszaq: We should make sure to include the message we're logging in the exception as a message as well.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatter.java
```diff
@@ -0,0 +1,18 @@
...
+public interface LogFormatter {
+
+    /**
+     * Takes in a log block and returns a formatted String to be logged.
+     *
+     * @param logBlock  the log block that represents a log and that will be formatted in to the needed format.
+     *
+     * @return a formatted String to be logged
+     */
+    String write(LogBlock logBlock);
```
cdeszaq: It's strange that this method is called `write` and not `format` when the interface is a Formatter. Can we change the name of the method to `format`? (it's also not supposed to be writing anything, only returning the LogBlock as a formatted string.)

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
```diff
@@ -472,13 +463,7 @@ public static String export() {
...
```
cdeszaq: Since there's now no longer any requirement that this return JSON, we should update the JavaDoc on this method to reflect that as well.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ Current
 
 ### Added:
 
+- [Customize Logging Format](https://github.com/yahoo/fili/pull/81)
```
cdeszaq: This should mention that it's the `RequestLog`, not all logs in general

fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
```diff
@@ -0,0 +1,54 @@
...
+                if (logFormatterImplementation == null) {
+                    logFormatterImplementation = SystemConfigProvider.getInstance().getStringProperty(
+                            SYSTEM_CONFIG.getPackageVariableName(LOG_FORMATTER_IMPLEMENTATION_KEY),
+                            DEFAULT_LOG_FORMATTER_IMPL
+                    );
+                }
+                return (LogFormatter) Class.forName(logFormatterImplementation).newInstance();
+            } catch (Exception exception) {
+                LOG.error("Exception while loading Log formatter: {}", exception);
+                throw new IllegalStateException(exception.getMessage());
```
cdeszaq: We should always include the causing exception as an exception's Cause when creating a new exception.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
```diff
@@ -0,0 +1,54 @@
...
+    private static final String LOG_FORMATTER_IMPLEMENTATION_KEY = SystemConfigProvider.getInstance().getStringProperty(
+            SYSTEM_CONFIG.getPackageVariableName("log_formatter_implementation"),
+            DEFAULT_LOG_FORMATTER_IMPL
+    );
+
+    /**
+     * The instance of the Log Formatter.
+     */
+    private static final LogFormatter LOG_FORMATTER = getInstance();
+
```
michael-mclawhorn: I'm not convinced that final is the correct behavior here.  SystemConfig was intended to allow bootstrapping to more sophisticated configurations at runtime.  I'm not sure that it wouldn't make sense to allow this to do the same.

michael-mclawhorn:
I'd like to see some tests.  Test that the Json log formatter produces something reasonable from a LogBlock at the very least.  Test that you get the expected LogFormatter based on configuration would also be nice.  (if you implement my suggestion with definalizing the LogFormatter, you could test with one and then with the other by tweaking configuration)

fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     *  Get an instance of LogFormatter.
+     *
+     *  @return an instance of LogFormatter
+     */
+    public static LogFormatter getInstance() {
+        if (LOG_FORMATTER == null) {
+            String logFormatterImplementation = System.getenv(LOG_FORMATTER_IMPLEMENTATION_KEY);
```
archolewa: We shouldn't be going to the environment directly at all. The default implementation of `SystemConfig` (`LayeredSystemConfig`) already goes to the environment first, and if a customer introduces their own custom system configuration that changes where the environment is used (or doesn't want to use it all) they'll get incorrect behavior here.

We should _only_ be going to the `SystemConfigProvider`.

QubitPi: 👍

archolewa:
Reafirrming my 👍

PR #82, opened 2016-10-28
Fix list environment variable not pulled in correctly

garyluoex:
Comma seperated list environment variables was pulled in as a single string before, this PR pulls environment variables through the `PropertiesConfiguration` so that the behavior is consistent with file based properties.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,10 @@ Current
 
 ### Fixed:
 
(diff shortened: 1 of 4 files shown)


cdeszaq:
👍  once a ChangeLog entry is added (should go under the `Fixed` section, since this is a bug-fix)

archolewa:
👍

PR #86, opened 2016-10-31
Populate the Dimension Cache from an avro file

RVRSKumar:
- Adding `AvroDimensionRowParser` class that parses an AVRO data file(.avro) after validating the AVRO schema (.avsc).

- Adding a functional Interface `DimensionFieldMapper` that maps field name based on the user implementation.

### DIFF
--- a/fili-core/pom.xml
+++ b/fili-core/pom.xml
@@ -208,10 +208,62 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-collections4</artifactId>
         </dependency>
+
+        <!-- Apache Avro Libraries -->
+        <dependency>
+            <groupId>org.apache.avro</groupId>
+            <artifactId>avro</artifactId>
+        </dependency>
     </dependencies>
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <version>2.10</version>
+                <executions>
+                    <execution>
+                        <id>copy</id>
+                        <phase>generate-test-resources</phase>
+                        <goals>
+                            <goal>copy</goal>
+                        </goals>
+                        <configuration>
+                            <artifactItems>
+                                <artifactItem>
+                                    <groupId>org.apache.avro</groupId>
+                                    <artifactId>avro-tools</artifactId>
+                                    <version>1.7.7</version>
+                                    <type>jar</type>
+                                    <outputDirectory>${project.basedir}/src/test/resources/jar</outputDirectory>
+                                    <destFileName>avro-tools.jar</destFileName>
+                                </artifactItem>
+                            </artifactItems>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <artifactId>maven-antrun-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <phase>generate-test-resources</phase>
+                        <configuration>
+                            <tasks>
+                                <exec
+                                        dir="${project.basedir}/src/test/scripts"
+                                        executable="./createAvro.bash"
+                                        failonerror="true">
+                                </exec>
+                            </tasks>
+                        </configuration>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
@@ -1,0 +1,194 @@
+// Copyright 2016 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.data.dimension;
+
+import com.yahoo.bard.webservice.data.cache.HashDataCache.Pair;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.core.JsonProcessingException;
+
+import org.apache.avro.Schema;
+import org.apache.avro.file.DataFileReader;
+import org.apache.avro.generic.GenericDatumReader;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.avro.io.DatumReader;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+/**
+ * Parses an AVRO file into Dimension Rows.
+ */
+public class AvroDimensionRowParser {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AvroDimensionRowParser.class);
+    private final ObjectMapper objectMapper;
+
+    private final DimensionFieldNameMapper dimensionFieldNameMapper;
+
+    /**
+     * Constructs an AvroDimensionRowParser object based on the DimensionFieldNameMapper object.
+     *
+     * @param dimensionFieldNameMapper Object that defines the dimension field name transformations
+     * @param objectMapper Object that is used to construct mapper objects
+     */
+    public AvroDimensionRowParser(DimensionFieldNameMapper dimensionFieldNameMapper, ObjectMapper objectMapper) {
+        this.dimensionFieldNameMapper = memoize(dimensionFieldNameMapper);
+        this.objectMapper = objectMapper;
+    }
+
+    /**
+     * Validates the schema of the given AVRO file with the Dimension schema configured by the user.
+     *
+     * <p>
+     * The sample schema is expected to be in the following format
+     * <pre><code>
+     * {
+     *    "type" : "record",
+     *    "name" : "TUPLE_0",
+     *    "fields" : [
+     *    {
+     *    "name" : "FOO_ID",
+     *    "type" : [ "null","int" ]
+     *    },
+     *    {
+     *    "name" : "FOO_DESC",
+     *    "type" : [ "null", "string" ]
+     *    }
+     *    ]
+     * }
+     * </code></pre>
+     *
+     * @param dimension The dimension object used to configure the dimension
+     * @param avroSchemaPath The path of the AVRO schema file (.avsc)
+     *
+     * @return true if the schema is valid, false otherwise
+     *
+     * @throws IllegalArgumentException thrown if JSON object `fields` is not present
+     */
+    private boolean doesSchemaContainAllDimensionFields(Dimension dimension, String avroSchemaPath)
+        throws IllegalArgumentException {
+
+        // Convert the AVRO schema file into JsonNode Object
+        JsonNode jsonAvroSchema = convertFileToJsonNode(avroSchemaPath);
+
+        jsonAvroSchema = Optional.ofNullable(jsonAvroSchema.get("fields")).orElseThrow(() -> {
+                String msg = "`fields` is a required JSON field in the avro schema";
+                LOG.error(msg);
+                return new IllegalArgumentException(msg);
+        });
+
+        // Populating the set of avro field names
+        Set<String> avroFields = StreamSupport.stream(jsonAvroSchema.spliterator(), false)
+                .map(jsonNode -> jsonNode.get("name"))
+                .map(JsonNode::asText)
+                .collect(Collectors.toSet());
+
+        // True only if all of the mapped dimension fields are present in the Avro schema
+        return dimension.getDimensionFields().stream()
+                .map(dimensionField -> dimensionFieldNameMapper.convert(dimension, dimensionField))
+                .allMatch(avroFields::contains);
+    }
+
+    /**
+     * Parses the avro file and populates the dimension rows after validating the schema.
+     *
+     * @param dimension The dimension object used to configure the dimension
+     * @param avroFilePath The path of the AVRO data file (.avro)
+     * @param avroSchemaPath The path of the AVRO schema file (.avsc)
+     *
+     * @return A set of dimension rows
+     *
+     * @throws IllegalArgumentException thrown if JSON object `fields` is not present
+     */
+    public Set<DimensionRow> parseAvroFileDimensionRows(Dimension dimension, String avroFilePath, String avroSchemaPath)
+        throws IllegalArgumentException {
+
+        if (doesSchemaContainAllDimensionFields(dimension, avroSchemaPath)) {
+
+            DataFileReader<GenericRecord> dataFileReader;
+            String filePath = avroSchemaPath;
+            try {
+                // Creates an AVRO schema object based on the parsed AVRO schema file (.avsc)
+                Schema schema = new Schema.Parser().parse(new File(filePath));
+
+                // Creates an AVRO DatumReader object based on the AVRO schema object
+                DatumReader<GenericRecord> datumReader = new GenericDatumReader<>(schema);
+
+                filePath = avroFilePath;
+                // Creates an AVRO DataFileReader object that reads the AVRO data file one record at a time
+                dataFileReader = new DataFileReader<>(new File(avroFilePath), datumReader);
+
+            } catch (IOException e) {
+                String msg = String.format("Unable to process the file, at the location %s", filePath);
+                LOG.error(msg, e);
+                throw new IllegalArgumentException(msg, e);
+            }
+
+            // Generates a set of dimension Rows after retrieving the appropriate fields
+            return StreamSupport.stream(dataFileReader.spliterator(), false)
+                    .map(genericRecord -> dimension.getDimensionFields().stream().collect(
+                             Collectors.toMap(
+                                 DimensionField::getName,
+                                 dimensionField -> genericRecord.get(
+                                     dimensionFieldNameMapper.convert(dimension, dimensionField)
+                                 ).toString()
+                             )
+                         )
+                    )
+                    .map(dimension::parseDimensionRow)
+                    .collect(Collectors.toSet());
+        } else {
+            String msg = "The AVRO schema file does not contain all the configured dimension fields";
+            LOG.error(msg);
+            throw new IllegalArgumentException(msg);
+        }
+    }
+
+    /**
+     * Constructs a JSON Node object from the avro schema file.
+     *
+     * @param avroSchemaPath The path of the AVRO schema file (.avsc)
+     *
+     * @return JsonNode object
+     *
+     * @throws IllegalArgumentException thrown if there is an error parsing the avro files
+     */
+    private JsonNode convertFileToJsonNode(String avroSchemaPath) throws IllegalArgumentException {
+
+        try {
+            return objectMapper.readTree(new File(avroSchemaPath));
+        } catch (JsonProcessingException e) {
+            String msg = "Unable to process the Json";
+            LOG.error(msg, e);
+            throw new IllegalArgumentException(msg, e);
+        } catch (IOException e) {
+            String msg = String.format("Unable to process the file, at the location %s", avroSchemaPath);
+            LOG.error(msg, e);
+            throw new IllegalArgumentException(msg, e);
+        }
+    }
+
+    /**
+     * Returns a memoized converter function for the dimension field name mapping.
+     *
+     * @param dimensionFieldNameMapper Object that defines the dimension field name transformations
+     * @return Memoized function that converts the dimension field name based on the user mapping
+     */
+    private DimensionFieldNameMapper memoize(DimensionFieldNameMapper dimensionFieldNameMapper) {
+        Map<Pair<Dimension, DimensionField>, String> cache = new HashMap<>();
+        return (dimension, dimensionField) -> cache.computeIfAbsent(
+                new Pair<>(dimension, dimensionField),
+                key -> dimensionFieldNameMapper.convert(key.getKey(), key.getValue())
+        );
+    }
+}
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParserSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParserSpec.groovy
@@ -1,0 +1,61 @@
+// Copyright 2016 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.data.dimension
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.yahoo.bard.webservice.data.dimension.impl.KeyValueStoreDimension
+import com.yahoo.bard.webservice.data.dimension.impl.ScanSearchProviderManager
+import spock.lang.Specification
+
+
+class AvroDimensionRowParserSpec extends Specification {
+    def LinkedHashSet<DimensionField> dimensionFields
+    def KeyValueStoreDimension dimension
+    def AvroDimensionRowParser avroDimensionRowParser
+
+    def setup() {
+        dimensionFields = [BardDimensionField.ID, BardDimensionField.DESC]
+        dimension = new KeyValueStoreDimension("foo", "desc-foo", dimensionFields, MapStoreManager.getInstance("foo"), ScanSearchProviderManager.getInstance("foo"))
+        avroDimensionRowParser = new AvroDimensionRowParser(DimensionFieldNameMapper.underscoreSeparatedConverter(), new ObjectMapper())
+    }
+
+    def "Schema file containing all the dimension fields and data parses to expected rows"() {
+        given:
+        DimensionRow dimensionRow1 = BardDimensionField.makeDimensionRow(dimension, "12345", "bar")
+        DimensionRow dimensionRow2 = BardDimensionField.makeDimensionRow(dimension, "67890", "baz")
+        Set<DimensionRow> dimSet = [dimensionRow1, dimensionRow2] as Set
+
+        expect:
+        avroDimensionRowParser.parseAvroFileDimensionRows(dimension, "src/test/resources/avroFilesTesting/sampleData.avro", "src/test/resources/avroFilesTesting/sampleData.avsc") == dimSet
+    }
+
+    def "Invalid schema - Schema file does not exist throws an IllegalArgumentException"() {
+        when:
+        avroDimensionRowParser.parseAvroFileDimensionRows(dimension, "src/test/resources/avroFilesTesting/sampleData.avro", "src/test/resources/avroFilesTesting/foo.avsc")
+
+        then:
+        IllegalArgumentException exception = thrown(IllegalArgumentException)
+        exception.message == "Unable to process the file, at the location src/test/resources/avroFilesTesting/foo.avsc"
+    }
+
+    def "Schema file does not contain all the dimension fields throws an IllegalArgumentException"() {
+        given:
+        dimensionFields.add(BardDimensionField.FIELD1)
+
+        when:
+        avroDimensionRowParser.parseAvroFileDimensionRows(dimension, "src/test/resources/avroFilesTesting/sampleData.avro", "src/test/resources/avroFilesTesting/sampleData.avsc")
+
+        then:
+        IllegalArgumentException exception = thrown(IllegalArgumentException)
+        exception.message == "The AVRO schema file does not contain all the configured dimension fields"
+    }
+
+    def "Schema file does not contain the required `fields` section in its JSON throws an IllegalArgumentException" () {
+        when:
+        avroDimensionRowParser.parseAvroFileDimensionRows(dimension, "src/test/resources/avroFilesTesting/sampleData.avro", "src/test/resources/avroFilesTesting/invalidSchema.avsc")
+
+        then:
+        IllegalArgumentException exception = thrown(IllegalArgumentException)
+        exception.message == "`fields` is a required JSON field in the avro schema"
+    }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,6 @@ snapshot-pom.xml
 report
 /target
 /checkstyle-style-local.xml
+
+# Binary avro files should be ignored
+*.avro
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,12 @@ Current
 
 ### Added:
 
+
 - [Customize Logging Format in RequestLog](https://github.com/yahoo/fili/pull/81)
+
+- [Populate Dimension Rows from an AVRO file](https://github.com/yahoo/fili/pull/86)
+    * Added `AvroDimensionRowParser` class that parses an AVRO data file (.avro) after validating the AVRO schema (.avsc).
+    * Added a functional Interface `DimensionFieldMapper` that maps field name based on the user implementation.
 
 - [Support for Druid's In Filter](https://github.com/yahoo/fili/pull/64)
     * The in-filter only works with Druid versions 0.9.0 and up.
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionFieldNameMapper.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionFieldNameMapper.java
@@ -1,0 +1,32 @@
+// Copyright 2016 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.data.dimension;
+
+import java.util.Locale;
+
+/**
+ * Maps a Dimension Field value to a string.
+ */
+@FunctionalInterface
+public interface DimensionFieldNameMapper {
+
+    /**
+     * A converter function for the dimension field Mapping.
+     *
+     * @param dimension The dimension object used to configure the Dimension
+     * @param dimensionField The dimension field object
+     *
+     * @return the converted string for the dimension field
+     */
+    String convert(Dimension dimension, DimensionField dimensionField);
+
+    /**
+     * A default implementation of the converter method which uses underscore as a separator.
+     *
+     * @return an instance of DimensionFieldNameMapper
+     */
+    static DimensionFieldNameMapper underscoreSeparatedConverter () {
+        return (dimension, dimensionField) ->
+                    (dimension.getApiName() + "_" + dimensionField.getName()).toUpperCase(Locale.ENGLISH);
+    }
+}
--- a/fili-core/src/test/resources/avroFilesTesting/invalidSchema.avsc
+++ b/fili-core/src/test/resources/avroFilesTesting/invalidSchema.avsc
@@ -1,0 +1,18 @@
+{
+    "type" : "record",
+    "name" : "TUPLE_0",
+    "foo" : [
+        {
+            "name" : "FOO_ID",
+            "type" : [ "null","int" ]
+        },
+        {
+            "name" : "FOO_DESC",
+            "type" : [ "null", "string" ]
+        },
+        {
+            "name" : "MISCELLANEOUS",
+            "type" : [ "null", "string" ]
+        }
+    ]
+}
--- a/fili-core/src/test/resources/avroFilesTesting/sampleData.avsc
+++ b/fili-core/src/test/resources/avroFilesTesting/sampleData.avsc
@@ -1,0 +1,18 @@
+{  
+    "type" : "record",
+    "name" : "TUPLE_0",
+    "fields" : [  
+        {  
+            "name" : "FOO_ID",
+            "type" : [ "null","int" ]
+        },
+        {  
+            "name" : "FOO_DESC",
+            "type" : [ "null", "string" ]
+        },
+        {  
+            "name" : "MISCELLANEOUS",
+            "type" : [ "null", "string" ]
+        }
+    ]
+}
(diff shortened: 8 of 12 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,185 @@
...
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * Parses an AVRO file into Dimension Rows.
+ */
+public class AvroDimensionRowParser {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AvroDimensionRowParser.class);
+    private static ObjectMapper MAPPER = null;
```
michael-mclawhorn: I just checked and this field does not appear to be used statically, so you can simply make it not static.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,185 @@
...
+
+            // Represents each record in the avro data file
+            GenericRecord avroData = null;
+
+            while (dataFileReader.hasNext()) {
+                avroData = dataFileReader.next(avroData);
+
+                // Creates a JsonObject from the AVRO GenericRecord object
+                JsonNode avroDataJson = MAPPER.readTree(new JsonFactory().createParser(avroData.toString()));
+
```
michael-mclawhorn: Is there a good reason to go to an intermediate JSON form here?

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,185 @@
...
+                Map<String, String> dimensionFieldKeyValues = dimension.getDimensionFields()
+                        .stream()
+                        .collect(Collectors.toMap(dimensionField -> dimensionField.getName(),
+                                                  // Looks up converted field name from the `convertedDimensionFields`
+                                                  // Parsed as text to avoid the extra quotes around string objects
+                                                  dimensionField -> avroDataJson
+                                                        .get(convertedDimensionFields
+                                                                .get(dimensionField.getName())
+                                                        ).asText()
+                                                )
```
michael-mclawhorn: This feels like we're doing way too many repeated steps in the tight loop.  Perhaps the necessary mapping needed to pull the details off the avro record can be done in advance so that it's a sequence of point lookups only inside the big loop.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,185 @@
...
+    /**
+     * Returns a converter function for the dimension field name mapping.
+     *
+     * @param dimension The dimension object used to configure the dimension
+     * @param dimensionFieldMapper Object that defines the dimension field name transformations
+     *
+     * @return Function that converts the dimension field name based on the user mapping
+     */
+    private Function<DimensionField, String> curry(Dimension dimension, DimensionFieldMapper dimensionFieldMapper) {
+        return dimensionField -> dimensionFieldMapper.converter(dimension, dimensionField);
```
michael-mclawhorn: The action of this method is to curry, but the meaning is to map the fieldNames of dimension fields for a given dimension.  The name would be better if it corresponded with the meaning of the method.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionFieldMapper.java
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * A converter function for the dimension field Mapping.
+     *
+     * @param dimension The dimension object used to configure the Dimension
+     * @param dimensionField The dimension field object
+     *
+     * @return the converted string for the dimension field
+     */
+    String converter(Dimension dimension, DimensionField dimensionField);
+
```
michael-mclawhorn: This is either 'map' if you generically want to indicate that this is a mapping operation, or, given the fact that this is a DimensionFieldMapper a more meaningful name like dimensionFieldName or something scoped to the meaning.

fili-core/src/main/resources/sampleData.avsc
```diff
@@ -0,0 +1,11 @@
+{
+  "type" : "record",
+  "name" : "TUPLE_0",
+  "fields" : [ {
+    "name" : "FOO_ID",
+    "type" : [ "null", "int" ]
+  }, {
+    "name" : "FOO_DESC",
```
michael-mclawhorn: This indentation looks wrong, since it aligns with 'fields' but is actually _inside_ the list that is the value of fields.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        Function<DimensionField, String> curriedConverter = curry(dimension, dimensionFieldMapper);
+
+        // Generates a map of configured dimension field names and the converted field names, for easier lookups in the
+        // `parseAvroFileDimensionRows()`
+        convertedDimensionFields = dimension.getDimensionFields()
+                .stream()
+                .collect(Collectors.toMap(dimensionField -> dimensionField.getName(), curriedConverter));
+
```
michael-mclawhorn: This is a major side effect that your downstream code depends on happening inside a validate method.  That is some major spooky-action-at-a-distance.  Either build the convertedDimensionFields explicitly, pass it as a variable, or throw away the copy you build for this validator and rebuild the convertedFields closer to where they'll be used later.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParserSpec.groovy
```diff
@@ -0,0 +1,41 @@
...
+        avroDimensionRowParser = new AvroDimensionRowParser(DimensionFieldMapper.underscoreSeparatedConverter(), new ObjectMapper())
+    }
+
+    def "isSchema valid"() {
+        expect:
+        avroDimensionRowParser.isSchemaValid(dimension, "src/main/resources/sampleData.avsc") == true
+    }
+
+    def "Schema data"() {
+        given:
```
michael-mclawhorn: A little more sentence structure would make this test clearer.  'Parse sample data with scheme' or some such.  @cdeszaq : you're the more opinionated on spock-esque names.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParserSpec.groovy
```diff
@@ -0,0 +1,41 @@
...
+
+        when:
+        avroDimensionRowParser.parseAvroFileDimensionRows(dimension,
+                "src/main/resources/sampleData.avro",
+                "src/main/resources/sampleData.avsc")
+
+        then:
+        dimension.searchProvider.findAllDimensionRows() == dimSet
+    }
+}
```
michael-mclawhorn: You should also test as many of your 'expected' error conditions as you can.  At the very least feed this thing an invalid schema or two, a schema that has the wrong dimensions, and a dataset that doesn't match it's schema.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
archolewa: Technically, this function is partially applying `dimensionFieldMapper::converter`, not performing any currying. Currying is an operation solely on a function, this is taking (effectively) a function and one of its arguments and applying the first argument.

An actual currying would look more like:

```java
private Function<DimensionField, Function<Dimension, String>> curry(DimensionFieldMappe mapper) {
     return dimensionField -> dimension -> mapper.converter(dimension, dimensionField);
}  
```

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionFieldMapper.java
archolewa: Alternatively, we could also have this interface extend `BiFunction<Dimension, DimensionField, String>`, and this method would just be called `apply`.

We could go _even farther_ and just do away with this interface entirely, and instead have `AvroDimensionRowParser` just take a `BiFunction<Dimension, DimensionField, String>` (its purpose can be made clear enough via a good variable name). One less wrapper concept in the codebase. Though I don't know enough about how we plan to integrate this into the rest of the codebase to know if that's a good idea.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
cdeszaq: One way to deal with the desire to not compute this twice would be to memoize the function so that it's not called for repeated values. Here's one way to do that:

```java
/**
 * Returns a memoized converter function for the dimension field name mapping.
 *
 * @param dimensionFieldMapper Object that defines the dimension field name transformations
 *
 * @return Memoized function that converts the dimension field name based on the user mapping
 */
private DimensionFieldMapper memoize(DimensionFieldMapper dimensionFieldMapper) {
    Map<Pair<Dimension, DimensionField>, String> cache = new HashMap<>();
    return (dimension, dimensionField) -> cache.computeIfAbsent(
            new Pair<>(dimension, dimensionField),
            key -> dimensionFieldMapper.converter(key.getKey(), key.getValue())
    );
}
```

Note that there's a risk of this getting to be a huge map, so we may want to switch to an LRU map from Apache Commons if we think there could be a huge number of different calls.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
cdeszaq: @michael-mclawhorn the challenge here is that "pull the info off of the avro record" is pretty much all this is doing. The only other task in here is to do that for each field on the dimension (so, another loop). Another way of looking at it would be:

```java
StreamSupport.stream(dataFileReader.spliterator(), false)
        .map(genericRecord -> dimension.getDimensionFields().stream().collect(
                        Collectors.toMap(
                                DimensionField::getName,
                                dimensionField -> genericRecord.get(
                                        dimensionFieldMapper.converter(dimension, dimensionField)
                                ).toString()
                        )
                )
        )
        .map(dimension::parseDimensionRow)
        .collect(Collectors.toSet());
```

One thing that _could_ be reduced (which the current code does via the shared data structure) would be to memoize the field conversions. (see my other comment)

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,185 @@
...
+            // Creates an AVRO DataFileReader object that reads the AVRO data file one record at a time
+            DataFileReader<GenericRecord> dataFileReader =
+                new DataFileReader<>(new File(avroFilePath), datumReader);
+
+            // Represents each record in the avro data file
+            GenericRecord avroData = null;
+
+            while (dataFileReader.hasNext()) {
+                avroData = dataFileReader.next(avroData);
```
cdeszaq: Is there a strong reason to use this mechanism instead of a more basic foreach loop? (eg. `for (GenericRecord genericRecord : dataFileReader) {`)

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParserSpec.groovy
```diff
@@ -0,0 +1,41 @@
...
+        DimensionRow dimensionRow1 = BardDimensionField.makeDimensionRow(dimension, "12345", "bar")
+        DimensionRow dimensionRow2 = BardDimensionField.makeDimensionRow(dimension, "67890", "baz")
+        Set<DimensionRow> dimSet = [dimensionRow1, dimensionRow2] as Set
+
+        when:
+        avroDimensionRowParser.parseAvroFileDimensionRows(dimension,
+                "src/main/resources/sampleData.avro",
+                "src/main/resources/sampleData.avsc")
+
+        then:
```
cdeszaq: This should be able to be converted into just an `expect` block if the parse method is made side-effect-free (ie. to return the collection of parsed rows)

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
RVRSKumar: @michael-mclawhorn The idea here is retrieval of the values would be easier if it is a JSON. Since, they are mapped as Key Value pairs.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,185 @@
...
+
+/**
+ * Parses an AVRO file into Dimension Rows.
+ */
+public class AvroDimensionRowParser {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AvroDimensionRowParser.class);
+    private static ObjectMapper MAPPER = null;
+
+    private DimensionFieldMapper dimensionFieldMapper;
```
RVRSKumar: We are passing in the value from the constructor, hence this can't be final here

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
michael-mclawhorn: To clarify, don't call the method just 'memoize' if you do.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
michael-mclawhorn: I think you'll find that values initialized in the constructor CAN be final if they're not static.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,176 @@
...
+                         )
+                    )
+                .map(dimension::parseDimensionRow)
+                .collect(Collectors.toSet());
+
+                // Adds all dimension rows to the dimension cache
+                dimension.addAllDimensionRows(dimensionRows);
+        }
+
+        else {
```
garyluoex: we should probably follow the `if`, `else` convention in fili `} else {`

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,185 @@
...
+     * @throws IOException thrown if there is error parsing the avro files
+     */
+    public boolean isSchemaValid(Dimension dimension, String avroSchemaPath) throws NullPointerException, IOException {
+
+        // Convert the AVRO schema file into JsonNode Object
+        JsonNode jsonAvroSchema = convertFileToJsonNode(avroSchemaPath);
+
+        // If the JSON schema file is corrupted/ empty, then return the schema is invalid
+        if (jsonAvroSchema == null) {
+            return false;
```
RVRSKumar: Logged it in the `parseDimensionRows` method

cdeszaq:
@RVRSKumar, please don't squash commits away until the PR is just about to merge, if possible. It's super helpful for reviewers to be able to see what changes were made in response to comments, and squashing hides that info. Just for future reference 😄

RVRSKumar:
@cdeszaq Yeah I saw @garyluoex comment last. So, I addressed it and squashed it.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionFieldMapper.java
```diff
@@ -0,0 +1,32 @@
...
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.data.dimension;
+
+import java.util.Locale;
+
+/**
+ * Maps a Dimension Field value to a string.
+ */
+@FunctionalInterface
+public interface DimensionFieldMapper {
```
cdeszaq: I think `DimensionFieldNameMapper` is a pretty solid option, since it's clear that this maps a `DimensionField` into some representation of it's name.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
cdeszaq: I'm generally a big fan of `Optional::orElseThrow`. In this case, however, I don't see how we could get a `null` back? It looks like (based on tracing the code a little) we'll either get back a `NullNode` (which is not `null`), or we'll get back some other sort of node. The only other cases I can see look like Exception flows, where an exception will be thrown instead of returning anything. So, based on that, do we even need this check at all?

cdeszaq: That said, I do think there's merit in making this a `void` method that throws validation exception of some sort. The advantage of that is that there's a channel by which we can know (and log) what the validation problem is. Simply returning a `boolean` doesn't give any indication of what's invalid, only that there's a problem.

A different way of dealing with this would be to make the method much more specific, rather than leave it a general "validate" method. In this particular case, the method name could be changed to something like `doesSchemaFullyCoverDimensionFields`, since it's clear then what a `false` response means. (and anything that prevents us from answering the question is an exception case where the argument is probably invalid, since we're documenting what the expected JSON should be.)

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        try {
+            jsonAvroSchema = jsonAvroSchema.get("fields");
+        } catch (NullPointerException e) {
```
cdeszaq: Good catch. This check actually needs to look to see if `jsonAvroSchema.get("fields")` is null, and throw on that, rather than looking for an NPE. Otherwise thNPE will pop out when we call `jsonAvroSchema.spliterator()` later.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    /**
+     * Parses the avro file and populates the dimension rows after validating the schema.
+     *
+     * @param dimension The dimension object used to configure the dimension
+     * @param avroFilePath The path of the AVRO data file (.avro)
+     * @param avroSchemaPath The path of the AVRO schema file (.avsc)
+     *
+     * @throws IllegalArgumentException thrown if JSON object `fields` is not present
```
cdeszaq: _where_ do we expect there to be a `fields` object? We should add that it needs to be there in the schema.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,192 @@
...
+     * @throws IOException thrown if there is error parsing the avro files
```
cdeszaq: We should trap this into an unchecked exception, unless we expect that the caller will be able to handle it somehow?

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,192 @@
...
+                             )
+                         )
+                    )
+                    .map(dimension::parseDimensionRow)
+                    .collect(Collectors.toSet());
+
+                // Adds all dimension rows to the dimension cache
+                dimension.addAllDimensionRows(dimensionRows);
+        } else {
+            String msg = "The AVRO schema file is corrupted / empty {}";
```
cdeszaq: I don't think this message is correct. We hit this block if the schema isn't _valid_ (ie. doesn't have the needed fields in it), not if the file is corrupted or empty.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParserSpec.groovy
```diff
@@ -0,0 +1,63 @@
...
+        expect:
+        avroDimensionRowParser.isSchemaValid(dimension, "src/test/resources/avroFilesTesting/sampleData.avsc")
+    }
+
+    def "a schema which doesn't have the `fields` section throws a NullPointerException"() {
+        when:
+        avroDimensionRowParser.isSchemaValid(dimension,  "src/test/resources/avroFilesTesting/illegalSchema.avsc")
+
+        then:
+        thrown NullPointerException
```
cdeszaq: We should never be throwing NPEs from our methods intentionally.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,192 @@
...
+     * Parses the avro file and populates the dimension rows after validating the schema.
+     *
+     * @param dimension The dimension object used to configure the dimension
+     * @param avroFilePath The path of the AVRO data file (.avro)
+     * @param avroSchemaPath The path of the AVRO schema file (.avsc)
+     *
+     * @throws IllegalArgumentException thrown if JSON object `fields` is not present
+     * @throws IOException thrown if there is error parsing the avro files
+     */
+    public void parseAvroFileDimensionRows(Dimension dimension, String avroFilePath, String avroSchemaPath)
```
cdeszaq: Rather than adding the rows to the dimension in here, it's better to have this method _return_ the dimension rows it parses and let the caller decide what to do with them. Generally, we want to avoid side effects for methods when we can, and returning void is often an indication we are doing that. (Also, its odd for a parser to also add them to a dimension, instead of just returning the things it parsed)

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,192 @@
...
+        try {
+            File file = new File(avroSchemaPath);
+            JsonNode jsonNode = objectMapper.readTree(file);
+            return jsonNode;
+        } catch (JsonProcessingException e) {
+            LOG.error("Unable to process the Json {}", e);
+            throw e;
+        } catch (IOException e) {
+            LOG.error("Unable to process the file {}", e);
+            throw e;
```
RVRSKumar: I segregated them because they are 2 different exceptions, and I thought they need to be logged differently. Let me, know if I can combine them into a single handler with a generic error message '''`Unable to process the file {}`'''

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,198 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+/**
+ * Parses an AVRO file into Dimension Rows.
+ */
+public class AvroDimensionRowParser {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AvroDimensionRowParser.class);
+    private ObjectMapper objectMapper;
```
cdeszaq: Should be `final`?

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,198 @@
...
+        JsonNode jsonAvroSchema = convertFileToJsonNode(avroSchemaPath);
+
+        try {
+            // The following method throws a NullPointerException, if the JSON schema doesn't have the `fields` object
+            // Avoiding the `Optional.of().orElseThrow()` to log the error details.
+            Optional.of(jsonAvroSchema.get("fields"));
+        } catch (NullPointerException e) {
+            String msg = "`fields` is a required JSON field in the avro schema {} ";
+            LOG.error(msg, e);
+            throw new IllegalArgumentException(msg);
```
cdeszaq: If we catch and throw a new exception, we should always be sure to include the initiating exception as the cause so that we don't lose the full trace.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,198 @@
...
+
+        // Convert the AVRO schema file into JsonNode Object
+        JsonNode jsonAvroSchema = convertFileToJsonNode(avroSchemaPath);
+
+        try {
+            // The following method throws a NullPointerException, if the JSON schema doesn't have the `fields` object
+            // Avoiding the `Optional.of().orElseThrow()` to log the error details.
+            Optional.of(jsonAvroSchema.get("fields"));
+        } catch (NullPointerException e) {
+            String msg = "`fields` is a required JSON field in the avro schema {} ";
```
cdeszaq: Having the `{}` in here does nothing, since we're not giving a templating variable to the log call, we're giving it an exception (which is a special case). So, no need for `{}` if we're logging just a pre-formatted string and an exception.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,198 @@
...
+        throws IllegalArgumentException {
+
+        // Convert the AVRO schema file into JsonNode Object
+        JsonNode jsonAvroSchema = convertFileToJsonNode(avroSchemaPath);
+
+        try {
+            // The following method throws a NullPointerException, if the JSON schema doesn't have the `fields` object
+            // Avoiding the `Optional.of().orElseThrow()` to log the error details.
+            Optional.of(jsonAvroSchema.get("fields"));
+        } catch (NullPointerException e) {
```
cdeszaq: Rather than allow the exception to be thrown, and then throwing a new exception, it would be better to use the `orElseThrow` method so that only 1 exception is generated. (ie. the NPE doesn't give us any real info). This also lets us drop the try-catch as well.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,198 @@
...
+            String msg = "`fields` is a required JSON field in the avro schema {} ";
+            LOG.error(msg, e);
+            throw new IllegalArgumentException(msg);
+        }
+
+        jsonAvroSchema = jsonAvroSchema.get("fields");
```
cdeszaq: We should just use the `Optional` we're making when we look for this field the 1st time to set this. Another reason to use `orElseThrow` 😉

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,198 @@
...
+                // Creates an AVRO DatumReader object based on the AVRO schema object
+                DatumReader<GenericRecord> datumReader = new GenericDatumReader<>(schema);
+
+                // Creates an AVRO DataFileReader object that reads the AVRO data file one record at a time
+                dataFileReader = new DataFileReader<>(new File(avroFilePath), datumReader);
+
+            } catch (IOException e) {
+                String msg = String.format("Unable to process the file, at the location %s {} ", avroSchemaPath);
+                LOG.error(msg, e);
+                throw new IllegalArgumentException(msg);
```
cdeszaq: Need to include caught exception as the cause.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,198 @@
...
+                Schema schema = new Schema.Parser().parse(new File(avroSchemaPath));
+
+                // Creates an AVRO DatumReader object based on the AVRO schema object
+                DatumReader<GenericRecord> datumReader = new GenericDatumReader<>(schema);
+
+                // Creates an AVRO DataFileReader object that reads the AVRO data file one record at a time
+                dataFileReader = new DataFileReader<>(new File(avroFilePath), datumReader);
+
+            } catch (IOException e) {
+                String msg = String.format("Unable to process the file, at the location %s {} ", avroSchemaPath);
```
cdeszaq: `{}` not needed, since the logging framework isn't doing any additional expansion.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,198 @@
...
+                                 dimensionField -> genericRecord.get(
+                                     dimensionFieldNameMapper.convert(dimension, dimensionField)
+                                 ).toString()
+                             )
+                         )
+                    )
+                    .map(dimension::parseDimensionRow)
+                    .collect(Collectors.toSet());
+        } else {
+            String msg = "The AVRO schema file does not contain all the configured dimension fields {} ";
```
cdeszaq: `{}` not needed, since no expansion is happening.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,198 @@
...
+     * @return JsonNode object
+     *
+     * @throws IllegalArgumentException thrown if there is an error parsing the avro files
+     */
+    private JsonNode convertFileToJsonNode(String avroSchemaPath) throws IllegalArgumentException {
+
+        try {
+            return objectMapper.readTree(new File(avroSchemaPath));
+        } catch (JsonProcessingException e) {
+            String msg = "Unable to process the Json {} ";
```
cdeszaq: no `{}`

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,198 @@
...
+            LOG.error(msg, e);
+            throw new IllegalArgumentException(msg);
```
cdeszaq: Include cause

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,198 @@
...
+        } catch (IOException e) {
+            String msg = String.format("Unable to process the file, at the location %s {} ", avroSchemaPath);
+            LOG.error(msg, e);
+            throw new IllegalArgumentException(msg);
```
cdeszaq: Include cause

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,198 @@
...
+    private JsonNode convertFileToJsonNode(String avroSchemaPath) throws IllegalArgumentException {
+
+        try {
+            return objectMapper.readTree(new File(avroSchemaPath));
+        } catch (JsonProcessingException e) {
+            String msg = "Unable to process the Json {} ";
+            LOG.error(msg, e);
+            throw new IllegalArgumentException(msg);
+        } catch (IOException e) {
+            String msg = String.format("Unable to process the file, at the location %s {} ", avroSchemaPath);
```
cdeszaq: No need for `{}`, since log framework isn't doing this expansion.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParserSpec.groovy
```diff
@@ -0,0 +1,61 @@
...
+        IllegalArgumentException exception = thrown()
+        exception.message == "The AVRO schema file does not contain all the configured dimension fields {} "
+    }
+
+    def "Schema file does not contain the required `fields` section in its JSON throws an IllegalArgumentException" () {
+        when:
+        avroDimensionRowParser.parseAvroFileDimensionRows(dimension, "src/test/resources/avroFilesTesting/sampleData.avro", "src/test/resources/avroFilesTesting/invalidSchema.avsc")
+
+        then:
+        IllegalArgumentException exception = thrown()
```
cdeszaq: Just `thrown()` doesn't assert about the _type_ of exception thrown. To make this assertion stronger, you can do `IllegalArgumentException exception = thrown(IllegalArgumentException)`.

Note: Applies to all other places like this in this file

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParserSpec.groovy
```diff
@@ -0,0 +1,61 @@
...
+        exception.message == "`fields` is a required JSON field in the avro schema {} "
```
cdeszaq: Message shouldn't have `{}`. (ie. it makes no sense)

Applies elsewhere in file as well.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParserSpec.groovy
garyluoex: Hmmm.. according to spock documentation [here](http://spockframework.org/spock/docs/1.0/spock_primer.html#_exception_conditions) this should work also right?

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
garyluoex: I don't see how we can use `orElseThrow` directly since this code relies on using the fact that calling `Optional.of` on a `null` value throws a `NullPointerException`. On the other hand, we can use `orElseThrow` if we make `convertFileToJsonNode` to return an `Optional` by mapping `JsonNode.isNull() ? null : JsonNode`.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParserSpec.groovy
cdeszaq: Ahh, good catch. I didn't realize they were the same thing 😄 . In that case, feel free to update them as I mention, but it's not a blocker by any means, since what's here is just fine.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,192 @@
...
+        JsonNode jsonAvroSchema = convertFileToJsonNode(avroSchemaPath);
+
+        Optional.ofNullable(jsonAvroSchema.get("fields")).orElseThrow( () -> {
+                String msg = "`fields` is a required JSON field in the avro schema";
+                LOG.error(msg);
+                throw new IllegalArgumentException(msg);
+        });
+
+        // Populating the set of avro field names
+        Set<String> avroFields = StreamSupport.stream(jsonAvroSchema.spliterator(), false)
```
cdeszaq: I don't think this will iterate over the right things, will it? I think you need to do `jsonAvroSchema = Optional.ofNullable(jsonAvroSchema.get("fields")).orElseThrow` so that you're in the `fields` collection

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,192 @@
...
+    private boolean doesSchemaContainAllDimensionFields(Dimension dimension, String avroSchemaPath)
+        throws IllegalArgumentException {
+
+        // Convert the AVRO schema file into JsonNode Object
+        JsonNode jsonAvroSchema = convertFileToJsonNode(avroSchemaPath);
+
+        Optional.ofNullable(jsonAvroSchema.get("fields")).orElseThrow( () -> {
+                String msg = "`fields` is a required JSON field in the avro schema";
+                LOG.error(msg);
+                throw new IllegalArgumentException(msg);
```
cdeszaq: This needs to `return`, not `throw`

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -0,0 +1,192 @@
...
+     *
+     * @throws IllegalArgumentException thrown if JSON object `fields` is not present
+     */
+    private boolean doesSchemaContainAllDimensionFields(Dimension dimension, String avroSchemaPath)
+        throws IllegalArgumentException {
+
+        // Convert the AVRO schema file into JsonNode Object
+        JsonNode jsonAvroSchema = convertFileToJsonNode(avroSchemaPath);
+
+        Optional.ofNullable(jsonAvroSchema.get("fields")).orElseThrow( () -> {
```
cdeszaq: Checkstyle has a whitespace issue here (shouldn't be a space between `(` and `(`)

michael-mclawhorn:
How was the avro file generated?
If possible it's better to generate binaries in test data rather than store in version control.

RVRSKumar:
The avro file was generated using an `avro-tools.jar`. So, do you want to include the `avro-tools` to maven dependency and build the avro files during compilation time?

fili-core/pom.xml
```diff
@@ -208,12 +208,65 @@
...
+                                    <destFileName>avro-tools.jar</destFileName>
+                                </artifactItem>
+                            </artifactItems>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <artifactId>maven-antrun-plugin</artifactId>
+                <version>1.7</version>
```
cdeszaq: Since the system config project depends on this as well, we should lift at least the configuration of the version of this plugin up to the `pluginManagement` section of the parent pom, to make sure we don't get version conflicts

RVRSKumar:
screen shot 2016-11-10 at 3 24 24 pm

PR #87, opened 2016-11-01
Fixes the LogFormatterProvider.

archolewa:
--Originally, we were requesting the LogFormatter implemnetation value twice, once at
static initialization time, and then again, _using the LogFormatter
class name as the key_. This has been corrected. We also added some
tests.

--We also made the initialization of the formatter lazy, largely because
my debugger didn't seem to be able to stop at methods that are invoked
at static initialization time, which made debugging this thing a hassle.
Making it lazy also made it a lot easier to clean up between tests.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
@@ -4,6 +4,7 @@ package com.yahoo.bard.webservice.logging;
 
 import com.yahoo.bard.webservice.config.SystemConfig;
 import com.yahoo.bard.webservice.config.SystemConfigProvider;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -15,17 +16,14 @@ public class LogFormatterProvider {
     private static final Logger LOG = LoggerFactory.getLogger(LogFormatterProvider.class);
     private static final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance();
 
-    private static final String DEFAULT_LOG_FORMATTER_IMPL = JsonLogFormatter.class.getCanonicalName();
+    public static final String LOG_FORMATTER_IMPLEMENTATION_SETTING_NAME = "log_formatter_implementation";
 
-    private static final String LOG_FORMATTER_IMPLEMENTATION_KEY = SystemConfigProvider.getInstance().getStringProperty(
-            SYSTEM_CONFIG.getPackageVariableName("log_formatter_implementation"),
-            DEFAULT_LOG_FORMATTER_IMPL
-    );
+    private static final String DEFAULT_LOG_FORMATTER_IMPL = JsonLogFormatter.class.getCanonicalName();
 
     /**
      * The instance of the Log Formatter.
      */
-    private static final LogFormatter LOG_FORMATTER = getInstance();
+    private static LogFormatter logFormatter;
 
     /**
      *  Get an instance of LogFormatter.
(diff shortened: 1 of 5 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/logging/LogFormatterProvider.java
```diff
@@ -15,38 +16,33 @@
     private static final Logger LOG = LoggerFactory.getLogger(LogFormatterProvider.class);
     private static final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance();
 
-    private static final String DEFAULT_LOG_FORMATTER_IMPL = JsonLogFormatter.class.getCanonicalName();
+    public static final String LOG_FORMATTER_IMPLEMENTATION_SETTING_NAME = "log_formatter_implementation";
```
QubitPi: Why do we use `public` for this particular field?

fili-core/src/test/groovy/com/yahoo/bard/webservice/logging/LogFormatterProviderSpec.groovy
```diff
@@ -0,0 +1,67 @@
...
+        assert LogFormatterProvider.logFormatter == null
+
+        when:
+        systemConfig.clearProperty(logFormatterKey)
+
+        then:
+        LogFormatterProvider.instance instanceof JsonLogFormatter
+    }
+
+}
```
QubitPi: Missing new line

fili-core/src/test/groovy/com/yahoo/bard/webservice/logging/TestLogFormatter.groovy
```diff
@@ -0,0 +1,11 @@
+package com.yahoo.bard.webservice.logging
+
+/**
+ * A Test implementation of the `LogFormatter` class for the purpose of testing the LogFormatterProvider.
+ */
+class TestLogFormatter implements LogFormatter {
+    @Override
+    String format(final LogBlock logBlock) {
```
cdeszaq: `final` doesn't buy us anything here

fili-core/src/test/groovy/com/yahoo/bard/webservice/logging/LogFormatterProviderSpec.groovy
```diff
@@ -0,0 +1,67 @@
...
+
+
+class LogFormatterProviderSpec extends Specification {
+
+    @Shared LogFormatter originalLogFormatter
+    String originalLogFormatterClassName
+    SystemConfig systemConfig = SystemConfigProvider.instance;
+    String logFormatterKey = systemConfig.getPackageVariableName(
+            LogFormatterProvider.LOG_FORMATTER_IMPLEMENTATION_SETTING_NAME
+    )
```
cdeszaq: Wrap not needed?

fili-core/src/test/groovy/com/yahoo/bard/webservice/logging/LogFormatterProviderSpec.groovy
```diff
@@ -0,0 +1,67 @@
...
+        if (originalLogFormatterClassName == null) {
+            systemConfig.clearProperty(logFormatterKey)
+        } else {
+            systemConfig.setProperty(logFormatterKey, originalLogFormatterClassName)
+        }
+    }
+
+    def "The LogFormatterProvider instantiates the configured logFormatter instance"() {
+        given: "We haven't instantiated the log formatter yet"
+        assert LogFormatterProvider.logFormatter == null
```
cdeszaq: Can make this an `expect` block I think, and then not need the `assert`

fili-core/src/test/groovy/com/yahoo/bard/webservice/logging/LogFormatterProviderSpec.groovy
```diff
@@ -0,0 +1,67 @@
...
+        when:
+        systemConfig.setProperty(logFormatterKey, TestLogFormatter.class.name)
+
+        then:
+        LogFormatterProvider.instance instanceof TestLogFormatter
+    }
+
+    def "When the LogFormatterProvider is not specified, we default to the JsonLogFormatter"() {
+        given: "We haven't instantiated the log formatter yet"
+        assert LogFormatterProvider.logFormatter == null
```
cdeszaq: Ditto re: `expect` block

fili-core/src/test/groovy/com/yahoo/bard/webservice/logging/LogFormatterProviderSpec.groovy
archolewa: The reason I'm using a `given` and not an `expect` is because I'm asserting a precondition rather than performing any actual tests. To me, when I see an `expect` block I see "here be tests of this feature" not "here be preconditions."

Also, "given" is often used when specifying formal preconditions in english, (well the english used by mathematicians) so this reads very naturally as "given the logFormatter is null."

I agree that the `assert` is annoying, but I think this more clearly calls out that this is merely asserting a precondition for the test, and not actually testing anything.

michael-mclawhorn:
👍

PR #88, opened 2016-11-02
Refactored TimeGrainDictionary out of StandardGranularityParser so as…

michael-mclawhorn:
… to expose the dictionary for extension and modification in binding as well as being able to be gotten outside of the GranularityParser interface.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
@@ -15,6 +15,17 @@ import com.yahoo.bard.webservice.application.healthchecks.DataSourceMetadataLoad
 import com.yahoo.bard.webservice.application.healthchecks.DruidDimensionsLoaderHealthCheck;
 import com.yahoo.bard.webservice.application.healthchecks.SegmentMetadataLoaderHealthCheck;
 import com.yahoo.bard.webservice.application.healthchecks.VersionHealthCheck;
+import com.yahoo.bard.webservice.async.broadcastchannels.BroadcastChannel;
+import com.yahoo.bard.webservice.async.broadcastchannels.SimpleBroadcastChannel;
+import com.yahoo.bard.webservice.async.jobs.jobrows.DefaultJobField;
+import com.yahoo.bard.webservice.async.jobs.jobrows.DefaultJobRowBuilder;
+import com.yahoo.bard.webservice.async.jobs.jobrows.JobRowBuilder;
+import com.yahoo.bard.webservice.async.jobs.payloads.DefaultJobPayloadBuilder;
+import com.yahoo.bard.webservice.async.jobs.payloads.JobPayloadBuilder;
+import com.yahoo.bard.webservice.async.jobs.stores.ApiJobStore;
+import com.yahoo.bard.webservice.async.jobs.stores.NoOpApiJobStore;
+import com.yahoo.bard.webservice.async.preresponses.stores.NoOpPreResponseStore;
+import com.yahoo.bard.webservice.async.preresponses.stores.PreResponseStore;
 import com.yahoo.bard.webservice.async.workflows.AsynchronousWorkflowsBuilder;
 import com.yahoo.bard.webservice.async.workflows.DefaultAsynchronousWorkflowsBuilder;
 import com.yahoo.bard.webservice.config.BardFeatureFlag;
@@ -43,6 +54,7 @@ import com.yahoo.bard.webservice.data.filterbuilders.DefaultDruidFilterBuilder;
 import com.yahoo.bard.webservice.data.filterbuilders.DruidFilterBuilder;
 import com.yahoo.bard.webservice.data.metric.MetricDictionary;
 import com.yahoo.bard.webservice.data.metric.TemplateDruidQueryMerger;
+import com.yahoo.bard.webservice.data.time.GranularityDictionary;
 import com.yahoo.bard.webservice.data.time.GranularityParser;
 import com.yahoo.bard.webservice.data.time.StandardGranularityParser;
 import com.yahoo.bard.webservice.data.volatility.DefaultingVolatileIntervalsService;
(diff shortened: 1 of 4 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/time/StandardGranularityParser.java
```diff
@@ -5,39 +5,58 @@
...
+    public StandardGranularityParser(GranularityDictionary dictionary) {
+        namedGranularities = dictionary;
+    }
 
     /**
      * Constructor.
+     *
+     * Use a default grain map with the enum name()s and the all granularity.
      */
+    @Inject
```
cdeszaq: There shouldn't be 2 `@Inject` constructors

fili-core/src/main/java/com/yahoo/bard/webservice/data/time/StandardGranularityParser.java
```diff
@@ -5,39 +5,58 @@
...
+     * @param dictionary  a dictionary containing names mapped to granularities
+     */
+    @Inject
+    public StandardGranularityParser(GranularityDictionary dictionary) {
+        namedGranularities = dictionary;
+    }
 
     /**
      * Constructor.
+     *
```
cdeszaq: `<p>` should be used to separate paragraphs in JavaDoc, not blank lines

fili-core/src/main/java/com/yahoo/bard/webservice/data/time/StandardGranularityParser.java
```diff
@@ -5,39 +5,58 @@
...
     /**
      * This method loads default grains and can be extended to add customer grain extensions.
      *
      * @return A map of time grain api name to time grain instances.
      */
-    protected Map<String, Granularity> getGrainMap() {
-        Map<String, Granularity> result = Arrays.stream(DefaultTimeGrain.values())
-                .collect(Collectors.toMap(DefaultTimeGrain::name, Function.identity()));
+    public static GranularityDictionary getDefaultGrainMap() {
+        GranularityDictionary result = Arrays.stream(DefaultTimeGrain.values())
```
cdeszaq: If we expect this method to be called per-request (or at least in a request scope), we should memoize this method so that it doesn't need to re-build the same map over and over.

fili-core/src/main/java/com/yahoo/bard/webservice/data/time/StandardGranularityParser.java
```diff
@@ -5,39 +5,58 @@
...
+                .collect(
+                        StreamUtils.toMap(
+                                DefaultTimeGrain::name,
+                                Function.identity(),
+                                GranularityDictionary::new
+                        )
+                );
         result.put(AllGranularity.ALL_NAME.toUpperCase(Locale.ENGLISH), AllGranularity.INSTANCE);
         return result;
```
cdeszaq: Since we're in here anyways, can we clean this up to be just a single stream operation?:

```java
public static GranularityDictionary getDefaultGrainMap(
    return Stream.concat(
            Stream.of(AllGranularity.INSTANCE),
            Arrays.stream(DefaultTimeGrain.values())
    ).collect(
            StreamUtils.toDictionary(
                    Granularity::getName,
                    GranularityDictionary::new
            )
    );
}

public Granularity parseGranularity(String granularityName) throws GranularityParseException {
    String key = granularityName.toLowerCase(Locale.ENGLISH);
    if (namedGranularities.containsKey(key)) {
        return namedGranularities.get(key);
    }
    throw new GranularityParseException(granularityName);
}
```
Note that I switched to use `getName()` instead of `name` and am using `toDictionary` instead of `toMap`

And yes, you could unwrap the `toDictionary` call, it's fine either way I think

fili-core/src/main/java/com/yahoo/bard/webservice/data/time/StandardGranularityParser.java
```diff
@@ -5,39 +5,58 @@
...
      *
      * @return A map of time grain api name to time grain instances.
      */
-    protected Map<String, Granularity> getGrainMap() {
-        Map<String, Granularity> result = Arrays.stream(DefaultTimeGrain.values())
-                .collect(Collectors.toMap(DefaultTimeGrain::name, Function.identity()));
+    public static GranularityDictionary getDefaultGrainMap() {
+        GranularityDictionary result = Arrays.stream(DefaultTimeGrain.values())
+                .collect(
+                        StreamUtils.toMap(
```
cdeszaq: Can be simplified if you use `toDictionary` instead of `toMap` (no need to pass `Function.identity()`

fili-core/src/main/java/com/yahoo/bard/webservice/data/time/StandardGranularityParser.java
michael-mclawhorn: It's only being used to initialize the granularity parser, which should be memoized by the hk2 framework.  (I think)

fili-core/src/main/java/com/yahoo/bard/webservice/data/time/StandardGranularityParser.java
michael-mclawhorn: Wrong.  getName doesn't return the hard upper case from the enum.

cdeszaq: Correct, it lower-cases it. Which is why I also lower-cased the lookup call in `parseGranularity`. Does this still not work?

cdeszaq:
👍

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -893,6 +895,15 @@ protected JobPayloadBuilder buildJobPayloadBuilder() {
 
 
     /**
+     * Extension point for building a dictionary of base granularities for use in the system.
+     *
+     * @return A map of time grain api name (forced to uppercase) to time grain instances.
```
archolewa: This documentation claims that the timegrain name is forced to uppercase, but in the [lookup logic](https://github.com/yahoo/fili/pull/88/files#diff-b285206fd3b789c0f559db575ccff632R73) we force it to lowercase instead.

archolewa:
A small inconsistency between some documentation and implementation, but once that's addressed, 👍

Issue #89, opened 2016-11-07
Overwriting keyvalue store's key


sriram-viswanathan:
Reference: https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/util/DimensionStoreKeyUtils.java#L49

By lower-casing the value part, there could be issues where both `FOO` and `foo` will end up having same key which causes the druid query to always return results for only `foo` and miss `FOO`

archolewa:
Looks like this issue has been resolved by #90.

PR #90, opened 2016-11-07
Case sensitive keys for key value store

sriram-viswanathan:
Fix for https://github.com/yahoo/fili/issues/89

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,9 @@ Current
 
 ### Deprecated:
 
+- [`DimensionStoreKeyUtils` now supports case sensitive row and column keys](https://github.com/yahoo/fili/pull/90)
+    - Case insensitive row and column keys will be deprecated going forward.
+
 - [All constructors of `ResponseException` that do not take an `ObjectWriter`](https://github.com/yahoo/fili/pull/70)
     * An `ObjectWriter` is required in order to ensure that the exception correctly serializes its associated Druid query
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ Current
 
 ### Changed:
 
+- [`DimensionStoreKeyUtils` now supports case sensitive row and column keys](https://github.com/yahoo/fili/pull/90)
+    - Wrapped this config in a feature flag `case_sensitive_keys_enabled` which is set to `false` by default for backwards compatibility. 
+        This flag will be set to `true` in future versions.
+
 - [The `getGrainMap` method in `StandardGranularityParser` class is renamed to `getDefaultGrainMap` and is made public static.](https://github.com/yahoo/fili/pull/88)
     - Created new class `GranularityDictionary` and bind getGranularityDictionary to it
 
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/util/DimensionStoreKeyUtilsSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/util/DimensionStoreKeyUtilsSpec.groovy
@@ -3,39 +3,62 @@
 package com.yahoo.bard.webservice.util
 
 import com.yahoo.bard.webservice.config.BardFeatureFlag
+import spock.lang.Shared
 import spock.lang.Specification
+import spock.lang.Unroll
 
 class DimensionStoreKeyUtilsSpec extends Specification {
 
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,10 @@ Current
 
 ### Changed:
 
+- [`DimensionStoreKeyUtils` now supports case sensitive row and column keys](https://github.com/yahoo/fili/pull/90)
```
cdeszaq: Could we add a blurb to the `### Deprecated` section to, calling out that case-insensitive row retrieval is deprecated? It's unlikely that anyone really cares, but having it in the deprecated section will call a bit more attention to the change as well

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/DimensionStoreKeyUtilsSpec.groovy
```diff
@@ -0,0 +1,41 @@
...
+        expect:
+        DimensionStoreKeyUtils.getRowKey("FOO","1") == "foo_1_row_key"
+    }
+
+    def "Test getColumnKey() with CASE_SENSITIVE_KEYS_ENABLED off"() {
+        expect:
+        DimensionStoreKeyUtils.getColumnKey("FOO") == "foo_column_key"
+    }
+
+    def "Test getRowKey() with CASE_SENSITIVE_KEYS_ENABLED on"() {
```
cdeszaq: Ideally, test names reflect something about the behavior they are testing / asserting. With that in mind, could we update the test names in this file to be a bit more informative about what condition / behavior they are asserting exists?

For this particular case, it might be something like `getRowKey retains the case of the rowName when CASE_SENSITIVE_KEYS_ENABLED is off`

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/DimensionStoreKeyUtilsSpec.groovy
```diff
@@ -0,0 +1,41 @@
...
+        setup:
+        BardFeatureFlag.CASE_SENSITIVE_KEYS.setOn(true)
+
+        expect:
+        DimensionStoreKeyUtils.getRowKey("FOO","1") == "FOO_1_row_key"
```
cdeszaq: We should add tests like this where the `rowValue` parameter is a string, just to cover that base as well.

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/DimensionStoreKeyUtilsSpec.groovy
```diff
@@ -0,0 +1,41 @@
...
+package com.yahoo.bard.webservice.util
+
+import com.yahoo.bard.webservice.config.BardFeatureFlag
+import spock.lang.Specification
+
+class DimensionStoreKeyUtilsSpec extends Specification {
+
+    def "Test getRowKey() with CASE_SENSITIVE_KEYS_ENABLED off"() {
+        expect:
+        DimensionStoreKeyUtils.getRowKey("FOO","1") == "foo_1_row_key"
```
cdeszaq: It would be good to explicitly set the flag to off to ensure that we're testing with the condition we want to test, rather than relying on default config. Not a requirement, but it would be nice.

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/DimensionStoreKeyUtilsSpec.groovy
```diff
@@ -0,0 +1,41 @@
...
+
+    def "Test getColumnKey() with CASE_SENSITIVE_KEYS_ENABLED on"() {
+        setup:
+        BardFeatureFlag.CASE_SENSITIVE_KEYS.setOn(true)
+
+        expect:
+        DimensionStoreKeyUtils.getColumnKey("FOO") == "FOO_column_key"
+
+        cleanup:
+        BardFeatureFlag.CASE_SENSITIVE_KEYS.setOn(false)
```
cdeszaq: It would be good to capture the initial state of the flag in the `setup` block, and then re-set to that initial state in the `cleanup` block, rather than just force back to `false`

Note: Applies to the other places we're flipping the flag as well.

codingwhatever:
👍

PR #94, opened 2016-11-14
Clean up too large to cache exception

cdeszaq:
Adding the ability to not even try caching Druid responses that seem like they will be too big. This is fully optional, and defaults to not caching anything longer than `MAX_LONG`.

Note: This is built on the pending release. Once that merges, this PR will be updated to be based on `master` directly.

### DIFF
--- a/fili-core/src/main/resources/moduleConfig.properties
+++ b/fili-core/src/main/resources/moduleConfig.properties
@@ -53,6 +53,10 @@ bard__memcached_expiration_seconds = 3600
 # Data Cache
 bard__druid_cache_enabled = true
 
+# Maximum Druid response size to cache, in bytes. Defaults to MAX_LONG (9223372036854775807)
+# If using memcached, make sure this aligns with the setting memcached is using, overridden with it's -I parameter
+bard__druid_max_response_length_to_cache = 9223372036854775807
+
 # Data Cache V2 (needs the above flag set as well)
 bard__druid_cache_v2_enabled = true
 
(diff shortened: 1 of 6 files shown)


michael-mclawhorn:
👍  seems pretty reasonable.

fili-core/src/main/resources/moduleConfig.properties
```diff
@@ -53,6 +53,10 @@ bard__memcached_expiration_seconds = 3600
 # Data Cache
 bard__druid_cache_enabled = true
 
+# Maximum Druid response size to cache, in bytes. Defaults to MAX_LONG
+# If using memcached, make sure this aligns with the setting memcached is using, overridden with it's -I parameter
+bard__druid_max_response_length_to_cache = 9223372036854775807
```
garyluoex: Might want to explain where this number comes from....? If it is MAX_LONG, I suggest simply use the default provided by the code and remove the constant here and simply include a commented out variable.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/responseprocessors/CacheV2ResponseProcessorSpec.groovy
```diff
@@ -152,7 +156,38 @@ class CacheV2ResponseProcessorSpec extends Specification {
         then:
         2 * next.getResponseContext() >> responseContext
         1 * next.processResponse(json, groupByQuery, null)
-        0 * dataCache.set(cacheKey, '[]')
+        0 * dataCache.set(*_)
+    }
+
+    def "Overly long data doesn't cache and then continues"() {
+        setup: "Save the old max-length-to-cache so we can restore it later"
+        String max_druid_response_length_to_cache_key = SYSTEM_CONFIG.getPackageVariableName(
```
garyluoex: If we decided to comment out the variable, might need to include a default MAX_LONG here.

fili-core/src/main/resources/moduleConfig.properties
cdeszaq: There are 2 ways to default a property:

1. Have the default be set in the module config that the property belongs to
2. Set the default every place the code accesses the property

Since we can't easily know every place that might access the property, it's safer and easier to manage to set the default here in the module properties. That said, we _do_ indicate that the default is `MAX_LONG` 2 lines up from here, but it may not be clear that the huge number _is_ a long, so I've added that number after in parenthesis to make it more clear.

garyluoex:
👍

PR #96, opened 2016-11-16
Generalizes `RequestValidationException::getStatus`

archolewa:
--`Response.Status` does not contain a remotely complete list of HTTP
status codes, including 422: Unprocessable Entity. 422 is quite useful
when users have correct syntax but incorrect semantics (which may happen
when a Fili customer imposes table-specific restrictions in their system).

--By having `RequestValidationException::getStatus` return
`Response.StatusType` (an interface) rather than `Response.Status` (an
enum), customers may use other HTTP status codes by providing the
appropriate implementation of `Response.StatusType`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,10 @@ Current
 
 ### Changed:
 
-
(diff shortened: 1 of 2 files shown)


cdeszaq:
👍

michael-mclawhorn:
👍  !!!

PR #98, opened 2016-11-16
Can now inject a subclass of `DruidQueryBuilder`.

archolewa:
--The original return type did not allow returning the class of a
subtype of `DruidQueryBuilder` because template types are much more
rigid than standard Java types by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ Current
 
 ### Fixed:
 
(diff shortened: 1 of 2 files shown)


michael-mclawhorn:
👍 
Not sure I approve of the smartass changelog comment, but I'll allow it.

archolewa:
I consider it ok because I was the one who wrote the original code being fixed.

PR #99, opened 2016-11-17
Fix and refactor role based filter to allow CORS

garyluoex:
* Fix `RoleBasedAuthFilter` to bypass `OPTIONS` request for CORS
* Discovered a bug where `user_roles` is declared but unset still reads as a list with empty string (included a temporary fix by deleting the variable declaration)
* Refactored `RoleBasedAuthFilter` and `RoleBasedAuthFilterSpec` for better testing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,11 +28,15 @@ Current
 
 ### Fixed:
 
+- [Fix and refactor role based filter to allow CORS](https://github.com/yahoo/fili/pull/99)
+    * Fix `RoleBasedAuthFilter` to bypass `OPTIONS` request for CORS
+    * Discovered a bug where `user_roles` is declared but unset still reads as a list with empty string (included a temporary fix by commenting the variable declaration)
+    * Refactored `RoleBasedAuthFilter` and `RoleBasedAuthFilterSpec` for better testing
+
 - [Made a few injection points not useless](https://github.com/yahoo/fili/pull/98)
     * Template types don't get the same subclass goodness that method invocation and
         dependencies get, so this method did not allow returning a subclass of
        `DruidQueryBuilder` or of `DruidResponseParser`.
-
 
 ### Known Issues:
 
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilter.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilter.java
@@ -21,6 +21,7 @@ import java.util.Map;
 import javax.annotation.Priority;
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import javax.ws.rs.HttpMethod;
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerRequestFilter;
 import javax.ws.rs.core.Response;
@@ -50,14 +51,50 @@ public class RoleBasedAuthFilter implements ContainerRequestFilter {
     @Override
     @SuppressWarnings("checkstyle:cyclomaticcomplexity")
     public void filter(ContainerRequestContext containerRequestContext) throws IOException {
-        List<String> allowedUserRoles = getAllowedUserRoles();
-        if (!allowedUserRoles.isEmpty() && !isUserInRole(allowedUserRoles, containerRequestContext)) {
-                abortRequest(
-                        containerRequestContext,
-                        Response.Status.FORBIDDEN,
-                        BouncerAuthorizationStatus.ACCESS_DENIED
-                );
+
+        if (!isAuthorized(containerRequestContext)) {
+            abortRequest(
+                    containerRequestContext,
+                    Response.Status.FORBIDDEN,
+                    BouncerAuthorizationStatus.ACCESS_DENIED
+            );
         }
+    }
+
+    /**
+     * Request authorization logic, bypassing OPTIONS request and authorize if user is in authorized role or
+     * authorized role is not specified.
+     *
+     * @param containerRequestContext containing the request context information
+     *
+     * @return true if request is authorized, false otherwise
+     */
+    private boolean isAuthorized(ContainerRequestContext containerRequestContext) {
+        return isBypassMethod(containerRequestContext) || isUserInRole(getAllowedUserRoles(), containerRequestContext);
+    }
+
+    /**
+     * Returns true to bypass the filter if the request is a Http OPTIONS method to allow CORS by default.
+     *
+     * @param containerRequestContext Request context containing request method information
+     *
+     * @return true to bypass the filter, and false otherwise
+     */
+    protected boolean isBypassMethod(ContainerRequestContext containerRequestContext) {
+        return containerRequestContext.getMethod().equals(HttpMethod.OPTIONS);
+    }
+
+    /**
+     * Checks if an user role belongs to the list of allowed roles.
+     *
+     * @param allowedUserRoles List of allowed roles
+     * @param containerRequestContext Request context contains the user role information
+     *
+     * @return boolean based on the user role
+     */
+    protected boolean isUserInRole (List<String> allowedUserRoles, ContainerRequestContext containerRequestContext) {
+        SecurityContext securityContext = containerRequestContext.getSecurityContext();
+        return allowedUserRoles.isEmpty() || allowedUserRoles.stream().anyMatch(securityContext::isUserInRole);
     }
 
     /**
@@ -73,7 +110,6 @@ public class RoleBasedAuthFilter implements ContainerRequestFilter {
             BouncerAuthorizationStatus reason
     ) {
         LOG.debug("The user is not authorized in an authorized role ", status);
-
         Map<String, Object> responseMap = new LinkedHashMap<>();
         responseMap.put("status", status);
         responseMap.put("reason", reason.getName());
@@ -101,18 +137,5 @@ public class RoleBasedAuthFilter implements ContainerRequestFilter {
         } catch (SystemConfigException ignored) {
             return Collections.emptyList();
         }
-    }
-
-    /**
-     * Checks if an user role belongs to the list of allowed roles.
-     *
-     * @param allowedUserRoles List of allowed roles
-     * @param containerRequestContext Request context contains the user role information
-     *
-     * @return boolean based on the user role
-     */
-    public boolean isUserInRole (List<String> allowedUserRoles, ContainerRequestContext containerRequestContext) {
-        SecurityContext securityContext = containerRequestContext.getSecurityContext();
-        return allowedUserRoles.stream().anyMatch(securityContext::isUserInRole);
     }
 }
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilterSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilterSpec.groovy
@@ -2,8 +2,11 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.web.filters;
 
-import com.yahoo.bard.webservice.application.JerseyTestBinder
-import com.yahoo.bard.webservice.web.endpoints.DataServlet
+import com.yahoo.bard.webservice.config.SystemConfig
+import com.yahoo.bard.webservice.config.SystemConfigException
+import com.yahoo.bard.webservice.config.SystemConfigProvider
+
+import com.fasterxml.jackson.databind.ObjectMapper
 
 import spock.lang.Specification
 import spock.lang.Timeout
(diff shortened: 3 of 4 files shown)


fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilterSpec.groovy
```diff
@@ -15,49 +15,55 @@ import javax.ws.rs.core.SecurityContext
...
-    }
+        containerRequestContext = Mock(ContainerRequestContext)
+        securityContext = Mock(SecurityContext)
+        containerRequestContext.getSecurityContext() >> securityContext
 
-    def cleanup() {
-        // Release the test web container
-        jtb.tearDown()
+        roleBasedAuthFilter = new RoleBasedAuthFilter()
+        SYSTEM_CONFIG.setProperty(SYSTEM_CONFIG.getPackageVariableName("user_roles"), "foo,bar");
```
cdeszaq: We need to make sure this gets cleaned up and re-set to it's original value. Property leaks are SUCH a pain to find.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilterSpec.groovy
```diff
@@ -15,49 +15,55 @@ import javax.ws.rs.core.SecurityContext
...
 
-    JerseyTestBinder jtb
+    private final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance()
 
-    Class<?>[] getResourceClasses() {
-        [RoleBasedAuthFilterSpec.class, DataServlet.class]
-    }
+    ContainerRequestContext containerRequestContext;
+    SecurityContext securityContext;
+    RoleBasedAuthFilter roleBasedAuthFilter;
```
cdeszaq: No semicolons in groovy

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilterSpec.groovy
```diff
@@ -15,49 +15,55 @@ import javax.ws.rs.core.SecurityContext
...
-        jtb = new JerseyTestBinder(getResourceClasses())
-    }
+        containerRequestContext = Mock(ContainerRequestContext)
+        securityContext = Mock(SecurityContext)
+        containerRequestContext.getSecurityContext() >> securityContext
 
-    def cleanup() {
-        // Release the test web container
-        jtb.tearDown()
+        roleBasedAuthFilter = new RoleBasedAuthFilter()
```
cdeszaq: This needs to take an `ObjectMapper`

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilterSpec.groovy
```diff
@@ -15,49 +15,55 @@ import javax.ws.rs.core.SecurityContext
...
-        given: "A request that does not have the security context of an user in role"
-        ContainerRequestContext context = Mock(ContainerRequestContext)
-        SecurityContext securityContext = Mock(SecurityContext)
-        securityContext.isUserInRole("foo") >> true
-        context.getSecurityContext() >> securityContext
-        List<String> allowedUserRoles = ["baz"]
+    def "GET request is not authorized if the user is not in an allowed role"() {
+        when: "received a request and that the user is not in the authorized roles"
+        containerRequestContext.getMethod() >> "GET"
+        securityContext.isUserInRole() >> false
```
cdeszaq: This mock won't work as you're expecting. It shouldn't ever trigger because it's not valid to call that method with no parameters. If you want the mock to trigger on all calls to that method, you should use the wildcard parameter syntax: `securityContext.isUserInRole(*_)`

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilterSpec.groovy
```diff
@@ -15,49 +15,55 @@ import javax.ws.rs.core.SecurityContext
...
-        ContainerRequestContext context = Mock(ContainerRequestContext)
-        SecurityContext securityContext = Mock(SecurityContext)
-        securityContext.isUserInRole("foo") >> true
-        context.getSecurityContext() >> securityContext
-        List<String> allowedUserRoles = ["foo","bar"]
+    def "GET request is authorized if the `user_roles` setting is not set"() {
+        when: "received a request and that the user is not in the authorized roles"
+        containerRequestContext.getMethod() >> "GET"
+        SYSTEM_CONFIG.clearProperty(SYSTEM_CONFIG.getPackageVariableName("user_roles"))
+        securityContext.isUserInRole() >> false
```
cdeszaq: See previous note on making this trigger the way I think you want it to.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilterSpec.groovy
```diff
@@ -15,49 +15,55 @@ import javax.ws.rs.core.SecurityContext
...
-        securityContext.isUserInRole("foo") >> true
-        context.getSecurityContext() >> securityContext
-        List<String> allowedUserRoles = ["baz"]
+    def "GET request is not authorized if the user is not in an allowed role"() {
+        when: "received a request and that the user is not in the authorized roles"
+        containerRequestContext.getMethod() >> "GET"
+        securityContext.isUserInRole() >> false
+
+        then: "the request is not authorized"
+        roleBasedAuthFilter.isAuthorized(containerRequestContext) == false
```
cdeszaq: These assertions can be simplified, since they are all pretty much boolean responses. Instead of `== false`, you can just prepend `!` to the call, and you don't need `== true` if you're asserting truth.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilterSpec.groovy
```diff
@@ -15,49 +15,55 @@ import javax.ws.rs.core.SecurityContext
...
-        SecurityContext securityContext = Mock(SecurityContext)
-        securityContext.isUserInRole("foo") >> true
-        context.getSecurityContext() >> securityContext
-        List<String> allowedUserRoles = ["baz"]
+    def "GET request is not authorized if the user is not in an allowed role"() {
+        when: "received a request and that the user is not in the authorized roles"
+        containerRequestContext.getMethod() >> "GET"
+        securityContext.isUserInRole() >> false
+
+        then: "the request is not authorized"
```
cdeszaq: None of these tests are "stimulus / assertion on the state of the world" tests, but instead they are all "given the state of the world, assert a behavior", which lends itself much more to the `given / expect` style of tests, rather than the `when / then` style. So, it would be good to convert all of these tests to `given / expect`

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilterSpec.groovy
```diff
@@ -15,49 +15,55 @@ import javax.ws.rs.core.SecurityContext
...
+
+        then: "the request is authorized"
+        roleBasedAuthFilter.isAuthorized(containerRequestContext) == true
+    }
 
-        and: "A role based auth filter"
-        RoleBasedAuthFilter roleBasedAuthFilter = new RoleBasedAuthFilter()
+    def "OPTIONS request is authorized which bypasses the filter"() {
+        when: "received an OPTIONS request and that the user is in the authorized roles"
+        securityContext.isUserInRole() >> false
```
cdeszaq: May not be triggering when you expect it to

fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilter.java
```diff
@@ -113,6 +124,17 @@ private void abortRequest(
      */
     public boolean isUserInRole (List<String> allowedUserRoles, ContainerRequestContext containerRequestContext) {
         SecurityContext securityContext = containerRequestContext.getSecurityContext();
-        return allowedUserRoles.stream().anyMatch(securityContext::isUserInRole);
+        return allowedUserRoles.isEmpty() ? true : allowedUserRoles.stream().anyMatch(securityContext::isUserInRole);
+    }
+
+    /**\
+     *  By default, returns true to bypass the filter if the request is a Http OPTIONS method to allow CORS.
```
garyluoex: Making it `protected` to allow for override

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilterSpec.groovy
garyluoex: Is it necessary to include ObjectMapper? We are not testing the objectMapper functionality and even if we pass it, we are gonna just pass a Mock right?

fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilter.java
```diff
@@ -50,17 +51,53 @@ public RoleBasedAuthFilter(ObjectMapper mapper) {
...
+     * Checks if an user role belongs to the list of allowed roles.
+     *
+     * @param allowedUserRoles List of allowed roles
+     * @param containerRequestContext Request context contains the user role information
+     *
+     * @return boolean based on the user role
+     */
+    protected boolean isUserInRole (List<String> allowedUserRoles, ContainerRequestContext containerRequestContext) {
+        SecurityContext securityContext = containerRequestContext.getSecurityContext();
+        return allowedUserRoles.isEmpty() ? true : allowedUserRoles.stream().anyMatch(securityContext::isUserInRole);
```
cdeszaq: Can simplify this using `||` like so: `allowedUserRoles.isEmpty() || allowedUserRoles.stream().anyMatch(securityContext::isUserInRole);`

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilterSpec.groovy
```diff
@@ -15,49 +17,61 @@ import javax.ws.rs.core.SecurityContext
...
+        containerRequestContext.getSecurityContext() >> securityContext
+
+        roleBasedAuthFilter = new RoleBasedAuthFilter(objectMapper)
+        SYSTEM_CONFIG.setProperty(SYSTEM_CONFIG.getPackageVariableName("user_roles"), "foo,bar");
     }
 
     def cleanup() {
-        // Release the test web container
-        jtb.tearDown()
+        SYSTEM_CONFIG.clearProperty(SYSTEM_CONFIG.getPackageVariableName("user_roles"))
```
cdeszaq: We can't assume that this was unset when the test started. We need to capture the value of the property (as a `String`) in the `setup`, and then use that captured value to `reset(capturedValue)` in the cleanup.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilterSpec.groovy
```diff
@@ -15,49 +18,71 @@ import javax.ws.rs.core.SecurityContext
...
-        jtb = new JerseyTestBinder(getResourceClasses())
+        objectMapper = Mock(ObjectMapper)
+        containerRequestContext = Mock(ContainerRequestContext)
+        securityContext = Mock(SecurityContext)
+        containerRequestContext.getSecurityContext() >> securityContext
+        roleBasedAuthFilter = new RoleBasedAuthFilter(objectMapper)
+        
+        try {
+            oldUserRoles = Optional.of(SYSTEM_CONFIG.getStringProperty(userRolesSetting))
+        } catch(SystemConfigException e) {
```
cdeszaq: For `catch` blocks where we're ignoring the exception, it's best to call the exception `ignored` (or something like that) to make it clear  that we're intentionally not doing anything with the exception.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilterSpec.groovy
```diff
@@ -15,49 +18,71 @@ import javax.ws.rs.core.SecurityContext
 @Timeout(30)
 class RoleBasedAuthFilterSpec extends Specification {
 
-    JerseyTestBinder jtb
+    private static final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance()
+    private static final String userRolesSetting = SYSTEM_CONFIG.getPackageVariableName("user_roles")
```
cdeszaq: Since this variable is ` static`, it should be in `STATIC_CASE`

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilterSpec.groovy
```diff
@@ -15,49 +18,71 @@ import javax.ws.rs.core.SecurityContext
...
+            oldUserRoles = Optional.empty()
+        }
+        SYSTEM_CONFIG.setProperty(userRolesSetting, "foo,bar");
     }
 
     def cleanup() {
-        // Release the test web container
-        jtb.tearDown()
+        // Restore pre-test setting value if exists
+        oldUserRoles.ifPresent({userRoles -> SYSTEM_CONFIG.resetProperty(userRolesSetting, userRoles)})
```
cdeszaq: This needs to clear the setting if the `Optional` is empty, which this doesn't do.

cdeszaq:
👍

CHANGELOG.md
```diff
@@ -28,12 +28,16 @@ Current
 
 ### Fixed:
 
+- [Fix and refactor role based filter to allow CORS](https://github.com/yahoo/fili/pull/99)
+    * Fix `RoleBasedAuthFilter` to bypass `OPTIONS` request for CORS
```
RVRSKumar: Can we expand what `CORS` is?

fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilter.java
```diff
@@ -50,17 +51,53 @@ public RoleBasedAuthFilter(ObjectMapper mapper) {
...
     /**
+     * Request authorization logic, bypassing OPTIONS request and authorize if user is in authorized role or
+     * authorized role is not specified.
+     *
+     * @param containerRequestContext containing the request context information
+     *
+     * @return true if request is authorized, false otherwise
+     */
+    private boolean isAuthorized(ContainerRequestContext containerRequestContext) {
+        return isBypassMethod(containerRequestContext) || isUserInRole(getAllowedUserRoles(), containerRequestContext);
```
RVRSKumar: In `getAllowedUserRoles()` what is the list that is returned when the `bard__user_roles` variable is uncommented. Is it not an empty list?

garyluoex: returns it is a list with an empty string

CHANGELOG.md
garyluoex: I think changelog is not the place to explain things, google CORS will give plenty of answers. What do you think? @cdeszaq

fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilter.java
garyluoex: I was thinking we should just make it a bug issue and fix it later on, because it should be obvious that this behavior is undesirable.

RVRSKumar: 👍

RVRSKumar:
👍

PR #100, opened 2016-11-17
Expose the request phase durations stored by RequestLog

clayreimann:
* Add FiliTimingFilter to allow applications to use custom logging filters while retaining the non-logging related behaviors of the BardLoggingFilter
* Refactored the RequestLog by spinning the model out into a seperate class and marking the existing class as a utility

## New methods
* `RequestLogUtils.startTimingRequest`
* `RequestLogUtils.stopTimingRequest`

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
@@ -144,7 +144,7 @@ public class RequestLog {
      * @param phase the phase to be timed
      */
     protected void putPhase(TimedPhase phase) {
-        times.put(phase.name, phase);
+        times.put(phase.getName(), phase);
     }
 
     /**
@@ -156,12 +156,11 @@ public class RequestLog {
         return times.values()
                 .stream()
                 .peek(phase -> {
-                    if (!phase.isRunning()) {
-                        return;
+                    if (phase.isRunning()) {
+                        LOG.warn("Timer running during export. Measurement might be wrong: {}", phase.getName());
                     }
-                    LOG.warn("Exporting duration while timer is running. Measurement might be wrong: {}", phase.name);
                 })
-                .collect(Collectors.toMap(phase -> phase.name, phase -> phase.duration));
+                .collect(Collectors.toMap(TimedPhase::getName, TimedPhase::getDuration));
     }
 
     /**
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
@@ -2,31 +2,18 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.logging;
 
-import static com.yahoo.bard.webservice.druid.client.impl.AsyncDruidWebServiceImpl.DRUID_QUERY_ALL_TIMER;
-import static com.yahoo.bard.webservice.druid.client.impl.AsyncDruidWebServiceImpl.DRUID_QUERY_MAX_TIMER;
-import static com.yahoo.bard.webservice.druid.client.impl.AsyncDruidWebServiceImpl.DRUID_QUERY_TIMER;
-import static com.yahoo.bard.webservice.util.StreamUtils.not;
-import static com.yahoo.bard.webservice.web.handlers.workflow.DruidWorkflow.REQUEST_WORKFLOW_TIMER;
-import static com.yahoo.bard.webservice.web.handlers.workflow.DruidWorkflow.RESPONSE_WORKFLOW_TIMER;
-
+import com.codahale.metrics.MetricRegistry;
 import com.yahoo.bard.webservice.application.MetricRegistryFactory;
-import com.yahoo.bard.webservice.config.SystemConfig;
-import com.yahoo.bard.webservice.config.SystemConfigProvider;
 import com.yahoo.bard.webservice.logging.blocks.Durations;
 import com.yahoo.bard.webservice.logging.blocks.Preface;
 import com.yahoo.bard.webservice.logging.blocks.Threads;
 import com.yahoo.bard.webservice.web.ErrorMessageFormat;
-
-import com.codahale.metrics.MetricRegistry;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 
-import java.util.Arrays;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
-import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.OptionalLong;
@@ -35,40 +22,37 @@ import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
+import static com.yahoo.bard.webservice.druid.client.impl.AsyncDruidWebServiceImpl.DRUID_QUERY_ALL_TIMER;
+import static com.yahoo.bard.webservice.druid.client.impl.AsyncDruidWebServiceImpl.DRUID_QUERY_MAX_TIMER;
+import static com.yahoo.bard.webservice.druid.client.impl.AsyncDruidWebServiceImpl.DRUID_QUERY_TIMER;
+
 /**
- * Represents the logging framework that provides timing capabilities of arbitrary phases on the handling lifecycle of a
- * request and accumulation of information for such a request in a single mega log line.
+ * The RequestLog holds data that we would like to log about the request. In particular, various timings are
+ * accumulated here.
  */
 public class RequestLog {
-
-    private static final Logger LOG = LoggerFactory.getLogger(RequestLog.class);
-    private static final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance();
-
-    private static final String ID_KEY = "logid";
-    private static final long MS_PER_NS = 1000000;
     private static final MetricRegistry REGISTRY = MetricRegistryFactory.getRegistry();
-    private static final ThreadLocal<RequestLog> RLOG = ThreadLocal.withInitial(RequestLog::new);
-    private static final String LOGINFO_ORDER_STRING = SYSTEM_CONFIG.getStringProperty(
-            SYSTEM_CONFIG.getPackageVariableName("requestlog_loginfo_order"), ""
-    );
-    private static final List<String> LOGINFO_ORDER = generateLogInfoOrder(LOGINFO_ORDER_STRING);
+    private static final Logger LOG = LoggerFactory.getLogger(RequestLog.class);
+    private static final long MS_PER_NS = 1000000;
+    protected static final String LOG_ID_KEY = "logid";
 
-    private String logId;
-    private LogBlock info;
-    private TimedPhase mostRecentTimer;
-    private final Map<String, TimedPhase> times;
-    private final Set<String> threadIds;
+    protected String logId;
+    protected LogBlock info;
+    protected TimedPhase mostRecentTimer;
+    protected final Map<String, TimedPhase> times;
+    protected final Set<String> threadIds;
 
     /**
      * This class has only static methods and is not supposed to be directly instantiated.
      */
-    private RequestLog() {
+    protected RequestLog() {
         logId = null;
         info = null;
         mostRecentTimer = null;
         times = new LinkedHashMap<>();
         threadIds = new LinkedHashSet<>();
-        MDC.remove(ID_KEY);
+        MDC.remove(LOG_ID_KEY);
+        init();
     }
 
     /**
@@ -76,13 +60,13 @@ public class RequestLog {
      *
      * @param  rl request log object to copy from
      */
-    private RequestLog(RequestLog rl) {
+    protected RequestLog(RequestLog rl) {
         logId = rl.logId;
         info = rl.info;
         mostRecentTimer = rl.mostRecentTimer;
         times = new LinkedHashMap<>(rl.times);
         threadIds = new LinkedHashSet<>(rl.threadIds);
-        MDC.put(ID_KEY, logId);
+        MDC.put(LOG_ID_KEY, logId);
     }
 
     /**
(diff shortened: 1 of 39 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLogUtils.java
```diff
@@ -0,0 +1,352 @@
...
+ * An interface for application code to interact with the {@link RequestLog}.
+ */
+public class RequestLogUtils {
+    private static final String TOTAL_TIMER = "TotalTime";
+    private static final Logger LOG = LoggerFactory.getLogger(RequestLog.class);
+    private static final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance();
+    private static final List<String> LOGINFO_ORDER = generateLogInfoOrder();
+
+    private static final ThreadLocal<RequestLog> RLOG = ThreadLocal.withInitial(RequestLog::new);
+
```
michael-mclawhorn: This is an interesting way to do this.  I wonder if SystemConfig should be tweaked in a similar way.

michael-mclawhorn:
I'm pretty excited about this PR.  I haven't had a chance to deep dive review in micro, but the broad strokes look very appealing.

michael-mclawhorn:
CHANGELOG definitely needs to be updated.  Also, this will break every piece of downstream code, which won't make us many friends.

michael-mclawhorn:
@mangesh-pardeshi @codingwhatever 
Could you guys give me an estimate of how many places you're calling RequestUtils in your code?  If the answer is very little, I'm inclined to let this change go through.  If not, I'd like to know how much work this will produce for you.

codingwhatever:
I can't find any uses in our code base.

michael-mclawhorn:
@clayreimann Could I request you to rebase this forward on master and see if you have any build errors.  Out of band conversations look like this will have low impact on known adopters.

clayreimann:
@michael-mclawhorn @archolewa Rebased on master

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
```diff
@@ -2,31 +2,18 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.logging;
 
-import static com.yahoo.bard.webservice.druid.client.impl.AsyncDruidWebServiceImpl.DRUID_QUERY_ALL_TIMER;
-import static com.yahoo.bard.webservice.druid.client.impl.AsyncDruidWebServiceImpl.DRUID_QUERY_MAX_TIMER;
-import static com.yahoo.bard.webservice.druid.client.impl.AsyncDruidWebServiceImpl.DRUID_QUERY_TIMER;
-import static com.yahoo.bard.webservice.util.StreamUtils.not;
-import static com.yahoo.bard.webservice.web.handlers.workflow.DruidWorkflow.REQUEST_WORKFLOW_TIMER;
-import static com.yahoo.bard.webservice.web.handlers.workflow.DruidWorkflow.RESPONSE_WORKFLOW_TIMER;
```
archolewa: Again, the imports got messed up. The static imports shuld remain at the top, and the `com.codahale` import should go back to where it was originally.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
```diff
@@ -35,79 +22,76 @@
...
     /**
      * Represents a phase that is timed.
      * TimedPhase is used to associate a Timer located in the registry with the exact duration of such a phase for a
      * specific request.
      */
-    private static class TimedPhase {
-        private final String name;
-        private long start;
-        private long duration;
+    protected static class TimedPhase {
```
archolewa: Since we're moving all of the stopwatch methods out of `RequestLog`, I think we should move `TimedPhase` out as well, since that's what the stopwatch methods are manipulating.

In fact, rather than having a `RequestLogUtils` class, I think we should have a `TimedPhase` class. That class has the definition of `TimedPhase`, along with all the static methods for manipulating them (i.e. starting, stopping, switching).

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLogUtils.java
```diff
@@ -0,0 +1,352 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import static com.yahoo.bard.webservice.druid.client.impl.AsyncDruidWebServiceImpl.DRUID_QUERY_TIMER;
+import static com.yahoo.bard.webservice.logging.RequestLog.LOG_ID_KEY;
+import static com.yahoo.bard.webservice.util.StreamUtils.not;
+import static com.yahoo.bard.webservice.web.handlers.workflow.DruidWorkflow.REQUEST_WORKFLOW_TIMER;
+import static com.yahoo.bard.webservice.web.handlers.workflow.DruidWorkflow.RESPONSE_WORKFLOW_TIMER;
```
archolewa: The imports here are incorrect:

1. `static` imports should go at the very top.
2. Then `com.yahoo` in a separate paragraph
3. Then `com.*` where `* != yahoo` in separate paragraph
4. Then `org.*` in a separate paragraph
5. Then `java.*`
6. Then `javax.*`

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
```diff
@@ -171,316 +155,56 @@ private void init() {
...
-                            if (phase.start != 0) {
-                                LOG.warn(
-                                        "Exporting duration while timer is running. Measurement might be wrong: {}",
-                                        phase.name
-                                );
-                            }
-                        }
-                )
+                .peek(phase -> {
+                    if (phase.start != 0) {
```
archolewa: If we move the `TimedPhase` out into its own class, then we'll need to make sure to use `TimedPhase::isStarted` instead.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLogUtils.java
```diff
@@ -0,0 +1,352 @@
...
+    private static final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance();
+    private static final List<String> LOGINFO_ORDER = generateLogInfoOrder();
+
+    private static final ThreadLocal<RequestLog> RLOG = ThreadLocal.withInitial(RequestLog::new);
+
+    /**
+     * Thou shalt not.
+     */
+    private RequestLogUtils() {
+    }
```
archolewa: If we merge this class with `TimedPhase`, then this won't be necessary.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLogUtils.java
```diff
@@ -0,0 +1,352 @@
...
+    public static void record(LogInfo logPhase) {
+        try {
+            RLOG.get().info.add(logPhase);
+        } catch (NullPointerException ignored) {
+            LOG.warn(
+                    "Attempted to append log info while request log object was uninitialized: {}",
+                    logPhase.getClass().getSimpleName()
+            );
+        }
+    }
```
archolewa: If we do merge `TimedPhase` and then this class then we'll probably want to move `record` back onto `RequestLog` (or have a `RequestLogUtils` class), because this really doesn't have anything to do with `TimedPhase`.

This applies to the rest of the methods in this class (I don't have a strong opinion one way or the other).

fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/BardLoggingFilter.java
```diff
@@ -66,13 +65,13 @@
     ClientResponseFilter, WriterInterceptor {
 
     private static final Logger LOG = LoggerFactory.getLogger(BardLoggingFilter.class);
-    private static final String PROPERTY_NANOS = LoggingFilter.class.getName() + ".nanos";
```
archolewa: Since this class is now basically the `FiliTimingFilter++`, does it make sense for this filter to either have a `FiliTimingFilter` or be a `FiliTimingFilter`?

That would reduce duplicate code, as well as make it clearer that the `BardLoggingFilter` is doing the same thing as the `FiliTimingFilter` plus more.

fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/FiliTimingFilter.java
```diff
@@ -0,0 +1,67 @@
...
+import javax.annotation.Priority;
+import javax.inject.Singleton;
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
+import javax.ws.rs.container.ContainerResponseContext;
+import javax.ws.rs.container.ContainerResponseFilter;
+import javax.ws.rs.container.PreMatching;
+import javax.ws.rs.core.MultivaluedMap;
+import java.io.IOException;
+import java.util.regex.Pattern;
```
archolewa: [ditto](https://github.com/yahoo/fili/pull/100/files#r96657389)

fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/FiliTimingFilter.java
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Starts/stops the timing for a request as well as stashing the request-id header in the RequestLog. This filter
+ * is useful when you <b>do not</b> want to use the {@link BardLoggingFilter}.
+ */
+@PreMatching
+@Singleton
+@Priority(1)
+public class FiliTimingFilter implements ContainerResponseFilter, ContainerRequestFilter {
```
archolewa: This could really use a better name, since it isn't just doing timing (it's also stashing the request-id header). Maybe `FiliInitializationFilter`?

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/CacheRequestHandler.java
```diff
@@ -2,32 +2,30 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.web.handlers;
 
-import static com.yahoo.bard.webservice.web.handlers.workflow.DruidWorkflow.REQUEST_WORKFLOW_TIMER;
-import static com.yahoo.bard.webservice.web.handlers.workflow.DruidWorkflow.RESPONSE_WORKFLOW_TIMER;
-
```
archolewa: [ditto](https://github.com/yahoo/fili/pull/100/files#r96657389)

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/CacheV2RequestHandler.java
```diff
@@ -3,35 +3,33 @@
 package com.yahoo.bard.webservice.web.handlers;
 
 
-import static com.yahoo.bard.webservice.web.handlers.workflow.DruidWorkflow.REQUEST_WORKFLOW_TIMER;
-import static com.yahoo.bard.webservice.web.handlers.workflow.DruidWorkflow.RESPONSE_WORKFLOW_TIMER;
-
```
archolewa: [ditto](https://github.com/yahoo/fili/pull/100/files#r96657389)

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/SplitQueryRequestHandler.java
```diff
@@ -2,33 +2,31 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.web.handlers;
 
-import static com.yahoo.bard.webservice.web.ErrorMessageFormat.EMPTY_INTERVAL_FORMAT;
-
```
archolewa: [ditto](https://github.com/yahoo/fili/pull/100/files#r96657389)

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/SplitQueryResponseProcessor.java
```diff
@@ -2,17 +2,16 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.web.responseprocessors;
 
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
```
archolewa: [ditto](https://github.com/yahoo/fili/pull/100/files#r96657389)

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
clayreimann: I disagree, the stopwatch methods all operate in the context of the tread's `RequestLog`

fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/BardLoggingFilter.java
clayreimann: Yeah, I don't want to marry the behavior of the two though.

clayreimann:
@michael-mclawhorn @archolewa Rebased again and addressed @archolewa's comments

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
archolewa: Well, erm, hmm. I actually have a PR out to address a separate issue that pulls the `TimedPhase` out: https://github.com/yahoo/fili/pull/143

Basically, to make starting and stopping timers cleaner, I had the `TimedPhase` implement auto-closeable, so that we could use Java's spiffy new try-with-resource block. But that required making the `TimedPhase` publicly available, and I'm not generally a fan of having public inner classes vs. just regular public classes. Don't really believe that they buy you much but code clutter.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
```diff
@@ -167,11 +156,10 @@ protected void init() {
...
                 .stream()
                 .peek(phase -> {
-                    if (phase.start != 0) {
-                        LOG.warn(
-                                "Exporting duration while timer is running. Measurement might be wrong: {}", phase.name
-                        );
+                    if (!phase.isRunning()) {
+                        return;
                     }
+                    LOG.warn("Exporting duration while timer is running. Measurement might be wrong: {}", phase.name);
```
archolewa: This can probably be simplified by using a positive conditional.

```java
if (phase.isRunning()) 
{
   LOG.warn("Exporting duration while timer is running. Measurement might be wrong: {}", phase.name);
}
```

fili-core/src/main/java/com/yahoo/bard/webservice/logging/TimedPhase.java
```diff
@@ -0,0 +1,64 @@
...
+     */
+    protected TimedPhase(String name) {
+        this.name = name;
+    }
+
+    /**
+     * Start the phase.
+     */
+    protected void start() {
+        if (start != 0) {
```
archolewa: Should probably replace `start != 0` with `isRunning()`.

PR #102, opened 2016-11-19
Including slf4j/logback

kkalyan:
wikipedia-example is failing because slf4j/logback are not included in the jar.


cdeszaq:
👍 Makes sense. In many cases, it's anticipated that the Fili application will be running inside a container like Jetty which likely will provide the dependency, but it doesn't make sense for the Wikipedia Example, which runs in embedded jetty.

Thanks for the fix!

cdeszaq:
This fix has been released in v0.7.4. Thanks again for the contribution!

PR #104, opened 2016-11-19
YAML-based configuration

jacobtolar:
This change is quite large (sorry about that...) but there are (almost) zero changes to the core code, so at least it shouldn't break any existing functionality; it just adds new broken functionality :).

* Adds a set of interfaces for setting up Fili via configuration (`ConfigProvider`). You can currently configure metrics, dimensions, physical and logical tables, and register custom metric makers. Notably, I've never had to set up dimension loading or search providers before, so those aren't supported yet but could presumably be added without too much trouble.

* There is also an implementation of AbstractBinderFactory that uses reflection to instantiate a specified ConfigProvider and provide all the necessary Fili configuration. With this, you can get Fili up and running in front of your Druid instance without writing a single line of code.

* There is an implementation of `ConfigProvider` for a YAML-driven configuration via Jackson's yaml support (jackson-dataformat-yaml).

* There's a new subproject, 'fili-yaml', that is set up to with the ConfigBinderFactory (this repo contains no actual java code). It just builds a war that can be configured easily with yaml. It also includes a couple scripts and a Dockerfile that can be used to stand up Fili via Docker, and a few instructions on how to do so. 

* Distinct from the YAML config, there is also a metric parser included (it's used by the YAML config provider, but could also be used independently of it). It lets you define metrics in a special DSL; for example, `(impressions + 3)/5 | gender == 2` would produce a filtered aggregate. There's not really any docs on the DSL included in this PR yet, but it's hopefully fairly intuitive. It doesn't support 100% of what you could do programmatically, but it's hopefully a good start.

* One of the few changes to existing code is that `ShardSpecMixIn` was updated to include 'linear' which is used for streaming.

* A couple assorted documentation updates.

* Updated the ClassScanner to handle array argument types.

The test coverage isn't perfect, but most things have at least one test :) (Notable exception: the metric parser; I've written tests for most of the classes it uses, just not the top-level parser). I have also tested this against our Druid instance, which was where I found the ShardSpecMixIn change. 

There's also a couple things in here (especially the metric parser) that I'm not 100% sure I have correct -- Fili/Druid-isms that I didn't dig into fully. I tried to make notes when I wasn't sure of something so there's a few FIXMEs scattered around.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/parser/MetricParser.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/parser/MetricParser.java
@@ -1,0 +1,297 @@
+// Copyright 2016 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+
+package com.yahoo.bard.webservice.data.config.metric.parser;
+
+import com.yahoo.bard.webservice.data.config.metric.parser.lexer.LexException;
+import com.yahoo.bard.webservice.data.config.metric.parser.lexer.Lexeme;
+import com.yahoo.bard.webservice.data.config.metric.parser.lexer.LexemeType;
+import com.yahoo.bard.webservice.data.config.metric.parser.operand.ConstantMetricNode;
+import com.yahoo.bard.webservice.data.config.metric.parser.operand.IdentifierNode;
+import com.yahoo.bard.webservice.data.config.metric.parser.operand.MetricNode;
+import com.yahoo.bard.webservice.data.config.metric.parser.operand.Operand;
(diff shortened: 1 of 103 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/parser/MetricParser.java
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    /**
+     * Generate a globally unique new name.
+     * <p>
+     * Note: this assumes that we've chosen a sufficiently unique prefix.
+     *
+     * @return the new name
+     */
+    public synchronized static String getNewName() {
```
jacobtolar: This is my least favorite line of code in the whole thing, but I forgot to add a FIXME.

jacobtolar:
Will be split into several separate pull requests.

PR #106, opened 2016-11-21
Fix SegmentMetaDataLoader to use provided headers for requests

garyluoex:
* `SegmentMetadataLoader` sends requests with the provided request headers in `AsyncDruidWebservice` now
* Refactored `AsyncDruidWebserviceSpec` test and added test for checking `getJsonData` includes request headers as well

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,10 @@ Current
 
 ### Fixed:
 
+- [SegmentMetadataLoader include provided request headers](https://github.com/yahoo/fili/pull/106)
+    * `SegmentMetadataLoader` sends requests with the provided request headers in `AsyncDruidWebservice` now
+    * Refactored `AsyncDruidWebserviceSpec` test and added test for checking `getJsonData` includes request headers as well
+
 - [Fix and refactor role based filter to allow CORS](https://github.com/yahoo/fili/pull/99)
     * Fix `RoleBasedAuthFilter` to bypass `OPTIONS` request for CORS
     * Discovered a bug where `user_roles` is declared but unset still reads as a list with empty string (included a temporary fix by commenting the variable declaration)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ Current
 
 ### Changed:
 
+- [SegmentMetadataLoader include provided request headers](https://github.com/yahoo/fili/pull/106)
+    * `SegmentMetadataLoader` sends requests with the provided request headers in `AsyncDruidWebservice` now
+    * Refactored `AsyncDruidWebserviceSpec` test and added test for checking `getJsonData` includes request headers as well
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -28,6 +28,10 @@ Current
 
 ### Fixed:
 
+- [SegmentMetadataLoader include provided request headers](https://github.com/yahoo/fili/pull/106)
```
cdeszaq: Should be a change, not a fix.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplSpec.groovy
```diff
@@ -14,38 +14,52 @@ import spock.lang.Specification
...
 
 class AsyncDruidWebServiceImplSpec extends Specification {
-    def "Ensure that headersToAppend are added"() {
+    def "Ensure that headersToAppend are added to request when calling postDruidQuery"() {
         setup:
         WeightEvaluationQuery weightEvaluationQuery = Mock(WeightEvaluationQuery)
         QueryContext queryContext = Mock(QueryContext)
-        weightEvaluationQuery.getContext() >> { queryContext }
-        queryContext.numberOfQueries() >> { 1 }
-        queryContext.getSequenceNumber() >> { 1 }
```
cdeszaq: Why can we lose these 2 lines?

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplSpec.groovy
```diff
@@ -14,38 +14,52 @@ import spock.lang.Specification
...
+        for (Map.Entry<String, String> header : expectedHeaders) {
+            assert actualHeaders.get(header.getKey()) == header.getValue()
         }
+    }
+
+    def "Ensure that headersToAppend are added to request when calling getJsonObject"() {
+        setup:
+        Map<String, String> expectedHeaders = new HashMap<>()
+        expectedHeaders.put("k1", "v1")
+        expectedHeaders.put("k2", "v2")
```
cdeszaq: It would be better to use Groovy's native Map literal syntax:

```groovy
Map<String, String> expectedHeaders = [
    k1: "v1"
]
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplSpec.groovy
garyluoex: Because of [default behavior of mocks](http://spockframework.org/spock/docs/1.0/interaction_based_testing.html#_default_behavior_of_mock_objects), so it will just return the default type of whatever it returns.

will-lauer:
:+1:

PR #107, opened 2016-11-21
Assorted fili cleanups

jacobtolar:
* Add support for linear ShardSpecs
* Add documentation to metadata loader settings
* Update ClassScanner to be able to instantiate arrays
* Add a new FilteredAggregationMaker

Note: this is the first child of #104

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/FilteredAggregationMaker.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/FilteredAggregationMaker.java
@@ -25,7 +25,7 @@ public class FilteredAggregationMaker extends MetricMaker {
 
     /**
      * Construct a metric maker for Filtered Aggregations.
-     *
+     * <p>
      * @param metricDictionary the metric dictionary
      * @param aggregation the underlying aggregation
      * @param filter the filter to filter the aggregation by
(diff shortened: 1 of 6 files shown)


michael-mclawhorn:
👍 
This stuff seems fine.  I'd like to get this out of the way and dig into the next PR somewhat more critically.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/FilteredAggregationMaker.java
```diff
@@ -0,0 +1,59 @@
...
+     * Construct a metric maker for Filtered Aggregations.
+     *
+     * @param metricDictionary the metric dictionary
+     * @param aggregation the underlying aggregation
+     * @param filter the filter to filter the aggregation by
+     */
+    public FilteredAggregationMaker(MetricDictionary metricDictionary, Aggregation aggregation, Filter filter) {
+        super(metricDictionary);
+        this.aggregation = aggregation;
+        this.filter = filter;
```
cdeszaq: Perhaps it would be better to pass in `ApiFilter` objects instead of Druid's filter model? The idea is that defining metrics is a logical operation, not a physical operation tied to Druid. Since the `FilteredAggregation` ultimately needs a Druid filter, however, this means that you'd also need to pass in a `DruidFilterBuilder` instance.

If it's the case that the API filters are not rich enough, passing in a `Collection<ApiFilter>` would likely also work as well.

Note: I'm also OK coming back to this when we re-work this maker, as long as this is mentioned in that ticket.

jacobtolar:
Added a bunch of <p> flags and rebased to latest master.

Filed https://github.com/yahoo/fili/issues/120 to address the FilteredAggregationMaker concerns.

cdeszaq:
Merged. Thanks for the contribution!

PR #108, opened 2016-11-21
Fili config provider interface

jacobtolar:
Adds the config provider interface (this is the second child of #104) to bard core. There are a couple FIXMEs in here that should probably be addressed. What @michael-mclawhorn was talking about would be a more extensive change, I think.

Because of github silliness, this actually includes #107 so you'll probably just want to look at the most recent commit (https://github.com/yahoo/fili/commit/37d48b66a797d897f308140433eb66a98e904d80 at the moment). I can rebase when/if things are merged.


jacobtolar:
rebased from master after #107 was merged

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/provider/MakerDictionary.java
```diff
@@ -0,0 +1,295 @@
...
+    /**
+     * Instantiate an object from a class name and parameters.
+     *
+     * @param type the class to instantiate
+     * @param args the constructor arguments
+     * @param <T> the type of object to return
+     * @return an instantiated object, or null
+     */
+    public static <T> T instantiateObjectFromArgs(Class<? extends T> type, Object... args) {
+
```
michael-mclawhorn: This is cool code. But it doesn't seem tightly bound to the concept of making Makers.  Maybe relocate the reflections into a Utility class.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/provider/MetricConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+import com.yahoo.bard.webservice.data.metric.LogicalMetric;
+import com.yahoo.bard.webservice.data.metric.MetricDictionary;
+
+/**
+ * A metric with a definition, description, and flag to exclude from external API
+ *
+ * FIXME: The isExcluded() flag is currently unused.
+ */
+public interface MetricConfiguration {
+
```
michael-mclawhorn: Perhaps this class doesn't need to know about temporary dictionaries.  Perhaps load everything into a single dictionary and then filter out things that need filtering as a clean up stage.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/provider/ConfigProvider.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    ConfigurationDictionary<DimensionConfig> getDimensionConfig();
+
+    /**
+     * Get the base metrics.
+     *
+     * @return a mapping of base metric names to their definitions
+     */
+    ConfigurationDictionary<MetricConfiguration> getBaseMetrics();
+
```
michael-mclawhorn: What is the use of distinguishing here between base and derived metrics?

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/provider/ConfiguredTableLoader.java
```diff
@@ -0,0 +1,111 @@
...
+
+            // Get the smallest granularity for this table
+            // TimeGrain smallestGrain = Collections.min(entry.getValue().timeGrains);
+
+            // Fixme: We should use 'apiName' vs 'asName' as intended
+            // FIXME: should be possible to override grains valid for?
+            Set<ApiMetricName> apiMetricNames = new HashSet<>();
+            for (String metric : logicalTable.getMetrics()) {
+                apiMetricNames.add(new ApiMetricName() {
+                    @Override
```
michael-mclawhorn: It might make sense to extract this ApiMetricName factory code into a method so as to make it more easily tested.  It's also not strictly necessary to give every ApiMetricName it's own anonymous class.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/provider/MakerDictionary.java
```diff
@@ -0,0 +1,295 @@
...
+     * @param dict MetricDictionary to use for every maker
+     * @param mDict Dictionary of function name to MetricMaker
+     * @param makers user-defined config listing custom MetricMakers
+     * @param skipDefault true to skip loading built-in metrics
+     */
+    public static void loadMetricMakers(
+            MetricDictionary dict,
+            MakerDictionary mDict,
+            Map<String, MakerConfiguration> makers,
+            boolean skipDefault
```
michael-mclawhorn: Is there a reason you're not using your custom class 'ConfigurationDictionary<T>' here instead of crafting a raw map?

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/provider/MakerDictionary.java
```diff
@@ -0,0 +1,295 @@
...
+     * Find the full package + class name for a maker.
+     *
+     * The class name chosen for the custom maker is one of:
+     *
+     * 1. The full package + class name specified in the className field
+     * 2. The name specified in the className field inside the MetricMaker package
+     * 3. If no className given, the Maker name specified (longSum), turned into a Maker class name
+     *    (LongSumMaker), inside the MetricMaker package
+     *
+     * @param suppliedMakerName The maker name the user specified
```
michael-mclawhorn: This is a good start.  Another option for number 3 would be to draw on a list of configured maker dictionaries from config.  Not needed in the first draft.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/provider/PhysicalTableConfiguration.java
```diff
@@ -0,0 +1,31 @@
...
+import com.yahoo.bard.webservice.data.config.names.FieldName;
+import com.yahoo.bard.webservice.data.config.table.PhysicalTableDefinition;
+
+import java.util.Set;
+
+/**
+ * A PhysicalTableConfiguration can create its PhysicalTableDefinition, given a dimension configuration.
+ */
+public interface PhysicalTableConfiguration {
+
```
michael-mclawhorn: Do you have any implementing code for this class? 
What's the set of assumptions about how the metrics will be provided and how the dimension map will be used?

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/provider/FuzzyQueryMatcher.groovy
```diff
@@ -0,0 +1,135 @@
...
+import com.yahoo.bard.webservice.druid.model.postaggregation.ArithmeticPostAggregation
+import com.yahoo.bard.webservice.druid.model.postaggregation.ConstantPostAggregation
+import com.yahoo.bard.webservice.druid.model.postaggregation.FieldAccessorPostAggregation
+import com.yahoo.bard.webservice.druid.model.postaggregation.PostAggregation
+
+/**
+ * Utility class to validate that TemplateDruidQueries have the same structure and values.
+ *
+ * It's ok if inner objects don't always have the same names.
+ */
```
michael-mclawhorn: What are the situations where names are non-deterministic?  Filtered aggregations?

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/provider/MakerDictionarySpec.groovy
```diff
@@ -0,0 +1,40 @@
...
+        dictionary.containsKey("longSum")
+    }
+
+    def "Should construct configured makers correctly"(){
+        setup:
+         def maker = MakerDictionary.buildCustomMaker("thetaMaker", ThetaSketchMaker.getName(), [1] as Object[], new MetricDictionary())
+
+        expect:
+        maker instanceof ThetaSketchMaker
+    }
```
michael-mclawhorn: You could go further and verify that the size of the sketch is correct as well.

MetaAshley:
This is an automatically generated message.

awolf@yahoo-inc.com is no longer with Yahoo! Inc.

Your message will not be forwarded.

If you have a sales inquiry, please email yahoosales@yahoo-inc.com and someone will follow up with you shortly.

If you require assistance with a legal matter, please send a message to legal-notices@yahoo-inc.com

Thank you!

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/provider/MakerDictionary.java
jacobtolar: No, except that I think I created `ConfigurationDictionary` after I wrote this code. Could be changed.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/provider/MakerDictionary.java
jacobtolar: Cool/horrifying, there's a fine line :) But you're right, it could obviously construct any type of object.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/provider/MetricConfiguration.java
jacobtolar: Hmm. I'll play around with it; I don't like how it works currently, either.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/provider/ConfigProvider.java
jacobtolar: I can't remember 😄 

I believe in the metric parsing code (not here) there's some assumptions that require them to be separate (same with the temporary metric dictionary somewhere else). I'll see if I can refactor it a bit so that this distinction isn't necessary.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/provider/ConfiguredTableLoader.java
jacobtolar: Whoops, I thought I did that already - thanks

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/provider/PhysicalTableConfiguration.java
jacobtolar: Yeah, just not in this PR.

The implementation I have for the YAML config provider: 
* Takes metric names, dimension names, and time grain in the constructor. 

* In buildPhysicalTable, you get a map of actual constructed dimension config objects. So the `dimensionMap` is used to resolve dimension names to DimensionConfig objects so a `PhysicalTableDefinition` can be built.

```
      @Override
      public PhysicalTableDefinition buildPhysicalTable(ConfigurationDictionary<DimensionConfig> dimensionMap) {
 
          Set<DimensionConfig> dimSet = dimensions
                  .stream()
                  .filter(dimensionMap::containsKey)
                  .map(dimensionMap::get)
                  .collect(Collectors.toSet());
 
          return new PhysicalTableDefinition(name, grain, dimSet);
      }
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/provider/FuzzyQueryMatcher.groovy
jacobtolar: It could be that there *shouldn't* be any, but the code I've written currently behaves that way (that code isn't in this PR). 

In the code I've written to parse/create logical metrics it creates temporary names for inner values (just increments a counter). Maybe there's a better way to do this, but it's simple/works.

So, something like `real_views = 2 * half_views + 7` would name the final logical metric `real_views` but would assign some temporary name to the inner nested metrics (e.g. `2*half_views`).

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/provider/MakerDictionarySpec.groovy
jacobtolar: Sure.

jacobtolar:
To address at least most of the discussion thus far (thanks @michael-mclawhorn):

* The global temporary metric dictionary is gone, replaced by just building makers on the fly
* The distinction between base/derived metrics is gone. Up to the user to supply them in the right order.
* Changed everything in the config provider interface to be lists of objects with a 'name' parameter.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/provider/PhysicalTableConfiguration.java
michael-mclawhorn: It looks like this interface is a a PhysicalDefinitionFactory.  But you're calling it a PhysicalTableConfiguration.
As best as I can infer, this is going to be backed by a 'smart bean', where the underlying bean is initialized from a configuration model (e.g. YAML file, xml, database query, etc), and this INTERFACE discusses using that model AS a factory.

Maybe this is a conceptual confusion, or maybe this is a naming issue.

If you have an object modeling a thing that provides name, grain, dim names, metric names, maybe that bean should be an explicit thing. (e.g. a PhysicalTableConfiguration) and there should be another thing that implements what you've described above pulling the parts out of the model object and building a definition out of it.

michael-mclawhorn: So I've thought about this some more.
I'm actually frustrated by some of the assumptions in the BaseTableLoader which this code has been forced to adapt to.

Are you imagining that in your config model that the list of field names will be strings in the tableDefinitionConfig, or that they'll be implicitly loaded in some other way?

jacobtolar: Hmm, some of the other classes are like this, too (e.g. the MetricConfiguration interface has a `build` method that returns a logical metric).

For physical tables, it would be straightforward to clear this up by exposing dimensions and remove this method. The `PhysicalTableDefinition` would be created by the consumer of this class (e.g. `ConfiguredTableLoader` or something like that).

For metrics, it's a little trickier. I didn't want to force use of the metric parser so I wanted implementers of `MetricConfiguration` to be able to specify how a metric is built. I guess I could split that out into a MetricConfiguration and some sort of MetricFactory.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/provider/ConfigBinderFactory.java
```diff
@@ -0,0 +1,93 @@
...
+        if (className == null) {
+            throw new ConfigurationError("Unable to read class name property from config: " + CONF_TYPE);
+        }
+
+        try {
+            // oof
+            LOG.info("Loading ConfigBinderFactory for type: {}", className);
+            Class<? extends ConfigProvider> providerClass = (Class<? extends ConfigProvider>) Class.forName(className);
+            Method build = providerClass.getDeclaredMethod("build", SystemConfig.class);
+            provider = (ConfigProvider) build.invoke(null, SYSTEM_CONFIG);
```
jacobtolar: I guess it's not strictly necessary to pass in the system config here. The config provider can get at it if it wants it.

jacobtolar:
@michael-mclawhorn @cdeszaq 

I moved this under `examples/fili-yaml-config-example` and rebased to the latest master.

This breaks the metric parser PR a bit because it depends on the "metric maker dictionary" stuff being in fili-core.

PR #110, opened 2016-11-28
Adds some additional timings.

archolewa:
--In order to better aid in performance optimizations, some additional
timing has been added to the request processing workflow.

--Based on past experiences, most slow queries spend the majority of
their time building a `DataApiRequest`, so timings have been added
around all the major steps in building a `DataApiRequest`.

--Furthermore, it's expected that the bulk of the time spent building a
`DataApiRequest` is spent resolving filters, so some additional timings
have been added to the `LuceneSearchProvider`, the provider most
commonly used for large dimensions.

--It's also possible that a lot of the request workflow is spent
serializing large Druid queries, so some timings have been added around
serializing Druid queries.


fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -200,40 +207,70 @@ public DataApiRequest(
...
+        this.intervals = timeGenerator(
+                () -> generateIntervals(intervals, this.granularity, dateTimeFormatter),
+                "GeneratingIntervals"
+        );
 
         this.filterBuilder = bardConfigResources.getFilterBuilder();
 
+        MetricDictionary metricDictionary = bardConfigResources
+                .getMetricDictionary()
+                .getScope(Collections.singletonList(tableName));
```
archolewa: This change was necessary to make `metricDictionary` effectively final for use in lambdas (also the previous approach was confusing).

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -545,8 +546,11 @@ private Query getFilterQuery(Set<ApiFilter> filters, int perPage) {
         TreeSet<DimensionRow> filteredDimRows;
         int documentCount;
         initializeIndexSearcher();
+        RequestLog.startTiming("LuceneReadLock");
         lock.readLock().lock();
+        RequestLog.stopTiming("LuceneReadLock");
```
cdeszaq: We shouldn't time this. A read-lock should be fast enough that the timings will actually be much slower than the lock. If we _really_ want to get a sense of how much time this takes, we should run a profiler on this running under an actual workload.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -631,8 +636,11 @@ private TopDocs getPageOfData(
             int perPage,
             int currentPage
     ) {
+        RequestLog.startTiming("QueryingLuceneForPage" + currentPage + "ReadLock");
         lock.readLock().lock();
+        RequestLog.stopTiming("QueryingLuceneForPage" + currentPage + "ReadLock");
```
cdeszaq: Don't time read-locks

archolewa:
Note that I tried to add timings just to steps of building the `DataApiRequest` that I felt was interesting or complex enough to be worth timing.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -587,6 +591,7 @@ private Query getFilterQuery(Set<ApiFilter> filters, int perPage) {
             throw new RuntimeException(e);
```
cdeszaq: Can't comment in the right place, but there are effectively 2 steps in this search operation:

1. Get the Lucene results for the right page
2. Hydrate the dim rows

I would wrap a timer around each of those high-level steps

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/ConjunctionDruidFilterBuilder.java
```diff
@@ -42,7 +43,10 @@ public Filter buildFilters(Map<Dimension, Set<ApiFilter>> filterMap) throws Dime
 
         List<Filter> dimensionFilters = new ArrayList<>(filterMap.size());
         for (Map.Entry<Dimension, Set<ApiFilter>> entry : filterMap.entrySet()) {
+            String filterTimerName = "Building" + entry.getKey().getApiName() + "Filters";
+            RequestLog.startTiming(filterTimerName);
```
cdeszaq: This is going to make a _ton_ of timers and _massively_ fill the request log. Don't do this per-event. Do it at the outer layer.

If you _really_ want per-element timings, use the raw Metrics timings framework (ie. not requestLog), or write down a LogInfo object that includes the count of how many filters are being built

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -631,8 +636,11 @@ private TopDocs getPageOfData(
             int perPage,
             int currentPage
     ) {
+        RequestLog.startTiming("QueryingLuceneForPage" + currentPage + "ReadLock");
         lock.readLock().lock();
+        RequestLog.stopTiming("QueryingLuceneForPage" + currentPage + "ReadLock");
         try {
+            RequestLog.startTiming("QueryingLuceneForPage" + currentPage);
```
cdeszaq: If there are many pages, that could create a large bloat to the request log. I would move this timing up a level. You can calculate the per-page timings using the pagination information that I _think_ should already be in the RequestLog.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -298,9 +298,12 @@ public void postDruidQuery(
     ) {
         String entityBody;
         try {
+            RequestLog.startTiming("DruidQuerySerialization");
             entityBody = writer.writeValueAsString(druidQuery);
         } catch (JsonProcessingException e) {
             throw new IllegalStateException(e);
+        } finally {
+            RequestLog.stopTiming("DruidQuerySerialization");
```
cdeszaq: This should include the sequence number (see the timing we do below). Otherwise, when stitching requests back together, we may have trouble.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
cdeszaq: How so? Couldn't you just mark the existing local variable `final` and make the compiler happy?

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -290,6 +327,24 @@ public DataApiRequest(
...
     /**
+     * Times the execution of the specified producer, making sure to stop timing in a finally clause.
+     *
+     * @param generator  The producer to be timed
+     * @param timerName  The name to use for the timer
+     * @param <T>  The type of the object produced
+     *
+     * @return The value returned by the generator
+     */
+    private <T> T timeGenerator(Producer<T> generator, String timerName) {
```
cdeszaq: While this solution is cute, it uglies up the calls everywhere, making the code much harder to understand. Instead, if we want to do this, we should use annotations on the "generator" methods. Take a look at the `@Timed` annotation in the Metrics library to see how they are doing it. Having a similar annotation for the `RequestLog` may be exactly what we're looking for.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/AsyncWebServiceRequestHandler.java
```diff
@@ -51,7 +51,12 @@ public void invoke(JsonNode rootNode) {
         HttpErrorCallback error = response.getErrorCallback(druidQuery);
         FailureCallback failure = response.getFailureCallback(druidQuery);
 
-        druidWebService.postDruidQuery(context, success, error, failure, druidQuery);
+        try {
+            RequestLog.startTiming("PostingDruidQuery");
```
cdeszaq: This timing should be moved into the actual druid client (and it may already be there). Also, I don't think it'll time what you want, since it's only going to time how long the request takes to _send_, not send and come back. (yay async!)

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
archolewa: Because we were originally assigning the `metricDictionary` variable at the very beginning of the method, but didn't do anything with it until we reassigned it to the above value just before actually using it.

Since we were reassigning it, it was no longer effectively final. So I just got rid of the unused initialization.

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/ConjunctionDruidFilterBuilder.java
archolewa: FYI, we already have a `LogInfo` object that includes a count of how many filters are being built (the `Filter` LogBlock).

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
archolewa: Why can't Java annotations be like Python decorators? Why???

cdeszaq:
Annotations don't require an aop framework at all. They are much simpler than that. Look at how the Timed annotation from Metrics does what it does. It should be pretty much what we want to do.

cdeszaq:
I'm not thinking we'd time all methods, just those that we annotate.

archolewa:
I _did_ look into how `@Timed` works. And everything I found says "You need AspectJ or Spring AOP to use them." 

Furthermore, all the examples I found for writing custom annotations seem to assume you have your own main method where you can scan files for annotations and do something, rather than doing something special on method invocation.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
cdeszaq: Fair enough. I dug in a bit deeper too and it seems you are indeed right. That said, we should open an issue to revisit this, since it would help the RequestLog piece immensely to be able to be annotation-driven, rather than code-driven.

That said, we should make sure the timings are internal to the methods that are getting timed, since callers don't really need this complexity, and the method is better suited to handling it. (note: I've not looked at your updates yet, so this may already be the case).

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/AsyncWebServiceRequestHandler.java
cdeszaq: Like I said, this should move into the methods doing the sending, rather than the caller. Also, the timer needs to take into account the sequence number of the query, I think, in order for log merging to work correctly.

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/ConjunctionDruidFilterBuilder.java
cdeszaq: It would be better to add another block, rather than combine them I think. Especially as we look to the possibility of breaking the Druid-specific aspects into it's own module...

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
archolewa: I agree 100% that making the RequestLog annotation-based would be a Good Thing(TM).

I haven't actually pushed the timings into the methods. I'll do that now.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
cdeszaq: I still don't see a timing differentiation between talking to Lucene and fetching the dimension rows using the Lucene results. Did you want to break those up?

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -562,7 +564,7 @@ private Query getFilterQuery(Set<ApiFilter> filters, int perPage) {
                     throw new PageNotFoundException(requestedPageNumber, perPage, 0);
                 }
             }
-
```
cdeszaq: Keep this whitespace

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
cdeszaq: Ahh, I see, this is just a straight refactor, no actual changes. got it.

archolewa:
#111 issue to make the `RequestLog` annotation-based.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
archolewa: 😆 I changed where I stopped the timer for querying Lucene to capture just the code that queries Lucene, but forgot to add a timer for hydrating the rows.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/AsyncWebServiceRequestHandler.java
michael-mclawhorn: This sounds like something tricky enough to justify a DruidServlet and a test case to make sure that all the phases get tested.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/blocks/DruidFilterInfo.java
```diff
@@ -0,0 +1,52 @@
+package com.yahoo.bard.webservice.logging.blocks;
```
cdeszaq: Copyright needed

archolewa:
I know. I died a little with every line I wrote. My higher-order timing function is still an option ;)

michael-mclawhorn:
What the status on this?  Is the job error being pursued?

archolewa:
@michael-mclawhorn The status is "Oh crap the quarter's almost over and I have other things that need to get done!"

archolewa:
@michael-mclawhorn The status is now "I addressed the things."

michael-mclawhorn:
👍  Seems reasonable.

fili-core/src/test/groovy/com/yahoo/bard/webservice/async/AsyncInvalidApiRequest.groovy
```diff
@@ -37,7 +37,8 @@ class AsyncInvalidApiRequest extends AsyncFunctionalSpec {
 
     @Override
     Map<String, Closure<Void>> getResultAssertions() {
-        [ data: { assert GroovyTestUtils.compareJson(it.readEntity(String), EXPECTED_ERROR_MESSAGE) } ]
+        [ data: { assert GroovyTestUtils.compareErrorPayload(it.readEntity(String), EXPECTED_ERROR_MESSAGE) } ]
```
cdeszaq: What impact does this change have?

fili-core/src/main/java/com/yahoo/bard/webservice/logging/blocks/DruidFilterInfo.java
```diff
@@ -0,0 +1,66 @@
...
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.logging.blocks;
+
+import com.yahoo.bard.webservice.druid.model.filter.MultiClauseFilter;
+import com.yahoo.bard.webservice.druid.model.filter.NotFilter;
+import com.yahoo.bard.webservice.logging.LogInfo;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+
+import com.yahoo.bard.webservice.druid.model.filter.Filter;
```
cdeszaq: Move up.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/blocks/DruidFilterInfo.java
```diff
@@ -0,0 +1,66 @@
...
+        }
+        Map<String, Long> filterTypeCounter = new LinkedHashMap<>();
+        Deque<Filter> filterStack = new ArrayDeque<>();
+        filterStack.add(filter);
+        while (!filterStack.isEmpty()) {
+            Filter currentFilter = filterStack.pop();
+            filterTypeCounter.put(
+                    currentFilter.getClass().getSimpleName(),
+                    filterTypeCounter.getOrDefault(currentFilter.getClass().getSimpleName(), 0L) + 1L
+            );
```
cdeszaq: Can be replaced with `Map::merge`:

```java
filterTypeCounter.merge(currentFilter.getClass().getSimpleName(), 1L, (old, increment) -> old + increment);
```

fili-core/src/main/java/com/yahoo/bard/webservice/logging/blocks/DruidFilterInfo.java
```diff
@@ -0,0 +1,66 @@
...
+            if (currentFilter instanceof MultiClauseFilter) {
+                filterStack.addAll(((MultiClauseFilter) currentFilter).getFields());
+            } else if (currentFilter instanceof NotFilter) {
+                filterStack.add(((NotFilter) currentFilter).getField());
+            }
```
cdeszaq: This smells a little like extensibility pain... It would be great if Java made it easy to ask if things had various methods, but about the best we can do is add interfaces for specific methods. To that end, do we want to make `HasFields` and `HasField` interfaces, apply those interfaces to the `MultiVlauseFilter` and `NotFilter` (respectively), and then update this conditional to depend on those interfaces instead of these more concrete classes?

Specifically, it's the `NotFilter` dependency that drew my attention, since the `MultiClauseFilter` is abstract already and (roughly) intended to serve the same purpose as the interface would.

archolewa: If we are going to have that kind of interface, I'd much rather just have `HasFields` (`Not` can just return a singleton set). Though I'd prefer a better name than `HasFields`. I found that kind of convention hideously ugly. Perhaps something like `ComplexFilter`? i.e. a Filter that's built out of other filters?

cdeszaq: To me, `Complex` isn't specific enough. `HasFilters` is nice, in a way, because it fits (roughly) the bean convention of `getFilters` / `setFilters` roughly translating to the `filters` property present in Groovy and C# for example.

That said, I'm happy with `ComplexFilter` meaning "has a `getFilters` method". In part, I think we're fighting the generality of Druid's filters and their decision to use "field" to mean "another filter".

cdeszaq:
Checkstyle needs fix... otherwise I think this is good to squash into logical commits and merge.

PR #112, opened 2016-11-29
Remove restriction for single physical dimension to multiple lookup dimension

garyluoex:
* Modified physical dimension name to logical dimension name mapping into a `Map<String, List<String>>` instead of `Map<String, String>` in `PhysicalTable.java`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Current
 
 ### Changed:
 
+- [Remove restriction for single physical dimension to multiple lookup dimensions](https://github.com/yahoo/fili/pull/112)
+    * Modified physical dimension name to logical dimension name mapping into a `Map<String, List<String>>` instead of `Map<String, String>` in `PhysicalTable.java`
+
 - [SegmentMetadataLoader include provided request headers](https://github.com/yahoo/fili/pull/106)
     * `SegmentMetadataLoader` sends requests with the provided request headers in `AsyncDruidWebservice` now
     * Refactored `AsyncDruidWebserviceSpec` test and added test for checking `getJsonData` includes request headers as well
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ Current
 ### Changed:
 
 - [Remove restriction for single physical dimension to multiple lookup dimensions](https://github.com/yahoo/fili/pull/112)
-    * Modified physical dimension name to logical dimension name mapping into a `Map<String, List<String>>` instead of `Map<String, String>` in `PhysicalTable.java`
+    * Modified physical dimension name to logical dimension name mapping into a `Map<String, Set<String>>` instead of `Map<String, String>` in `PhysicalTable.java`
 
 - [SegmentMetadataLoader include provided request headers](https://github.com/yahoo/fili/pull/106)
     * `SegmentMetadataLoader` sends requests with the provided request headers in `AsyncDruidWebservice` now
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
@@ -23,6 +23,7 @@ import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
@@ -143,16 +144,16 @@ public class PhysicalTable extends Table {
 
                 getLogicalColumnNames(physicalName).stream()
                         .map(dimensionDictionary::findByApiName)
-                        .filter(dimension -> dimension != null)
+                        .filter(Objects::nonNull)
                         .forEach(
-                                dimension -> {
-                                    DimensionColumn dimensionColumn = DimensionColumn.addNewDimensionColumn(
-                                            this,
-                                            dimension
-                                    );
-                                    workingIntervals.put(dimensionColumn, nameIntervals.getValue());
-                                }
-                        );
+                        dimension -> {
+                            DimensionColumn dimensionColumn = DimensionColumn.addNewDimensionColumn(
+                                    this,
+                                    dimension
+                            );
+                            workingIntervals.put(dimensionColumn, nameIntervals.getValue());
+                        }
+                );
             }
             for (Map.Entry<String, Set<Interval>> nameIntervals : metricIntervals.entrySet()) {
                 MetricColumn metricColumn = MetricColumn.addNewMetricColumn(this, nameIntervals.getKey());
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
@@ -16,10 +16,8 @@ import org.joda.time.ReadablePeriod;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.Map;
@@ -146,14 +144,14 @@ public class PhysicalTable extends Table {
                         .map(dimensionDictionary::findByApiName)
                         .filter(Objects::nonNull)
                         .forEach(
-                        dimension -> {
-                            DimensionColumn dimensionColumn = DimensionColumn.addNewDimensionColumn(
-                                    this,
-                                    dimension
-                            );
-                            workingIntervals.put(dimensionColumn, nameIntervals.getValue());
-                        }
-                );
+                                dimension -> {
+                                    DimensionColumn dimensionColumn = DimensionColumn.addNewDimensionColumn(
+                                            this,
+                                            dimension
+                                    );
+                                    workingIntervals.put(dimensionColumn, nameIntervals.getValue());
+                                }
+                        );
             }
             for (Map.Entry<String, Set<Interval>> nameIntervals : metricIntervals.entrySet()) {
                 MetricColumn metricColumn = MetricColumn.addNewMetricColumn(this, nameIntervals.getKey());
@@ -247,7 +245,7 @@ public class PhysicalTable extends Table {
      * @return Translated physicalName if applicable
      */
     private Set<String> getLogicalColumnNames(String physicalName) {
-        return physicalToLogicalColumnNames.getOrDefault(physicalName, new HashSet<>(Arrays.asList(physicalName)));
+        return physicalToLogicalColumnNames.getOrDefault(physicalName, Collections.singleton(physicalName));
     }
 
     /**
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
@@ -141,8 +141,7 @@ public class PhysicalTable extends Table {
             for (Map.Entry<String, Set<Interval>> nameIntervals : dimensionIntervals.entrySet()) {
                 String physicalName = nameIntervals.getKey();
 
-                getLogicalColumnName(physicalName).stream()
-                        .sequential()
+                getLogicalColumnNames(physicalName).stream()
                         .map(dimensionDictionary::findByApiName)
                         .filter(dimension -> dimension != null)
                         .forEach(
@@ -246,7 +245,7 @@ public class PhysicalTable extends Table {
      * @param physicalName  Physical name to lookup in physical table
      * @return Translated physicalName if applicable
      */
-    private Set<String> getLogicalColumnName(String physicalName) {
+    private Set<String> getLogicalColumnNames(String physicalName) {
         return physicalToLogicalColumnNames.getOrDefault(physicalName, new HashSet<>(Arrays.asList(physicalName)));
     }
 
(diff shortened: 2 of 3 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -38,7 +40,7 @@
     private Map<Column, Set<Interval>> workingIntervals;
     private final Object mutex = new Object();
     private final Map<String, String> logicalToPhysicalColumnNames;
-    private final Map<String, String> physicalToLogicalColumnNames;
+    private final Map<String, List<String>> physicalToLogicalColumnNames;
```
archolewa: Any particular reason we're making this a `List` rather than a `Collection`?

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -137,14 +141,17 @@ public synchronized void resetColumns(SegmentMetadata segmentMetadata, Dimension
...
+                List<String> apiNames = getLogicalColumnName(physicalName);
+
+                for (String apiName : apiNames) {
+                    Dimension dimension = dimensionDictionary.findByApiName(apiName);
+                    // Schema evolution may lead to unknown dimensions, skip these
+                    if (dimension == null) {
+                        continue;
+                    }
+                    DimensionColumn dimensionColumn = DimensionColumn.addNewDimensionColumn(this, dimension);
+                    workingIntervals.put(dimensionColumn, nameIntervals.getValue());
```
archolewa: You can streamify this easily enough, I think:

```java
 getLogicalColumnName(physicalName).stream()
     .map(dimensionDictionary::findByApiName)
     .filter(dimension -> dimension != null)
     .forEach(dimension -> {
           DimensionColumn dimensionColumn = DimensionColumn.addNewDimensionColumn(this, dimension);
           workingIntervals.put(dimensionColumn, nameIntervals.getValue()); 
     });
```

Note that you could put `DimensionColumn.addNewDimensionColumn(this, dimension)` in a `map` rather than the first statement of a `forEach`, but `addNewDimensionColumn` has side-effects, and side-effecting operations in a `map` operation feels ugly to me.

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -237,8 +244,8 @@ public String getPhysicalColumnName(String logicalName) {
      * @param physicalName  Physical name to lookup in physical table
      * @return Translated physicalName if applicable
      */
-    private String getLogicalColumnName(String physicalName) {
-        return physicalToLogicalColumnNames.getOrDefault(physicalName, physicalName);
+    private List<String> getLogicalColumnName(String physicalName) {
```
archolewa: Again, this could probably be a `Collection` rather than a `List`.

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
garyluoex: Does `Set` fit better (since it should be a unique set of logical dimensions)? I'm not sure what are the advantages of just using `Collection`, fill me in if you have time, thanks!

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -58,10 +60,12 @@ public PhysicalTable(
         this.workingIntervals = Collections.synchronizedMap(new HashMap<>());
         this.logicalToPhysicalColumnNames = Collections.unmodifiableMap(logicalToPhysicalColumnNames);
         this.physicalToLogicalColumnNames = Collections.unmodifiableMap(
-                this.logicalToPhysicalColumnNames
-                        .entrySet()
-                        .stream()
-                        .collect(StreamUtils.toLinkedMap(Map.Entry::getValue, Map.Entry::getKey))
+                this.logicalToPhysicalColumnNames.entrySet().stream().sequential().collect(
```
garyluoex: I think removing the `sequential()` here is fine right? Will remove it in the next commit if no one complains.

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
archolewa: It's really more about avoiding overspecification. The places we use it only needs a `Collection` (we're iterating over every element). And at least in the current implementation we don't seem overly concerned with uniqueness. Tough having the same dimension twice could lead to weirdness if `addNewDimensionColumn` isn't idempotent.

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
archolewa: I believe `sequential` just forces the stream to be sequential, even if it was originally parallel. However, by default the stream of a set is sequential, so `sequential` doesn't actually do anything. 

So yes, I would suggest removing the sequential.

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
michael-mclawhorn: I don't think we want to have any headaches around nested types.  Let's just use set for now.

garyluoex: Ok, will make it `Set` since it makes sense to add a uniqueness constraint and doesn't make sense that there are duplicates.

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -244,8 +246,8 @@ public String getPhysicalColumnName(String logicalName) {
      * @param physicalName  Physical name to lookup in physical table
      * @return Translated physicalName if applicable
      */
-    private List<String> getLogicalColumnName(String physicalName) {
-        return physicalToLogicalColumnNames.getOrDefault(physicalName, Arrays.asList(physicalName));
+    private Set<String> getLogicalColumnName(String physicalName) {
+        return physicalToLogicalColumnNames.getOrDefault(physicalName, new HashSet<>(Arrays.asList(physicalName)));
     }
```
michael-mclawhorn: Pluralize this method name.  `getLogicalColumnNames`

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ Current
 
 ### Changed:
 
+- [Remove restriction for single physical dimension to multiple lookup dimensions](https://github.com/yahoo/fili/pull/112)
+    * Modified physical dimension name to logical dimension name mapping into a `Map<String, List<String>>` instead of `Map<String, String>` in `PhysicalTable.java`
+
```
michael-mclawhorn: Keep in synch with the code change.

michael-mclawhorn:
The fact that this didn't break a test worries me.  Please add a test to see that the correct set of logical column names is returned from the method in both the single-dimension-to-physical-name situation and the multiple-dimension-to-physical-name situation.

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -137,14 +140,20 @@ public synchronized void resetColumns(SegmentMetadata segmentMetadata, Dimension
...
-                Dimension dimension = dimensionDictionary.findByApiName(apiName);
-                // Schema evolution may lead to unknown dimensions, skip these
-                if (dimension == null) {
-                    continue;
-                }
-                DimensionColumn dimensionColumn = DimensionColumn.addNewDimensionColumn(this, dimension);
-                workingIntervals.put(dimensionColumn, nameIntervals.getValue());
+
+                getLogicalColumnName(physicalName).stream()
+                        .sequential()
```
garyluoex: It seems like this is extra again, will remove in the next commit again.

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -137,14 +140,19 @@ public synchronized void resetColumns(SegmentMetadata segmentMetadata, Dimension
...
-                // Schema evolution may lead to unknown dimensions, skip these
-                if (dimension == null) {
-                    continue;
-                }
-                DimensionColumn dimensionColumn = DimensionColumn.addNewDimensionColumn(this, dimension);
-                workingIntervals.put(dimensionColumn, nameIntervals.getValue());
+
+                getLogicalColumnNames(physicalName).stream()
+                        .map(dimensionDictionary::findByApiName)
+                        .filter(dimension -> dimension != null)
```
cdeszaq: `Objects::isNotNull` simplifies this I think

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -137,14 +140,19 @@ public synchronized void resetColumns(SegmentMetadata segmentMetadata, Dimension
...
+                        .forEach(
+                                dimension -> {
+                                    DimensionColumn dimensionColumn = DimensionColumn.addNewDimensionColumn(
```
cdeszaq: This can be a `.map()` call on the stream, since you're just transforming the `Dimension` into a `DimensionColumn`

garyluoex: But the operations will cause side-affects right (adding things to existing things)? so if the stream runs in parallel things will get synchronization problem. `forEach` is a reduce operation that allows side-affects I believe. Let me know if I'm wrong or not making sense, thanks!

archolewa: I'm with @garyluoex on this one. It's not a big deal, especially since we don't run around randomly making streams parallel (and we shouldn't, not without carefully examining the operations in the stream first), but it just feels deeply and fundamentally __wrong__ to me to put a side-effectful operation in a `map`.

archolewa: To elaborate on why it feels wrong to me, I generally prefer to be hit across the head with a big neon sign when an operation has side-effects, _especially_ when the operation returns a value as well. forEach does that, while a `map` operation does not.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/PhysicalTableSpec.groovy
```diff
@@ -150,4 +150,14 @@ class PhysicalTableSpec extends Specification {
         expect:
         physicalTable.getDimensions() == [dimension] as Set
     }
+
+    def "test physical to logical mapping is constructed correctly"() {
+        setup:
+        def oneDimPhysicalTable = new PhysicalTable("test table", DAY.buildZonedTimeGrain(UTC), ['dimension':'druidDim'])
+        def twoDimPhysicalTable = new PhysicalTable("test table", DAY.buildZonedTimeGrain(UTC), ['dimension1':'druidDim', 'dimension2':'druidDim'])
```
archolewa: Unless we have good reason to be taking advantage of Groovy's duck-typing, we should be explicitly specifying the types of the variables.

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -237,8 +246,8 @@ public String getPhysicalColumnName(String logicalName) {
      * @param physicalName  Physical name to lookup in physical table
      * @return Translated physicalName if applicable
      */
-    private String getLogicalColumnName(String physicalName) {
-        return physicalToLogicalColumnNames.getOrDefault(physicalName, physicalName);
+    private Set<String> getLogicalColumnNames(String physicalName) {
+        return physicalToLogicalColumnNames.getOrDefault(physicalName, new HashSet<>(Arrays.asList(physicalName)));
```
archolewa: Should use `Collections.singleton`.

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -137,14 +141,19 @@ public synchronized void resetColumns(SegmentMetadata segmentMetadata, Dimension
...
-                    continue;
-                }
-                DimensionColumn dimensionColumn = DimensionColumn.addNewDimensionColumn(this, dimension);
-                workingIntervals.put(dimensionColumn, nameIntervals.getValue());
+
+                getLogicalColumnNames(physicalName).stream()
+                        .map(dimensionDictionary::findByApiName)
+                        .filter(Objects::nonNull)
+                        .forEach(
+                        dimension -> {
```
archolewa: The indentation seems off here. `dimension ->` should be indented eight spaces, and everything inside the lambda should be indented four spaces past that.

archolewa:
Had a few small comments, 👍 once they've been addressed.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/PhysicalTableSpec.groovy
archolewa: At least in the context of our project, it's considered good practice (I don't know enough about the larger Groovy community to know if that's true more generally). 

If nothing else, it improves IDE support, because the IDE knows what the variable type is. 

It can also make things more readable, because it's easy for readers to see what the variable type this (though this can be partially offset because `def` is more concise, so which is more readable is going to be a matter of opinion). 

Duck typing has its advantages, but it has its disadvantages too. Here we're not using any of the advantages, but we're still saddled with the disadvantages.

garyluoex: ok will change it then, thanks!

PR #113, opened 2016-12-01
Warn for empty segment metadata load

### LINKED ISSUE #109: `NoMatchFoundException` should be a 500 error, not a 400 error
The `NoMatchFoundException` gets thrown if we're not able to find a matching physical table for a request. Right now, that results in a `400 Bad Request` response, but it really isn't the client's fault, since at that point we should already have verified the request against the `LogicalTable`.

The exception is really saying that there wasn't a `PhysicalTable` that satisfied the request, which is really a server-side problem. Typically, it means that there was a misconfiguration, or that Fili isn't detecting the columns needed to satisfy the request from Druid.

Here's an example of the problematic response I'm talking about:

```json
{
    "status": 400,
    "statusName": "Bad Request",
    "reason": "com.yahoo.bard.webservice.table.resolver.NoMatchFoundException",
    "description": "Table 'unsatisfiedTable' is incompatible with the dimensions '[logicalDimension, missingDimensions]', metrics '[someMetric, missingMetric]' and granularity 'day' requested.",
    "druidQuery": null,
    "requestId": "117c68bd-f9f7-41a9-9869-1374fe2e66f9"
}
```

cdeszaq:
There were some issues around segment metadata loading and table selection if the segment metadata that got loaded was empty. These changes should make that issue more obvious.

Also, these changes convert the response in that case from a `400 BAD REQUEST` into a `500 INTERNAL SERVER ERROR`. This is a more correct response, since this isn't the user's fault, but is the application's fault. 

**NOTE THAT THIS IS AN API CHANGE**, though a small one, since it's only on the errors. (fixes #109)

There is also some general cleanup in here, as well as some more stream utility methods that I built while playing in this code. I didn't end up using them, but felt it was a shame to chuck them, so I tossed them into the `StreamUtil` anyways.


fili-core/src/main/java/com/yahoo/bard/webservice/util/StreamUtils.java
```diff
@@ -145,4 +146,137 @@
...
+    public static <A, B, K, V, R extends Map<A, B>, T extends Map<K, V>> Function<R, T> mapMap(
+            Function<? super Map.Entry<A, B>, ? extends K> keyMapper,
+            Function<? super Map.Entry<A, B>, ? extends V> valueMapper,
+            Supplier<T> resultSupplier
+    ) {
+        return (R input) -> input.entrySet().stream()
+                .collect(
+                        resultSupplier,
+                        (result, entry) -> result.put(keyMapper.apply(entry), valueMapper.apply(entry)),
+                        Map::putAll
```
michael-mclawhorn: The problem with this is that if the keymapper maps keys into a smaller space, then the ordering of the entries will potentially alter the results on the mapping, somewhat unpredictably.  A more general case of this should probably take a collision managing joiner.

michael-mclawhorn:
I'm a little confused by some of the inclusions in this PR.

Is the new mapMap code being used anywhere?
Is there a reason this one includes the big checkstyle tweaks on the ErrorMessageFormat?

If not, just break these up into separate PRs from the fix, please.

cdeszaq:
@michael-mclawhorn As I mentioned in the PR description (last paragraph), [the additions to the `StreamUtil` class](https://github.com/yahoo/fili/pull/113/commits/cae53a57ce1fe797a5ee26e0f1cc8356d0523758) (like `mapMap`) are not currently being used anywhere:

> ... as well as some more stream utility methods that I built while playing in this code. I didn't end up using them, but felt it was a shame to chuck them, so I tossed them into the `StreamUtil` anyways.

I did try to break up the discontinuous changes in this PR into their own [commits](https://github.com/yahoo/fili/pull/113/commits), including putting all of the [cleanup changes](https://github.com/yahoo/fili/pull/113/commits/ecd0be7631c9d9b5b9c17ac8e0a7a63982b055c7)  into their own commit, but if you'd like even more separation, I'm happy to do so, just let me know which commits you'd like to be in their own PRs.

michael-mclawhorn:
Enh.  I can live with the other cleanups, but put the stream utils change into it's own PR, please.  I agree with you that it's cool, but I think it muddies the seriousness of an API change.  The rest of the PR is cohesive enough.

fili-core/src/main/java/com/yahoo/bard/webservice/util/StreamUtils.java
cdeszaq: Good catch. Part of the point of this method is to smooth over the "common parts" of mapping the keys or the values of a map, like processing the entry set, and then collecting back into a map. It's not really intended for doing any reducing (that's why it's a `Function` instead of a `Collector`), but that's not really knowable from the method name or javadoc.

So, do you think it's sufficient to just add a bit to the doc indicating that this uses a "putAll" combiner (and what that means)? And do you think I need a version that allows for providing a combiner?

michael-mclawhorn: It's not the putAll that's potentially an error, it's the result.put(...) that can create collisions.

If you explicitly throw an exception in the closure on key existence, you can prevent accidental combining.
Otherwise, the safe operation, if reduction is to be prevented, is to map and/or reduce keys in one step and map values in a second pass.

cdeszaq:
Ripped mapMap commit off into it's own PR: https://github.com/yahoo/fili/pull/114. Moving discussion on that line of code over there manually.

fili-core/src/main/java/com/yahoo/bard/webservice/metadata/SegmentMetadata.java
```diff
@@ -92,9 +80,13 @@ private SegmentMetadata(Map<String, Set<Interval>> dimensionIntervals, Map<Strin
         return metricIntervals;
     }
 
+    public boolean isEmpty() {
+        return isEmpty;
+    }
+
```
michael-mclawhorn: Was there a rationale for calculating this at construction time rather than using the expression here?

michael-mclawhorn:
👍  I don't see anything particularly worrying.

fili-core/src/main/java/com/yahoo/bard/webservice/metadata/SegmentMetadata.java
cdeszaq: Initially, I was thinking it was going to be called possibly many times (each time the data was accessed), but now that I think about it, it will only be called once when the `SegmentMetadataLoader` runs.

I'd be happy to move the calculation into the method call, but if we know it will be called at least once, why not do it in the constructor and more easily cache the value? (yes, it's a trivial calculation, but java lacks easy memoization primitives...)

archolewa:
👍

PR #114, opened 2016-12-01
Add some stream helpers for dealing with maps

cdeszaq:
Some more stream utility methods that I built while playing in this code. I didn't end up using them, but felt it was a shame to chuck them, so I tossed them into the `StreamUtil` anyways.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ Current
 
 ### Added:
 
+- [Stream utils for dealing with `Map`s](https://github.com/yahoo/fili/pull/114)
+
 - [Added MetricField accessor to the interface of LogicalMetric](https://github.com/yahoo/fili/pull/124)
     * Previously accessing the metric field involved using three method calls
 
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/util/StreamUtils.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/util/StreamUtils.java
@@ -2,6 +2,7 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.util;
 
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.function.BinaryOperator;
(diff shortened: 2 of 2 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/util/StreamUtils.java
```diff
@@ -145,4 +146,137 @@
                 return castedOperand;
         };
     }
+
+    /**
+     * Build a function that can map a Map from one map to another.
+     *
+     * @param keyMapper  Function to map from the input Map's Entries to the result Map's keys
+     * @param valueMapper  Function to map from the input Map's Entries to the result Map's values
```
michael-mclawhorn: keyMapper is the issue with avoiding reduction.

fili-core/src/main/java/com/yahoo/bard/webservice/util/StreamUtils.java
```diff
@@ -145,4 +146,137 @@
...
+    public static <A, B, K, V, R extends Map<A, B>, T extends Map<K, V>> Function<R, T> mapMap(
+            Function<? super Map.Entry<A, B>, ? extends K> keyMapper,
+            Function<? super Map.Entry<A, B>, ? extends V> valueMapper,
+            Supplier<T> resultSupplier
+    ) {
+        return (R input) -> input.entrySet().stream()
+                .collect(
+                        resultSupplier,
+                        (result, entry) -> result.put(keyMapper.apply(entry), valueMapper.apply(entry)),
+                        Map::putAll
```
michael-mclawhorn: https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#merge-K-V-java.util.function.BiFunction-

Does what I believe you want it to do.  Basically, have a defaulting BiFunction that just picks left or right.  And then create a version that takes the remapping function as an argument.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ Current
 
 ### Added:
 
+- [Stream utils for dealing with `Map`s](https://github.com/yahoo/fili/pull/114)
```
QubitPi: Shall we add test for `StreamUtils`?

cdeszaq:
I'm not going to get to this any time soon. If I resurface it, it will be later, but not now.

PR #115, opened 2016-12-01
Default DimensionColumn name to use apiName instead of physicalName

garyluoex:
* Change `DimensionColumn.java` to use dimension api name instead of physical name as its name
* Modified tests according to the above change

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/PartialDataHandler.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/PartialDataHandler.java
@@ -3,7 +3,6 @@
 package com.yahoo.bard.webservice.data;
 
 import static com.yahoo.bard.webservice.config.BardFeatureFlag.PERMISSIVE_COLUMN_AVAILABILITY;
-import static com.yahoo.bard.webservice.util.TableUtils.getColumnNames;
 
 import com.yahoo.bard.webservice.druid.model.query.AllGranularity;
 import com.yahoo.bard.webservice.druid.model.query.DruidAggregationQuery;
@@ -11,6 +10,7 @@ import com.yahoo.bard.webservice.druid.model.query.Granularity;
 import com.yahoo.bard.webservice.table.PhysicalTable;
 import com.yahoo.bard.webservice.util.IntervalUtils;
 import com.yahoo.bard.webservice.util.SimplifiedIntervalList;
+import com.yahoo.bard.webservice.util.TableUtils;
 import com.yahoo.bard.webservice.web.DataApiRequest;
 
 import org.slf4j.Logger;
@@ -95,7 +95,7 @@ public class PartialDataHandler {
             Granularity granularity
     ) {
         SimplifiedIntervalList availableIntervals = physicalTables.stream()
-                .map(table -> getAvailability(table, getColumnNames(apiRequest, query, table)))
+                .map(table -> getAvailability(table, TableUtils.getColumnNames(apiRequest, query)))
                 .flatMap(SimplifiedIntervalList::stream)
                 .collect(SimplifiedIntervalList.getCollector());
 
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionColumn.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionColumn.java
@@ -17,13 +17,23 @@ public class DimensionColumn extends Column {
 
     /**
      * Constructor.
-     * Uses dimension DruidName for column name.
+     * Uses dimension api name for column name.
      *
-     * @param dimension  The column name
-     * @param physicalName  Physical column name backing dimension
+     * @param dimension  The column's corresponding dimension
      */
-    protected DimensionColumn(@NotNull Dimension dimension, @NotNull String physicalName) {
-        super(physicalName);
+    public DimensionColumn(@NotNull Dimension dimension) {
+        this(dimension, dimension.getApiName());
+    }
+
+    /**
+     * Constructor.
+     * Uses the given name as  for column name.
+     *
+     * @param dimension  The column's corresponding dimension
+     * @param columnName  Physical column name backing dimension
+     */
+    protected DimensionColumn(@NotNull Dimension dimension, @NotNull String columnName) {
+        super(columnName);
         this.dimension = dimension;
     }
 
@@ -32,15 +42,17 @@ public class DimensionColumn extends Column {
     }
 
     /**
-     * Method to create a DimensionColumn tied to a PhysicalTable.
+     * Method to create a DimensionColumn tied to a schema.
      *
-     * @param physicalTable  Physical table associated with dimension column
+     * @param schema  The schema to which the column needs to be added
      * @param d  The dimension the column encapsulates
      *
      * @return The dimension column created
      */
-    public static DimensionColumn addNewDimensionColumn(PhysicalTable physicalTable, Dimension d) {
-        return addNewDimensionColumn(physicalTable, d, physicalTable);
+    public static DimensionColumn addNewDimensionColumn(Schema schema, Dimension d) {
+        DimensionColumn col = new DimensionColumn(d);
+        schema.addColumn(col);
+        return col;
     }
 
     /**
@@ -51,9 +63,12 @@ public class DimensionColumn extends Column {
      * @param physicalTable  Physical table associated with dimension column
      *
      * @return The dimension column created
+     *
+     * @deprecated in favor of addNewDimensionColumn(Schema, Dimension) which stores api name instead of physical name
      */
+    @Deprecated
     public static DimensionColumn addNewDimensionColumn(Schema schema, Dimension d, PhysicalTable physicalTable) {
-        DimensionColumn col = new DimensionColumn(d, d.getApiName());
+        DimensionColumn col = new DimensionColumn(d, physicalTable.getPhysicalColumnName(d.getApiName()));
         schema.addColumn(col);
         return col;
     }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionColumn.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionColumn.java
@@ -53,7 +53,7 @@ public class DimensionColumn extends Column {
      * @return The dimension column created
      */
     public static DimensionColumn addNewDimensionColumn(Schema schema, Dimension d, PhysicalTable physicalTable) {
-        DimensionColumn col = new DimensionColumn(d, physicalTable.getPhysicalColumnName(d.getApiName()));
+        DimensionColumn col = new DimensionColumn(d, d.getApiName());
         schema.addColumn(col);
         return col;
     }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionColumn.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionColumn.java
@@ -17,7 +17,7 @@ public class DimensionColumn extends Column {
 
     /**
      * Constructor.
-     * Uses dimension api name for column name.
+     * Uses the given dimension's api name for column name.
      *
      * @param dimension  The column's corresponding dimension
      */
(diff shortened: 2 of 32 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/util/TableUtils.java
```diff
@@ -36,8 +36,7 @@
     ) {
         return Stream.<Stream<String>>of(
                 getDimensions(request, query)
-                        .map(Dimension::getApiName)
-                        .map(table::getPhysicalColumnName),
+                        .map(Dimension::getApiName),
```
archolewa: According to the JavaDoc, this method returns fact-store column names (i.e. physical column names). Note that the `AggregatableDimensionsMatcher::test` expects it to return fact-store column names. But it looks like we're now returning api names.

fili-core/src/main/java/com/yahoo/bard/webservice/data/PartialDataHandler.java
```diff
@@ -3,14 +3,14 @@
 package com.yahoo.bard.webservice.data;
 
 import static com.yahoo.bard.webservice.config.BardFeatureFlag.PERMISSIVE_COLUMN_AVAILABILITY;
-import static com.yahoo.bard.webservice.util.TableUtils.getColumnNames;
```
cdeszaq: Can still keep this statically imported I think? Or does it complain?

fili-core/src/main/java/com/yahoo/bard/webservice/util/TableUtils.java
```diff
@@ -43,6 +46,25 @@
...
+     *
+     * @param request  A request which supplies grouping dimensions and filtering dimensions
+     * @param query  A query model which has metric column and possibly dimension column names
+     *
+     * @return a set of strings representing table column names
+     */
+    public static Set<String> getColumnNames(
+            DataApiRequest request,
+            DruidAggregationQuery<?> query
+    ) {
```
cdeszaq: Doesn't need to wrap

fili-core/src/main/java/com/yahoo/bard/webservice/util/TableUtils.java
```diff
@@ -43,6 +46,25 @@
...
+        return Stream.<Stream<String>>of(
+                getDimensions(request, query)
+                        .map(Dimension::getApiName),
```
cdeszaq: Should just go on a single line, so each line is a single stream

fili-core/src/main/java/com/yahoo/bard/webservice/util/TableUtils.java
```diff
@@ -43,6 +46,25 @@
...
+     * @param request  A request which supplies grouping dimensions and filtering dimensions
+     * @param query  A query model which has metric column and possibly dimension column names
+     *
+     * @return a set of strings representing table column names
+     */
+    public static Set<String> getColumnNames(
+            DataApiRequest request,
+            DruidAggregationQuery<?> query
+    ) {
+        return Stream.<Stream<String>>of(
```
cdeszaq: Type arguments are not needed I don't think

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/TableUtilsSpec.groovy
```diff
@@ -87,19 +82,6 @@ class TableUtilsSpec extends  Specification {
         query.dependentFieldNames >> ([metric1, metric2, metric3] as Set)
 
         expect:
-        TableUtils.getColumnNames(request, query, new PhysicalTable("", DefaultTimeGrain.DAY.buildZonedTimeGrain(DateTimeZone.UTC), [:])) == [d1Name, metric1, metric2, metric3] as Set
-    }
-
-    def "logicalName to physicalName mapping not required for metrics" () {
```
cdeszaq: Why are we dropping this test?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/TableFullViewProcessorSpec.groovy
```diff
@@ -68,7 +68,7 @@ class TableFullViewProcessorSpec extends Specification {
...
+        String expectedResponse = "[description:The pets all grain, dimensions:[[category:General, name:species, longName:species, uri:http://localhost:9998/v1/species, cardinality:0], [category:General, name:sex, longName:sex, uri:http://localhost:9998/v1/sex, cardinality:0], [category:General, name:breed, longName:breed, uri:http://localhost:9998/v1/breed, cardinality:0]], longName:All, metrics:[[category:General, name:rowNum, longName:rowNum, uri:http://localhost:9998/v1/rowNum], [category:General, name:limbs, longName:limbs, uri:http://localhost:9998/v1/limbs], [category:General, name:dayAvgLimbs, longName:dayAvgLimbs, uri:http://localhost:9998/v1/dayAvgLimbs]], name:all, retention:P1Y]"
```
garyluoex: Because the name was stored as physical name in inside the table, but now it is stored as it's api name. (apiName: species, physicalName: class, and class is after sex alphabetically)

cdeszaq: The ordering should have been configuration-based (ie. insertion-order), not sorted, since the underlying collections, in general, are all `HashSet`s. It's not a big deal, but things changing order generally _shouldn't_ be happening, so I want to make sure we're not breaking something accidentally.

garyluoex: Let me check again, I think I'm not making sense on the last comment.

garyluoex: So the order is not consistent is due to using `HashMap` [here](https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java#L58), the order is not preserved when inserting two different names into this `HashTable` (i.e. `species`, `class`)

cdeszaq: Ahh, I see. Perhaps we should flip that to be a `LinkedHashMap` so that order will be preserved? At the very least, it makes testing easier... We've been bitten by different hashing on different platforms before causing tests to break for no reason.

garyluoex: Ok, changing all `Map` implementation in `PhysicalTable` to use `LinkedHashMap` instead of `HashMap`

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/AggregatableDimensionsMatcher.java
```diff
@@ -43,7 +43,7 @@ public AggregatableDimensionsMatcher(DataApiRequest request, TemplateDruidQuery
 
     @Override
     public boolean test(PhysicalTable table) {
-        Set<String> columnNames = TableUtils.getColumnNames(request, query.getInnermostQuery(), table);
+        Set<String> columnNames = TableUtils.getColumnNames(request, query.getInnermostQuery());
```
garyluoex: After looking into this.... I discovered that not only that the test was no sufficient, in addition... `AggregatableDimensionsMatcher` never worked the way it supposed to work but instead worked the opposite way it supposed to work =)

garyluoex: Should I go ahead and fix it in this PR?

archolewa: Fun! I know that we have the `AggregatableDimensionMatcher` in our matching chain, but none of our dimensions are non-aggregatable.

fili-core/src/main/java/com/yahoo/bard/webservice/data/PartialDataHandler.java
garyluoex: I don't see why we want to statically import this method here. Let me know if there is a reason for it. (the reason I removed it is because the IDE does not know exactly which version of `getColumnNames` the static import refers to and thinks that all versions of this function are used in this file)

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/TableUtilsSpec.groovy
garyluoex: We are not converting logicalName to physicalName for both dimension and metrics now, originally dimensions were. So testing metric do not get converted into physicalName loses its meaning I think.

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/AggregatableDimensionsMatcher.java
garyluoex: It turns out that this is not a bug, will add javadoc to clarify it.

fili-core/src/main/java/com/yahoo/bard/webservice/data/PartialDataHandler.java
cdeszaq: If the IDE gets confused, that's a good reason not to be a static import, but in general, static imports reduce verbosity and improve code clarity.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionColumn.java
```diff
@@ -17,13 +17,23 @@
...
-     * Uses dimension DruidName for column name.
+     * Uses dimension api name for column name.
      *
-     * @param dimension  The column name
-     * @param physicalName  Physical column name backing dimension
+     * @param dimension  The column's corresponding dimension
      */
-    protected DimensionColumn(@NotNull Dimension dimension, @NotNull String physicalName) {
-        super(physicalName);
+    public DimensionColumn(@NotNull Dimension dimension) {
```
cdeszaq: This should be `protected` I think, since we still want to force people to use the `static` factory method, right?

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/AggregatableDimensionsMatcher.java
```diff
@@ -43,12 +43,12 @@ public AggregatableDimensionsMatcher(DataApiRequest request, TemplateDruidQuery
...
 
+        // If table contains non-agg dimensions, query must contain all these non-agg dimensions to use this table.
```
cdeszaq: This is a beautiful, perfect comment. Could we add it up in the class's JavaDoc though, so that it's more clear and easier to see what this class is requiring?

fili-core/src/main/java/com/yahoo/bard/webservice/util/TableUtils.java
```diff
@@ -43,6 +46,21 @@
...
+     * @param request  A request which supplies grouping dimensions and filtering dimensions
+     * @param query  A query model which has metric column and possibly dimension column names
+     *
+     * @return a set of strings representing schema column names
+     */
+    public static Set<String> getColumnNames(DataApiRequest request, DruidAggregationQuery<?> query) {
+        return Stream.of(
+                getDimensions(request, query).map(Dimension::getApiName),
+                query.getDependentFieldNames().stream()
+        ).flatMap(Function.identity()).collect(Collectors.toSet());
```
archolewa: This can be simplified using the `Stream.concat` function:

```java
return Stream.concat(
         getDimensions(request, query).map(Dimension::getApiName),
         query.getDependentFieldNames().stream
)
         .collect(Collectors.toSet());
```

archolewa:
👍 Had a small suggestion to clean up a Stream, but other than that I'm satisfied.

garyluoex:
Squash can be done with the squash and merge button right?

archolewa:
@garyluoex I believe so. I believe that squash and merge essentially does the following:

1. Rebase onto master
2. Squash
3. Fast-forward merge.

As opposed to rebase and merge, which does:

1. Rebase onto master
2. Fast-forward merge.

fili-core/src/main/java/com/yahoo/bard/webservice/util/TableUtils.java
cdeszaq: Nice catch here @archolewa

PR #117, opened 2016-12-06
Reverse apache configuration update.

archolewa:


archolewa:
Looks like this is more of a case of dependency hell. If apps using Fili have an apache configuration of version 1.6, then they get a whole lot of NoClassDef found errors. Removing that dependency (if possible) or upgrading to 1.10 solves the problem.

cdeszaq:
Can this branch be deleted?

PR #119, opened 2016-12-07
The NoOpResultSetMapper is now O(1).

archolewa:
--Previously, the `NoOpResultSetMapper` iterated through every result
and did nothing to it before adding it to a new `ResultSet`, resulting
in a no-op that took linear time and space. Now, the `NoOpResultSetMapper`
just returns the original `resultSet` it received, resulting in a
constant space and time no-op.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ Current
 
 ### Changed:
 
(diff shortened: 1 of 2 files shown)


cdeszaq:
:+1: merge it

michael-mclawhorn:
👍

PR #121, opened 2016-12-08
Add <p> tags for cdeszaq

jacobtolar:
😈

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/PreResponseDeserializer.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/PreResponseDeserializer.java
@@ -56,7 +56,7 @@ import java.util.stream.StreamSupport;
  * a. We can reconstruct the dimension rows for each result with the help of dimension dictionary.
  * b. Schema object de-serialization is more efficient then the default one.
  * c. Provides the flexibility of optimizing serialization
- *
+ * <p>
  * Sample serialized PreResponse string
  * <pre>
  * <code>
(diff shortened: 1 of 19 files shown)


jacobtolar:
Since I had the script to fix up my changes in #107 I figured I might as well run it on everything.

The simple bash version I used on #107 - 
```
find . \( -name '*.java' -o -name '*.groovy' \)  -exec gsed -i 's/^\(\s*\)\*\s*$/\1* <p>/g' {} \;
```

Slightly better version that I used here which updates only comments that look like javadocs (probably not perfect, but seems to be sufficient):

https://gist.github.com/jacobtolar/46f03ced47ce6f15c005e8201ba601e1

cdeszaq:
So, one interesting note for this: It's not _all_ whitespace in JavaDoc that needs this, only vertical whitespace _contained in_ the description section that needs it. So, this means that there _should_ be a blank line before the `@params` block, as well as between any other `@blah` sections (like `@returns`, for example).

jacobtolar:
This is what I get for not reading the javadocs for javadoc. :)

jacobtolar:
Updated to a significantly smaller changeset -- made a few small changes to the script and then manually staged the valid changes.

jacobtolar:
And -- right, that would be nice, although this particular script would need to be a bit smarter. e.g., it would add `<p>` tags inside of `<pre>` blocks, which isn't what you want. Don't think it would be too difficult to make the necessary changes, though.

cdeszaq:
Merging. These are Javadoc-only changes, and trivial cleanup ones at that.

PR #122, opened 2016-12-09
Module Test code to support module setup is correct.

michael-mclawhorn:


fili-core/src/test/groovy/com/yahoo/bard/systemconfig/ModuleSetupSpec.groovy
```diff
@@ -0,0 +1,31 @@
...
+package com.yahoo.bard.systemconfig
+
+import com.yahoo.bard.webservice.config.ConfigResourceLoader
+import com.yahoo.bard.webservice.config.SystemConfig
+
+import org.apache.commons.configuration.Configuration
+
+import spock.lang.Specification
+
+abstract class ModuleSetupSpec extends Specification {
```
cdeszaq: Description of what this base spec is for and how to use it would be good.

fili-core/src/test/groovy/com/yahoo/bard/systemconfig/ModuleSetupSpec.groovy
```diff
@@ -0,0 +1,31 @@
...
+
+    def "Current module is loaded by ConfigResourceLoader"() {
+        setup:
+        List<Configuration> configurations = new ConfigResourceLoader().loadConfigurations(MODULE_RESOURCE_NAME)
+
+        expect:
+        configurations.stream().map( {it.getString(MODULE_NAME_KEY)}).anyMatch({it.equals(getModuleName())})
+    }
+
+    abstract String getModuleName()
```
cdeszaq: Javadoc on this is needed so that usage and contract is clear.

fili-core/src/test/groovy/com/yahoo/bard/systemconfig/ModuleSetupSpec.groovy
```diff
@@ -0,0 +1,31 @@
...
+import org.apache.commons.configuration.Configuration
+
+import spock.lang.Specification
+
+abstract class ModuleSetupSpec extends Specification {
+
+    public static final String MODULE_NAME_KEY = "moduleName"
+    public static final String MODULE_RESOURCE_NAME = "/moduleConfig.properties"
+
+    def "Test that SystemConfig loads for this module"() {
```
cdeszaq: 1. "Test that" isn't helpful in the test name.
2. It would be cool if the module name could be injected into here somehow? Perhaps via a `where` block that gets it's value from the `getModuleName` method?

fili-core/src/test/groovy/com/yahoo/bard/systemconfig/ModuleSetupSpec.groovy
```diff
@@ -0,0 +1,31 @@
...
+        expect:
+        SystemConfig.properties.getOrDefault("foo", "foo") == "foo"
```
cdeszaq: What is this testing?

fili-core/src/test/groovy/com/yahoo/bard/systemconfig/ModuleSetupSpec.groovy
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    def "Current module is loaded by ConfigResourceLoader"() {
```
cdeszaq: Would be good to replace "Current module" with an unrolled value from a `where` block that could source from the `getModuleName()` method.

fili-core/src/test/groovy/com/yahoo/bard/systemconfig/ModuleSetupSpec.groovy
```diff
@@ -0,0 +1,31 @@
...
+        setup:
+        List<Configuration> configurations = new ConfigResourceLoader().loadConfigurations(MODULE_RESOURCE_NAME)
```
cdeszaq: Should `MODULE_RESOURCE_NAME` be fixed or should it be dynamic based on what module is being tested?

fili-core/src/test/groovy/com/yahoo/bard/systemconfig/ModuleSetupSpec.groovy
```diff
@@ -0,0 +1,31 @@
...
+
+        expect:
+        configurations.stream().map( {it.getString(MODULE_NAME_KEY)}).anyMatch({it.equals(getModuleName())})
```
cdeszaq: spacing inside the calls is a little inconsistent.

fili-core/src/test/groovy/com/yahoo/bard/systemconfig/BardModuleSetupSpec.groovy
```diff
@@ -0,0 +1,10 @@
+// Copyright 2016 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.systemconfig
+
+class BardModuleSetupSpec extends ModuleSetupSpec {
+
+    String getModuleName() {
```
cdeszaq: Needs `@Override`

fili-core/src/test/groovy/com/yahoo/bard/systemconfig/ModuleSetupSpec.groovy
michael-mclawhorn: It was my first test, it tested that the applicationConfig or some other piece of conflictable-config didn't prevent SystemConfig from working
I'm sure there's a better way to do this.

fili-core/src/test/groovy/com/yahoo/bard/systemconfig/ModuleSetupSpec.groovy
michael-mclawhorn: it's baked into ModuleLoader as a private constant

fili-core/src/test/groovy/com/yahoo/bard/systemconfig/ModuleSetupSpec.groovy
```diff
@@ -0,0 +1,49 @@
...
+        where:
+        moduleName << [getModuleName()]
+    }
+
+    /**
+     * Implement this test by supplying a subclass that provides the name of the module
+     *
+     * @return  The name of the module as configured in moduleConfig.properties
+     */
+    abstract String getModuleName()
```
garyluoex: Maybe we should move this to higher in the file so that people see this first.

PR #124, opened 2016-12-15
Added getMetricField to LogicalMetric

michael-mclawhorn:
Simplified calls in maker code so as to not use TDQ where unnecessary
Streamified for loop for making arithmetic operations

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
@@ -61,15 +61,13 @@ public class AggregationAverageMaker extends MetricMaker {
     @Override
     protected LogicalMetric makeInner(String metricName, List<String> dependentMetrics) {
         // Get the Metric that is being averaged over
-        String dependantMetricName = dependentMetrics.get(0);
-        TemplateDruidQuery innerDependentQuery = getDependentQuery(dependantMetricName);
-        MetricField sourceMetric = innerDependentQuery.getMetricField(dependantMetricName);
+        LogicalMetric dependentMetric = metrics.get(dependentMetrics.get(0));
 
-        // Convert if needed
-        sourceMetric = convertToSketchEstimateIfNeeded(sourceMetric);
+        // Get the field being subtotalled in the inner query
+        MetricField sourceMetric = convertToSketchEstimateIfNeeded(dependentMetric.getMetricField());
 
         // Build the TemplateDruidQuery for the metric
-        TemplateDruidQuery innerQuery = buildInnerQuery(sourceMetric, innerDependentQuery);
+        TemplateDruidQuery innerQuery = buildInnerQuery(sourceMetric, dependentMetric.getTemplateDruidQuery());
         TemplateDruidQuery outerQuery = buildOuterQuery(metricName, sourceMetric, innerQuery);
 
         return new LogicalMetric(outerQuery, new NoOpResultSetMapper(), metricName);
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
@@ -131,6 +131,21 @@ public abstract class MetricMaker {
     protected abstract int getDependentMetricsRequired();
 
     /**
+     * Fetch the TemplateDruidQuery of the dependent metric from the Metric Dictionary.
+     *
+     * @param name  Name of the metric to fetch the template druid query from
+     *
+     * @return The template druid query of the metric
+     *
+     * @deprecated Instead get the metric in the calling function and then get the TDQ out only if necessary
+     */
+    @Deprecated
+    protected TemplateDruidQuery getDependentQuery(String name) {
+        LogicalMetric dependentMetric = metrics.get(name);
+        return dependentMetric.getTemplateDruidQuery();
+    }
+
+    /**
      * A helper function returning the resulting aggregation set from merging one or more template druid queries.
      *
      * @param names  Names of the metrics to fetch and merge the aggregation clauses from
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
@@ -131,17 +131,6 @@ public abstract class MetricMaker {
     protected abstract int getDependentMetricsRequired();
 
     /**
-     * Fetch the TemplateDruidQuery of the dependent metric from the Metric Dictionary.
-     *
-     * @param name  Name of the metric to fetch the template druid query from
-     * @return The template druid query of the metric
-     */
-    protected TemplateDruidQuery getDependentQuery(String name) {
-        LogicalMetric dependentMetric = metrics.get(name);
-        return dependentMetric.getTemplateDruidQuery();
-    }
-
-    /**
      * A helper function returning the resulting aggregation set from merging one or more template druid queries.
      *
      * @param names  Names of the metrics to fetch and merge the aggregation clauses from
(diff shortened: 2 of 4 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
```diff
@@ -131,41 +131,24 @@ public LogicalMetric make(String metricName, String dependentMetric) {
     protected abstract int getDependentMetricsRequired();
 
     /**
-     * Fetch the TemplateDruidQuery of the dependent metric from the Metric Dictionary.
-     *
-     * @param name  Name of the metric to fetch the template druid query from
-     * @return The template druid query of the metric
-     */
-    protected TemplateDruidQuery getDependentQuery(String name) {
```
cdeszaq: We can't just remove this, since it was `protected` and in a class that we _intended_ to be extended. We need to `@Deprecate` it instead.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
```diff
@@ -131,41 +131,24 @@ public LogicalMetric make(String metricName, String dependentMetric) {
...
-            TemplateDruidQuery additionalQuery = getDependentQuery(names.get(i));
-            query = query.merge(additionalQuery);
-        }
-
-        // Return the merged query
-        return query;
+        return names.stream()
+                .map(metrics::get)
+                .map(LogicalMetric::getTemplateDruidQuery)
+                .reduce(TemplateDruidQuery::merge)
```
cdeszaq: Does merge handle `null` correctly? I know we're collapsing the error check to the end of this stream (which I'm a fan of), but I just want to make sure that the merge will still be working correctly.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
```diff
@@ -187,7 +170,7 @@ protected TemplateDruidQuery getMergedQuery(List<String> names) {
      */
     protected PostAggregation getNumericField(String fieldName) {
         // Get the field
-        MetricField field = getDependentQuery(fieldName).getMetricField(fieldName);
+        MetricField field = metrics.get(fieldName).getMetricField();
```
cdeszaq: This is not the same semantics, is it? In the previous code, it was retrieving an aggregation with a specific name from a metric with that name. Admittedly, the previous code had many assumptions, and I think the new semantics are better, but we should be at least a little careful here, since this is a `protected` method, meaning that others could be relying on it...

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
```diff
@@ -212,7 +195,7 @@ protected PostAggregation getNumericField(String fieldName) {
      */
     protected PostAggregation getSketchField(String fieldName) {
         // Get the field
-        MetricField field = getDependentQuery(fieldName).getMetricField(fieldName);
+        MetricField field = metrics.get(fieldName).getMetricField();
```
cdeszaq: ditto my comments on semantic changes

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
```diff
@@ -61,15 +61,13 @@ public AggregationAverageMaker(MetricDictionary metrics, ZonelessTimeGrain inner
     @Override
     protected LogicalMetric makeInner(String metricName, List<String> dependentMetrics) {
         // Get the Metric that is being averaged over
-        String dependantMetricName = dependentMetrics.get(0);
-        TemplateDruidQuery innerDependentQuery = getDependentQuery(dependantMetricName);
-        MetricField sourceMetric = innerDependentQuery.getMetricField(dependantMetricName);
```
cdeszaq: Is this really the only maker in which we were doing this?

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
michael-mclawhorn: boo.  Ok.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
michael-mclawhorn: The previous implementation calls names.size first and thus blows up on names being null.
Similarly, getDependentQuery wasn't nullsafeing on it's MetricDictionary.get().getTemplateDruidQuery()
So the only open question is whether reduce with a binary function does the thing it's spec says it does.

cdeszaq: I didn't look at `reduce`'s contract, so if it's protecting us, we're good 😄

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
michael-mclawhorn: The previous method call did:
foo = metrics.get(fieldName)
bar = return foo.getTemplateDruidQuery
bar.getMetricField(fieldName)

How is this not the same?

cdeszaq: Ahh, I see. Makes sense then. Clearly, the previous code was worse, since it was harder to trace 😉

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
cdeszaq: Fair enough.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
```diff
@@ -134,8 +134,12 @@ public LogicalMetric make(String metricName, String dependentMetric) {
      * Fetch the TemplateDruidQuery of the dependent metric from the Metric Dictionary.
      *
      * @param name  Name of the metric to fetch the template druid query from
+     *
      * @return The template druid query of the metric
+     *
+     * @deprecated Instead get the metric in the calling function and then get the TDQ out only if necessary
      */
+    @Deprecated
```
cdeszaq: Needs to be called out in change log.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
archolewa: Yay for less variety than you might suspect!

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
```diff
@@ -149,23 +153,17 @@ protected TemplateDruidQuery getDependentQuery(String name) {
...
-            query = query.merge(additionalQuery);
-        }
-
-        // Return the merged query
-        return query;
+        return names.stream()
+                .map(metrics::get)
+                .map(LogicalMetric::getTemplateDruidQuery)
+                .reduce(TemplateDruidQuery::merge)
+                .orElseGet(() -> {
```
archolewa: What you actually want to use here is `orElseThrow`:

```java
.orElseThrow(() -> {
          String message = "At least 1 name is needed to merge aggregations";
          LOG.error(message);
          return new IllegalArgumentException(message);
}
```

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
```diff
@@ -149,23 +153,17 @@ protected TemplateDruidQuery getDependentQuery(String name) {
...
+                            String message = "At least 1 name is needed to merge aggregations";
+                            LOG.error(message);
+                            throw new IllegalArgumentException(message);
+                        }
```
archolewa: Is this bracket indented correctly? Typically, when the opening of a lambda is on the same line as a method, I put the closing bracket on the same line as the closing parens. i.e.:

```java
  method(() -> {
      doStuf();
  });
```

and the closing bracket is on its own line only if the start of the lambda is on its own line:


```java
method(
          () -> {
              doStuf();
          }
);
```

I don't have a strong opinion either way, I just wanted to know if we had a convention around this.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
```diff
@@ -187,7 +185,7 @@ protected TemplateDruidQuery getMergedQuery(List<String> names) {
      */
     protected PostAggregation getNumericField(String fieldName) {
         // Get the field
```
archolewa: This comment is now redundant.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
cdeszaq: Ah, yes, that's the one. I knew there was something like that...

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
cdeszaq: It's not wrong, just a little different. I think all of the options are pretty much equal, though I do like the more compressed form of stacking things where it's obvious what's closing and isn't stacked too deep if I think hard about it.

michael-mclawhorn: Autoformat thinks that the closure shouldn't be double indented because it's a sub block not an argument line continuation.  What do you think of what I'm about to post?

archolewa:
👍

PR #125, opened 2016-12-15
Fix `DataSource::getNames`.

archolewa:
--Currently, `DataSource::getNames` returns the fact names of its
physical tables. However, everywhere `getNames` is used for anything but
logging purposes, the names are resolved against the
`physicalTableDictionary`, which is based on Fili's name of the table,
not Druid's.

--Therefore, `DataSource::getNames` should really be returning the Fili
names, not the fact store names.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -162,6 +162,8 @@ Current
 
 ### Fixed:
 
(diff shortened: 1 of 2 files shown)


cdeszaq:
👍

michael-mclawhorn:
👍

PR #127, opened 2016-12-19
Reduce makers

michael-mclawhorn:
Big simplification of makers

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
@@ -159,7 +159,10 @@ public class AggregationAverageMaker extends MetricMaker {
      * @param originalAggregations  The read-only original aggregations
      *
      * @return The new aggregation set
+     *
+     * @deprecated This will become unnecessary when the old sketch library is removed
      */
+    @Deprecated
     private Set<Aggregation> convertSketchesToSketchMerges(Set<Aggregation> originalAggregations) {
         return originalAggregations.stream()
                 .map(agg -> agg.isSketch() ? sketchConverter.asInnerSketch((SketchAggregation) agg) : agg)
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
@@ -2,12 +2,13 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.data.config.metric.makers;
 
-import static com.yahoo.bard.webservice.druid.model.postaggregation.ArithmeticPostAggregation.ArithmeticPostAggregationFunction.DIVIDE;
+import static com.yahoo.bard.webservice.druid.model.postaggregation.ArithmeticPostAggregation
+        .ArithmeticPostAggregationFunction.DIVIDE;
+import static com.yahoo.bard.webservice.druid.util.FieldConverterSupplier.sketchConverter;
 
 import com.yahoo.bard.webservice.data.metric.LogicalMetric;
 import com.yahoo.bard.webservice.data.metric.MetricDictionary;
 import com.yahoo.bard.webservice.data.metric.TemplateDruidQuery;
-import com.yahoo.bard.webservice.data.metric.mappers.NoOpResultSetMapper;
 import com.yahoo.bard.webservice.data.time.ZonelessTimeGrain;
 import com.yahoo.bard.webservice.druid.model.MetricField;
 import com.yahoo.bard.webservice.druid.model.aggregation.Aggregation;
@@ -18,13 +19,15 @@ import com.yahoo.bard.webservice.druid.model.postaggregation.ArithmeticPostAggre
 import com.yahoo.bard.webservice.druid.model.postaggregation.ConstantPostAggregation;
 import com.yahoo.bard.webservice.druid.model.postaggregation.FieldAccessorPostAggregation;
 import com.yahoo.bard.webservice.druid.model.postaggregation.PostAggregation;
-import com.yahoo.bard.webservice.druid.util.FieldConverterSupplier;
+
+import com.google.common.collect.ImmutableSet;
 
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.validation.constraints.NotNull;
 
@@ -35,6 +38,10 @@ import javax.validation.constraints.NotNull;
  * takes aggregated data from a finer time grain (i.e. DefaultTimeGrain.DAY) and computes an average across a coarser
  * time grain (i.e. DefaultTimeGrain.WEEK). For example, given the total number of visitors to www.example.com for each
  * day of the week, we can compute the average number of daily visitors to example.com for the entire week.
+ * <p>
+ * A nested average requires the following columns, an inner query constant, and outer query sum of that constant,
+ * an inner query aggregation or post aggregation field to sum, and outer query sum of the inner field, and
+ * finally a post aggregation in the outer query taking the division of the sum and the constant sum.
  */
 public class AggregationAverageMaker extends MetricMaker {
 
@@ -70,7 +77,7 @@ public class AggregationAverageMaker extends MetricMaker {
         TemplateDruidQuery innerQuery = buildInnerQuery(sourceMetric, dependentMetric.getTemplateDruidQuery());
         TemplateDruidQuery outerQuery = buildOuterQuery(metricName, sourceMetric, innerQuery);
 
-        return new LogicalMetric(outerQuery, new NoOpResultSetMapper(), metricName);
+        return new LogicalMetric(outerQuery, NO_OP_MAPPER, metricName);
     }
 
     /**
@@ -113,59 +120,53 @@ public class AggregationAverageMaker extends MetricMaker {
      * @return A template query representing the inner aggregation
      */
     private TemplateDruidQuery buildInnerQuery(MetricField sourceMetric, TemplateDruidQuery innerDependentQuery) {
-
         Set<Aggregation> newInnerAggregations = convertSketchesToSketchMerges(innerDependentQuery.getAggregations());
-        Set<PostAggregation> newInnerPostAggregations = new LinkedHashSet<>();
 
-        // If the sourceMetric is a Post Aggregator, we need to preserve it in the inner query
-        if (sourceMetric instanceof PostAggregation) {
-            newInnerPostAggregations.add((PostAggregation) sourceMetric);
-        }
+        Set<PostAggregation> newInnerPostAggregations = !(sourceMetric instanceof PostAggregation) ?
+                Collections.emptySet() : ImmutableSet.of((PostAggregation) sourceMetric);
 
         // Build the inner query with the new aggregations and with the count
-        TemplateDruidQuery innerQuery = innerDependentQuery;
-        innerQuery = innerQuery.withAggregations(newInnerAggregations);
-        innerQuery = innerQuery.withPostAggregations(newInnerPostAggregations);
-        innerQuery = innerQuery.merge(buildTimeGrainCounterQuery());
-        return innerQuery;
+        return innerDependentQuery.withAggregations(newInnerAggregations)
+                .withPostAggregations(newInnerPostAggregations)
+                .merge(buildTimeGrainCounterQuery());
     }
 
     /**
-     * If the aggregation being averaged is a sketch, it will become a sketch merge, and need to be estimated for
-     * summing to work at the outer level.
+     * If the aggregation being averaged is a sketch, the inner query must convert it to a numerical type so that it
+     * can be summed.
      *
      * @param originalSourceMetric  The metric being target for sums
      *
      * @return Either the original MetricField, or a new SketchEstimate post aggregation
      */
     private MetricField convertToSketchEstimateIfNeeded(MetricField originalSourceMetric) {
-        if (originalSourceMetric instanceof SketchAggregation) {
-            return FieldConverterSupplier.sketchConverter.asSketchEstimate((SketchAggregation) originalSourceMetric);
-        }
-        return originalSourceMetric;
+        return !(originalSourceMetric instanceof SketchAggregation) ?
+                originalSourceMetric :
+                sketchConverter.asSketchEstimate((SketchAggregation) originalSourceMetric);
     }
 
     /**
      * Copy a set of aggregations, replacing any sketch aggregations with sketchMerge aggregations.
+     * This is an artifact of earlier sketch code which didn't have a single sketch type that could be used without
+     * finalization in inner queries.
      *
      * @param originalAggregations  The read-only original aggregations
      *
      * @return The new aggregation set
      */
     private Set<Aggregation> convertSketchesToSketchMerges(Set<Aggregation> originalAggregations) {
-        Set<Aggregation> result = new LinkedHashSet<>();
-        for (Aggregation agg : originalAggregations) {
-            if (agg.isSketch()) {
-                result.add(FieldConverterSupplier.sketchConverter.asInnerSketch((SketchAggregation) agg));
-            } else {
-                result.add(agg);
-            }
-        }
-        return result;
+        return originalAggregations.stream()
+                .map(agg -> ! agg.isSketch() ? agg : sketchConverter.asInnerSketch((SketchAggregation) agg))
+                .collect(Collectors.toCollection(LinkedHashSet::new));
     }
 
     /**
      * Create an Aggregation for summing on a metric from an inner query.
+     * The base metric for this maker has a name in the request schema.  Daily average aggregates over the summand it
+     * creates in the inner query, which is based on the original metric name.  If the original metric field is also a
+     * sum, the outer aggregation of that metric and the outer aggregation of the daily average do not need to be
+     * distinguished.  However if the original metric field was aggregating on a non numeric type, the daily average
+     * will appeaed '_sum' to the original name for it's outer sum.
      *
      * @param innerMetric  The metric on the inner query being summed
      *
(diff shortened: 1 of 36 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
```diff
@@ -35,6 +38,10 @@
  * takes aggregated data from a finer time grain (i.e. DefaultTimeGrain.DAY) and computes an average across a coarser
  * time grain (i.e. DefaultTimeGrain.WEEK). For example, given the total number of visitors to www.example.com for each
  * day of the week, we can compute the average number of daily visitors to example.com for the entire week.
+ * <p>
+ * A nested average requires the following columns, an inner query constant, and outer query sum of that constant,
```
archolewa: Should there be a colon after `following columns` instead of a comma?

Also this paragraph is extremely dense, and I couldn't keep my eyes from glazing as I looked at it. I would suggest something like:

1. An inner query constant `c` to sum, and the outer query sum of `c`: `cs`.
2. An inner query (post) aggregation `a` to sum, and an outer query sum of `a`: `as`
3. An outer query post aggregation that computes `as / cs`.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
```diff
@@ -115,80 +120,69 @@ private TemplateDruidQuery buildOuterQuery(
...
-
         Set<Aggregation> newInnerAggregations = convertSketchesToSketchMerges(innerDependentQuery.getAggregations());
-        Set<PostAggregation> newInnerPostAggregations = new LinkedHashSet<>();
 
-        // If the sourceMetric is a Post Aggregator, we need to preserve it in the inner query
-        if (sourceMetric instanceof PostAggregation) {
-            newInnerPostAggregations.add((PostAggregation) sourceMetric);
-        }
+        Set<PostAggregation> newInnerPostAggregations = !(sourceMetric instanceof PostAggregation) ?
+                Collections.emptySet() : ImmutableSet.of((PostAggregation) sourceMetric);
```
archolewa: I would flip these conditions, and make the check a positive check rather than a negative one.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
```diff
@@ -115,80 +120,69 @@ private TemplateDruidQuery buildOuterQuery(
...
      * @return Either the original MetricField, or a new SketchEstimate post aggregation
      */
     private MetricField convertToSketchEstimateIfNeeded(MetricField originalSourceMetric) {
-        if (originalSourceMetric instanceof SketchAggregation) {
-            return FieldConverterSupplier.sketchConverter.asSketchEstimate((SketchAggregation) originalSourceMetric);
-        }
-        return originalSourceMetric;
+        return !(originalSourceMetric instanceof SketchAggregation) ?
+                originalSourceMetric :
+                sketchConverter.asSketchEstimate((SketchAggregation) originalSourceMetric);
```
archolewa: I would flip these conditions, and make the check a positive check rather than a negative one.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
```diff
@@ -115,80 +120,69 @@ private TemplateDruidQuery buildOuterQuery(
...
-        for (Aggregation agg : originalAggregations) {
-            if (agg.isSketch()) {
-                result.add(FieldConverterSupplier.sketchConverter.asInnerSketch((SketchAggregation) agg));
-            } else {
-                result.add(agg);
-            }
-        }
-        return result;
+        return originalAggregations.stream()
+                .map(agg -> ! agg.isSketch() ? agg : sketchConverter.asInnerSketch((SketchAggregation) agg))
```
archolewa: I would flip these conditions, and make the check a positive check rather than a negative one.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
```diff
@@ -115,80 +120,69 @@ private TemplateDruidQuery buildOuterQuery(
...
+                .collect(Collectors.toCollection(LinkedHashSet::new));
     }
 
     /**
      * Create an Aggregation for summing on a metric from an inner query.
```
archolewa: Should probably toss in a blank line and a `<p>` tag here, to separate the summary from the more detailed description.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/ArithmeticMaker.java
```diff
@@ -28,26 +30,49 @@
...
-     * @param resultSetMapper  The function to be applied to the result that is returned by the query
+     * @param resultSetMapperBuilder  A builder for a function to be applied to the result that is returned by the query
     that is built from the LogicalMetric which is built by this maker.
      */
-    public ArithmeticMaker(
+    protected ArithmeticMaker(
             MetricDictionary metricDictionary,
             ArithmeticPostAggregationFunction function,
-            ColumnMapper resultSetMapper
+            Function<String, ResultSetMapper> resultSetMapperBuilder
```
archolewa: Awwww yeah...

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/ConstantMaker.java
```diff
@@ -41,17 +39,25 @@ public LogicalMetric make(String metricName, List<String> dependentMetrics) {
...
+            return new LogicalMetric(
+                    new TemplateDruidQuery(Collections.emptySet(), postAggregations),
+                    NO_OP_MAPPER,
+                    metricName
+            );
+        } catch (NumberFormatException nfe) {
+            String message = String.format(
+                    "%s value '%s' does not parse to a number",
+                    metricName,
+                    dependentMetrics.get(0));
```
archolewa: `);` should be moved down a line and indented to the same level as `String message...`

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/CountMaker.java
```diff
@@ -2,22 +2,20 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.data.config.metric.makers;
 
+import static java.util.Collections.emptySet;
```
archolewa: Why the static import? A static import makes sense to me if the class name is really long and we're using a function/constant a lot, but `Collections` isn't that long, and `emptySet` is only used once. So I would argue that the readability benefits of `Collections.emptySet()` outweighs the benefits from reducing verbosity.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/CountMaker.java
```diff
@@ -32,14 +30,16 @@ public CountMaker(MetricDictionary metricDictionary) {
...
+        TemplateDruidQuery query = new TemplateDruidQuery(
+                Collections.singleton(new CountAggregation(metricName)),
+                emptySet()
+        );
+
+        return new LogicalMetric(
+                query,
+                NO_OP_MAPPER,
+                metricName
+        );
```
archolewa: Does this need to be wrapped?

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/FilteredAggregationMaker.java
```diff
@@ -6,54 +6,90 @@
...
     }
 
+    /**
+     * Test that the metric is associated with an Aggregation Metric and return that Aggregation.
+     *
+     * @param sourceMetric  The source metric
+     *
+     * @return The aggregation used by this metric
+     */
+    private Aggregation assertDependentIsAggregationMetric(final LogicalMetric sourceMetric) {
```
archolewa: `final` not necessary on `final LogicalMetric sourceMetric`.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
```diff
@@ -149,23 +155,17 @@ protected TemplateDruidQuery getDependentQuery(String name) {
...
-            query = query.merge(additionalQuery);
-        }
-
-        // Return the merged query
-        return query;
+        return names.stream()
+                .map(metrics::get)
+                .map(LogicalMetric::getTemplateDruidQuery)
+                .reduce(TemplateDruidQuery::merge)
+                .orElseGet(() -> {
```
archolewa: Should use `orElseThrow` here instead.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/RawAggregationMetricMaker.java
```diff
@@ -2,24 +2,39 @@
...
 
 /**
  * Metric maker which only creates aggregations.  They do not depend on other metrics and cannot be
  * validated against the metric dictionary
  */
 public abstract class RawAggregationMetricMaker extends MetricMaker {
 
+    private static final int DEPENDENT_METRICS_REQUIRED = 1;
+
+    private final BiFunction<String, String, Aggregation> aggregationFactory;
```
archolewa: Awww yeah...

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
michael-mclawhorn: That hasn't been our rule.  The summary ends with a period and newline.  Further commentary may be offset, but need not be.  The sentence end will be caught by the javadoc tool and pulled out to a summary.

archolewa:
👍

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
michael-mclawhorn: Fair enough

archolewa:
👍 Reafirming my plus one.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/metric/makers/RawAggregationMetricMakerImplsSpec.groovy
```diff
@@ -0,0 +1,81 @@
...
+class RawAggregationMetricMakerImplsSpec extends Specification {
+
+    public static String NAME = "FOO"
+    public static String FIELD_NAME = "BAR"
+
+    @Unroll
+    def "Expected numeric aggregation is produced for #makerClass.getSimpleName()"() {
+        setup:
+        RawAggregationMetricMaker maker = makerClass.newInstance()
+        expect:
```
cdeszaq: Blank line between blocks

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/metric/makers/RawAggregationMetricMakerImplsSpec.groovy
```diff
@@ -0,0 +1,81 @@
...
+        LongSumMaker   | new LongSumAggregation(NAME, FIELD_NAME)
+        LongMaxMaker   | new LongMaxAggregation(NAME, FIELD_NAME)
+        LongMinMaker   | new LongMinAggregation(NAME, FIELD_NAME)
+    }
+
+    @Unroll
+    def "Expected sketch aggregation is produced for #makerClass.getSimpleName()"() {
+        setup:
+        RawAggregationMetricMaker maker = makerClass.newInstance((MetricDictionary) null, 5)
+        expect:
```
cdeszaq: Blank line between blocks

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/metric/makers/RawAggregationMetricMakerImplsSpec.groovy
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * Created by mclawhor on 12/19/16.
+ */
+class RawAggregationMetricMakerImplsSpec extends Specification {
+
+    public static String NAME = "FOO"
+    public static String FIELD_NAME = "BAR"
+
+    @Unroll
+    def "Expected numeric aggregation is produced for #makerClass.getSimpleName()"() {
```
cdeszaq: I don't think the call to `getSimpleName()` is needed, since I believe Groovy defaults to that. That said, I also don't think method calls will work correctly with unrolling, only property-style accessors, so this should be `#makerClass.simpleName`

(Applies to other test as well)

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/ArithmeticMaker.java
michael-mclawhorn: I'll call it a supplier, even if it takes a single argument

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/ArithmeticMaker.java
```diff
@@ -28,26 +30,49 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param metricDictionary  The dictionary used to resolve dependent metrics when building the LogicalMetric
+     * @param function  The arithmetic operation performed by the LogicalMetrics constructed by this maker
+     * @param resultSetMapper  The mapping function to be applied to the result that is returned by the query that is
+     * built from the LogicalMetric which is built by this maker.
+     *
+     * @deprecated to override default mapping, use the Function constructor
```
michael-mclawhorn: Deprecating

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/ArithmeticMaker.java
```diff
@@ -58,32 +83,32 @@ public ArithmeticMaker(
      */
     public ArithmeticMaker(MetricDictionary metricDictionary, ArithmeticPostAggregationFunction function) {
         // TODO: Deprecate me, mappers should always be specified at creation time, not implicitly
```
michael-mclawhorn: I don't know if I actually agree with this TODO.  Does anyone know why this is asked for?

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/ConstantMaker.java
```diff
@@ -41,17 +39,26 @@ public LogicalMetric make(String metricName, List<String> dependentMetrics) {
...
+                    NO_OP_MAPPER,
+                    metricName
+            );
+        } catch (NumberFormatException nfe) {
+            String message = String.format(
+                    "%s value '%s' does not parse to a number",
+                    metricName,
+                    dependentMetrics.get(0)
+            );
+            LOG.error(message);
```
michael-mclawhorn: So that I wouldn't have to roll a logger down into every subclass? :)

cdeszaq: While I understand the motivation, I don't agree with it 😉 . I think there are only a small number of places the parent logger is being used.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/ArithmeticMaker.java
cdeszaq: The idea behind this TODO is that we shouldn't apply a `SketchRoundUpMapper` to everything that uses the `ArithmeticMaker`, and instead require those who _want_ that to add the mapper they want explicitly.

For example, someone might want to add 2 estimated sketches and have the result be a floating-point, rather than an integer. The Maker should not be opinionated.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/ArithmeticMaker.java
```diff
@@ -57,33 +81,29 @@ public ArithmeticMaker(
      * @param function  The arithmetic operation performed by the LogicalMetrics constructed by this maker
      */
     public ArithmeticMaker(MetricDictionary metricDictionary, ArithmeticPostAggregationFunction function) {
-        // TODO: Deprecate me, mappers should always be specified at creation time, not implicitly
-        this(metricDictionary, function, new SketchRoundUpMapper());
+        this(
+                metricDictionary,
+                function,
+                (Function<String, ResultSetMapper>) (ignore) -> NO_OP_MAPPER
```
cdeszaq: Definitely need to call this change out in the ChangeLog, and I'm pretty sure there are some that will need to change due to this.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
```diff
@@ -115,80 +125,72 @@ private TemplateDruidQuery buildOuterQuery(
...
-        return originalSourceMetric;
+        return originalSourceMetric instanceof SketchAggregation ?
+                sketchConverter.asSketchEstimate((SketchAggregation) originalSourceMetric) :
+                originalSourceMetric;
     }
 
     /**
      * Copy a set of aggregations, replacing any sketch aggregations with sketchMerge aggregations.
+     * This is an artifact of earlier sketch code which didn't have a single sketch type that could be used without
+     * finalization in inner queries.
```
michael-mclawhorn: Not until we remove that deprecated code, I believe.

cdeszaq: If that's the case, can we `@Deprecate` this method as well, so that when we do the next sweep to remove deprecated code, this will be flagged and we can remove it?

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/ArithmeticMaker.java
cdeszaq: Parens around `ignore` (ie. the lambda input variable) are not technically needed, since there's no type on it and there's only 1 variable.

/NotABlocker

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/ConstantMaker.java
cdeszaq: Note: This has been fixed.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/FilteredAggregationMaker.java
```diff
@@ -6,54 +6,95 @@
...
+
+    /**
+     * The filter being applied to the metrics created.
+     */
     protected final Filter filter;
+
+    /**
+     * The metric dictionary used to resolve dependent metrics.
+     */
     protected final MetricDictionary metricDictionary;
```
cdeszaq: Can/should we kick these over to be `private` instead of `protected`? My gut says we should generally not encourage extension, since it makes it harder to change internal variables like these...

/NotABlocker

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/SketchSetOperationMaker.java
```diff
@@ -27,7 +27,7 @@
 
     private static final int DEPENDENT_METRICS_REQUIRED = 2;
 
-    final SketchSetOperationPostAggFunction function;
+    private final SketchSetOperationPostAggFunction function;
```
cdeszaq: Probably should call this change out in the change log. It's unlikely this was extended, but just in case...

PR #129, opened 2016-12-21
Log filter query

QubitPi:


fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -521,6 +521,7 @@ private Query getFilterQuery(Set<ApiFilter> filters, int perPage) {
             filterQueryBuilder.add(new BooleanClause(new MatchAllDocsQuery(), BooleanClause.Occur.MUST));
         }
 
+        LOG.debug("Filter Query: {}", filterQueryBuilder.build().toString());
```
michael-mclawhorn: This will result in the build being done twice every time.  
Perhaps this is not the outcome you want.

archolewa: Yeah, we should be saving the query in a local variable.

archolewa: And then logging and returning that local variable.

archolewa: Also, I don't think the `toString` is necessary. It'll be called automatically.

archolewa:
👍

michael-mclawhorn:
👍

PR #130, opened 2016-12-21
Remove unused parameter

QubitPi:


michael-mclawhorn:
👍

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -365,9 +365,7 @@ private void refreshCardinality() {
             Set<ApiFilter> filters,
             PaginationParameters paginationParameters
     ) {
-        int perPage = paginationParameters.getPerPage();
-        validatePerPage(perPage);
```
cdeszaq: I don't think we should lose the validation step. I realize that we're not using `perPage` directly, but we still want to validate it.

cdeszaq: Also, the fact that removing this didn't break any tests means that we're missing some tests around this...

QubitPi: We could put validation in places where `perPage` is used, in this case `getResultsPage`, couldn't we?

cdeszaq: The methods called do use the `paginationParameters` object, so we still need to do the validation. If the `getResultsPage` method _also_ does the validation, we can drop it from here, but if not, we should keep this.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -543,6 +540,7 @@ private Query getFilterQuery(Set<ApiFilter> filters, int perPage) {
     private Pagination<DimensionRow> getResultsPage(Query query, PaginationParameters paginationParameters)
             throws PageNotFoundException {
         int perPage = paginationParameters.getPerPage();
+        validatePerPage(perPage);
```
cdeszaq: If we're adding it here, we should remove it from `findAllDimensionRowsPaged` as well, so we don't validate twice 😉

PR #137, opened 2017-01-14
Dimension Field Tagging and Dynamic Dimension Field Serilization

garyluoex:
* Added a new module `fili-navi` for components added to support for Navi
* Added `TaggedDimensionField` and related components in `fili-navi`
* Changed `fili-core` dimension endpoint `DimensionField` serialization strategy from hard coded static attributes to dynamic serialization based on `jackson` serializer 
* Deprecated `DimensionsServlet::getDimensionFieldListSummaryView` and `DimensionsServlet::getDimensionFieldSummaryView`
    since there is no need for it anymore due to the change in serialization of `DimensionField`


fili-navi/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/DefaultDimensionFieldTag.java
```diff
@@ -0,0 +1,28 @@
...
+    PRIMARY_KEY
+    ;
+
+    private final String tagName;
+
+    /**
+     * Constructor, build name using camel cased enum name.
+     */
+    DefaultDimensionFieldTag() {
+        this.tagName = EnumUtils.camelCase(name());
```
garyluoex: CamelCase or just LowerCase? (Does LowerCase feel more like tags?)

michael-mclawhorn: A quick survery of online tag systems suggests that they are mostly flat cased and searched case insensitively.  I agree that all-lower is the right impulse.

fili-navi/src/main/java/com/yahoo/bard/webservice/data/dimension/TaggedDimensionField.java
```diff
@@ -0,0 +1,18 @@
...
+ * Dimension field with tags.
+ */
+public interface TaggedDimensionField extends DimensionField {
+
+    /**
+     * Get a list of tags associated to the current field.
+     *
+     * @return a list of tags
+     */
+    List<? extends Tag> getTags();
```
michael-mclawhorn: I think tags should probably be a set.  Maybe even a sorted set to encourage canonical display.

fili-navi/src/test/java/com/yahoo/bard/webservice/data/dimension/TestTaggedDimensionField.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.List;
+
+/**
+ * Test implementation of TaggedDimensionField used in testings.
+ */
+public enum TestTaggedDimensionField implements TaggedDimensionField {
+    TEST_PRIMARY_KEY(Collections.singletonList(PRIMARY_KEY)),
+    TEST_DISPLAY_NAME(Arrays.asList(PRIMARY_KEY, PRIMARY_KEY)),
+    TEST_DESCRIPTION(Collections.<Tag>emptyList())
+    ;
```
michael-mclawhorn: If these fields exist only to illustrate behavior on single tag, multiple tags and no tags, you might want to just rename the dimensions to reflect that purpose.  

Also, would it be clearer to instantiate these tags as singletons in the test class?  That would make it easer to see what you're testing by putting it next to the test itself, if the test is the only place these are being consumed.

fili-navi/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/DefaultDimensionFieldTag.java
michael-mclawhorn: If we do make them sortable, I'd make DimensionTag comparable on getName()

fili-navi/src/main/java/com/yahoo/bard/webservice/data/dimension/Tag.java
```diff
@@ -0,0 +1,19 @@
+// Copyright 2017 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.data.dimension;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/**
+ * An interface for tag.
```
garyluoex: I was thinking maybe one day we might need tag for other things, so this serves as a general purpose tag with a name interface. Let me know if you think we should indicate that it is for dimension field for now.

cdeszaq: Since this is tucked under the dimension package, and we only have 1 use for tags at the moment, I think it's OK to keep it tight. If we need to make it more widely used later, that should be easy enough to do.

fili-navi/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/DefaultDimensionFieldTag.java
```diff
@@ -0,0 +1,29 @@
...
+    PRIMARY_KEY
+    ;
+
+    private final String tagName;
+
+    /**
+     * Constructor, build name using camel cased enum name.
+     */
+    DefaultDimensionFieldTag() {
+        this.tagName = name().toLowerCase(Locale.ENGLISH);
```
garyluoex: Well not all, the physical names of things are all* lower cased, the logical names of things are all* camel cased. Let me know if you still want it to be camel case, thanks! 

* Well, all that I know of

cdeszaq: Right, I meant all the _external_ names. Internal names, in general, _do_ use `snake_case`, but external uses `camelCase`. (that's one distinction that actually helps identify where an identifier is coming from)

I'd say we should still go with `camelCase` for consistency

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/SerializationResources.groovy
```diff
@@ -334,6 +334,6 @@ class SerializationResources extends Specification {
     }
 
     String getSerializedReponseContext1() {
-        """["com.yahoo.bard.webservice.web.responseprocessors.ResponseContext",{"randomHeader":"someHeader","apiMetricColumnNames":["java.util.LinkedHashSet",["metric1, metric2"]],"requestedApiDimensionFields":["java.util.LinkedHashMap",{"ageBracket":["java.util.LinkedHashSet",[["com.yahoo.bard.webservice.data.dimension.BardDimensionField","ID"]]]}]}]"""
+        """["com.yahoo.bard.webservice.web.responseprocessors.ResponseContext",{"randomHeader":"someHeader","apiMetricColumnNames":["java.util.LinkedHashSet",["metric1, metric2"]],"requestedApiDimensionFields":["java.util.LinkedHashMap",{"ageBracket":["java.util.LinkedHashSet",[["com.yahoo.bard.webservice.data.dimension.BardDimensionField",{"name":"id","description":"Dimension ID"}]]]}]}]"""
```
garyluoex: It seems like it is a custom serialization done in [PreResponseSerializationProxy](https://github.com/yahoo/fili/blob/369787a86097e19c672072c8fe4e5b76e4fc012a/fili-core/src/main/java/com/yahoo/bard/webservice/data/PreResponseSerializationProxy.java#L60)

cdeszaq: The piece I'm most concerned about is the deserialization of this thing. That's what usually breaks when serialization / deserialization don't line up. Since this capability is lightly (if ever) used over a duration that spans more than a single request, it's probably OK to set this asside.

michael-mclawhorn:
👍

PR #138, opened 2017-01-18
Initial commit of metric parser.

jacobtolar:
For comments/suggestions (obviously, it's a little short on tests at the moment...I will write some once the approach is finalized).

It's gone through a few iterations: 
1. There was an initial version of this in #104, (and versions before that) that were all hand-written parsing code.
2. There was a version that used this antlr grammar for lexing and implemented the visitor interface, but kept around the `MetricNode`s used in #104. I tested this against our Druid cluster and it seemed to be working well.
3. This version: all of the custom parsing code was ripped out and replaced with a parser that  directly produces native Bard objects (`Filter`s and `LogicalMetric`s). It wasn't that hard, I should have just done it like that in the first place...

I've done some basic tests of (2) against Druid, so the antlr grammar here is...at least reasonably correct. (3) (this PR) I haven't built or tested against a real cluster, so I may have missed a couple little things when refactoring...

This does depend on the`MakerBuilder` code in #108 (I was thinking they were totally independent, but not quite), so it won't build with maven as-is.

### DIFF
--- a/fili-metric-parser/README.md
+++ b/fili-metric-parser/README.md
@@ -1,0 +1,1 @@
+Metric grammar and parser for Fili metrics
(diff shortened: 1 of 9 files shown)


jacobtolar:
Moved [here](https://github.com/jacobtolar/fili-config-utils/tree/master/fili-metric-parser) for now, although this is a much smaller and simpler change than #108.

PR #141, opened 2017-01-18
Fix some timers.

archolewa:
--We forgot to stop the `DataApiRequest` timer.

--When timing the posting of the Druid Query we
failed to account for the thread hopping during
asynchronous programming. As a result, not only
were we failing to stop the timer in the correct
instance of the RequestLog (the one that is used
by the response handling thread), but we also
attempted to stop a timer in a RequestLog that
has since been cleared. So we end up with two
warnings: One for attempting to stop a timer that
doesn't exist, and one for logging a timer that
was never stopped.

Since there isn't a good way to time just the
posting of the druid query, and it probably
isn't necessary anyway, the timing has been
removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,9 +49,13 @@ Current
 
 ### Changed:
 
+- [The druid query posting timer has been removed](https://github.com/yahoo/fili/pull/141)
+    * There wasn't really a good way of stopping timing only the posting itself. Since the timer is 
+       probably not that useful, it has been removed. 
+
 - [Dimension Field Tagging and Dynamic Dimension Field Serilization](https://github.com/yahoo/fili/pull/137)
     * Changed `fili-core` dimension endpoint `DimensionField` serialization strategy from hard coded static attributes to dynamic serialization based on `jackson` serializer
-
+        
 - [MetricMaker cleanup and simplification](https://github.com/yahoo/fili/pull/127)
     * Simplified raw aggregation makers
     * `ConstantMaker` now throws an `IllegalArgumentException` wrapping the raw NumberFormatException on a bad argument
@@ -208,6 +212,9 @@ Current
      so `LogicalDimensionColumn` is no longer needed
 
 ### Fixed:
+
+- [Stopped `DataApiRequest` timer](https://github.com/yahoo/fili/pull/141):
+    * Stopping the `DataApiRequest` timer that we forgot to stop
 
 - [Added missing coverage for `ThetaSketchEstimate` unwrapping in `MetricMaker.getSketchField`](https://github.com/yahoo/fili/pull/128) 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,13 +49,18 @@ Current
 
 ### Changed:
 
+- [`RequestLog::switchTiming` has been removed]((https://github.com/yahoo/fili/pull/141)
+    - `RequestLog::switchTiming` is very context-dependent, and therefore brittle. In particular, adding any
+        additional timers inside code called by a timed block may result in the original timer not stopping
+        properly. All usages of `switchTiming` have been replaced with explicit starts and stops.
+
 - [The druid query posting timer has been removed](https://github.com/yahoo/fili/pull/141)
     * There wasn't really a good way of stopping timing only the posting itself. Since the timer is 
        probably not that useful, it has been removed. 
 
 - [Dimension Field Tagging and Dynamic Dimension Field Serilization](https://github.com/yahoo/fili/pull/137)
     * Changed `fili-core` dimension endpoint `DimensionField` serialization strategy from hard coded static attributes to dynamic serialization based on `jackson` serializer
-        
+
 - [MetricMaker cleanup and simplification](https://github.com/yahoo/fili/pull/127)
     * Simplified raw aggregation makers
     * `ConstantMaker` now throws an `IllegalArgumentException` wrapping the raw NumberFormatException on a bad argument
@@ -212,9 +217,6 @@ Current
      so `LogicalDimensionColumn` is no longer needed
 
 ### Fixed:
-
-- [Stopped `DataApiRequest` timer](https://github.com/yahoo/fili/pull/141):
-    * Stopping the `DataApiRequest` timer that we forgot to stop
 
 - [Added missing coverage for `ThetaSketchEstimate` unwrapping in `MetricMaker.getSketchField`](https://github.com/yahoo/fili/pull/128) 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,10 +49,11 @@ Current
 
 ### Changed:
 
-- [`RequestLog::switchTiming` has been removed]((https://github.com/yahoo/fili/pull/141)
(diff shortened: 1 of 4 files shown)


michael-mclawhorn:
Reading this code makes me all sad inside.

archolewa:
Writing this code a while back filled me with RAGE. :rage4:

michael-mclawhorn:
👍

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/DataServlet.java
```diff
@@ -365,6 +365,7 @@ public void getData(
                     uriInfo,
                     this
             );
+            RequestLog.stopTiming("DataApiRequest");
```
cdeszaq: Why was this needed? The call on line 374 should switch from the timer that was open to the next one. If anything, perhaps we should wrap a timing phase around the request mapping (which would also stop this one). I guess my question is: Under what conditions was this timer not being stopped, since it looks like this shouldn't be a problem based on the code in this file.

archolewa: Oh, that's right, I forgot that switching also stops timers.

archolewa: Hmm. I'll need to do some more digging. I know that we're having a similar problem with the hydrating dimension rows timer in Lucene. Despite the code ostensibly being there, it doesn't seem like we are actually executing the line.

archolewa: I know why `switchTimer` isn't working. Here's the implementation:

```java
stopMostRecentTimer();
startTiming(nextPhase);
```
However, the `DataApiRequest` timer is no longer the most recent timer, because of all the timings we added to the `DataApiRequest`. So, `switchTimer` doesn't do what you think it should if you have timings happening inside any method calls between `startTimer` and `switchTimer`.

michael-mclawhorn: *growls*  I always hated this behavior.

archolewa: I'm telling you, when someone says "X is convenient" you should hear "X will cause unexpected bugs a few refactors from now."

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
```diff
@@ -278,17 +278,6 @@ public static void startTiming(String timePhaseName) {
     }
 
     /**
-     * Stop the most recent stopwatch and start this one.
-     * Time is accumulated if the stopwatch is already registered.
-     *
-     * @param nextPhase  the name of the stopwatch to be started
-     */
-    public static void switchTiming(String nextPhase) {
```
cdeszaq: Should deprecate and call out in changelog, not directly remove.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/DataServlet.java
```diff
@@ -346,41 +345,60 @@ public void getData(
...
+            }
 
             if (requestMapper != null) {
                 apiRequest = (DataApiRequest) requestMapper.apply(apiRequest, containerRequestContext);
             }
 
-            RequestLog.switchTiming("DruidQueryMerge");
+            RequestLog.startTiming("DruidQueryMerge");
+            TemplateDruidQuery templateQuery;
             // Build the query template
```
cdeszaq: You're gonna hate me for this, but in these refactor spots, can we float the comment to the start of the "block" (ie. above the start of the timer)? Having it in the middle is slightly confusing.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/DataServlet.java
```diff
@@ -408,10 +426,13 @@ public void getData(
                     httpResponseMaker
             );
 
-            RequestLog.switchTiming("logRequestMetrics");
-            logRequestMetrics(apiRequest, readCache, druidQuery);
-            RequestLog.record(new DruidFilterInfo(apiRequest.getFilter()));
-            RequestLog.switchTiming(REQUEST_WORKFLOW_TIMER);
```
cdeszaq: We've lost the start to the request workflow timer I think.

archolewa: So we did.

michael-mclawhorn: Do we want to have tests to verify that the request log is correctly stimulated?

CHANGELOG.md
```diff
@@ -49,10 +49,11 @@ Current
 
 ### Changed:
 
-- [`RequestLog::switchTiming` has been removed]((https://github.com/yahoo/fili/pull/141)
+- [`RequestLog::switchTiming` has been deprecated]((https://github.com/yahoo/fili/pull/141)
```
cdeszaq: Need to call out deprecation in deprecation section of change log

CHANGELOG.md
```diff
@@ -49,6 +49,16 @@ Current
 
 ### Changed:
 
+- [`RequestLog::switchTiming` has been deprecated]((https://github.com/yahoo/fili/pull/141)
```
cdeszaq: Not sure where the other comment I left went, but this note should move into the `deprecated` section

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/DataServlet.java
archolewa: We probably do. Especially since I looked through the logs and it looks like there are a few places where we aren't properly stopping timers (the `JobsServlet` and I don't think our `TestDruidWebService` is stopping the request workflow timer correctly).

I tried poking at fixing those problems, but my quick fix in the `TestDruidWebService` didn't seem to stop all the warnings about the request workflow timer, and the `JobsServlet` desperately needs to be cleaned up. We start and stop timers all over the place. Since that was more than I wanted to take on right now, I dropped the effort.

The tricky part is that I'm not sure how to do that test. Most methods that run timers, we could probably check easily enough that all the appropriate timers have been stopped. Not sure how to verify that a timer has been started. Especially in the `DataServlet`. 

Ideally we'd use a mock, but `RequestLog` is statically accessible.

michael-mclawhorn: Well, we know a technique for fixing the static dependency, but it's a bigger effort than this PR.
Do we care about timers that have started but not stopped?

archolewa: Well, if the timers haven't been stopped, then when we go to log their durations, we'll get incorrect numbers.

michael-mclawhorn:
reaffirming my 👍

PR #142, opened 2017-01-18
Removes warning in getExtractionFunction

archolewa:
--`ModelUtil::getExtractionFunction` is a part of an optional feature,
but it only works on Dimensions that Fili supports directly in a very
non-extensible way. Therefore, if customers are using Fili with their
own implementation of `Dimension`, the customers end up with spurious
warnings about not being able to build an `ExtractionFunction`.

Since the feature isn't critical and can't be integrated with custom
dimensions right now, the log level has been reduced to debug until it
can be reworked to be more extensible.


michael-mclawhorn:
👍

PR #143, opened 2017-01-23
Makes TimedPhase auto-closeable.

archolewa:
--This would allow us to wrap timed phases in a try-with-resources
block:

```java
   try(TimedPhase phase = RequestLog.startTiming("timerAllTheThings") {
                  doAllTheThings();
    }
```

--This has several advantages over the current approach:
        1. We always want to wrap the code that stops a timer in a finally block.
            Otherwise, the timer won't be stopped if we have an
            exception. try-with-resources makes that clean.
        2. It reduces the chances of typos when starting and stopping
            timers. It also reduces the chances of people staring a timer
            and then forgetting to stop it.
        3. It is more concise.

--This has several limitations:
        1. The minimum effort code to implement it is awkward. We'll
            need to do a much more significant (and not yet clear) refactor
            of RequestLog to handle this cleanly.
        2. Timers that cross threads can't be handled this way. Those
            still need to be stopped and started manually.

--Consequences that are more mixed:
        1. It exposes a lot more of the inner workings of RequestLog.
            This could make it easier to customers to perform their own
            custom timings, but it also exposes code that is very easy
            to use incorrectly if you don't have a strong understanding
            of it.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
@@ -55,6 +55,7 @@ public class RequestLog {
 
     private String logId;
     private LogBlock info;
+    @Deprecated
     private TimedPhase mostRecentTimer;
     private final Map<String, TimedPhase> times;
     private final Set<String> threadIds;
@@ -271,7 +272,11 @@ public class RequestLog {
 
     /**
      * Stop the most recent stopwatch.
+     *
+     * @deprecated  Stopping a timer based on context is brittle, and prone to unexpected changes. Timers should be
+     * stopped explicitly, or started in a try-with-resources block
      */
+    @Deprecated
     public static void stopMostRecentTimer() {
         try {
             stopTiming(RLOG.get().mostRecentTimer.getName());
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
@@ -86,54 +86,6 @@ public class RequestLog {
     }
 
     /**
-     * Represents a phase that is timed.
-     * TimedPhase is used to associate a Timer located in the registry with the exact duration of such a phase for a
-     * specific request.
-     */
-    private static class TimedPhase {
-        private final String name;
-        private long start;
-        private long duration;
-
-        /**
-         * Constructor.
-         *
-         * @param name  Name of the phase
-         */
-        private TimedPhase(String name) {
-            this.name = name;
-        }
-
-        /**
-         * Start the phase.
-         */
-        private void start() {
-            if (start != 0) {
-                LOG.warn("Tried to start timer that is already running: {}", name);
-                return;
-            }
-            start = System.nanoTime();
-        }
-
-        /**
-         * Stop the phase.
-         */
-        private void stop() {
-            if (start == 0) {
-                LOG.warn("Tried to stop timer that has not been started: {}", name);
-                return;
-            }
-            duration += System.nanoTime() - start;
-            REGISTRY.timer(name).update(duration, TimeUnit.NANOSECONDS);
-            start = 0;
-        }
-
-        private boolean isStarted() {
-            return start != 0;
-        }
-    }
-
-    /**
      * Resets the contents of a request log at the calling thread.
      */
     private void clear() {
@@ -180,19 +132,7 @@ public class RequestLog {
      * @return the map containing all the recorded times per phase in milliseconds
      */
     private Map<String, Long> getDurations() {
-        return times.values()
-                .stream()
-                .peek(
-                        phase -> {
-                            if (phase.start != 0) {
-                                LOG.warn(
-                                        "Exporting duration while timer is running. Measurement might be wrong: {}",
-                                        phase.name
-                                );
-                            }
-                        }
-                )
-                .collect(Collectors.toMap(phase -> phase.name, phase -> phase.duration));
+        return times.values().stream().collect(Collectors.toMap(TimedPhase::getName, TimedPhase::getDuration));
     }
 
     /**
@@ -221,15 +161,15 @@ public class RequestLog {
     }
 
     /**
-     * Check if a stopwatch is started.
+     * Check if a stopwatch is currently running.
      *
      * @param caller  the caller to name this stopwatch with its class's simple name
      *
      * @return whether this stopwatch is started
      */
 
-    public static boolean isStarted(Object caller) {
-        return isStarted(caller.getClass().getSimpleName());
+    public static boolean isRunning(Object caller) {
+        return isRunning(caller.getClass().getSimpleName());
     }
 
     /**
@@ -239,10 +179,10 @@ public class RequestLog {
      *
      * @return whether this stopwatch is started
      */
-    public static boolean isStarted(String timePhaseName) {
+    public static boolean isRunning(String timePhaseName) {
         RequestLog current = RLOG.get();
         TimedPhase timePhase = current.times.get(timePhaseName);
-        return timePhase != null && timePhase.isStarted();
+        return timePhase != null && timePhase.isRunning();
     }
 
     /**
@@ -250,9 +190,11 @@ public class RequestLog {
      * Time is accumulated if the stopwatch is already registered
      *
      * @param caller  the caller to name this stopwatch with its class's simple name
+     *
+     * @return The stopwatch
      */
-    public static void startTiming(Object caller) {
-        startTiming(caller.getClass().getSimpleName());
+    public static TimedPhase startTiming(Object caller) {
+        return startTiming(caller.getClass().getSimpleName());
     }
 
     /**
@@ -260,8 +202,10 @@ public class RequestLog {
      * Time is accumulated if the stopwatch is already registered
      *
      * @param timePhaseName  the name of this stopwatch
+     *
+     * @return The stopwatch
      */
-    public static void startTiming(String timePhaseName) {
+    public static TimedPhase startTiming(String timePhaseName) {
         RequestLog current = RLOG.get();
         TimedPhase timePhase = current.times.get(timePhaseName);
         if (timePhase == null) {
@@ -274,7 +218,7 @@ public class RequestLog {
             current.times.put(timePhaseName, timePhase);
         }
         current.mostRecentTimer = timePhase;
-        timePhase.start();
+        return timePhase.start();
     }
 
     /**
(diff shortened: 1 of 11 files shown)


archolewa:
There are a couple of changes here:

1. I moved `TimedPhase` out into its own class, and made it beanlike. This isn't strictly necessary, but makes it a little bit cleaner IMO since it can now show up in the code independent of the RequestLog.

2. Stopping a timer and storing its duration in the RequestLog have been split into two steps:
    1. Stop the timer.
    2. Store the timer's final time in the RequestLog's registery.

This was necessary because `TimedPhase` no longer has insight into the guts of `RequestLog`.

 Both of these steps have been encapsulated in `TimedPhase::close`, which is automatically invoked at the end of a try-with-resources block with the `finally` semantics.

michael-mclawhorn:
I like this approach a lot

fili-core/src/main/java/com/yahoo/bard/webservice/logging/TimedPhase.java
```diff
@@ -0,0 +1,88 @@
...
+            return;
+        }
+        start = System.nanoTime();
+    }
+
+    /**
+     * Stop the phase.
+     */
+    public void stop() {
+        if (start == 0) {
```
cdeszaq: Would be better to convert to use the `isStarted` method

fili-core/src/main/java/com/yahoo/bard/webservice/logging/TimedPhase.java
```diff
@@ -0,0 +1,88 @@
...
+     */
+    public TimedPhase(String name) {
+        this(name, TimeUnit.NANOSECONDS);
+    }
+
+    /**
+     * Start the phase.
+     */
+    public void start() {
+        if (start != 0) {
```
cdeszaq: Would be better to convert to use the `isStarted` method

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
```diff
@@ -275,6 +231,7 @@ public static void startTiming(String timePhaseName) {
         }
         current.mostRecentTimer = timePhase;
         timePhase.start();
+        return timePhase;
```
cdeszaq: Just make `TimedPhase::start` return itself?

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
```diff
@@ -86,54 +86,6 @@ private RequestLog(RequestLog rl) {
...
-        /**
-         * Stop the phase.
-         */
-        private void stop() {
-            if (start == 0) {
-                LOG.warn("Tried to stop timer that has not been started: {}", name);
-                return;
-            }
-            duration += System.nanoTime() - start;
-            REGISTRY.timer(name).update(duration, TimeUnit.NANOSECONDS);
```
cdeszaq: I think this is getting lost. It may very well be that `TimedPhase::stop` needs to do the registration that `TimedPhase::close` is doing? Otherwise, I think there's a gap where things could be lost.

Or, to put it a different way, `stop` did the registering before, and now only `close` does it. Since `stop` was private, I don't know how much impact it will really have, but it's worth noting.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
```diff
@@ -308,15 +277,15 @@ public static void stopTiming(String timePhaseName) {
             LOG.warn("Tried to stop non-existent phase: {}", timePhaseName);
             return;
         }
-        timePhase.stop();
+        timePhase.close();
```
cdeszaq: If we make "stop" convert to doing the registering, calling "close" is likely not needed, but "stop" will be.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
archolewa: I'm not sure I understand your concern about registering "getting lost."  It's now invoked as part of the `TimedPhase::close`, which is being called by `RequestLog::stopTiming`. So the behavior of the public interface hasn't changed. So this change will only affect people if they are doing Unsupported Things.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
archolewa: `close` has to do the registering anyway, because that's the method invoked when the `try-with-resources` block ends. So even if we didn't change how `close` works, and had `close` delegate to `stop`, we still wouldn't need to change this invocation.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/TimedPhase.java
```diff
@@ -0,0 +1,88 @@
+// Copyright 2016 Yahoo Inc.
```
archolewa: *fist shake* JANUARRYY!!!!

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
archolewa: Plus we shouldn't change this from `stop` to `close`. In my opinion, `RequestLog::stopTiming` should have the same semantics as what the try-with-resources block does (with the exception of the existence check that the try-with-resources doesn't have to worry about).

archolewa:
Since there seems to be strong support for this change, I'm going to flesh it out into a proper PR tomorrow (updating timers that could benefit from it, adding a CHANGELOG, etc).

I'm also going to squash the commits that are currently here down into logical commits to facilitate a proper review.

michael-mclawhorn:
👍

fili-core/src/main/java/com/yahoo/bard/webservice/logging/TimedPhase.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public String getName() {
+        return name;
+    }
+
+    public TimeUnit getUnit() {
+        return TimeUnit.NANOSECONDS;
+    }
+
+    public boolean isStarted() {
```
cdeszaq: This may seem stupid, but should we change the name of this method to `isRunning`? We seem to be using it to indicate if this `TimedPhase` is currently _running_ or not, and `isStarted` feels more like a "once I've been started, I've always been started" sort of thing, where as running / stopped feels more like a toggleable state.

Obviously this is a small thing, but every time I see `isStarted`, I noticed I had to mentally pause and convert to `isRunning`.

archolewa: 👍

fili-core/src/main/java/com/yahoo/bard/webservice/logging/TimedPhase.java
```diff
@@ -0,0 +1,79 @@
...
+        if (!isStarted()) {
+            LOG.warn("Tried to stop timer that has not been started: {}", name);
+            return;
+        }
+        duration += System.nanoTime() - start;
+        start = 0;
+    }
+
+    public long getDuration() {
+        return duration;
```
cdeszaq: There are two places in `RequestLog` where the duration is retrieved, and those places are just after an `isStarted` check. Could we move that `isStarted` check / warning log into this getter so that the `RequestLog` doesn't have to worry about it?

fili-core/src/main/java/com/yahoo/bard/webservice/logging/TimedPhase.java
```diff
@@ -0,0 +1,79 @@
...
+ * Represents a phase that is timed.
+ * TimedPhase is used to associate a Timer located in the registry with the exact duration of such a phase for a
+ * specific request.
+ */
+public class TimedPhase implements AutoCloseable {
+    private static final Logger LOG = LoggerFactory.getLogger(TimedPhase.class);
+
+    private final String name;
+    private long start;
+    private long duration;
```
cdeszaq: I think `start` and `duration` need to be made thread-safe (likely with AtomicLongs)... It wasn't an issue before, because `TimedPhase` was encapsulated within the `RequestLog`, which was a `ThreadLocal` and was never able to exist on more than 1 thread at a time, but `TimedPhase` is now it's own _thing_, and there's a possibility of sharing them across threads.

I tried playing around with using `AtomicLong` for these variables a little bit, and it's not hard to retain the existing semantics, though it did make me think about if we _wanted_ to maintain the existing semantics around multiple consecutive starts / stops, especially in the new world of being across threads. No good solutions to this came to mind immediately, so it may make sense to add a cautionary note to the class-level doc for `TimedPhase` indicating the _possible_ race condition of starting and stopping the same phase on multiple threads.

Anyway, here's the relevant results of my exploration, in case they help:

```java
private static final long NOT_RUNNING = 0;

private final String name;
private final AtomicLong start = new AtomicLong();
private final AtomicLong duration = new AtomicLong();

/**
 * Start the phase.
 *
 * @return This phase after being started
 */
public TimedPhase start() {
    if (!start.compareAndSet(NOT_RUNNING, System.nanoTime())) {
        LOG.warn("Tried to start timer that is already running: {}", name);
    }
    return this;
}

/**
 * Stop the phase.
 */
public void stop() {
    long started = start.getAndUpdate(ignoredCurrent -> NOT_RUNNING);
    if (started == 0) {
        LOG.warn("Tried to stop timer that has not been started: {}", name);
        return;
    }
    duration.addAndGet(System.nanoTime() - started);
}

public long getDuration() {
    return duration.get();
}

public boolean isStarted() {
    return start.get() != 0;
}
```

cdeszaq:
A few small things, and medium thing (the warning logger), and a large thing (the thread safety). I'm OK punting on the `TimedPhase` thread safety (I think) as long as we _clearly_ note that it's _not_ a thread-safe class if both classes are starting / stopping. It would be great to close that gap, but I'm happy to take the clarity offered by it being closable, and coming back to making TimedPhase thread-safe. (A techdebt ticket might be a good idea)

Other than the thread safety issue, if everything else is updated, this is good to merge.

fili-core/src/main/java/com/yahoo/bard/webservice/logging/TimedPhase.java
archolewa: I'm ok with making it thread safe, especially since many of these timers are stopped on a different thread than they're started (the request vs response thread). Rethinking the semantics is outside the scope of this PR though, I think.

archolewa: Actually, after playing with it a little bit, I think I'm going to punt on this, largely because validating whether a timer is running outside of this class or not becomes a lot trickier in a multi-threaded scenario.

archolewa:
Ticket to revisit making timers thread-safe: https://github.com/yahoo/fili/issues/185

PR #144, opened 2017-01-23
Replacing regexp based filter parsing with ANTLR grammer

will-lauer:
Adding ANTLR based grammer to simplify and sanitize the parsing of 
filter expressions. 

Several backward incompatible changes are occuring because of this,
including:

	- whitespace before and after filter expressions is now stripped
	- ] and , are now valid characters within a quoted string
	- double quote can be escaped within a quoted string as ""
	- error messages in the event of bad filters have changed
	- filters must terminate correctly now instead of allowing an open [ 
    to be terminated by the end of the string in addition to ]


archolewa:
We should add some documentation somewhere describing what commands people should run to generate the Java files, so that people know what to run to get their IDE's to stop screaming at them about unresolved classes.

will-lauer:
The best way to generate the necessary files is to run maven. The command `mvn generate-sources` is typically how this is done. Some IDEs know how to do this automatically for you. For Eclipse, there is an ANTLR plugin that can do it, and I've writing an Eclipse/Maven integration that detects it automatically. For IntelliJ, I suggest looking at https://github.com/antlr/intellij-plugin-v4, although that won't integrate cleanly with maven.

archolewa:
@will-lauer `mvn generate sources` won't work for us, I don't think, since we have modules that depend on the jars built by other modules, and those jars aren't built with `mvn generate sources`. Same reason `mvn compile` doesn't seem to work:

```
Failed to execute goal on project fili-core: Could not resolve dependencies for project com.yahoo.fili:fili-core:jar:0.7-SNAPSHOT: Could not find artifact com.yahoo.fili:fili-system-config:jar:0.7-SNAPSHOT in artifactory-snapshots (http://edge.artifactory.yahoo.com:8000/artifactory/maven-local-snapshot)
```

will-lauer:
That's a different error, for which we do know the solution but just haven't implemented it yet. mvn generate-sources is still the prefered method of generating the files. You can run it within fili-core if you need to instead of from the top level.
Will Will Lauer
Tech Yahoo, Software Dev Eng, Principal
P: 217.255.4262  M: 508.561.6427 
1908 S First St. Champaign IL 61820

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -992,45 +1001,39 @@ protected DateTime getCurrentDate(DateTime dateTime, TimeGrain timeGrain) {
...
+            parser.addErrorListener(ExceptionErrorListener.INSTANCE);
 
-                    }
-                }
-                Dimension dim = newFilter.getDimension();
-                if (!generated.containsKey(dim)) {
-                    generated.put(dim, new LinkedHashSet<>());
+            try {
+                try {
+                    FiltersParser.FiltersContext tree = parser.filters();
```
archolewa: [ditto](https://github.com/yahoo/fili/pull/144/files#r98962547).

Also, it looks like this is pretty much identical to the linked code in the `ApiFilter`. We may want to pull this logic out into a utility method. Ideally, it'd be something that could take a `ParseTreeListener`, and an AntlrInputStream and that populates the listener with the appropriate values. That should make it easier to integrate other ANTRL grammars as well.

fili-core/src/main/java/com/yahoo/bard/webservice/web/filterparser/ApiFilterListener.java
```diff
@@ -0,0 +1,82 @@
...
+        super(table, dimensionDictionary);
+    }
+
+    /* (non-Javadoc)
+     * @see com.yahoo.bard.webservice.web.filterparser.ApiFiltersListListener#exitFilter(
+     * com.yahoo.bard.webservice.web.filterparser.FiltersParser.FilterContext)
+     */
+    @Override
+    public void exitFilter(FilterContext ctx) {
+        FiltersParser.FilterComponentContext f = ctx.filterComponent();
```
archolewa: Could we get a better variable name than just `f`? Something like `filterComponent` would be much readable in my opinion.

fili-core/src/main/java/com/yahoo/bard/webservice/web/filterparser/ApiFilterListener.java
```diff
@@ -0,0 +1,82 @@
...
+        dimension = extractDimension(f);
+        if (dimension == null) {
```
archolewa: Was our old `ApiFilter` supposed to be able to support a `null` `dimension`? The code that tries to extract the dimension out of an `ApiFilter` certainly seems to assume that `dimension` will be not null. So I feel like we should either throw an exception here if `dimension` is `null` (so that we don't get a vague `NullPointerException` later), or the getters on these listeners (and even the `ApiFilter`?) should be returning `Optional`.

fili-core/src/main/java/com/yahoo/bard/webservice/web/filterparser/ApiFilterListener.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * Simple getter for the Dimension that also checks for any pending errors and throws exceptions for them.
+     * @return the Dimension
+     * @throws BadFilterException invalid filter expression
+     * @throws BadApiRequestException filter expression that doesn't match the current table or dimensions
+     */
+    public Dimension getDimension() throws BadFilterException, BadApiRequestException {
+        processErrors();
```
archolewa: I'm not a particularly big fan of this "delay on error reporting until we actually try to use the object." If we want that, we should just ditch the whole type system and go duck ;). 

Is there a reason we can't call `processErrors` in the `exitFilter` method where we build out the listener's fields?

archolewa: Also, if we _do_ stick with this behavior, we really should be invoking `processErrors` in every getter. There's no guarantee that future code will always be invoking `getDimension` before/along with the other getters.

fili-core/src/main/java/com/yahoo/bard/webservice/web/filterparser/ApiFiltersListListener.java
```diff
@@ -0,0 +1,219 @@
...
+                // there was some error and it has been saved in the errors list
+                break;
+            }
+            Dimension dimension = filter.getDimension();
+            if (!dimensionFiltersMap.containsKey(dimension)) {
+                dimensionFiltersMap.put(dimension, new LinkedHashSet<>());
+            }
+            Set<ApiFilter> filterSet = dimensionFiltersMap.get(dimension);
+            filterSet.add(filter);
+            dimensionFiltersMap.put(dimension, filterSet);
```
archolewa: Everything from the if statement on can be consolidated into a single statement with `computeIfAbsent`:

```java
dimensionFiltersMap.computeIfAbsent(filter.getDimension(), ignored -> new LinkedHashSet<>()).add(filter);
```

fili-core/src/main/java/com/yahoo/bard/webservice/web/filterparser/ApiFiltersListListener.java
```diff
@@ -0,0 +1,219 @@
...
+        }
+
+        FilterOperation operation = extractOperation(f);
+        if (operation == null) {
+            return;
+        }
+
+        Set<String> values = extractValues(f);
+
+        ApiFilter filter = new ApiFilter(dimension, dimensionField, operation, values);
```
archolewa: Can inline this.

fili-core/src/main/java/com/yahoo/bard/webservice/web/filterparser/ApiFiltersListListener.java
```diff
@@ -0,0 +1,219 @@
...
+
+        filters.put(f, filter);
+    }
+
+    /**
+     * Extract values from the filter.
+     * @param f The filter context
+     * @return the list of values
+     */
+    protected Set<String> extractValues(FiltersParser.FilterComponentContext f) {
```
archolewa: A better variable name than `f` would be really nice. This applies to the other methods of this style in this class as well.

fili-core/src/main/java/com/yahoo/bard/webservice/web/filterparser/ApiFiltersListListener.java
```diff
@@ -0,0 +1,219 @@
...
+     * @param f The filter context
+     * @param dimension The dimension for the field
+     * @return the DimensionField
+     */
+    protected DimensionField extractField(FiltersParser.FilterComponentContext f, Dimension dimension) {
+        String dimensionFieldName;
+        dimensionFieldName = f.field().getText();
+        DimensionField dimensionField;
+        try {
+            dimensionField = dimension.getFieldByName(dimensionFieldName);
```
archolewa: You could probably just have a `return` statement here, and save yourself the extra variable.

fili-core/src/main/java/com/yahoo/bard/webservice/web/filterparser/ApiFiltersListListener.java
```diff
@@ -0,0 +1,219 @@
...
+        return dimensionFiltersMap;
+    }
+
+    /**
+     * Throw any pending exceptions.
+     *
+     * @throws BadFilterException Thrown when the filter is invalid
+     * @throws BadApiRequestException Thrown for valid filtes that are invalid for the current table or dimensions
+     */
+    protected void processErrors() throws BadFilterException, BadApiRequestException {
```
archolewa: If we're going to delay throwing errors, then I think we should attempt to consolidate all the errors into a single error message with all the problems we found with the filters.

Otherwise, we could just throw the exceptions as they happen, and save ourselves all the nulls flying around in the interim.

archolewa:
@will-lauer Perhaps that'd be a great thing to add to the [README](https://github.com/yahoo/fili/blob/master/README.md#building) ;)

Though it still fails if I try to run `generate-sources` from inside `fili-core` instead of at the top, even if I haven't cleaned anything with the same error.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -992,45 +1001,39 @@ protected DateTime getCurrentDate(DateTime dateTime, TimeGrain timeGrain) {
...
-                                ErrorMessageFormat.FILTER_SUBSTRING_OPERATIONS_DISABLED.format()
-                        );
+            ANTLRInputStream input = new ANTLRInputStream(filterQuery);
+            FiltersLex lexer = new FiltersLex(input);
+            lexer.removeErrorListeners();
+            lexer.addErrorListener(ExceptionErrorListener.INSTANCE);
+            CommonTokenStream tokens = new CommonTokenStream(lexer);
+            FiltersParser parser = new FiltersParser(tokens);
+            parser.removeErrorListeners();
+            parser.addErrorListener(ExceptionErrorListener.INSTANCE);
```
michael-mclawhorn: Per discussion, move this code to a helper method to avoid redundancy.

fili-core/src/main/java/com/yahoo/bard/webservice/web/ApiFilter.java
```diff
@@ -113,74 +112,33 @@ public ApiFilter(
     ) throws BadFilterException {
         LOG.trace("Filter query: {}\n\n DimensionDictionary: {}", filterQuery, dimensionDictionary);
 
-        /*  url filter query pattern:  (dimension name)|(field name)-(operation)[?(value or comma separated values)]?
```
michael-mclawhorn: This constructor is now dead outside of test code.
Everywhere it's currently used in test code, the filter is put into a collection anyway.  Use the helper method created for DataApiRequest so there isn't a bunch of duplicated only-slightly-different code in service to tests.  Simply change the generators for those tests to expect a collection.

fili-core/src/main/java/com/yahoo/bard/webservice/web/filterparser/ApiFilterListener.java
```diff
@@ -0,0 +1,84 @@
+// Copyright 2017 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.web.filterparser;
```
michael-mclawhorn: This whole class can go away once the unnecessary constructor is retired.

fili-core/src/main/java/com/yahoo/bard/webservice/web/filterparser/ApiFilterListener.java
michael-mclawhorn: Oh, I think I see.  In the method called, the exception is trapped and this is a bailout. Correct?

michael-mclawhorn: Another way to structure this, which I'm not passionate about, but might be more idiomatic, would be to let each of the extract methods bubble the exception they create and wrap this whole extract, extract, extract, construct into a try that catches whichever exception bubbles and returns out of that catch block.

fili-core/src/main/java/com/yahoo/bard/webservice/web/filterparser/ApiFiltersListListener.java
```diff
@@ -0,0 +1,222 @@
...
+
+    /**
+     * Throw any pending exceptions.
+     *
+     * @throws BadFilterException Thrown when the filter is invalid
+     * @throws BadApiRequestException Thrown for valid filters that are invalid for the current table or dimensions
+     */
+    protected void processErrors() throws BadFilterException, BadApiRequestException {
+        for (Exception ex : errors) {
+            if (ex instanceof BadFilterException) {
```
michael-mclawhorn: If we're only returning the first error, why not bail out of parsing?  Should we be collecting all the errors into a new error and then returning it for maximum user feedback?

michael-mclawhorn:
I think this PR is dead now and picked up in other work.

Issue #146, opened 2017-01-26
GroovyTestUtils::compareErrorPayload does not handle nested queries


cdeszaq:
For single-layer queries, it finds the context and rips out the UUID pieces so that they compare cleanly. For queries with more than 1 layer, however, it doesn't clean context blocks for deeper layers, making the method essentially useless.

To fix, we need to make the cleaning process much smarter, so that it dives down into nested queries (when present) and cleans them as well, rolling up the values the right way.

/cc
@will-lauer (impacted by this bug)
@archolewa (implemented the original solution)

cdeszaq:
@archolewa did this get fixed?

archolewa:
Yes. https://github.com/yahoo/fili/pull/147

PR #147, opened 2017-01-27
Compare error payload handles nested queries

archolewa:
This addresses issue #146.


archolewa:
@cdeszaq Could always write a test for `GroovyTestUtils::compareErrorPayload`.

We could test our test code!

PR #148, opened 2017-01-27
Fixing required parameter for overriding default ArithmeticMaker roun…

michael-mclawhorn:
…ding mapper

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -249,12 +249,13 @@ Current
         synchronized. Since threads need to acquire a write lock before
         doing anything else anyway, the method is still effectively 
         synchronized.
-        
+- [Made now required constructor for ArithmeticMaker with rounding public](https://github.com/yahoo/fili/pull/148)
+
 ### Known Issues:
 
 
-
 ### Removed:
+- [Removed invalid constructor from SketchRoundUpMappepr](https://github.com/yahoo/fili/pull/148) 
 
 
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -249,12 +249,13 @@ Current
         synchronized. Since threads need to acquire a write lock before
         doing anything else anyway, the method is still effectively 
         synchronized.
-        
-### Known Issues:
+- [Made now required constructor for ArithmeticMaker with rounding public](https://github.com/yahoo/fili/pull/148)
```
cdeszaq: How is this a fix? Isn't this just a change?

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/metric/makers/ArithmeticMakerSpec.groovy
```diff
@@ -54,7 +53,7 @@ class ArithmeticMakerSpec extends Specification {
         ArithmeticMaker roundUpDivisionMaker = new ArithmeticMaker(
                 metricDictionary,
                 ArithmeticPostAggregation.ArithmeticPostAggregationFunction.DIVIDE,
-                (Function) { String it -> return new SketchRoundUpMapper() }
+                (Function) { String name -> return new SketchRoundUpMapper(name) }
```
cdeszaq: Can just be a method reference? `SketchRoundUpMapper::new`?

CHANGELOG.md
michael-mclawhorn: Because the constructor was protected so it was actually unusable despite being mandatory?

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/metric/makers/ArithmeticMakerSpec.groovy
michael-mclawhorn: SketchRoundUpMapper has two single argument constructors.  It needs a type to choose one.

PR #149, opened 2017-01-30
Clean up CHANGELOG for v0.7.36 release

cdeszaq:
The changes to the CHANGELOG are all minor cleanup and re-ordering, along with a bit of explanation about what's in this release. I also included a doc on how to do the release.

I would like to get this in today if possible so that some of the other larger changes that are in the wings can start to land. I anticipate v0.8.x including some major changes, so getting v0.7.x out as stable is important.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,44 +8,87 @@ pull request if there was one.
 Current
 -------
 
+### Changed:
+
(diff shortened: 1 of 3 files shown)


michael-mclawhorn:
👍 good enough

PR #153, opened 2017-01-30
Better message for healthcheck filter reject

cdeszaq:
Based off of previous branch. Only last commit is relevant

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/HealthCheckFilter.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/HealthCheckFilter.java
@@ -66,7 +66,7 @@ public class HealthCheckFilter implements ContainerRequestFilter {
      */
     public Optional<Map.Entry<String, Result>> getFirstUnhealthy() {
         return HealthCheckRegistryFactory.getRegistry().runHealthChecks().entrySet().stream()
-                .filter(entry -> entry.getValue().isHealthy())
+                .filter(entry -> !entry.getValue().isHealthy())
                 .findFirst();
     }
 }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/HealthCheckFilter.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/HealthCheckFilter.java
@@ -6,15 +6,13 @@ import com.yahoo.bard.webservice.application.HealthCheckRegistryFactory;
 import com.yahoo.bard.webservice.logging.RequestLog;
 
 import com.codahale.metrics.health.HealthCheck.Result;
-import com.codahale.metrics.health.HealthCheckRegistry;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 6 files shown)


fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/HealthCheckFilterSpec.groovy
```diff
@@ -166,7 +166,7 @@ class HealthCheckFilterSpec extends Specification {
     def "Healthy /data check passes"() {
         setup:
         HealthCheckFilter filter = Spy(HealthCheckFilter)
-        filter.isHealthy() >> { true }
+        filter.getFirstUnhealthy() >> { true }
```
michael-mclawhorn: Why does this even work?

cdeszaq: Interestingly enough, it doesn't. Once the tests are made to work (silly logic bugs...), there's some cast exceptions that show up. Fix coming shortly.

cdeszaq:
@michael-mclawhorn Promised update delivered. Test no longer broken, and also now cleaner, better, faster, stronger!

michael-mclawhorn:
https://www.youtube.com/watch?v=GDpmVUEjagg

fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/HealthCheckFilter.java
```diff
@@ -36,45 +34,39 @@
...
-                LOG.error(String.format("%s check failed: %s", checkEntry.getKey(), checkEntry.getValue().toString()));
-                return false;
-            }
-        }
-
-        return true;
+    public Optional<Map.Entry<String, Result>> getFirstUnhealthy() {
+        return HealthCheckRegistryFactory.getRegistry().runHealthChecks().entrySet().stream()
+                .filter(entry -> !entry.getValue().isHealthy())
+                .findFirst();
```
michael-mclawhorn: Is findFirst better than findAny here?

cdeszaq: Not really. It's unlikely to hurt anything, since the expectation is that this collection will be too small to paralelize, but in theory, this could just return _all_ failed healthchecks, and log all that are failing. If I get time before this gets a 2nd +1, I'll update this PR with those changes, otherwise I'll take it as a techdebt ticket and circle back to it.

michael-mclawhorn: I would LOVE to see all the health check failures logged.

michael-mclawhorn:
👍

archolewa:
👍 To the stuff I didn't add.

@michael-mclawhorn @cdeszaq Decided just to do what Mike asked because I was here reviewing it, and it was simple (famous last words).

PR #155, opened 2017-01-30
Add dim fields to table full view

### LINKED ISSUE #151: Request for dimension field tags tags to be added to the table fullView response format

cdeszaq:
Fixes #151 

Note: Built on top of the v0.8.x line, so only the last commit is relevant.

/cc @jkusa

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/TableFullViewProcessor.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/TableFullViewProcessor.java
@@ -2,6 +2,7 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.web;
 
+import com.yahoo.bard.webservice.data.dimension.Dimension;
 import com.yahoo.bard.webservice.table.LogicalTable;
 import com.yahoo.bard.webservice.web.endpoints.DimensionsServlet;
 import com.yahoo.bard.webservice.web.endpoints.MetricsServlet;
@@ -9,10 +10,14 @@ import com.yahoo.bard.webservice.web.endpoints.MetricsServlet;
 import org.apache.commons.lang.StringUtils;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.ws.rs.core.UriInfo;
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -116,7 +116,7 @@ dependencies {
 
 The most bleeding-edge version is: [![Bleeding-edge](https://api.bintray.com/packages/yahoo/maven/fili/images/download.svg)](https://bintray.com/yahoo/maven/fili/_latestVersion)
 
-The most recent stable version is: [![Stable](https://img.shields.io/badge/Stable-0.6.29-blue.svg)](https://bintray.com/yahoo/maven/fili/0.6.29)
+The most recent stable version is: [![Stable](https://img.shields.io/badge/Stable-0.7.36-blue.svg)](https://bintray.com/yahoo/maven/fili/0.7.36)
```
garyluoex: Do you think we should put the stable version "pill" (shape thing) thing on the top along with the other ones? Since there is a download "pill" there also.

fili-core/src/main/java/com/yahoo/bard/webservice/web/TableFullViewProcessor.java
```diff
@@ -95,14 +98,44 @@ public TableGrainView formatTableGrain(LogicalTable logicalTable, String grain,
         resultRow.put("longName", StringUtils.capitalize(grain));
         resultRow.put("description", "The " + logicalTable.getName() + " " + grain + " grain");
         resultRow.put("retention", logicalTable.getRetention().toString());
-        resultRow.put(
-                "dimensions",
-                DimensionsServlet.getDimensionListSummaryView(logicalTable.getDimensions(), uriInfo)
-        );
+        resultRow.put("dimensions", getDimensionListFullView(logicalTable.getDimensions(), uriInfo));
```
garyluoex: What is the reason we make this change here instead of in `DimensionsServlet`? If we make change in `DimensionsServlet` things are in sync which creates less duplication and complication that feels nicer unless there is a reason for not doing so.

cdeszaq: The reason the change is here, rather than in the dim servlet, is to keep the change to the external API as small as possible. If we change it in the dim servlet, then we end up changing the dimension view as well, which I didn't want to do.

What _really_ needs to happen is we need to put an abstraction in place so that these formats can be their own extensions and are pluggable, but I didn't want to tackle that in this PR.

README.md
cdeszaq: Tucking the stable version ID down here, rather than at the top, is very much intentional. We'd _rather_ users are on the bleeding-edge, since that lets us get feedback quickly, but we also want a path for those who are more cautious and want to absorb fewer changes.

Also, having 2 version numbers / download links in the top would be a bit confusing as well 😄

cdeszaq:
@michael-mclawhorn In theory, showing usage of field tags was added back [when the capability was added](https://github.com/yahoo/fili/commit/2ae0605ea97659cc79fe5423b52869bd8a24c810). If there are not sufficient tests or documentation or examples, we should certainly address that (please open an issue with what you find lacking). Since @garyluoex implemented the dim field tags initially, I'm sure he wants them to thrive and be used, so he'll probably be happy to fill in any gaps we identify 😉

cdeszaq:
@michael-mclawhorn what did you mean by this:

> I wish we had a more scalable way to update the acceptance blocks for the containerized test, but then we'd be comparing code to code.

If you mean "I wish this PR didn't have the same block of copy-paste code in the expected JSON", I'm with you. And there _are_ smarter / more automated ways to have that sort of thing generated, we just haven't used them much. For example, we could easily embed fairly simple control structures in the JSON, and "build" it much more than we do. We've just not taken the time to do so 😢 

So, if you're saying what I think you are, then yes, there _are_ better ways. We just don't use them enough, and I didn't want to totally re-work this test for this PR to keep the changes sufficiently small.

michael-mclawhorn:
👍

PR #158, opened 2017-02-01
RFC and design for composite table availability design do not merge

michael-mclawhorn:
A design for producing synthetic availabilities for tables with more than one backing datasource.

### DIFF
--- a/rfc/composite_table_availability.pseudocode
+++ b/rfc/composite_table_availability.pseudocode
@@ -10,10 +10,10 @@ DataSource:
 	Set<FieldName> getFields() //union of all keys
 	List<Interval> getIntervals(FieldName fieldName)
 	Map<FieldName,List<Interval>> getIntervals(Set<FieldName>) // map slice from squashedMap
+	List<Interval> getAllIntervals()
 	Set<SegmentId> getSegments(List<Interval>)  // Filter all segments by overlap on time
 
 DataSourceConstraints // Simple bean parameter object
-
 	getGrain
 	Set<FieldName> getMetrics
 	Set<Dimension> getDimensions
@@ -22,8 +22,7 @@ DataSourceConstraints // Simple bean parameter object
 	DataSourceConstraints(ApiRequest)
 	DataSourceConstraints (DruidQuery, ApiFilter apifilters)
 
-	withMetrics(Set<FieldName> metrics) // Copy constructor
-
+	withMetrics(Set<FieldName> metrics) // Copy builder
 
 DataSourceAvailabilityResolver:
 	List<String> getDataSourceNames(DataSourceConstraints)
@@ -32,7 +31,6 @@ DataSourceAvailabilityResolver:
 ConcreteAvailabilityResolver:
 	DataSource dataSource
     Map<LogicalFieldName, PhysicalFieldName> fieldMap // may just be string/string
-    joiner = IntersectJoiner
 
 	ConcreteAvailability(DataSource, fieldMap)
 
@@ -41,27 +39,27 @@ ConcreteAvailabilityResolver:
 
     getAvailability(dsConstraints):
         fieldNames = Union(dsConstraints.metrics, dsConstraints.dimensions.map(fieldMap.get(it.name))
-        return dataSource.getIntervals(fieldNames).reduce(joiner)
+        return dataSource.getIntervals(fieldNames).reduce(intersectJoin)
 
     Description:
         For a single datasource -> select intervals by requested field names -> join intervals
 
-PermissiveAvailability: extends ConcreteAvailabilityResolver
-    joiner = UnionJoiner  // Hippodrome uses this logic
-
+PermissiveAvailabilityResolver: extends ConcreteAvailabilityResolver
+    getAvailability(dsConstraints):
+        return dataSource.getAllIntervals()
 
 MetricUnionAvailabilityResolver
-	Map<ConcreteAvailabilityResolver, Set<Metric>> availabilityMetrics
+	Map<AvailabilityResolver, Set<Metric>> availabilityMetrics
 	Set<Metric> metrics // optional optimization
 
-	MetricUnionAvailabilityResolver(Map<ConcreteAvailability, Set<Metric> availMetrics)
+	MetricUnionAvailabilityResolver(Map<Availability, Set<Metric> availMetrics)
 		// Validate total schema
 		metrics = availMetrics.values.reduce(Union)
 		metrics.allMatch(
 			availabilityMetrics.entries.filter(entry.value.contains(it)).count() == 1
 		)
 
-    private Stream<Map.Entry<CAR, Set<Metric>>>> ConcreteAvailabilityResolver getDependentAvails(dsConstraints):
+    private Stream<Map.Entry<AvailabilityResolver, Set<Metric>>>> getDependentAvailabilities(dsConstraints):
         validate(metrics.containsAll(dsConstraints.metrics))
         availabilityMetrics.entries
                 .filter(entry.value intersect aReq.metrics not empty)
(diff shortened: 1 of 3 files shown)


archolewa:
I don't see anything objectionable. The only thing that I think is lacking, is where the resolvers will live, and how we associate them with physical tables.

michael-mclawhorn:
An issue I left out because it continues to be personally controversial.

rfc/composite_table_availability.pseudocode
```diff
@@ -0,0 +1,123 @@
+
+PhysicalTableSchema:
+	getFieldName(Column c)
+
+DataSource:
+	String getName()
+	Map<FieldName, Set<Segment>> raw segment data
+	Map<FieldName, List<Interval>> squashedFromSegments  // optional optimization
```
cdeszaq: This is a SimplifieldIntervalList, right?

rfc/composite_table_availability.pseudocode
```diff
@@ -0,0 +1,123 @@
...
+
+	Set<FieldName> getFields() //union of all keys
+	List<Interval> getIntervals(FieldName fieldName)
+	Map<FieldName,List<Interval>> getIntervals(Set<FieldName>) // map slice from squashedMap
+	Set<SegmentId> getSegments(List<Interval>)  // Filter all segments by overlap on time
```
cdeszaq: This method here confuses me... The rest make sense, since they are really just helpers for getting at the internal `Map<FieldName, Set<Segment>>` in a clean, simple, non-map way, but this method is less clear as to why it's here.

I'm guessing it's here to serve the cache fingerprint use-case? If so, it's not clear to me how it will do it's job and be sure it's correct, since there's no top-level `Set<Segment>` structure in here, only one broken out by `FieldName`, which doesn't preclude having different segments for the same time ranges for different `FieldName`s?

rfc/composite_table_availability.pseudocode
```diff
@@ -0,0 +1,123 @@
...
+
+DataSourceConstraints // Simple bean parameter object
+
+	getGrain
+	Set<FieldName> getMetrics
+	Set<Dimension> getDimensions
```
cdeszaq: For these, what level are we assuming the they are at? Logical level? Physical level? Are these already mapped down to their underlying Logical columns? Underlying _physical_ columns?

rfc/composite_table_availability.pseudocode
```diff
@@ -0,0 +1,123 @@
...
+	getApiFilters
+
+	DataSourceConstraints(ApiRequest)
+	DataSourceConstraints (DruidQuery, ApiFilter apifilters)
```
cdeszaq: I don't think this can take a `DruidQuery`, since there shouldn't be a physical table selected yet when this is used?

rfc/composite_table_availability.pseudocode
```diff
@@ -0,0 +1,123 @@
...
+
+	withMetrics(Set<FieldName> metrics) // Copy constructor
+
+
+DataSourceAvailabilityResolver:
+	List<String> getDataSourceNames(DataSourceConstraints)
```
cdeszaq: I'm still not sure I like returning strings instead of objects here (following the idea of "move to string land as late as possible"), but until we see this in action, I don't have any specific objections, since it may be that the only thing after this is called is serialization

rfc/composite_table_availability.pseudocode
```diff
@@ -0,0 +1,123 @@
...
+
+    getAvailability(dsConstraints):
+        fieldNames = Union(dsConstraints.metrics, dsConstraints.dimensions.map(fieldMap.get(it.name))
+        return dataSource.getIntervals(fieldNames).reduce(joiner)
+
+    Description:
+        For a single datasource -> select intervals by requested field names -> join intervals
+
+PermissiveAvailability: extends ConcreteAvailabilityResolver
+    joiner = UnionJoiner  // Hippodrome uses this logic
```
cdeszaq: What's Hippodrome?

rfc/composite_table_availability.pseudocode
michael-mclawhorn: As a matter of practice, SimplifiedIntervalList should be used to coalesce intervals into a simpler form.  I'm not convinced it won't go away in favor of a more standard library, so I've been reluctant to put it into interfaces more than necessary.

SimplifiedIntervalList is actually a little costly to build, so I'd rather not force people to build them more often than necessary.  It is /advisable/ to simplify, but carrying simplified data around in a list should be reasonable.

rfc/composite_table_availability.pseudocode
michael-mclawhorn: The raw data from druid is in the form of a set of Segments where Segment contains an interval and a set of fields.  Segments should be atomic over their unit of time, at any given instant there is one 'winning' segment, and the schema it has is correct for that instant.  So Segments aren't a function on fields, only on time.  

I was imagining a crude way of doing this, going over every segment in values and filtering on time overlap.  An optimizing data structure could be used to cache this data.

michael-mclawhorn: One consideration, I suppose, is that this segment data model may be very specialized to druid.  So it's possible this should be a DruidDataSource, but it should be a gentle refactor if we ever have a different availability model to use.

rfc/composite_table_availability.pseudocode
michael-mclawhorn: Field names always point at the underlying query fact schema.

Metric field names are currently only physical.  If that changes, they'll need to be filtered through the table field name map.

Dimension field names are quasi-physical, needing to be filtered through the table field name map.

rfc/composite_table_availability.pseudocode
michael-mclawhorn: There are two use cases.
One is during query planning.
The other is during partial data evaluation during request processing.
The latter could happen AFTER the query has undergone transformations which alter it's field signature.

rfc/composite_table_availability.pseudocode
michael-mclawhorn: I can sub in 'DataSourceName' as a class, but to the best of my intuition, all the uses are as an identity and for connecting to serialized usages.

rfc/composite_table_availability.pseudocode
michael-mclawhorn: I'm guess I'm not sure, I'll clean this up.

rfc/composite_table_availability.pseudocode
cdeszaq: makes sense.

michael-mclawhorn: Talking to @garyluoex today another reason to shun SimplifiedIntervalList is that it consciously doesn't support all list operations.  Using primarily immutable lists may be the smarter choice.

ps - unmodifiableCollections don't observe the collections equality contract, instead use the Object equality contract.  fun fact.

rfc/composite_table_availability.pseudocode
```diff
@@ -0,0 +1,116 @@
+
+PhysicalTableSchema:
+	getFieldName(Column c)
+
+DataSource:
```
garyluoex: The class [DataSource](https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/datasource/DataSource.java) already exists to hold `PhysicalTables` in `DruidQuery`, we might have to find a different name?

rfc/composite_table_availability.pseudocode
```diff
@@ -0,0 +1,116 @@
+
+PhysicalTableSchema:
+	getFieldName(Column c)
```
garyluoex: The `FieldName` mentioned here refers to things with a `asName()` ? like `Columns`?

michael-mclawhorn: Per side-discussion, I believe that Fieldname per the existing class references a column in the fact store (in this case, in Druid)

rfc/composite_table_availability.pseudocode
michael-mclawhorn: I think DataSourceMetadata has all the state we need for this interface, so we should repurpose it for implementation, yes.

PR #168, opened 2017-02-14
Replacing use of UriBuilder to avoid potential template issues

will-lauer:
Replacing UriBulder with Apache Httpcomponents URIBuilder because JAX-RS
UriBuilder allows template parameter substitution behind the scenes that
causes some invalid URIs to fail to parse.


fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/QueryParameterNormalizationFilter.java
```diff
@@ -73,16 +74,18 @@ public void filter(ContainerRequestContext containerRequestContext) throws IOExc
...
         // Set normalized values
         getNormalizedQueryParameters(uriInfo)
-                .forEach(keyValue -> builder.queryParam(keyValue.getLeft(), keyValue.getRight()));
+                .forEach(keyValue -> builder.addParameter(keyValue.getLeft(), keyValue.getRight()));
 
-        return builder.build();
+        try {
+            return builder.build();
+        } catch (URISyntaxException e) {
+            throw new RuntimeException("failed to build uri", e);
```
cdeszaq: We should log the exception here (probably at the `error` level), otherwise there's a risk we won't log the exception, or at least no the way we do everywhere else. (I don't think errors in filters have a top-level catch wrapped around them the same way that other places do)

We may even want to go so far as to kill the request by responding with a 500 error here, rather than bubbling the exception and having Jersey deal with the response, since it gives us more control over the response payload / message, etc.

cdeszaq:
I know it's been a while since this PR was opened, but I have 2 additional thoughts:

1. Can this PR get a few tests that illustrate the sorts of naughty queries it's protecting against? Not only would that show the problem, but it would also show that it's fixed.
2. Apache Http Components is a _huge_ library, and duplicates what we already use AsyncHttpClient for (for the most part). I would _love_ to be able to find a solution with what we already have, or possibly using a much smaller dependency. If we can't, or it doesn't make sense, that's fine, but some due diligence on this would be amazing. (and again, tests would help validate that)

PR #170, opened 2017-02-15
dateTime specified as sortable field in sorting clause

dayamr:
This PR is first one among the 'dateTime' based sort feature.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/orderby/OrderByColumn.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/orderby/OrderByColumn.java
@@ -51,6 +51,17 @@ public class OrderByColumn {
     }
 
     /**
+     * Constructor.
+     *
+     * @param dateTime  a dateTime
+     * @param direction  sort direction
+     */
+    public OrderByColumn(String dateTime, SortDirection direction) {
+        this.dimension = dateTime;
+        this.direction = direction;
+    }
+
+    /**
      * Getter for sort column dimension.
      *
      * @return dimension A dimension name, a metric name, an aggregation name or a post aggregation name
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/orderby/OrderByColumn.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/orderby/OrderByColumn.java
@@ -51,13 +51,14 @@ public class OrderByColumn {
     }
 
     /**
-     * Constructor.
+     * Constructor which accepts generic column with direction. For example: dateTime column is not part of aggregation
+     * or posAggregation. But still allowed to sort the resultSet based on dateTime value.
      *
-     * @param dateTime  a dateTime
+     * @param column  a column needs to be associated with the direction
      * @param direction  sort direction
      */
-    public OrderByColumn(String dateTime, SortDirection direction) {
-        this.dimension = dateTime;
+    public OrderByColumn(String column, SortDirection direction) {
+        this.dimension = column;
         this.direction = direction;
     }
 
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
@@ -245,9 +245,13 @@ public class DataApiRequest extends ApiRequest {
         String sortApiValue = sorts;
         if (sorts != null && !sorts.isEmpty() && sorts.contains(DATE_TIME_STRING)) {
             //Requested sort on dateTime - It has to be the first field in string
-            this.dateTimeSort = generateDateTimeSortColumn(sorts.substring(0, sorts.indexOf(",")));
-            //Separating dateTimeSort from the complete api sort value string
-            sortApiValue = sorts.substring(sorts.indexOf(",") + 1, sorts.length());
+            if (sorts.contains(",")) {
+                this.dateTimeSort = generateDateTimeSortColumn(sorts.substring(0, sorts.indexOf(",")));
+                sortApiValue = sorts.substring(sorts.indexOf(",") + 1, sorts.length());
+            }
+            else {
+                sortApiValue = null;
+            }
         }
 
         // Requested sort on metrics - optional, can be empty Set
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
@@ -116,7 +116,7 @@ public class DataApiRequest extends ApiRequest {
 
     private final DruidFilterBuilder filterBuilder;
 
-    private Optional<OrderByColumn> dateTimeSort = Optional.ofNullable(null);
+    private Optional<OrderByColumn> dateTimeSort = Optional.empty();
 
     /**
      * Parses the API request URL and generates the Api Request object.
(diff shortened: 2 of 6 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/orderby/OrderByColumn.java
```diff
@@ -51,6 +51,17 @@ public OrderByColumn(PostAggregation postAggregation, SortDirection direction) {
     }
 
     /**
+     * Constructor.
+     *
+     * @param dateTime  a dateTime
```
cdeszaq: Since this constructor is quite different from the rest, and it takes very generic parameters, it should get more description as to why it exists and what it's intended to be used for. Additionally, the comment for this parameter is not great... if I give it, for example, `2017-02-01T10:25`, something tells me it wouldn't do the right thing.

If we could have a way to create this that _didn't_ require the user to specify the column name, that would be great. A static factory method that takes a direction, for example.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -238,6 +242,18 @@ public DataApiRequest(
 
         this.having = DruidHavingBuilder.buildHavings(this.havings);
 
+
+        if (
+                sorts != null &&
+                        !sorts.isEmpty() &&
+                        sorts.toLowerCase(Locale.ENGLISH).contains(DATE_TIME_STRING.toLowerCase(Locale.ENGLISH))
```
cdeszaq: Can we simplify this conditional, or isolate it's complexity into a helper method?

cdeszaq: Also, why are we lower-casing this keyword? Do we do that anywhere else in the API for keywords?

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -114,6 +116,8 @@
 
     private final DruidFilterBuilder filterBuilder;
 
+    private Optional<OrderByColumn> dateTimeSort = Optional.ofNullable(null);
```
cdeszaq: Rather than defaulting to a `null` `Optional`, we should use `Optional.emtpy()`

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
cdeszaq: Also also, why use `contains`, why not use `startsWith`?

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -1098,6 +1115,35 @@ protected DateTime getCurrentDate(DateTime dateTime, TimeGrain timeGrain) {
...
+                        SortDirection.valueOf(dateTimeWithDirection.get(1).toUpperCase(Locale.ENGLISH)) :
+                        SortDirection.DESC;
+            } catch (IllegalArgumentException ignored) {
+                String sortDirectionName = dateTimeWithDirection.get(1);
+                LOG.debug(SORT_DIRECTION_INVALID.logFormat(sortDirectionName));
+                throw new BadApiRequestException(SORT_DIRECTION_INVALID.format(sortDirectionName));
+            }
+            return Optional.of(new OrderByColumn(dateTimeWithDirection.get(0), sortDirection));
+        }
+        else {
```
cdeszaq: Should be up a line

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -1098,6 +1115,35 @@ protected DateTime getCurrentDate(DateTime dateTime, TimeGrain timeGrain) {
...
+        if (DATE_TIME_STRING.equals(dateTimeWithDirection.get(0))) {
+            SortDirection sortDirection;
+            try {
+                sortDirection = dateTimeWithDirection.size() == 2 ?
+                        SortDirection.valueOf(dateTimeWithDirection.get(1).toUpperCase(Locale.ENGLISH)) :
+                        SortDirection.DESC;
+            } catch (IllegalArgumentException ignored) {
+                String sortDirectionName = dateTimeWithDirection.get(1);
+                LOG.debug(SORT_DIRECTION_INVALID.logFormat(sortDirectionName));
+                throw new BadApiRequestException(SORT_DIRECTION_INVALID.format(sortDirectionName));
```
cdeszaq: Is this code duplicated with the existing code that deals with sort ordering?

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -238,6 +242,18 @@ public DataApiRequest(
...
+
+        if (
+                sorts != null &&
+                        !sorts.isEmpty() &&
+                        sorts.toLowerCase(Locale.ENGLISH).contains(DATE_TIME_STRING.toLowerCase(Locale.ENGLISH))
+                ) {
+            //Requested sort on dateTime - It has to be first field in the string
+            this.dateTimeSort = generateDateTimeSortColumn(sorts.substring(0, sorts.indexOf(",")));
+            //Separating dateTimeSort from the complete api sort value string
+            sorts = sorts.substring(sorts.indexOf(",") + 1, sorts.length());
```
cdeszaq: If `sorts` is a method parameter, it's generally a bad idea to modify method parameters. They should be treated as immutable.

fili-core/src/main/java/com/yahoo/bard/webservice/web/ErrorMessageFormat.java
```diff
@@ -73,6 +73,8 @@
     SORT_METRICS_NOT_SORTABLE_FORMAT("Sorting not possible on metric(s) '%s'."),
     SORT_METRICS_UNDEFINED("Metric(s) in sort expression '%s' do not exist."),
 
+    DATE_TIME_SORT_VALUE_INVALID("dateTime must always be the first field in the sort list"),
```
cdeszaq: Technically, it doesn't _always_ _need_ to be the first field in the list. It only needs to be the 1st one when it's being sorted on...

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/ErrorDataServletSpec.groovy
```diff
@@ -444,6 +445,43 @@ class ErrorDataServletSpec extends Specification {
...
+        r.getStatus() == Response.Status.BAD_REQUEST.getStatusCode()
+        GroovyTestUtils.compareErrorPayload(r.readEntity(String.class), jsonFailure)
+    }
+
+    def "Successful execution if dateTime is first field in sort list"() {
+        when:
+        Response r = jtb.getHarness().target("data/shapes/day/color")
+                .queryParam("metrics","height")
+                .queryParam("dateTime","2014-09-01%2F2014-09-10")
+                .queryParam("sort","dateTime|DESC,height|ASC")
```
cdeszaq: I feel like this "success" test doesn't belong in the `ErrorDataServletSpec`, I feel like it belongs in a Spec that's focused on sorting. If there isn't already a place that's testing that, we should make one 😉

cdeszaq: Also, we should add a test for sorting by `dateTime` that _doesn't_ specify the direction, and uses the default.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
dayamr: if we use contains, we can educate user with proper error message. If we use starts with, error message would be 'dateTime is not metric......'. It may restrict the user to think sort is allowed only for metrics.

fili-core/src/main/java/com/yahoo/bard/webservice/web/ErrorMessageFormat.java
dayamr: I thought we have decided to have dateTime as first value for the sort param if we want to sort based on dateTime.

cdeszaq: My point is that if it's not sorted on at all, or there are no sorts, it doesn't need to be there.

dayamr: This error message is displayed only when `dateTime` is part of the sort list but not the first one. If there is no sort list or if there is no request for `dateTime` sort, we will not reach till this error.

cdeszaq: Right, and that's my point... the error message indicates (emphasis added):

> dateTime must **always** be the first field

By the error message indicating _always_, it may be confusing. A more accurate error message would remove that confusion.

cdeszaq:
Does #178 supercede this PR? Can this one be closed, or is it different from #178?

dayamr:
PR https://github.com/yahoo/fili/pull/178 considers all these changes. Hence closing this PR.

PR #176, opened 2017-02-17
Support underscores in request ids.

archolewa:
--Some clients want to have underscores in their unique request ids.
Since it's kind of arbitrary what characters we allow or don't allow in
the request header, we've added it.

### DIFF
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/BardLoggingFilterSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/BardLoggingFilterSpec.groovy
@@ -1,0 +1,93 @@
+// Copyright 2017 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.web.filters
+
+import com.yahoo.bard.webservice.logging.RequestLog
+
+import spock.lang.Specification
+import org.slf4j.MDC
+
+import spock.lang.Subject
+import spock.lang.Unroll
+
+import javax.ws.rs.container.ContainerRequestContext
+import javax.ws.rs.core.MultivaluedHashMap
+import javax.ws.rs.core.UriInfo
+
+class BardLoggingFilterSpec extends Specification {
+
+    @Subject BardLoggingFilter filter = new BardLoggingFilter()
+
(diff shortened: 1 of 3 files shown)


michael-mclawhorn:
I agree with Rick's comment.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/BardLoggingFilterSpec.groovy
```diff
@@ -0,0 +1,94 @@
...
+        filter.filter(requestContext)
+
+        then:
+        notThrown(NullPointerException)
+    }
+
+    @Unroll
+    def "The BardLoggingFilter does not add a malformed requestId '#requestId' to MDC"() {
+        given: "A ContainerRequestContext with the specified malformed requestId"
+        MultivaluedHashMap<String, String> fakeHeaders = new MultivaluedHashMap<>()
```
cdeszaq: This shadows the class-level `fakeHeaders` map. Do we need this declaration in this test?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/BardLoggingFilterSpec.groovy
```diff
@@ -0,0 +1,94 @@
...
+        fakeHeaders.put(BardLoggingFilter.X_REQUEST_ID_HEADER, [requestId])
+
+        when:
+        filter.filter(requestContext)
+
+        then:
+        !MDC.get(RequestLog.ID_KEY) || !MDC.get(RequestLog.ID_KEY).startsWith(requestId)
```
cdeszaq: You _could_ simplify this by using the null-safe coalescence operator:

```groovy
!MDC.get(RequestLog.ID_KEY)?.startsWith(requestId)
```

If `MDC.get(RequestLog.ID_KEY)` comes back `null`, it returns false and doesn't call `.startsWith(...)`, preventing the NPE.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/filters/BardLoggingFilterSpec.groovy
archolewa: Could catch. It should be using the class level `fakeHeaders`

PR #177, opened 2017-02-17
Create a default base class for all dimension types

asifmansoora:
Create a common base class for Dimension Config , Lookup Dimension Config , Registered Lookup Dimension Config and DimensionName Interface
Made the changes to Wikipedia example to include the new Default Dimension Config class


fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/DefaultDimensionConfig.java
```diff
@@ -0,0 +1,134 @@
+// Copyright 2016 Yahoo Inc.
```
cdeszaq: 2017. (May apply to other new files, I won't comment in those places)

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/DefaultDimensionConfig.java
```diff
@@ -0,0 +1,134 @@
...
+
+  public String getCategory() {
+    return category;
+  }
+
+  public String getPhysicalName() {
+    return physicalName;
+  }
+
+  public String getDescription() {
```
cdeszaq: All of these methods, if they override a method declared in the interface, should have an `@Override` annotation.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/DefaultDimensionConfig.java
```diff
@@ -0,0 +1,134 @@
...
+import com.yahoo.bard.webservice.data.dimension.DimensionField;
+import com.yahoo.bard.webservice.data.dimension.KeyValueStore;
+import com.yahoo.bard.webservice.data.dimension.SearchProvider;
+
+import java.util.LinkedHashSet;
+import javax.validation.constraints.NotNull;
+
+
+/**
+ * A Dimension instance holds all of the information needed to construct a Dimension.
```
cdeszaq: This thing is not a dimension instance, since it's not a `Dimension`, right?

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/DefaultDimensionConfig.java
```diff
@@ -0,0 +1,134 @@
...
+     * @param physicalName  The internal, physical name for the dimension.
+     * @param description  A description of the dimension and its meaning.
+     * @param longName  The Long Name is the external, end-user-facing long  name for the dimension.
+     * @param category  The Category is the external, end-user-facing category for the dimension.
+     * @param fields  The set of fields for this dimension.
+     * @param defaultDimensionFields  The default set of fields for this dimension to be shown in the response.
+     * @param keyValueStore  The key value store holding dimension row data.
+     * @param searchProvider  The search provider for field value lookups on this dimension.
+     */
+    public DefaultDimensionConfig(
```
cdeszaq: Since this includes a `KeyValueStore` and a `SearchProvider`, it's actually not the config for a _general_ dimension, but it's the config for a `KeyValueStoreDimension` (which happens to be Fili's default).

So, perhaps a better name is `DefaultKeyValueStoreDimensionConfig`?

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/LookupDefaultDimensionConfig.java
```diff
@@ -0,0 +1,102 @@
...
+   * @param apiName  The API Name is the external, end-user-facing name for the dimension.
+   * @param physicalName  The internal, physical name for the dimension.
+   * @param description  A description of the dimension and its meaning.
+   * @param longName  The Long Name is the external, end-user-facing long  name for the dimension.
+   * @param category  The Category is the external, end-user-facing category for the dimension.
+   * @param defaultDimensionFields  The default set of fields for this dimension to be shown in the response.
+   * @param keyValueStore  The key value store holding dimension row data.
+   * @param searchProvider  The search provider for field value lookups on this dimension.
+   * @param namespaces  A list of namespaces used to configure the Lookup dimension.
+   */
```
cdeszaq: These javadocs need to call out the differences between these two constructors. It's not obvious what the differences are at a quick glance.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/RegisteredLookupDefaultDimensionConfig.java
```diff
@@ -0,0 +1,105 @@
...
+   *
+   * @param apiName  The API Name is the external, end-user-facing name for the dimension.
+   * @param physicalName  The internal, physical name for the dimension.
+   * @param description  A description of the dimension and its meaning.
+   * @param longName  The Long Name is the external, end-user-facing long  name for the dimension.
+   * @param category  The Category is the external, end-user-facing category for the dimension.
+   * @param defaultDimensionFields  The default set of fields for this dimension to be shown in the response.
+   * @param keyValueStore  The key value store holding dimension row data.
+   * @param searchProvider  The search provider for field value lookups on this dimension.
+   * @param lookups  A list of lookups used to configure the Lookup dimension.
```
cdeszaq: Again, need to document the differences in these constructors

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/names/DimensionName.java
```diff
@@ -0,0 +1,16 @@
+// Copyright 2017 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.data.config.names;
+
+/**
+ * Defines the name of a Dimension.
```
cdeszaq: _what_ name? Physical? Logical? etc.

fili-wikipedia-example/src/main/java/com/yahoo/wiki/webservice/data/config/names/WikiApiDimensionName.java
```diff
@@ -38,4 +49,16 @@
     public String asName() {
         return camelName;
     }
+
+    public String getDescription() {
```
cdeszaq: If this thing is a `DimensionName` object, why are we stuffing all of these other parameters in there as well? They don't seem to have anything to do with names.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/DefaultDimensionConfig.java
```diff
@@ -0,0 +1,134 @@
...
+        @NotNull DimensionName apiName,
+        @NotNull String physicalName,
+        @NotNull String description,
+        @NotNull String longName,
+        @NotNull String category,
```
cdeszaq: Not all of these are required I don't think. In particular, `description`, `physicalName`, `longName`, and `category` are optional and have defaults provided when the actual dimensions are made I think.

I'm not necessarily opposed to having the default config objects be stricter or better encouraging of best practices, but I just wanted to point these differences out.

fili-wikipedia-example/src/main/java/com/yahoo/wiki/webservice/data/config/names/WikiApiDimensionName.java
```diff
@@ -2,32 +2,43 @@
...
+    USER("Description of User", "wikipedia user", "General"),
+    UNPATROLLED("Description of Unpatrolled", "wikipedia unpatrolled", "General"),
+    NEW_PAGE("Description of New page", "wikipedia newPage", "General"),
+    ROBOT("Description of Robot", "wikipedia robot", "General"),
+    ANONYMOUS("Description of Anonymous", "wikipedia anonymous", "General"),
+    NAMESPACE("Description of Namespace", "wikipedia namespace", "General"),
+    CONTINENT("Description of Continent", "wikipedia continent", "General"),
+    COUNTRY("Description of Country", "wikipedia country", "General"),
+    REGION("Description of Region", "wikipedia region", "General"),
+    CITY("Description of City", "wikipedia city", "General");
```
cdeszaq: Since this is an example for others to look at for how to set up a good Fili instance, it would be great to show something more than default-style placeholder text for these descriptions, long names, etc.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/DefaultDimensionConfig.java
```diff
@@ -0,0 +1,134 @@
...
+import com.yahoo.bard.webservice.data.dimension.SearchProvider;
+
+import java.util.LinkedHashSet;
+import javax.validation.constraints.NotNull;
+
+
+/**
+ * A Dimension instance holds all of the information needed to construct a Dimension.
+ */
+public class DefaultDimensionConfig implements DimensionConfig {
```
garyluoex: I would prefer this to be an `abstract` class with the name `BaseDimensionConfig` since we have a `BasePhysicalTable`

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/LookupDefaultDimensionConfig.java
```diff
@@ -0,0 +1,102 @@
...
+import com.yahoo.bard.webservice.data.dimension.SearchProvider;
+
+import java.util.LinkedHashSet;
+import java.util.List;
+import javax.validation.constraints.NotNull;
+
+/**
+ * A Lookup Dimension instance holds all of the information needed to construct a Lookup Dimension.
+ */
+public class LookupDefaultDimensionConfig extends DefaultDimensionConfig implements LookupDimensionConfig {
```
garyluoex: I do not see the need to have an default implementation for every implementation of `DimensionConfig` since most likely users needs to extend or implement something in fili application anyways. So simply extending `BaseDimensionConfig` would do the job I think. Convince me otherwise.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/DefaultDimensionConfig.java
cdeszaq: I think the point of this change is so that an app-specific class isn't _needed_, so making this abstract defeats that purpose. This work originated from some folks standing up a new Fili instance that had also done a previous one, and noticing that they had to implemented the exact same class for each instance in order to implement the `DimensionConfig` interface.

Can you call out a specific reason for wanting this abstract? What additions would you expect a concrete version of that abstract one to add?

garyluoex: I thought the purpose is to get rid of redundant methods between different types of `DimensionConfig`. This PR is originated from Will's comment [here](https://git.corp.yahoo.com/Digits/Hippodrome/pull/701) saying 
`SHouldn't this be a common bard class rather than reimplementing it here and in UAD?`.

garyluoex: Also, how do we know what Dimension to build if a user passes a `DefaultDimensionConfig` around?

garyluoex: Remember, we have the `TypeAwareDimensionLoader` that does instances of in order to build `Dimensions` from `DimensionConfig`.

garyluoex: Ok, I misunderstood what this class is. The name should be changed as rick suggested. Also, I think it is a good idea to have a `BaseDimensionConfig` abstract class so that we can implement `DruidBackedDimensionConfig` and so on in the future.

garyluoex:
One more thing to take into consideration: I am guessing the reason that there are no implementation of `DimensionConfig` inside Fili is because that originally we want user to directly configure `DimensionConfig` instead of configuring `Enums` and then passing those information into `DimensionConfig`. The reason why we think implementation of `DimensionConfig` should be in Fili is mostly because the way how UAD2 and Digits configures dimensions, which is constructing `Enums` first then passing into `DimensionConfig`, but there is not reason why, and even more reasonable for users to directly configure (i.e. implement) `DimensionConfig` directly. (in which I plan to try to change Bank into if no one disagrees)

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/DefaultDimensionConfig.java
asifmansoora: Sure. Changing it in all the new files

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/DefaultDimensionConfig.java
asifmansoora: Yes, you are right, changing the description of `Default..`classes

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/DefaultDimensionConfig.java
asifmansoora: Sure.

fili-wikipedia-example/src/main/java/com/yahoo/wiki/webservice/data/config/dimension/WikiDimensions.java
```diff
@@ -74,9 +78,10 @@ public WikiDimensions() {
      *
      * @return set of dimension configurations
      */
-    public LinkedHashSet<DimensionConfig> getDimensionConfigurationsByApiName(WikiApiDimensionName... dimensionNames) {
+    public LinkedHashSet<DimensionConfig> getDimensionConfigurationsByApiName(
+        WikiApiDimensionConfigInfo... dimensionNames) {
```
cdeszaq: `) {` should be on the next line down, aligned with the start of the method declaration line

fili-wikipedia-example/src/main/java/com/yahoo/wiki/webservice/data/config/names/WikiApiDimensionConfigInfo.java
```diff
@@ -0,0 +1,64 @@
+// Copyright 2016 Yahoo Inc.
```
cdeszaq: 2017

fili-wikipedia-example/src/main/java/com/yahoo/wiki/webservice/data/config/names/WikiApiDimensionConfigInfo.java
```diff
@@ -0,0 +1,64 @@
...
+    USER("User is a person who generally use or own wiki services", "wiki user", "General"),
+    UNPATROLLED("Unpatrolled are class of pages that are not been patrolled", "wiki unpatrolled", "General"),
+    NEW_PAGE("New Page is the first page that is created in wiki ", "wiki newPage", "General"),
+    ROBOT("Robot is an tool that carries out repetitive and mundane tasks", "wiki robot", "General"),
+    ANONYMOUS("Anonymous are individual or entity whose identity is unknown", "wiki anonymous", "General"),
+    NAMESPACE("Namespace is a set of wiki pages that begins with a reserved word", "wiki namespace", "General"),
+    CONTINENT("Continent refers to one of the seven region", "wiki continent", "General"),
+    COUNTRY("Country is identified as a distinct national entity in political geography", "wiki country", "General"),
+    REGION("Regions are areas that are broadly divided by several distinct characteristics ", "wiki region", "General"),
+    CITY("City is a large and permanent human settlement", "wiki city", "General");
```
cdeszaq: No need to change them, since they are accurate, but generally descriptions of dimensions are likely to connect to the domain of the facts for which they are dimensions (usually). For example, while a city _is_ where humans live, in this context it is the city in which the user making the edit was detected to be located. (ie. it's where the user was when they made the edit).

fili-wikipedia-example/src/main/java/com/yahoo/wiki/webservice/data/config/names/WikiApiDimensionConfigInfo.java
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * View the dimension name.
+     *
+     * @return The camelCase version of the dimension.
+     */
+    public String asName() {
+        return camelName;
+    }
+
+    public String getDescription() {
```
cdeszaq: Should all have `@Override` annotations

asifmansoora: Interface `DimensionName` has only `asName()` method to be implemented

asifmansoora:
Addressed all the comments

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/DefaultKeyValueStoreDimensionConfig.java
```diff
@@ -0,0 +1,145 @@
...
+   * @param physicalName  The internal, physical name for the dimension.
+   * @param description  A description of the dimension and its meaning.
+   * @param longName  The Long Name is the external, end-user-facing long  name for the dimension.
+   * @param category  The Category is the external, end-user-facing category for the dimension.
+   * @param defaultDimensionFields  The default set of fields for this dimension to be shown in the response.
+   * @param keyValueStore  The key value store holding dimension row data.
+   * @param searchProvider  The search provider for field value lookups on this dimension.
+   */
+  public DefaultKeyValueStoreDimensionConfig(
+      DimensionName apiName,
```
garyluoex: Does it make sense that the other constructor requires some of these arguments to be `@NotNull` but this constructor does not require? And we should just put `@NotNull` for all the arguments since thats easier. Apply to other files too.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/DefaultRegisteredLookupDimensionConfig.java
```diff
@@ -0,0 +1,106 @@
...
+import com.yahoo.bard.webservice.data.config.names.DimensionName;
+import com.yahoo.bard.webservice.data.dimension.DimensionField;
+import com.yahoo.bard.webservice.data.dimension.KeyValueStore;
+import com.yahoo.bard.webservice.data.dimension.SearchProvider;
+
+import javax.validation.constraints.NotNull;
+import java.util.LinkedHashSet;
+import java.util.List;
+
+
```
garyluoex: Extra line.

fili-core/src/test/java/com/yahoo/bard/webservice/data/config/names/TestDimensionName.java
```diff
@@ -0,0 +1,20 @@
...
+import com.yahoo.bard.webservice.util.EnumUtils;
+
+/**
+ * Bard test dimensions.
+ */
+public enum TestDimensionName implements DimensionName {
+    NAME,
+    AGE,
+    ETHNICITY,
+    GENDER;
```
garyluoex: Move the semicolon to the next line so that when we add things to the enum list in the future, the last item won't be shown as deleted and then re-added due to moving semicolon to the next item. Apply to other files also.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/DefaultKeyValueStoreDimensionConfig.java
```diff
@@ -0,0 +1,145 @@
...
+      String physicalName,
+      String description,
+      String longName,
+      String category,
+      LinkedHashSet<DimensionField> defaultDimensionFields,
```
garyluoex: I think this should be `fields` instead of `defaultDimensionFields`, default one is the one can be inferred from fields, not the other way around. Apply to other files too.

fili-wikipedia-example/src/main/java/com/yahoo/wiki/webservice/data/config/dimension/WikiDimensions.java
```diff
@@ -74,9 +78,11 @@ public WikiDimensions() {
      *
      * @return set of dimension configurations
      */
-    public LinkedHashSet<DimensionConfig> getDimensionConfigurationsByApiName(WikiApiDimensionName... dimensionNames) {
+    public LinkedHashSet<DimensionConfig> getDimensionConfigurationsByApiName(
```
garyluoex: Should change the name of the method since it does not take `ApiName` now.

fili-wikipedia-example/src/main/java/com/yahoo/wiki/webservice/data/config/names/WikiApiDimensionConfigInfo.java
```diff
@@ -0,0 +1,65 @@
...
+    private final String category;
+
+    /**
+     * Constructor of Wikipedia Dimension Name.
+     *
+     * @param description  A description of the dimension and its meaning.
+     * @param longName  The Long Name is the external, end-user-facing long  name for the dimension.
+     * @param category  The Category is the external, end-user-facing category for the dimension.
+     */
+    WikiApiDimensionConfigInfo(String description, String longName, String category) {
```
garyluoex: Consider get rid of the `category` property from the enum and simply return in the method using default category.

```java
    @Override
    public String getCategory() {
        return Dimension.DEFAULT_CATEGORY;
    }
```

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/DefaultKeyValueStoreDimensionConfig.java
cdeszaq: Re: Making all parameters not null: Not all parameters are required, so no, we should not make them all `@NotNull`

Re: Making the constructors have the same "nullability" for corresponding parameters: I think that's probably a good idea, yes.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/DefaultKeyValueStoreDimensionConfig.java
cdeszaq: This is a good idea. And the parameter comment in the JavaDoc should be updated to reflect that this set of fields will _also_ be used for the default fields (ie. all fields on the Dimension will be the default)

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/DefaultKeyValueStoreDimensionConfig.java
garyluoex: So we should take a look at its `super` class to determine what arguments needs to be `@NotNull`.

fili-wikipedia-example/src/main/java/com/yahoo/wiki/webservice/data/config/names/WikiApiDimensionConfigInfo.java
garyluoex: Can we make this change? @asifmansoora

asifmansoora: addressed it

PR #178, opened 2017-02-21
 Added dateTime based sort feature

dayamr:
When dateTime is present in the sort field, irrespective of the query type, result should be sorted based on dateTime. To achieve this, 

- Added `DateTimeSortMapper` to reverse the time buckets and `DateTimeSortRequestHandler` to inject to the workflow.
- Added groupBy query type results to test the sorting.

This PR created on top of https://github.com/yahoo/fili/pull/170/


fili-core/src/test/groovy/com/yahoo/bard/webservice/web/DataApiRequestSortSpec.groovy
```diff
@@ -0,0 +1,80 @@
...
+                .queryParam("dateTime","2014-09-01%2F2014-09-10")
+                .queryParam("sort","dateTime|DESC,height|ASC")
+                .request().get()
+
+        then:
+        r.getStatus() == 200
+    }
+
+    def "Successful execution if dateTime is only field in the sort list"() {
+        when:
```
dayamr: This test case is added separately with generateDateTimeSortColumn( method.

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/mappers/DateTimeSortMapper.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        List<DateTime> dateTimeList = new ArrayList(bucketizedResultsMap.keySet());
+
+        Collections.sort(dateTimeList, new DateTimeComparator());
+
+        if (direction.equals(SortDirection.DESC)) {
+            Collections.sort(dateTimeList, Collections.reverseOrder());
+        }
+
```
michael-mclawhorn: This doesn't make much sense.  You're doing a comparison based sort.  Then doing another one.

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/mappers/DateTimeSortMapper.java
```diff
@@ -0,0 +1,79 @@
...
+     *
+     * @param direction Sort direction
+     */
+    public DateTimeSortMapper(SortDirection direction) {
+        this.direction = direction;
+    }
+
+    /**
+     *  Reverses the resultSet if the requested order is DESC.
+     *
```
michael-mclawhorn: This javadoc is now out of date.
Should reflect that it will sort in ascending or descending order.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -296,6 +304,56 @@ public DataApiRequest(
...
+                );
+            } else {
+                LOG.debug(DATE_TIME_SORT_VALUE_INVALID.logFormat());
+                throw new BadApiRequestException(DATE_TIME_SORT_VALUE_INVALID.format());
+            }
+        } else {
+            return Optional.empty();
+        }
+
+    }
```
michael-mclawhorn: extra whitespace

CHANGELOG.md
```diff
@@ -10,6 +10,15 @@ Current
 
 ### Added:
 
+
+- [dateTime based sort feature for the final ResultSet added](https://github.com/yahoo/fili/pull/178)
+   * Now we can sort the final result set in DESC order. Druid provided results are in ASC order by default.
+   * Added `DateTimeSortMapper` to reverse the time buckets and `DateTimeSortRequestHandler` to inject to the workflow
```
michael-mclawhorn: Might want to generalize this to say that we not support sorting time in ascending or descending order.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -238,8 +242,12 @@ public DataApiRequest(
 
         this.having = DruidHavingBuilder.buildHavings(this.havings);
 
+        this.dateTimeSort = generateDateTimeSortColumn(sorts);
+
+        String sortApiValue = generateApiSortValue(sorts);
+
```
michael-mclawhorn: I wouldn't call this 'generate...'  Maybe 'filterTimeSort' or 'trimTimeSort'.

Then I'd consider putting this inline in the call to generateSortColumns(...) and save a local variable

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/mappers/DateTimeComparator.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.Comparator;
+
+/**
+ * Comparator to compare two dateTime.
+ */
+public class DateTimeComparator implements Comparator<DateTime> {
+
+    private SortDirection directin;
+
```
michael-mclawhorn: misspelled

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/mappers/DateTimeSortMapper.java
```diff
@@ -0,0 +1,74 @@
...
+
+        Map<DateTime, List<Result>> bucketizedResultsMap = new LinkedHashMap<>();
+
+        for (Result result: resultSet) {
+            bucketizedResultsMap.computeIfAbsent(result.getTimeStamp(), ignored -> new ArrayList<>()).add(result);
+        }
+
+        List<DateTime> dateTimeList = new ArrayList(bucketizedResultsMap.keySet());
+        Collections.sort(dateTimeList, new DateTimeComparator(direction));
+
```
michael-mclawhorn: Collections.sort(list, a == SortDirection.ASC ? Comparator.naturalOrder() : Comparator.reverseOrder());

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/mappers/DateTimeComparator.java
```diff
@@ -0,0 +1,39 @@
+// Copyright 2016 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.data.metric.mappers;
+
```
michael-mclawhorn: get rid of this class if the code change below works

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -296,6 +303,55 @@ public DataApiRequest(
     }
 
     /**
+     * Update sort field string by truncating the dateTime if it is present.
+     *
+     * @param sorts  String of sort columns
+     *
+     * @return Updated sort column if dateTime exists. Otherwise  return as it is received.
```
michael-mclawhorn: Otherwise followed by two spaces.

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/mappers/DateTimeComparator.java
dayamr: Removed it

michael-mclawhorn:
👍

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -296,6 +303,55 @@ public DataApiRequest(
...
+     * @param sorts  String of sort columns
+     *
+     * @return Updated sort column if dateTime exists. Otherwise return as it is received.
+     */
+    protected String truncateTimeSort(String sorts) {
+
+        if (sorts != null && !sorts.isEmpty() && sorts.contains(DATE_TIME_STRING)) {
+            if (sorts.contains(",")) {
+                //Truncate the dateTime sort column from the api sort string
+                return sorts.substring(sorts.indexOf(",") + 1, sorts.length());
```
will-lauer: This isn't safe. It doesn't work correctly if a dimension exists that starts with dateTime but isn't exactly dateTime. For example, if I define a dimension called `dateTimeInOtherTz` and included it at the _end_ of my sorts string, this code would incorrectly truncate the first dimension from the sorts clause, even if it wasn't dateTime.

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/mappers/DateTimeSortMapper.java
```diff
@@ -0,0 +1,78 @@
...
+        Map<DateTime, List<Result>> bucketizedResultsMap = new LinkedHashMap<>();
+
+        for (Result result: resultSet) {
+            bucketizedResultsMap.computeIfAbsent(result.getTimeStamp(), ignored -> new ArrayList<>()).add(result);
+        }
+
+        List<DateTime> dateTimeList = new ArrayList(bucketizedResultsMap.keySet());
+
+        Collections.sort(dateTimeList, direction == SortDirection.ASC ?
```
cdeszaq: Rather than sorting the keyset, and then getting those out, can you just dump the LHM into a `TreeMap`, and then stream over the tree map's entry set? I think it might be simpler, and should be just as performant I think?

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/mappers/DateTimeSortMapper.java
```diff
@@ -0,0 +1,78 @@
...
+                Comparator.<DateTime>naturalOrder() :
+                Comparator.<DateTime>reverseOrder());
+
+        return new ResultSet(
+                dateTimeList.stream()
+                        .map(e -> bucketizedResultsMap.get(e))
```
cdeszaq: Replace with method reference

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/mappers/DateTimeSortMapper.java
```diff
@@ -0,0 +1,78 @@
...
+                        .flatMap(List::stream).collect(Collectors.toList()),
```
cdeszaq: Pop `collect` down a line

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/orderby/OrderByColumn.java
```diff
@@ -51,6 +51,18 @@ public OrderByColumn(PostAggregation postAggregation, SortDirection direction) {
     }
 
     /**
+     * Constructor which accepts generic column with direction. For example: dateTime column is not part of aggregation
+     * or posAggregation. But still allowed to sort the resultSet based on dateTime value.
```
cdeszaq: s/posAggregation/postAggregation

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -114,6 +116,8 @@
 
     private final DruidFilterBuilder filterBuilder;
 
+    private Optional<OrderByColumn> dateTimeSort = Optional.empty();
```
cdeszaq: Why isn't this `final`? Can it be?

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -296,6 +303,55 @@ public DataApiRequest(
...
+     *
+     * @param sorts  String of sort columns
+     *
+     * @return instance of OrderByColumn for dateTime
+     */
+    protected Optional<OrderByColumn> generateDateTimeSortColumn(String sorts) {
+
+        if (sorts != null && !sorts.isEmpty() && sorts.contains(DATE_TIME_STRING)) {
+            //Requested sort on dateTime - It has to be the first field in string
+            String dateTime = sorts.contains(",") ? sorts.substring(0, sorts.indexOf(",")) : sorts;
```
cdeszaq: This sure looks like duplicate work that the `truncateTimeSort` method is doing. Can we _not_ do the work twice?

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -296,6 +303,55 @@ public DataApiRequest(
...
+            List<String> dateTimeWithDirection = Arrays.asList(dateTime.split("\\|"));
```
cdeszaq: This here seems like work that we're duplicating with the "normal" sort processing. Again, can we not duplicate the work?

One idea would be to generate the sort _objects_ (as a list) for the entire clause (so this method will need to handle dateTime not being a column but just a String), and then process that list of sort objects to pull off the dateTime sort from the front, and otherwise validate the sorts.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -1098,6 +1155,28 @@ protected DateTime getCurrentDate(DateTime dateTime, TimeGrain timeGrain) {
...
+     */
+    protected SortDirection getSortDirection(List<String> columnWithDirection) {
+
+        SortDirection sortDirection;
+        try {
+            sortDirection = columnWithDirection.size() == 2 ?
+                    SortDirection.valueOf(columnWithDirection.get(1).toUpperCase(Locale.ENGLISH)) :
+                    SortDirection.DESC;
+        } catch (IllegalArgumentException ignored) {
+            String sortDirectionName = columnWithDirection.get(1);
```
cdeszaq: Just pull this string out once, rather than twice if there's an error.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/DateTimeSortRequestHandler.java
```diff
@@ -0,0 +1,42 @@
+// Copyright 2016 Yahoo Inc.
```
cdeszaq: 2017

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/DateTimeSortRequestHandler.java
```diff
@@ -0,0 +1,42 @@
...
+import com.yahoo.bard.webservice.web.responseprocessors.ResponseProcessor;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * Injects the result set mapper for dateTime based sort if an API request asks for result in descending order.
+ * The time sort mapper should be second last mappers to execute on the result set.
+ */
+public class DateTimeSortRequestHandler implements DataRequestHandler {
+    private final @NotNull DataRequestHandler next;
```
cdeszaq: Move the `@NotNull` to the constructor so it's more visible to callers.

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/mappers/DateTimeSortMapper.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     *  Sorting the resultSet based on dateTime column sort direction.
+     *
+     * @param resultSet  The result set need to be sorted in ascending or descending order
+     *
+     * @return sorted ResultSet
+     */
+    @Override
+    public ResultSet map(ResultSet resultSet) {
```
cdeszaq: Can we wrap a RequestLog timer around this, so we can get a sense of how long sorting is taking?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/DataApiRequestSortSpec.groovy
```diff
@@ -0,0 +1,97 @@
+// Copyright 2016 Yahoo Inc.
```
cdeszaq: 2017

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/DataApiRequestSortSpec.groovy
```diff
@@ -0,0 +1,97 @@
...
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.web
+
+import com.yahoo.bard.webservice.application.JerseyTestBinder
+import com.yahoo.bard.webservice.druid.model.orderby.OrderByColumn
+import com.yahoo.bard.webservice.druid.model.orderby.SortDirection
+import com.yahoo.bard.webservice.web.endpoints.DataServlet
+import spock.lang.Specification
```
cdeszaq: Blank line

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/DataApiRequestSortSpec.groovy
```diff
@@ -0,0 +1,97 @@
...
+        Optional<OrderByColumn> dateTimeSort = new DataApiRequest().generateDateTimeSortColumn("dateTime|asc")
+
+        then:
+        dateTimeSort.get().direction == SortDirection.ASC
+        dateTimeSort.get().dimension == "dateTime"
+    }
+
+    def "If dateTime is not the first value in the string, then throw an error"() {
+        setup:
+        // mis spelled dateTime as dateTiem.
```
cdeszaq: This comment doesn't seem to be valid for this test?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/DataApiRequestSortSpec.groovy
```diff
@@ -0,0 +1,97 @@
...
+    def "Generate dateTime sort column when it is requested with other columns"() {
+        when:
+        Optional<OrderByColumn> dateTimeSort = new DataApiRequest().generateDateTimeSortColumn("dateTime|desc,height|ASC")
+
+        then:
+        dateTimeSort.get().direction == SortDirection.DESC
+        dateTimeSort.get().dimension == "dateTime"
+    }
+
+    def "Generate dateTime sort direction with default value"() {
```
cdeszaq: A better name would indicate what the expected default is: "...with default value of DESC if not direction is not specified"

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/DataApiRequestSortSpec.groovy
```diff
@@ -0,0 +1,97 @@
...
+        dateTimeSort.get().direction == SortDirection.DESC
+        dateTimeSort.get().dimension == "dateTime"
+    }
+
+    def "Generate api sort string after truncating the dateTime column"() {
+        when:
+        String apiSortValue = new DataApiRequest().truncateTimeSort("dateTime|desc,height|ASC")
+
+        then:
+        apiSortValue == 'height|ASC'
```
cdeszaq: Add a version of this test with the default dateTime sort (ie. no direction specified)

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/DateTimeSortAscOrderSpec.groovy
```diff
@@ -0,0 +1,250 @@
+// Copyright 2016 Yahoo Inc.
```
cdeszaq: 2017

cdeszaq:
Does this PR supercede the other one like this? (#170) If that's the case, can we close that one?

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
dayamr: Added this check and more test cases.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
dayamr: Nope. It can't be final. Because, its value will be changed if there is a sort request for dateTime column. After the code optimization, we don't need to initialize this.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
dayamr: To pull this out, we need to make sure the size of the list should be two.  If there is only column in the list (Need to consider the default direction) then we need to have default value for the variable which suppose to contain the pulled string from columnWithDirection.get(1). 

We need to do this to make catch part happy. Rather then that, would like to keep as it is. 

    SortDirection sortDirection;                                                                      
    String requestedDirection = null;                                                                 
    try {                                                                                             
       if (columnWithDirection.size() == 2)     {                                                    
           requestedDirection = columnWithDirection.get(1);                                          
           sortDirection =  SortDirection.valueOf(requestedDirection.toUpperCase(Locale.ENGLISH));    
                                                                                                  
       }                                                                                             
       else {                                                                                        
           sortDirection = SortDirection.DESC;                                                       
       }                                                                                             
    } catch (IllegalArgumentException ignored) {                                                      
          LOG.debug(SORT_DIRECTION_INVALID.logFormat(requestedDirection));                              
        throw new BadApiRequestException(SORT_DIRECTION_INVALID.format(requestedDirection));          
    }                                                                                                 
    return sortDirection;

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/mappers/DateTimeSortMapper.java
dayamr: It would have been useful if there is only natural order sort. But now we need to allow reverse order sort too. So, I don't feel it would make anything better then then this.

michael-mclawhorn:
@cdeszaq Rick, I think your feedback is partially contradicting mine.  I didn't want string parsing in the main DataApiRequest processing.  Which means the two options would be to parse ALL the sort order fields in one go, and then create two post-processing/validation stages to pick apart bucket sorting and validate that buckets are not mixed in with non buckets.  

Or to keep the parsing stage for the non-bucket sorts unchanged and add another lifecycle parsing off the same text field a second time.  Less efficient, but it retains the current code for bucket sorting.

Are you suggesting a compile-everything-and-then-filter-and-validate?  Or some other path?

dayamr:
I have used the map to store column names and its respective direction. That will avoid all the repetitions.

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/mappers/DateTimeSortMapper.java
dayamr: Done

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/mappers/DateTimeSortMapper.java
```diff
@@ -0,0 +1,85 @@
...
+
+        RequestLog.startTiming("sortResultSet");
+        try {
+            for (Result result : resultSet) {
+                bucketizedResultsMap.computeIfAbsent(result.getTimeStamp(), ignored -> new ArrayList<>()).add(result);
+            }
+
+            List<DateTime> dateTimeList = new ArrayList(bucketizedResultsMap.keySet());
+
+            Collections.sort(dateTimeList, direction == SortDirection.ASC ?
```
cdeszaq: Can just call `myList.sort(...)` on this, rather than rely on the static method. That also allows the line to be unwrapped.

/CanBeInASubsequentPr

cdeszaq: This ternary can actually get moved up into the constructor, since all we really need to hold onto is the comparator, not the direction.

/CanBeInASubsequentPr

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/mappers/DateTimeSortMapper.java
```diff
@@ -0,0 +1,85 @@
...
+                    Comparator.<DateTime>naturalOrder() :
+                    Comparator.<DateTime>reverseOrder());
+
+            return new ResultSet(
+                    dateTimeList.stream()
```
cdeszaq: Since `sorted` is an operation you can provide do to a stream, you can actually move the sorting down into the stream, and greatly simplify the bulk of the method by doing almost all of it as a stream pipeline:

```java
return new ResultSet(
        bucketizedResultsMap.keySet().stream()
                .sorted(direction)
                .map(bucketizedResultsMap::get)
                .flatMap(List::stream)
                .collect(Collectors.toList()),
        resultSet.getSchema()
);
```

/CanBeInASubsequentPr

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -114,6 +117,8 @@
 
     private final DruidFilterBuilder filterBuilder;
 
+    private Optional<OrderByColumn> dateTimeSort;
```
cdeszaq: Can this be `final`?

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -93,6 +95,7 @@
     private static final Logger LOG = LoggerFactory.getLogger(DataApiRequest.class);
     private static final String RATIO_METRIC_CATEGORY = "Ratios";
     public static final String REQUEST_MAPPER_NAMESPACE = "dataApiRequestMapper";
+    private static final String DATE_TIME_STRING = "dateTime";
```
cdeszaq: It's obvious from the type signature and the value that this is a date time string. Could we get a better name for this variable that indicates the _meaning_ of it's contents?

/CanBeInASubsequentPr

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -296,6 +309,87 @@ public DataApiRequest(
...
+     * @param sortColumns  map of columns and its direction
+     *
+     * @return True if dateTime column is first one in the sort list. False otherwise
+     */
+    protected Boolean isDateTimeFirstSortField(Map<String, SortDirection> sortColumns) {
+        if (sortColumns != null) {
+            List<String> columns = new ArrayList(sortColumns.keySet());
+            return columns.get(0).equals(DATE_TIME_STRING);
+        }
+        else {
```
cdeszaq: `else` should go up a line, onto the close brace of the `if` block

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -296,6 +309,87 @@ public DataApiRequest(
...
     /**
+     * To check whether dateTime column request is first one in the sort list or not.
+     *
+     * @param sortColumns  map of columns and its direction
+     *
+     * @return True if dateTime column is first one in the sort list. False otherwise
+     */
+    protected Boolean isDateTimeFirstSortField(Map<String, SortDirection> sortColumns) {
+        if (sortColumns != null) {
+            List<String> columns = new ArrayList(sortColumns.keySet());
```
cdeszaq: Needs type indicator so that the compiler doesn't think we're doing raw types.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -296,6 +309,87 @@ public DataApiRequest(
     }
 
     /**
+     * To check whether dateTime column request is first one in the sort list or not.
+     *
+     * @param sortColumns  map of columns and its direction
+     *
+     * @return True if dateTime column is first one in the sort list. False otherwise
+     */
+    protected Boolean isDateTimeFirstSortField(Map<String, SortDirection> sortColumns) {
```
cdeszaq: It's strange that we're talking about this method doing things in a list-y way, yet it takes a map. We should probably make this method take a list, since it cares about position.

At the very least, if order matters (since we convert the map to a list), we _definitely_ need to make the Map ordered in the method signature, rather than just taking a Map. (eg. HashMap will _not_ preserve order here and will not work as expected, causing irritating bugs)

cdeszaq: When we do this change, we can likely collapse this method body into just a single `return` statement, since I don't think the condition will be that complex.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -296,6 +309,87 @@ public DataApiRequest(
...
+     * @param sortColumns  map of columns and its direction
+     *
+     * @return  Map of columns and its direction without dateTime sort column
+     */
+    protected Map<String, SortDirection> removeDateTimeSortColumn(Map<String, SortDirection> sortColumns) {
+        if (sortColumns != null && sortColumns.containsKey(DATE_TIME_STRING)) {
+            sortColumns.remove(DATE_TIME_STRING);
+            return sortColumns;
+        }
+        else {
```
cdeszaq: else up a line

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -296,6 +309,87 @@ public DataApiRequest(
...
+    /**
+     * Method to remove the dateTime column from map of columns and its direction.
+     *
+     * @param sortColumns  map of columns and its direction
+     *
+     * @return  Map of columns and its direction without dateTime sort column
+     */
+    protected Map<String, SortDirection> removeDateTimeSortColumn(Map<String, SortDirection> sortColumns) {
+        if (sortColumns != null && sortColumns.containsKey(DATE_TIME_STRING)) {
+            sortColumns.remove(DATE_TIME_STRING);
```
cdeszaq: Since we only care about dateTime being the 1st element, we can probably use a list here as well, instead of a map, and only remove the 1st element.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -296,6 +309,87 @@ public DataApiRequest(
...
+            return sortColumns;
+        }
+        else {
+            return sortColumns;
+        }
+    }
+
+    /**
+     * Method to generate DateTime sort column from the map of columns and its direction.
```
cdeszaq: Note: Having "Method to" in a javadoc on a method is not useful info, since it's clear from context (ie. the comment is on a method) that the doc is describing the method. Obviously not a _big_ issue, but it's always good to be as concise as possible 😄 

/NotABigDeal

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -296,6 +309,87 @@ public DataApiRequest(
...
+    /**
+     * Method to convert sort list to column and direction map.
+     *
+     * @param sorts  String of sort columns
+     *
+     * @return Map of columns and its direction
+     */
+    protected Map<String, SortDirection> generateSortColumns(String sorts) {
+        if (sorts != null && !sorts.isEmpty()) {
+            return Arrays.asList(sorts.split(",")).stream()
```
cdeszaq: `Arrays.asList(sorts.split(",")).stream()` can be replaced with `Arrays.stream(sorts.split(","))`

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
dayamr: Since it is not a constructor parameter and no initial initialization, we don't think we can make it as final.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -296,6 +309,87 @@ public DataApiRequest(
...
+                    .map(e -> Arrays.asList(e.split("\\|")))
+                    .collect(Collectors.toMap(
+                            e -> e.get(0), e -> getSortDirection(e),
```
cdeszaq: These two parameters should be wrapped on the `,`, and the 2nd one can be replaced with a method reference.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
dayamr: Updated the method signature to preserve the order by using LinkedHashMap

cdeszaq: Please make it take a `List` in a follow-on PR. It really doesn't feel like we're doing Map operations for most of this logic, it feels much more list-y (and the JavaDoc reflects that as well). It can stay a LHM for now, but we should correct the mis-match.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
cdeszaq: It looks like we're setting it in the constructors to me. The only place I think we're _not_ is in a copy-constructor, to which we should probably add it as a parameter. Once that's done, it _should_ be able to be `final` :-)

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -1098,52 +1193,56 @@ protected DateTime getCurrentDate(DateTime dateTime, TimeGrain timeGrain) {
     }
 
     /**
+     * Extract valid sort direction.
+     *
+     * @param columnWithDirection  Column and its sorting direction
```
cdeszaq: Since this is a String, we should definitely add more detail to this comment that indicates the expected format(s).

/CanBeInAFollowOnPr

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
dayamr: During the creation of `<OrderByColumn> dateTime`, It is assured that dateTime is first in the list. If not, it will show an exception. If the code reaches this point, it assures dateTime is first in the sort list.

Reason for not considering List is, 
- I need to split each element in a list for validations like 'dateTime' is first in the list
- Need to split to check if it is requested middle of the list. So i can throw valid error to the user
- Split during regular sorting.

Instead of all the different splits, I have done it once and associating with its direction. We can avoid all the other splits unnecessarily.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
cdeszaq: I better understand things now. This isn't a String, but a `List<String>` which is expected to have 2 elements, a column name and a sort direction. It's a little odd, but I realize now that it's part of parsing.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -1098,52 +1193,56 @@ protected DateTime getCurrentDate(DateTime dateTime, TimeGrain timeGrain) {
...
+     *
+     * @return Sorting direction. If no direction provided then the default one will be DESC
+     */
+    protected SortDirection getSortDirection(List<String> columnWithDirection) {
+
+        SortDirection sortDirection;
+        try {
+            sortDirection = columnWithDirection.size() == 2 ?
```
cdeszaq: Rather than save to a variable and then return it, just return here directly.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
cdeszaq: Ahh, I get it. Makes sense.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
dayamr: Done

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
dayamr: Sure. Have taken a note for follow up PR.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -303,55 +309,87 @@ public DataApiRequest(
...
-     * @param sorts  String of sort columns
+     * @param sortColumns  map of columns and its direction
      *
-     * @return Updated sort column if dateTime exists. Otherwise return as it is received.
+     * @return True if dateTime column is first one in the sort list. False otherwise
      */
-    protected String truncateTimeSort(String sorts) {
+    protected Boolean isDateTimeFirstSortField(Map<String, SortDirection> sortColumns) {
+        if (sortColumns != null) {
+            List<String> columns = new ArrayList(sortColumns.keySet());
```
will-lauer: I really don't like this. The contract for the input here is a Map, which is inherently unsorted, so assuming that you can get the keySet and take the first item in the list to get the first item from the sort list is incorrect. I understand that behind the scenes, sortColumns is a LinkedHashMap, which causes this to actually return the desired result, but that isn't captured here in the contract. If you really expect that, then I think you need to change the signature here to REQUIRE a LinkedHashMap. Otherwise, you need to change this to some other Collection that does enforce ordering that isn't a Map.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -296,6 +309,87 @@ public DataApiRequest(
...
+     * @return Map of columns and its direction
+     */
+    protected Map<String, SortDirection> generateSortColumns(String sorts) {
+        if (sorts != null && !sorts.isEmpty()) {
+            return Arrays.asList(sorts.split(",")).stream()
+                    .map(e -> Arrays.asList(e.split("\\|")))
+                    .collect(Collectors.toMap(
+                            e -> e.get(0), e -> getSortDirection(e),
+                                    (u, v) -> {
+                                        LOG.debug(SORT_REQUESTED_ON_DUPLICATE_COLUMN.logFormat(u));
```
cdeszaq: I don't think we can actually make this error check without much more indication in the change log, nor am I sure we want to add it... Previously, we were happily allowing the same sort to show up more than once (I think), and were probably doing whatever Druid did for sorting (so probably either first or last wins). Since this would technically be a breaking API change, we can't just add it in here, in case someone is relying on how this works today.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/DateTimeSortRequestHandler.java
```diff
@@ -0,0 +1,42 @@
...
+import com.yahoo.bard.webservice.druid.model.query.DruidAggregationQuery;
+import com.yahoo.bard.webservice.web.DataApiRequest;
+import com.yahoo.bard.webservice.web.responseprocessors.MappingResponseProcessor;
+import com.yahoo.bard.webservice.web.responseprocessors.ResponseProcessor;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * Injects the result set mapper for dateTime based sort if an API request asks for result in descending order.
+ * The time sort mapper should be second last mappers to execute on the result set.
```
cdeszaq: A couple of things here:

1. It's not only activated for descending order, but _any_ order, right?
2. It says that the mapper should be 2nd to last, yet the `add` call in this class seems to be putting it at the end. Something doesn't seem to be correct?

cdeszaq: nvm, I understand now. It's the static ordering of the mappers (as set in the Workflow), _not_ the mapper. My bad 😄

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/DataApiRequestSortSpec.groovy
```diff
@@ -0,0 +1,142 @@
...
+        ["dinga":SortDirection.DESC] as Map                                                      | Optional.empty()
+    }
+
+    @Unroll
+    def "Remove dateTime sort column from columnDirection map "() {
+        when:
+        Map<String, SortDirection> actual = new DataApiRequest().removeDateTimeSortColumn(columnDirection)
+
+        then:
+        actual == expected
```
cdeszaq: Can be simplified to just an `expected` block, since we're just comparing a value to a single call.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/DataApiRequestSortSpec.groovy
```diff
@@ -0,0 +1,142 @@
...
+        e.getMessage() == expectedMessage
+    }
+
+    @Unroll
+    def "Validate the sort column and direction map from #sortString string"() {
+        when:
+        Map<String, SortDirection> actual = new DataApiRequest().generateSortColumns(sortString)
+
+        then:
+        actual == expected
```
cdeszaq: Can be simplified to just an `expected` block, since we're just comparing a value to a single call.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/DataApiRequestSortSpec.groovy
```diff
@@ -0,0 +1,142 @@
...
+        "xyz|DESC,dateTime|DESC,abc|ASC"  | ["xyz":SortDirection.DESC,"dateTime":SortDirection.DESC, "abc":SortDirection.ASC] as Map
+        "xyz|DESC,dateTime,abc|DESC"      | ["xyz":SortDirection.DESC,"dateTime":SortDirection.DESC, "abc":SortDirection.DESC] as Map
+        "xyz|DESC,dateTime"               | ["xyz":SortDirection.DESC,"dateTime":SortDirection.DESC] as Map
+        ""                                | null
+        "dinga"                           | ["dinga":SortDirection.DESC] as Map
+
+    }
+
+    @Unroll
+    def "Generate dateTime sort column from columnDirection map "() {
```
cdeszaq: Should have some sort of expanded variable in the test name when we have an `@Unroll` test, to differentiate the cases.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/DataApiRequestSortSpec.groovy
```diff
@@ -0,0 +1,142 @@
...
+        when:
+        Optional<OrderByColumn> actual = new DataApiRequest().generateDateTimeSortColumn(columnDirection)
+
+        then:
+        actual == expected
```
cdeszaq: Can be simplified to just an `expected` block, since we're just comparing a value to a single call.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
dayamr: I have updated the contract to make sure argument type is LinkedHashMap. Will commit the changes after addressing all the review comments.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/DateTimeSortRequestHandler.java
dayamr: Updated the javadoc.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -493,8 +484,8 @@ private DataApiRequest(
             int topN,
             long asyncAfter,
             DateTimeZone timeZone,
-            DruidFilterBuilder filterBuilder
-    ) {
+            DruidFilterBuilder filterBuilder,
+            Optional<OrderByColumn> dateTimeSort) {
```
cdeszaq: This wrapping is wrong.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/DateTimeSortAscOrderSpec.groovy
```diff
@@ -1,15 +1,12 @@
 // Copyright 2017 Yahoo Inc.
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.web.endpoints
-
```
cdeszaq: Keep

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/DateTimeSortDescOrderSpec.groovy
```diff
@@ -1,15 +1,12 @@
 // Copyright 2017 Yahoo Inc.
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.web.endpoints
-
```
cdeszaq: Keep

cdeszaq:
Just trivial things. Tweak those, and this looks good

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
dayamr: Got it. Fixed.

cdeszaq:
In the follow-on PR to do some final cleanup things, we should also document this capability in the API doc.

PR #179, opened 2017-02-22
3. CompositePhsyicalTable Core Components Refactor

garyluoex:
***Motivation***
This PR is created to refactor the existing table availability structure to be more extensible such that it can support composite tables i.e. union datasource in druid. Composite tables depends on other concrete physical tables, and therefore its availability is a result of merging the availability of the underlying concrete physical table's availability. The details of the merging process is implemented in various different kinds of availability, such as metricUnionAvailability. The only implementation in this PR is ConcreteAvailability which resembles the basic table backed by druid directly that is supported in fili already but implemented in a new extensible structure. In addition, availability directly sources segment information from DataSourceMetadataService now instead of SegmentMetadataLoader which is now deprecated.

***Detailed Changes***
 * Added `ConcretePhysicalTable` and `ConcreteAvailability` to model table in druid datasource and its availabillity in the new table availability structure
 * Added method `containsColumn` to `Schema`
 * `ConfigurationLoader` now takes an additional constructor argument `DataSourceMetadataService` for creating tables
 * `findMissingRequestTimeGrainIntervals` method in `PartialDataHandler` now takes `DataSourceConstraint`
 * Removed deprecated method `findMissingRequestTimeGrainIntervals` from `PartialDataHandler`
 * Removed `permissive_column_availability` feature flag support and corresponding functionality in `PartialDataHandler`
 * Removed `getIntersectSubintervalsForColumns` and `getUnionSubintervalsForColumns` from `PartialDataHandler` since the logic is pushed into `Availability` now
 * Removed `getIntervalsByColumnName`, `resetColumns` and `hasLogicalMapping` methods in `PhysicalTable` since no longer needed with the availability structure
 * Removed deprecated `buildTableGroup` method in `BaseTableLoader`
 * `getAvailability` method on `PartialDataHandler` is now deprecated since the logic is pushed into `Availability`
* Removed `getAvailability` method on `PartialDataHandler` since the logic is pushed into `Availability`


fili-core/src/main/java/com/yahoo/bard/webservice/metadata/SegmentMetadataLoader.java
```diff
@@ -156,7 +156,6 @@ public void invoke(JsonNode rootNode) {
                 if (segmentMetadata.isEmpty()) {
                     LOG.warn("Empty segment metadata detected when loading table '{}'", table.getFactTableName());
                 }
-                table.resetColumns(segmentMetadata, dimensionDictionary);
```
garyluoex: Can I just remove the deprecated SegmentMetadata?

archolewa:
Taking a look now.

fili-core/src/main/java/com/yahoo/bard/webservice/data/PartialDataHandler.java
```diff
@@ -82,21 +40,21 @@ public SimplifiedIntervalList findMissingRequestTimeGrainIntervals(
      * present for a given combination of request metrics and dimensions (pulled from the API request and generated
      * druid query) at the specified granularity.
      *
-     * @param columnNames  all the column names the request depends on
+     * @param constraints all the column names the request depends on
```
archolewa: This JavaDoc should be updated, since we are now passing in something more general than just column names.

fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseSchema.java
```diff
@@ -33,6 +33,22 @@ protected BaseSchema(Granularity granularity, Iterable<Column> columns) {
...
+     *
+     * @return the column if found
+     */
+    public Column getColumn(String columnName) {
+        for (Column column : columns) {
+            if (column.getName().equals(columnName)) {
+                return column;
+            }
+        }
+        return null;
```
garyluoex: Will fix this null in next commit.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
```diff
@@ -57,33 +60,23 @@ protected BaseTableLoader() {
...
+    public abstract void loadTableDictionary(ResourceDictionaries dictionaries);
+
+    @Override
+    public void loadTableDictionary(
+            ResourceDictionaries dictionaries,
+            DataSourceMetadataService metadataService
     ) {
-        return buildTableGroup(apiMetrics, druidMetrics, tableDefinitions, dictionaries);
+        this.metadataService = metadataService;
+        this.loadTableDictionary(dictionaries);
```
archolewa: This setup is awkward.

We've deprecated the old version, which doesn't take the `metadataService`. This implies that we shouldn't be using the old method at all. Meanwhile, this version of the function just assigns the class level `metadataService` and then calls into the old version. That's fine as far as preserving backwards compatibility. 

However, near as  I can tell, the "bridge to preserve backwards compatibility" is now the new version of this class is _intended_ to be used. People are meant to override `loadTableDictionary`, and then call into the various helper functions to do the heavy lifting for them. However, none of those helper functions take the `metadataService` as an argument, and instead rely on the `metadataService` field.

So in other words, if people want to do the "right" thing, and stop using the deprecated method, and they want to do the "right" thing and use the various helper methods, they have to:

1. Set a field in the super class.
2. Call their old code and let magic happen.

Instead, I think we should deprecate _all_ the helper methods that rely on `metadataService` but don't take an argument (`loadPhysicalTable`, `buildPhysicalTable`, `buildTableGroup`), in favor of a version that does take the `metadataService` as an argument. It will be intended then, that people will call the helper methods that take the `metadataService` as an argument, and pass it through to them from their version of `loadTableDictionary`.

The deprecated versions can then call into the new versions, passing in the `segmentMetadata` field.

That way, the blessed way of using the `BaseTableLoader` doesn't have to do any awkward superclass field manipulation, and we still have the same backwards compatibility as we do now.

@michael-mclawhorn Does this make sense to you?

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
```diff
@@ -57,33 +60,23 @@ protected BaseTableLoader() {
...
+     * @deprecated should use the same method that takes an additional DataSourceMetadataService argument instead
      */
     @Deprecated
-    public TableGroup buildTableGroup(
-            String logicalTableName,
-            Set<ApiMetricName> apiMetrics,
-            Set<FieldName> druidMetrics,
-            Set<PhysicalTableDefinition> tableDefinitions,
-            ResourceDictionaries dictionaries
+    public abstract void loadTableDictionary(ResourceDictionaries dictionaries);
```
archolewa: If we're going to be deprecating this, then we shouldn't make it abstract. We really shouldn't be _forcing_ people to implement a deprecated method. 

If we still want to force people to implement _something_, then we should make the new `loadTableDictionary` abstract, and include a small snippet in the migration path that basically tells users to do what we do below.

It's a little bit of pain for people upgrading to the new version, but I would argue it's better than forcing people to implement cruft, and it's a very minor compile-time pain so whatever.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
```diff
@@ -268,7 +261,11 @@ protected PhysicalTable loadPhysicalTable(
         PhysicalTable existingTable = dictionaries.getPhysicalDictionary().get(definition.getName().asName());
         if (existingTable == null) {
             // Build the physical table
-            existingTable = buildPhysicalTable(definition, metricNames, dictionaries.getDimensionDictionary());
+            existingTable = buildPhysicalTable(
```
archolewa: Doesn't need to be wrapped.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/TableLoader.java
```diff
@@ -14,7 +15,8 @@
      * Load all of the logical tables, their table groups, and the physical tables for those groups.
      *
      * @param dictionaries  The resource dictionaries that will be loaded with table configuration
+     * @param metadataService datasource metadata service containing segments for building table
```
archolewa: Needs an extra space between `metadataService` and `datasource`

fili-core/src/main/java/com/yahoo/bard/webservice/metadata/SegmentMetadataLoader.java
archolewa: Well, seeing as how removing this line means that this class _will not work_, we have two choices:

1. Remove the class completely (so people relying on it get compile time rather than runtime errors).

2. Find some way of getting the segments pulled out by `SegmentMetadataLoader` and shoved into the tables in the new scheme to preserve backwards compatibility.

I haven't dug deeply enough into the code yet to see how feasible number 2 is. Number 1 will make for some migration pain, but I think migration pain is unavoidable at this point. The only question is whether that pain should be up-front and found during migration, or subtle and found during production.

fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
```diff
@@ -3,26 +3,29 @@
...
 
 import java.util.Map;
+import java.util.Set;
 
 import javax.validation.constraints.NotNull;
 
 /**
- * Base Physical Table implements common PhysicalTable capabilities.
+ * Physical Table represents a druid table.
+ * Working availabilities a transactional sensibility to changing availabilities.
```
archolewa: I have no idea what this is trying to say.

fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
```diff
@@ -3,26 +3,29 @@
...
 import org.slf4j.LoggerFactory;
 
 import java.util.Map;
+import java.util.Set;
 
 import javax.validation.constraints.NotNull;
 
 /**
- * Base Physical Table implements common PhysicalTable capabilities.
+ * Physical Table represents a druid table.
```
archolewa: This comment is also slightly off. 

Now, a `PhysicalTable` _may_ represent a concrete druid (actually we should call it `fact`) table, but it may also represent a view/union/whatevers. It's the `ConcretePhysicalTable` that represents an honest-to-goodness table that you can get your hands on.

fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseSchema.java
```diff
@@ -33,6 +33,22 @@ protected BaseSchema(Granularity granularity, Iterable<Column> columns) {
...
     }
 
+    /**
+     * Retrieve a column from schema given the name of the column.
+     *
+     * @param columnName  The column name in string
+     *
+     * @return the column if found
+     */
+    public Column getColumn(String columnName) {
```
archolewa: Since this is a brand new method, we should have it return `Optional<Column>` rather than potentially `null`. Optionals are nice because they bake into the type system the fact that there may not be a column, and so it forces the callers to deal with that eventuality.

Also, you can stream this very easily:

```public Optional<Column> getColumn(String columnName) {
    return columns.stream()
              .filter(column -> column.getName().equals(columnName))
              .findFirst();
}
```
The beautiful thing is that `findFirst` returns an `Optional`, so you can kill two birds with one stone!

fili-core/src/main/java/com/yahoo/bard/webservice/table/ConcretePhysicalTable.java
```diff
@@ -26,19 +27,21 @@
...
      * @param columns  The columns for this table
      * @param logicalToPhysicalColumnNames  Mappings from logical to physical names
+     * @param metadataService datasource metadata service containing availability data for the table
      */
     public ConcretePhysicalTable(
             @NotNull TableName name,
             @NotNull ZonedTimeGrain timeGrain,
-            @NotNull Iterable<Column> columns,
-            @NotNull Map<String, String> logicalToPhysicalColumnNames
+            @NotNull Set<Column> columns,
```
archolewa: Why the change from an `Iterable` to a `Set`? I believe @michael-mclawhorn made it an iterable, precisely to try to get across the idea that "Iteration order doesn't really matter, but we'll preserve it, so don't worry if it does."

`Set` doesn't really get that across.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * An availability which guarantees immutability on its contents.
+ */
+public class ConcreteAvailability implements Availability {
```
archolewa: Why are we replacing `ImmutableAvailability` with this?

fili-core/src/main/java/com/yahoo/bard/webservice/table/ConcretePhysicalTable.java
```diff
@@ -26,19 +27,21 @@
...
+            @NotNull Map<String, String> logicalToPhysicalColumnNames,
+            @NotNull DataSourceMetadataService metadataService
     ) {
         super(
                 name,
                 timeGrain,
                 columns,
                 logicalToPhysicalColumnNames,
-                new ImmutableAvailability(name, Collections.emptyMap())
+                new ConcreteAvailability(name, columns, metadataService)
```
archolewa: Do we want to be building the availability, or passing it in?

If we inject the `Availability` directly, it'll make it a lot easier for customers to provide their own custom availability rules...

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -4,18 +4,29 @@
...
+ * An interface describing a config level physical table.
+ * It may be backed by a single concrete data source or not.
  */
 public interface PhysicalTable extends Table {
 
+    @Override
+    String getName();
+
+    @Override
+    PhysicalTableSchema getSchema();
```
archolewa: These overrides are not necessary.

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -58,23 +75,19 @@
...
-     * Determine whether or not this PhysicalTable has a mapping for a specific logical name.
+     * Getter for active column intervals.
      *
-     * @param logicalName  Logical name to check
+     * @param constraints data constraint containing columns and api filters
      *
-     * @return True if contains a non-default mapping for the logical name, false otherwise
-     *
-     * @deprecated This may no longer be needed
+     * @return tableEntries map of column to set of available intervals
```
archolewa: This JavaDoc isn't accurate. The method is returning a `SimplifiedIntervalList`, not a map.

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -58,23 +75,19 @@
...
      *
      * @param logicalName  Logical name to lookup in physical table
+     *
      * @return Translated logicalName if applicable
      */
     String getPhysicalColumnName(String logicalName);
 
     /**
-     * Determine whether or not this PhysicalTable has a mapping for a specific logical name.
+     * Getter for active column intervals.
```
archolewa: I would suggest expanding on this a bit. Something like:

```java
Return all the available intervals that satisfy the given constraints.
```

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -0,0 +1,86 @@
...
+
+    private final TableName name;
+    private final Set<Column> columns;
+    private final DataSourceMetadataService metadataService;
+
+    /**
+     * Constructor.
+     *
+     * @param tableName The name of the data source associated with this ImmutableAvailability
```
archolewa: This javadoc needs to be updated. This class is called `ConcreteAvailability`, not `ImmutableAvailability`.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Override
+    public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constraints) {
+
+        Set<String> configuredColumns = columns.stream()
+                .map(Column::getName).collect(Collectors.toSet());
+
+        Set<String> requestColumns = constraints.getAllColumnNames().stream()
+                .filter(requestColumn -> configuredColumns.contains(requestColumn)).collect(Collectors.toSet());
```
archolewa: The filter can be simplified to a method reference: `configuredColumns::contains`

fili-core/src/main/java/com/yahoo/bard/webservice/data/PartialDataHandler.java
```diff
@@ -82,21 +40,21 @@ public SimplifiedIntervalList findMissingRequestTimeGrainIntervals(
...
      *
-     * @param columnNames  all the column names the request depends on
+     * @param constraints  Constraint containing all the column names the request depends on
      * @param physicalTables  the tables whose column availabilities are checked
      * @param requestedIntervals  The intervals that may not be fully satisfied
      * @param granularity  The granularity at which to find missing intervals
      *
      * @return subintervals of the requested intervals with incomplete data
      */
     public SimplifiedIntervalList findMissingTimeGrainIntervals(
```
garyluoex: Can this be a static method? This class got strip away a lot of responsibility such that it only have this method left.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
garyluoex: Good point, I will simply not deprecate this method since I think it is the best way without any breaking change but still adding the metadataService to it.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
garyluoex: I think simply removing the deprecated tag to the `loadTableDictionary` method that people are using today should fix all the problem, since user should not care and does not have to touch the `metadataService`, instead fili takes care of it in its `AbstractBinderFactory`.

fili-core/src/main/java/com/yahoo/bard/webservice/metadata/SegmentMetadataLoader.java
garyluoex: I will remove it since it was already `Deprecated` and are using the new `DataSourceMetadataLoader` now.

fili-core/src/main/java/com/yahoo/bard/webservice/table/ConcretePhysicalTable.java
garyluoex: I am making it a set because I do not yet to see any place requiring the behavior "Iteration order doesn't really matter, but we'll preserve it, so don't worry if it does.", if this behavior is not required, I prefer using a more standard approach of `Set`. Let me know if there is a case this is needed and I will change it back.

fili-core/src/main/java/com/yahoo/bard/webservice/table/ConcretePhysicalTable.java
garyluoex: Speaking of this...... I am thinking if we should just make one type of table and let user define what type of availability it has. If we decided to go this route, we might need to re-think the configuration of tables entirely. @cdeszaq @michael-mclawhorn

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
garyluoex: Yep, any question or concern?

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/PartialDataHandlerSpec.groovy
```diff
@@ -69,73 +57,47 @@ class PartialDataHandlerSpec extends Specification {
          * dim1 is missing four days of data internally, dim2 and dim3 are complete over the period and page_views is
          * starts inside the dim1 hole and goes to the end of the period.
          */
-        SegmentMetadata segments = new SegmentMetadata(
-                [("userDeviceType"): buildIntervals(["2014-07-01/2014-07-09","2014-07-11/2014-07-29"]) as LinkedHashSet,
-                 ("property"): buildIntervals(["2014-07-01/2014-07-29"]) as LinkedHashSet,
-                 ("os"): buildIntervals(["2014-07-01/2014-07-29"]) as LinkedHashSet],
-                [("page_views"): buildIntervals(["2014-07-04/2014-07-29"]) as Set]
+        Map<Column, Set<Interval>> segmentIntervals = [
+                (new Column("userDeviceType")): buildIntervals(["2014-07-01/2014-07-09","2014-07-11/2014-07-29"]).toSet(),
```
archolewa: It's more idiomatic to use the `as` operator, rather than the `toSet()`method:

```groovy
 (new Column("userDeviceType")): buildIntervals(["2014-07-01/2014-07-09","2014-07-11/2014-07-29"]) as Set
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/QueryBuildingTestingResources.groovy
```diff
@@ -104,6 +108,8 @@ public class QueryBuildingTestingResources {
         DateTimeZone.setDefault(UTC)
         def ages = ["1": "0-10", "2": "11-14", "3": "14-29", "4": "30-40", "5": "41-59", "6": "60+"]
 
+        metadataService = new TestDataSourceMetadataService(Collections.emptyMap())
```
archolewa: In Groovy, it's more idiomatic to use `[:]` rather than `Collections.emptyMap()`

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/table/BaseTableLoaderSpec.groovy
```diff
@@ -13,6 +13,7 @@ import com.yahoo.bard.webservice.data.config.names.TestApiMetricName
 import com.yahoo.bard.webservice.data.config.names.TestDruidMetricName
 import com.yahoo.bard.webservice.data.dimension.Dimension
 import com.yahoo.bard.webservice.data.dimension.impl.KeyValueStoreDimension
+import com.yahoo.bard.webservice.metadata.DataSourceMetadataService
```
archolewa: Should be deleted.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/query/TimeSeriesQuerySpec.groovy
```diff
@@ -40,8 +41,10 @@ class TimeSeriesQuerySpec extends Specification {
         vars.dataSource = vars.dataSource ?: new TableDataSource(new ConcretePhysicalTable(
                 "table_name",
                 DAY.buildZonedTimeGrain(DateTimeZone.UTC),
-                [] as Set,
-                [:]
+                [] as Set
+                ,
```
archolewa: Should be moved up onto the previous line.

archolewa:
Finished my first round of reviews.

garyluoex:
Thanks @archolewa

fili-core/src/main/java/com/yahoo/bard/webservice/table/ConcretePhysicalTable.java
```diff
@@ -70,7 +75,6 @@ public String getFactTableName() {
 
     @Override
     public String toString() {
-        return super.toString() + " factTableName: " + getAvailability().getDataSourceNames() + " alignment: " +
-                getTableAlignment();
+        return super.toString() + " factTableName: " + getAvailability().getDataSourceNames();
```
garyluoex: Simplify testing, class scanner checks toString method and blows up with this thing.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/Availability.java
```diff
@@ -22,33 +22,21 @@
...
-    List<Interval> get(Column c);
+    Set<TableName> getDataSourceNames();
 
     /**
      * The availability of all columns.
      *
      * @return The intervals, by column, available.
      */
-    Map<Column, List<Interval>> getAvailableIntervals();
+    Map<Column, Set<Interval>> getAllAvailableIntervals();
```
garyluoex: The question I was thinking is why is `SimplifiedIntervalList` not `SimplifiedIntervalSet`

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -0,0 +1,90 @@
...
+        this.name = tableName;
+        this.columns = ImmutableSet.copyOf(columns);
+        this.metadataService = metadataService;
+
+        this.columnNames = columns.stream().map(Column::getName).collect(Collectors.toSet());
+    }
+
+    @Override
+    public Set<TableName> getDataSourceNames() {
+        return Collections.singleton(name);
```
garyluoex: I argue that doing it here makes logic clear and easier to maintain, might lose maintainability by trying to do a tiny optimization that might not be needed.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -316,12 +305,13 @@ protected void configure() {
...
                 bind(getClock()).to(Clock.class);
 
+                DruidDimensionsLoader druidDimensionsLoader = buildDruidDimensionsLoader(
+                        nonUiDruidWebService,
+                        loader.getPhysicalTableDictionary(),
+                        loader.getDimensionDictionary()
+                );
+
                 if (DRUID_DIMENSIONS_LOADER.isOn()) {
-                    DruidDimensionsLoader druidDimensionsLoader = buildDruidDimensionsLoader(
```
garyluoex: So that Bank can get a hold of the dimensionDictionary since `builtSegmentMetadata is removed`. Temporary fix...

fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
```diff
@@ -3,23 +3,25 @@
...
 import java.util.Map;
+import java.util.Set;
 
 import javax.validation.constraints.NotNull;
 
 /**
- * Base Physical Table implements common PhysicalTable capabilities.
+ * Base implementation of physical table that are shared across various types of physical tables.
  */
 public abstract class BasePhysicalTable implements PhysicalTable {
```
garyluoex: It is weird if move into schema, then we would have to call `containsLogicalName` just to invoke the log.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
cdeszaq: In almost every other place in the codebase, we've generally pushed conversions like this (ie. from an instance into an immutable collection due to a collection getter) into the constructor, rather than call it multiple times. So, from that standpoint, doing so maintains consistency with the codebase.

The other perspective on this is that a `ConcreteAvailability` will generally live for the life of the system, and so this method will be called at least once per request. While I'm sure it's not a huge impact, creating more objects than needed causes more garbage overhead, which introduces less consistency in the performance of the system. So, having more performant code (especially in terms of consistency), for _slightly_ less maintainability, is likely a worthwhile reason.  In theory, this is something the JIT _should_ be able to optimize away, so if you can show that the JIT _will_ optimize the repeated object away, I'm all for letting the JIT do the work.

Lastly, while your point about having a collection as a private field _could_ be slightly confusing, a good name (perhaps like `cachedDataSourceNames`) mitigates that concern significantly.

fili-core/src/main/java/com/yahoo/bard/webservice/table/Schema.java
```diff
@@ -15,6 +15,15 @@
 public interface Schema {
 
     /**
+     * Retrieve a column from schema given the name of the column.
+     *
+     * @param columnName  The column name in string
+     *
+     * @return an optional of the matching name column
+     */
+    Optional<Column> getColumn(String columnName);
```
garyluoex: It doesn't hurt and seems like a natural method on schema right?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/Availability.java
cdeszaq: It's a List because it makes collapsing the intervals down _much_ more performant. It's the same general idea that searching through an ordered collection is _much_ faster than an unordered collection (which essentially requires a scan). @michael-mclawhorn can go deeper on this if you'd like 😉

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/Availability.java
```diff
@@ -22,33 +22,21 @@
...
-
-    /**
-     * The availability for a given column.
-     *
-     * @param c  A column
-     *
-     * @return The list of intervals that column is available for.
-     */
-    List<Interval> get(Column c);
+    Set<TableName> getDataSourceNames();
```
garyluoex: I believe SortedSet was accidentally pushed, I try not to enforce ordering when there is no actually meaning for the ordering.

fili-core/src/main/java/com/yahoo/bard/webservice/table/ConcretePhysicalTable.java
cdeszaq: Rather than change the code, we should make sure we understand _why_ the class scanner is blowing up. In general, changing code that causes tests to break should first be well understood. Since `getTableAlignment` is a generally innocuous method, and critical for some aspects of the system (like the granularity logic engine), we should make sure we understand this. I'm not convinced that changing `toString` is the right impulse here.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
cdeszaq: 👎 If we want a hook to get at the dimensionDictionary, it would be better to just add that hook in a more long-term fashion, rather than subvert a feature flag and cause an object to be built when it's not supposed to be.

fili-core/src/main/java/com/yahoo/bard/webservice/table/Schema.java
cdeszaq: While I don't disagree that it seems natural, if we don't have a clear use for a method beyond tests, we should be very careful with making the public surface of classes any larger than is required. Once there, it becomes _much_ harder to change it in a backwards-compatible way, and any extra surface area is extra complexity. (even if it seems small)

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/Availability.java
```diff
@@ -22,33 +22,21 @@
...
+     * @param constraints  Data constraint containing columns and api filters
      *
-     * @return Set of intervals associated with a column, empty if column is missing
+     * @return Set of intervals associated with each corresponding column in a map, empty if column is missing
      */
-    default List<Interval> getIntervalsByColumnName(String columnName) {
-        List<Interval> result = get(new Column(columnName));
-        return result == null ? Collections.emptyList() : result;
-    }
+    SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constraints);
```
garyluoex: I do not see a need to have single column case except for testing purposes at this point.

garyluoex: Because it is a `SimplifedIntervalList`, let me know if `List` makes more sense in this case or should we even try to convert it into SIL here.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -0,0 +1,90 @@
...
+                        )
+                );
+    }
+
+    @Override
+    public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constraints) {
+
+        Set<String> requestColumns = constraints.getAllColumnNames().stream()
+                .filter(columnNames::contains)
+                .collect(Collectors.toSet());
```
garyluoex: Your choice, you might not be able to see the changes again in the next PR so if it is this code that you want to comment, most likely comment it here would be better.

fili-core/src/main/java/com/yahoo/bard/webservice/web/SlicesApiRequest.java
```diff
@@ -165,7 +164,7 @@ protected SlicesApiRequest() {
             throw new BadApiRequestException(msg);
         }
 
-        Map<Column, List<Interval>> columnCache = table.getAvailability().getAvailableIntervals();
+        Map<Column, Set<Interval>> columnCache = table.getAllAvailableIntervals();
```
garyluoex: Not sure how the ordering of interval is used in the response, but I do want to argue changing the core code just for less painful tests is not worth it, and it would be great if the actual case that make this hard to test could be pointed out so that I can check if it make sense to make it a list just for testing.

fili-core/src/test/groovy/com/yahoo/bard/webservice/application/AbstractBinderFactorySpec.groovy
```diff
@@ -288,34 +283,6 @@ public class AbstractBinderFactorySpec extends Specification {
         capturedHealthCheck.dimensionDictionary.is(dimensionDictionary)
     }
 
-    def "Test Setup Partial Data"() {
```
garyluoex: There is a "Test Setup Datasource Metadata" that does the same thing

fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
```diff
@@ -93,20 +108,13 @@ public String getPhysicalColumnName(String logicalName) {
...
-        setAvailability(new ImmutableAvailability(
-                name.asName(),
-                segmentMetadata.getDimensionIntervals(),
-                segmentMetadata.getMetricIntervals(),
-                dimensionDictionary
-        ));
-    }
-
+    @Deprecated
     protected void setAvailability(Availability availability) {
```
garyluoex: Some how in `AvailbilityTestingUtils`, it can't use it if I make it private.

fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
cdeszaq: Right now, `BasePhysicalTable` is calling `containsLogicalName` just to invoke the log, so that's not really a change. When possible, we should push complexity lower in the stack.

For reference, here the current code in `BasePhysicalTable`:

```java
/**
 * Translate a logical name into a physical column name. If no translation exists (i.e. they are the same),
 * then the logical name is returned.
 * <p>
 * NOTE: This defaulting behavior <em>WILL BE REMOVED</em> in future releases.
 * <p>
 * The defaulting behavior shouldn't be hit for Dimensions that are serialized via the default serializer and are
 * not properly configured with a logical-to-physical name mapping. Dimensions that are not "normal" dimensions,
 * such as dimensions used for DimensionSpecs in queries to do mapping from fact-level dimensions to something else,
 * should likely use their own serialization strategy so as to not hit this defaulting behavior.
 *
 * @param logicalName  Logical name to lookup in physical table
 *
 * @return Translated logicalName if applicable
 */
public String getPhysicalColumnName(String logicalName) {
    if (!schema.containsLogicalName(logicalName)) {
        LOG.warn(
                "No mapping found for logical name '{}' to physical name on table '{}'. Will use logical name as " +
                        "physical name. This is unexpected and should not happen for properly configured " +
                        "dimensions.",
                logicalName,
                getName()
        );
    }
    return schema.getPhysicalColumnName(logicalName);
}
```

and the code in `PhysicalTableSchema`:

```java
/**
 * Translate a logical name into a physical column name. If no translation exists (i.e. they are the same),
 * then the logical name is returned.
 * <p>
 * NOTE: This defaulting behavior <em>WILL BE REMOVED</em> in future releases.
 * <p>
 * The defaulting behavior shouldn't be hit for Dimensions that are serialized via the default serializer and are
 * not properly configured with a logical-to-physical name mapping. Dimensions that are not "normal" dimensions,
 * such as dimensions used for DimensionSpecs in queries to do mapping from fact-level dimensions to something else,
 * should likely use their own serialization strategy so as to not hit this defaulting behavior.
 *
 * @param logicalName  Logical name to lookup in physical table
 *
 * @return Translated logicalName if applicable
 */
public String getPhysicalColumnName(String logicalName) {
    return logicalToPhysicalColumnNames.getOrDefault(logicalName, logicalName);
}
```

fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
cdeszaq: As long as AvailabilityTestingUtils is a Groovy class, it should be able to see it. IntelliJ may _warn_ you (or pretend that it can't see it), but Groovy ignores it.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/Availability.java
cdeszaq: In general, using more generic types is best, but for this use-case, I think using the SIL is probably fine.

fili-core/src/main/java/com/yahoo/bard/webservice/web/SlicesApiRequest.java
cdeszaq: By "less painful tests" I mean "tests that work on more than 1 system". The behavior we've seen is that passing things through a `HashSet` allows for ordering to be different on different platforms / machines, etc. and the tests that it breaks are often test where we're comparing an expected output JSON with an API response, and it's ordering of things within those JSON blobs that matters for comparison.

So, while I agree that we should generally not let the test needs force interface direction, consistent ordering is one of those places that we've not had much choice. If Java had an `OrderedSet` interface, we'd use that in a heart-beat, but it doesn't 😢

garyluoex: I am trying to push towards the conclusion that we do have a choice. [this](https://github.com/yahoo/fili/blob/master/fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/BaseDataServletComponentSpec.groovy#L120) thing exists for this kind of reason right? If ordering matters for a component, the test should be written accordingly to enforce ordering. If ordering does not matter, the tests should be written in a way that order does not matter. The ordering is a part of the test right?

garyluoex: Or are we just trying to delay the work to fix this problem since we are loaded with too much work right now?

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
cdeszaq: Still needs addressing

fili-core/src/main/java/com/yahoo/bard/webservice/data/PartialDataHandler.java
cdeszaq: We should leave this method as-is for now I think. We can come back to it later if needed.

fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
cdeszaq: Still needs addressing

fili-core/src/main/java/com/yahoo/bard/webservice/table/ConcretePhysicalTable.java
cdeszaq: We should put this back to `Iterable`. An `Iterable` is a smaller interface than `Set` and is certainly smaller than `Collection`. The `Set` contract (ie. that there is only 1 of each column) is not needed, and the fact that a `PhysicalTable` will hold a set of columns is part of the fact that a `Table` has a `Schema`. The general property of retaining order is a good one to keep, for the same reason that unsorted arrays are harder to deal with than sorted or ordered arrays. If it's minimal extra work to keep things ordered, it's generally best to do so.

fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseSchema.java
```diff
@@ -29,6 +30,13 @@ protected BaseSchema(Granularity granularity, Iterable<Column> columns) {
     }
 
     @Override
+    public Optional<Column> getColumn(String columnName) {
+        return getColumns().stream()
+                .filter(column -> column.getName().equals(columnName))
+                .findFirst();
+    }
```
garyluoex: I will throw an exception to it, since we only support druid right now so it should be fine for now. If not at least we hit an exception.

fili-core/src/main/java/com/yahoo/bard/webservice/table/ConcretePhysicalTable.java
```diff
@@ -20,25 +22,51 @@
...
+     * @param columns  The columns for this table
+     * @param logicalToPhysicalColumnNames  Mappings from logical to physical names
+     * @param availability  Availability that serves interval availability for columns
+     */
+    public ConcretePhysicalTable(
+            @NotNull TableName name,
+            @NotNull ZonedTimeGrain timeGrain,
+            @NotNull Set<Column> columns,
+            @NotNull Map<String, String> logicalToPhysicalColumnNames,
+            @NotNull Availability availability
```
cdeszaq: This should restrict this to a `ConcreteAvailability` to ensure type safety, right?

fili-core/src/main/java/com/yahoo/bard/webservice/table/ConcretePhysicalTable.java
cdeszaq: Still needs to be addressed.

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -39,12 +45,11 @@
     DateTime getTableAlignment();
 
     /**
-     * Get the schema for this physical table.
-     * Schemas contain granularity and column definitions.
+     * Getter for all possible column available intervals.
      *
-     * @return A physical table schema.
+     * @return tableEntries map of column to set of available intervals
```
cdeszaq: This comment doesn't make sense... what's `tableEntries`?

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -39,12 +45,11 @@
     DateTime getTableAlignment();
 
     /**
-     * Get the schema for this physical table.
-     * Schemas contain granularity and column definitions.
+     * Getter for all possible column available intervals.
```
cdeszaq: Drop `possible`. There _could_ be data from the year 5billion, so this isn't "possible", but rather a "Getter for all of the available intervals for the columns in this table"

fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseSchema.java
cdeszaq: This has nothing to do with Druid, it has to do with valid states of an object, and what it means to only get the 1st one if there are multiple columns somehow. 

The extra work of going through all the columns to see if there _may_ be others every time a single column is requested seems like massive overhead (ie. if we want that constraint, we should enforce it on construction), and throwing an exception is rather drastic behavior for what seems like an otherwise simple method (to the callers).

Since it's only used in tests, and not production code, and there are clearly more questions than answers here, I'd still vote for simply removing it.

fili-core/src/main/java/com/yahoo/bard/webservice/web/SlicesApiRequest.java
cdeszaq: It's amazingly hard to write tests involving non-trivial JSON objects where some parts are order-agnostic, while others are order-sensitive. If you can think of a simple solution, please do share it because it would save so much pain.

Without that, however, the more straight-forward solution is to be generally stricter around enforcing ordering outbound. Given how much pain these issues have caused, and the difficulty in debugging them (since it's different behavior on different hosts), I'm not willing to risk dropping the ordering if there isn't a compelling reason, especially for already existing interfaces.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/Availability.java
cdeszaq: Still needs addressing

cdeszaq: Specifically, this (currently) goes to 2 places: The slices endpoint (so, back into the response ordering world), and the alignment determination. Currently, alignment determination _isn't_ order-aware, but keeping it a List means that it _could_ be. So, not the strongest arguments, but definitely ones that matter.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Override
+    public Set<TableName> getDataSourceNames() {
+        return cachedDataSourceNames;
+    }
+
+    @Override
+    public Map<Column, Set<Interval>> getAllAvailableIntervals() {
+
+        Map<String, Set<Interval>> allAvailableIntervals = metadataService.getAvailableIntervalsByTable(name);
```
cdeszaq: `metadataService.getAvailableIntervalsByTable(name)` is a common piece that's shown up more than once in this class, and while it's simple logic, it's probably worth pulling into a helper method. I think it's also one of the things that one of the follow-on PRs also does, making the case to make it a helper even stronger.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -0,0 +1,92 @@
...
+
+        if (requestColumns.isEmpty()) {
+            return new SimplifiedIntervalList();
+        }
+
+        Map<String, Set<Interval>> allAvailableIntervals = metadataService.getAvailableIntervalsByTable(name);
+        return new SimplifiedIntervalList(
+                requestColumns.stream()
+                        .map(columnName -> allAvailableIntervals.getOrDefault(columnName, Collections.emptySet()))
+                        .reduce(null, IntervalUtils::getOverlappingSubintervals)
```
cdeszaq: We should add a comment here about why this won't NPE. Since the `reduce` will return `null` if (somehow) `requestColumns` is empty, there's an NPE risk, but as-written, this method doesn't have it, and it's not immediately clear why that's true, so we should make it more clear.

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/SchemaPhysicalTableMatcher.java
```diff
@@ -56,11 +56,22 @@ public NoMatchFoundException noneFoundException() {
...
+                )
         );
         return new NoMatchFoundException(
-                MESSAGE_FORMAT.format(logicalTableName, dimensions, logicalMetrics, grainName)
+                MESSAGE_FORMAT.format(
+                        logicalTableName,
+                        dimensions,
+                        logicalMetrics,
+                        grainName
+                )
```
cdeszaq: These can both unwrap

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/ConcreteAvailabilitySpec.groovy
```diff
@@ -0,0 +1,102 @@
...
+                ])
+        )
+    }
+
+    def "getAllAvailability returns the correct availabilities for each columns configured to the table"() {
+        expect:
+        concreteAvailability.getAllAvailableIntervals() == [
+                (column1): [interval1].toSet(),
+                (column2): [interval2].toSet(),
+                (column3): [].toSet(),
```
cdeszaq: /NotABlocker

s/`.toSet()`/` as Set`

fili-core/src/test/java/com/yahoo/bard/webservice/application/TestBinderFactory.java
```diff
@@ -172,7 +176,13 @@ public ConfigurationLoader getConfigurationLoader() {
      * @return the datasource metadata service
      */
     public DataSourceMetadataService getDataSourceMetaDataService() {
-        return buildDataSourceMetadataService();
+        return new TestDataSourceMetadataService();
+    }
+
+    @Override
+    protected DataSourceMetadataService getDataSourceMetadataService() {
+        this.dataSourceMetadataService = new TestDataSourceMetadataService();
```
cdeszaq: We should only have 1 of these methods, not both (likely the latter one, not the former)

fili-core/src/test/java/com/yahoo/bard/webservice/application/TestBinderFactory.java
```diff
@@ -84,7 +86,9 @@ public void loadMetricDictionary(MetricDictionary dictionary) {
         };
         tableLoader = new TableLoader() {
             @Override
-            public void loadTableDictionary(ResourceDictionaries dictionaries) {
+            public void loadTableDictionary(
+                    ResourceDictionaries dictionaries
+            ) {
```
cdeszaq: Can unwrap I think

fili-core/src/test/java/com/yahoo/bard/webservice/application/TestBinderFactory.java
```diff
@@ -201,7 +211,11 @@ protected ConfigurationLoader buildConfigurationLoader(
             TableLoader tableLoader
     ) {
         // Store the config loader so that we can get access to it, and then return it
-        this.configurationLoader = super.buildConfigurationLoader(dimensionLoader, metricLoader, tableLoader);
+        this.configurationLoader = super.buildConfigurationLoader(
+                dimensionLoader,
+                metricLoader,
+                tableLoader
+        );
```
cdeszaq: Can unwrap?

CHANGELOG.md
```diff
@@ -48,15 +50,18 @@ Current
 - [Support timeouts for lucene search provider](https://github.com/yahoo/fili/pull/183)
 
 ### Changed:
-- [Refactor DatasourceMetadataService to fit composite table needs](https://github.com/yahoo/fili/pull/173)
+- [CompositePhsyicalTable Core Components Refactor](https://github.com/yahoo/fili/pull/179)
+    * `ConfigurationLoader` now takes an additional constructor argument `DataSourceMetadataService` for creating tables
```
cdeszaq: This change isn't accurate any more. The change is actually to the constructor of `BaseTableLoader`.

fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
garyluoex: If we log it in schema, we lose the ability to report the table name in the log. Right?

fili-core/src/main/java/com/yahoo/bard/webservice/table/ConcretePhysicalTable.java
garyluoex: Blows up because `getTableAlignment` calls `getAllAvailableIntervals` which will in turn go to `DatasourceMetadataservice` to look for a non-existing datasource for segment data which will throw an exception directly.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
```diff
@@ -310,7 +295,8 @@ protected PhysicalTable buildPhysicalTable(
                 definition.getName(),
                 definition.getGrain(),
                 columns,
-                definition.getLogicalToPhysicalNames()
+                definition.getLogicalToPhysicalNames(),
+                metadataService
```
garyluoex: Well... the sub-class should simply call `super()` with their own `metadataService`?

fili-core/src/main/java/com/yahoo/bard/webservice/table/ConcretePhysicalTable.java
garyluoex: Well, that would force people to extend `ConcreteAvailability` is this ok?

fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
cdeszaq: Ahhhh, I see. I _completely_ missed that before! good catch!

So, yes, this warning is in the right place then 😄

fili-core/src/main/java/com/yahoo/bard/webservice/table/ConcretePhysicalTable.java
cdeszaq: Thanks for hunting down what's happening. I'm OK not tackling it right now, but we should add a techdebt issue that explains the change we made, why we made it, and what we should do to fix it.

To fix this, we should put the `getTableAlignment()` back into the toString for ConcretePhysicalTable, and then fix up the ClassScanner so that it's able to build a valid instance of all of the things in that chain you walked through so that it works.

fili-core/src/main/java/com/yahoo/bard/webservice/table/ConcretePhysicalTable.java
cdeszaq: Yes, I think so. I can't think of a reason for someone to have a ConcretePhysicalTable that does not also have a ConcreteAvailability, since the _primary_ contract of a ConcreteAvailability is that it is providing availability info for a _single_ backing data source.

archolewa: I'd be more willing to buy that if there was something special that `ConcreteAvailability` contributes that `Availability` doesn't have. However, that's not the case. Therefore, by making it a `ConcreteAvailability`, you're _forcing_ people to extend a class (which you can only extend one of) vs _implementing_ an interface (which people can implement many of) if they want to add their own custom availability rules.

So I would argue that making it a `ConcreteAvailability` is making it the framework more rigid than it needs to be.

cdeszaq: @archolewa There actually _is_ an additional contract that a `ConcreteAvailability` offers over an `Availability`, but it's not currently used (and perhaps it should be, and may be indicating a deeper problem). A `ConcretePhysicalTable` has this same contract extension, so let's look at what that is:

A `ConcretePhysicalTable` exposes the additional `getFactTableName`, which is used by the `DataSourceMetadataLoader` and _should_ also be used by the `TableDataSource` when we build one of those for the query. This additional method, which is there to not only indicate that there is, but also to _simplify_ retrieval of, the name of the 1 and only 1 backing data source. Non-concrete physical tables cannot make this assertion, and it's an important one to have (1 vs. many).

Now, the possibly _smelly_ part of the current implementation is that `ConcretePhysicalTable` uses _it's_ name as the name of it's backing data source, rather than getting it from it's contained `ConcreteAvailability`. This is probably incorrect (ie. it _should_ call down to the concrete availability to get it), and the following use case illustrates it:

Consider that we have a single Druid data source that contains both hourly and daily data, where it's hourly for the 1st 30 days, and then switches to daily from then on. If we wanted to be able to have Fili's table selection logic see that table as both an hourly table _and_ a daily table, we need to have 2 different physical tables. Here's where the problem arises: Only _one_ of those two tables (based on the current implementation) can have it's name match the name of Druid's data source. In order for both tables to have the same underlying data source, the only way to get there would be to have it sourced from the same `ConcreteAvailability`, which would need to have the equivalent singleton method.

So, given the above, not only should we _keep_ `ConcreteAvailability`, but we should also fix the abstraction that `ConcretePhysicalTable` is giving us, by making it source it's `factTableName` from it's underlying `ConcreteAvailability`.

<sub>P.S. Sorry for the short saga, I'm in a storytelling mood I guess 😉 </sub>

PR #180, opened 2017-02-22
Remove duplicate method in ErrorMessageFormat

QubitPi:
Since `format(Object...values)` in `ErrorMessageFormat` is doing the same thing that the default method in `MessageFormatter` interface is doing, we drop the method


cdeszaq:
👍

michael-mclawhorn:
👍

michael-mclawhorn:
👍

PR #183, opened 2017-02-28
Support timeouts for lucene search provider

aklish:
…ger'

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
@@ -10,6 +10,7 @@ import com.yahoo.bard.webservice.data.dimension.DimensionField;
 import com.yahoo.bard.webservice.data.dimension.DimensionRow;
 import com.yahoo.bard.webservice.data.dimension.KeyValueStore;
 import com.yahoo.bard.webservice.data.dimension.SearchProvider;
+import com.yahoo.bard.webservice.data.dimension.TimeoutException;
 import com.yahoo.bard.webservice.logging.RequestLog;
 import com.yahoo.bard.webservice.util.DimensionStoreKeyUtils;
 import com.yahoo.bard.webservice.util.Pagination;
@@ -36,6 +37,7 @@ import org.apache.lucene.search.PrefixQuery;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.TermQuery;
+import org.apache.lucene.search.TimeLimitingCollector;
 import org.apache.lucene.search.TopDocs;
 import org.apache.lucene.search.WildcardQuery;
 import org.apache.lucene.store.Directory;
@@ -565,6 +567,8 @@ public class LuceneSearchProvider implements SearchProvider {
         TreeSet<DimensionRow> filteredDimRows;
         int documentCount;
         initializeIndexSearcher();
+        LOG.trace("Lucene Query {}", query);
+
         lock.readLock().lock();
         try {
             RequestLog.startTiming("QueryingLucene");
@@ -676,6 +680,9 @@ public class LuceneSearchProvider implements SearchProvider {
             String errorMessage = "Unable to find dimension rows for page " + currentPage;
             LOG.error(errorMessage);
             throw new RuntimeException(errorMessage);
+        } catch (TimeLimitingCollector.TimeExceededException e) {
+            LOG.warn("Lucene query timeout: {}. {}", query, e.getMessage());
+            throw new TimeoutException(e.getMessage());
         } finally {
             lock.readLock().unlock();
         }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
@@ -2,6 +2,8 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.data.dimension.impl;
 
+import com.yahoo.bard.webservice.config.SystemConfig;
+import com.yahoo.bard.webservice.config.SystemConfigProvider;
 import com.yahoo.bard.webservice.data.cache.HashDataCache.Pair;
 import com.yahoo.bard.webservice.data.dimension.Dimension;
 import com.yahoo.bard.webservice.data.dimension.DimensionField;
@@ -29,22 +31,27 @@ import org.apache.lucene.index.IndexWriterConfig;
 import org.apache.lucene.index.Term;
 import org.apache.lucene.search.BooleanClause;
 import org.apache.lucene.search.BooleanQuery;
+import org.apache.lucene.search.CollectorManager;
 import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.MatchAllDocsQuery;
 import org.apache.lucene.search.PrefixQuery;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.TermQuery;
+import org.apache.lucene.search.TimeLimitingCollector;
 import org.apache.lucene.search.TopDocs;
+import org.apache.lucene.search.TopScoreDocCollector;
 import org.apache.lucene.search.WildcardQuery;
 import org.apache.lucene.store.Directory;
 import org.apache.lucene.store.MMapDirectory;
+import org.apache.lucene.util.Counter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.nio.file.Paths;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
@@ -68,6 +75,11 @@ public class LuceneSearchProvider implements SearchProvider {
     private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
     private final String luceneIndexPath;
 
+    private static final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance();
+    public static final int LUCENE_SEARCH_TIMEOUT_MS = SYSTEM_CONFIG.getIntProperty(
+            SYSTEM_CONFIG.getPackageVariableName("lucene_search_timeout_ms"), 600000
+    );
+
     /**
      * The maximum number of results per page.
      */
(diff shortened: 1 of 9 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -648,11 +674,58 @@ private TopDocs getPageOfData(
...
+            public AccessibleTimeLimitingCollector newCollector() throws IOException {
+                return new AccessibleTimeLimitingCollector(
+                        TopScoreDocCollector.create(perPage, lastEntry), Counter.newCounter(false), searchTimeout
+                );
+            }
+
+            @Override
+            public TopDocs reduce(Collection<AccessibleTimeLimitingCollector> collectors) throws IOException {
+                final TopDocs[] docs = collectors.stream()
+                        .map((collector) -> collector.getWrappedCollector().topDocs())
```
cdeszaq: Can be decomposed into two method reference-based mappings:

```java
.map(AccessibleTimeLimitingCollector::getWrappedCollector)
.map(TopDocsCollector::topDocs)
```

Decomposing allows for more flexible streams and reduces complexity by making each step simpler.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -648,11 +674,58 @@ private TopDocs getPageOfData(
...
+            @Override
+            public AccessibleTimeLimitingCollector newCollector() throws IOException {
+                return new AccessibleTimeLimitingCollector(
+                        TopScoreDocCollector.create(perPage, lastEntry), Counter.newCounter(false), searchTimeout
+                );
+            }
+
+            @Override
+            public TopDocs reduce(Collection<AccessibleTimeLimitingCollector> collectors) throws IOException {
+                final TopDocs[] docs = collectors.stream()
```
cdeszaq: `final` is not needed here

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -648,11 +674,58 @@ private TopDocs getPageOfData(
             int perPage,
             int currentPage
     ) {
+        /**
+         * Works around the lack of a way to get at the underlying Collector in TimeLimitingCollector
+         */
+        class AccessibleTimeLimitingCollector extends TimeLimitingCollector {
+            private TopScoreDocCollector wrappedCollector;
```
cdeszaq: Would be nice to make this `final` to better provide immutability where we can.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -648,11 +674,58 @@ private TopDocs getPageOfData(
...
+             */
+            public TopScoreDocCollector getWrappedCollector() {
+                return wrappedCollector;
+            }
+        }
+
+        /**
+         * This parallelizes the lucene search - each thread using a time limited collector.
+         */
+        final CollectorManager<AccessibleTimeLimitingCollector, TopDocs> manager
```
cdeszaq: `final` not needed here.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -648,11 +674,58 @@ private TopDocs getPageOfData(
...
+            @Override
+            public TopDocs reduce(Collection<AccessibleTimeLimitingCollector> collectors) throws IOException {
+                final TopDocs[] docs = collectors.stream()
+                        .map((collector) -> collector.getWrappedCollector().topDocs())
+                        .collect(Collectors.toList())
+                        .toArray(new TopDocs[collectors.size()]);
+
+                return TopDocs.merge(perPage, docs);
+            }
+        };
```
cdeszaq: It would be great to pull this anonymous class (and the helper wrapper class) out into it's own stand-alone class, for a number of reasons.

1. I think the capability is generally useful, especially if another part of the application needs to deal with Lucene in a different way. Having time-bounded searches is generally a good idea
2. Having these classes defined in-line in the method will incur more performance overhead, since a new class is generated every method call, rather than in a more static manner that will happen once per class-load
3. The classes are very hard to unit-test when they are embedded in the method. Making them top-level makes them much easier to test on their own as a small, focused unit with little environment / setup overhead
4. The `LuceneSearchProvider` is already rather large. Adding more responsibilities to it adds more complexity, but having this responsibility (ie. time-bounding TopDoc searches) encapsulated into it's own top-level class keeps the complexity better in-check

It probably makes sense to have a single top-level `TimeLimitingTopDocsCollectorManager` class, with the `AccessibleTimeLimitingCollector` as an inner-class, since it's pretty tightly associated with time-limiting top docs.

I would also leave the `searchTimeout` in the `LuceneSearchProvider`, and pass it to the `TimeLimitingTopDocsCollectorManager` as a constructor param, since we've generally found it better to keep the interactions with `SystemConfig` as high in the stack as we can to reduce coupling and keep responsibilities small for pieces low in the stack.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -648,11 +674,58 @@ private TopDocs getPageOfData(
...
+
         lock.readLock().lock();
         try {
-            return lastEntry == null ?
-                    indexSearcher.search(query, perPage) :
-                    indexSearcher.searchAfter(lastEntry, query, perPage);
+            return indexSearcher.search(query, manager);
```
archolewa: It looks like we're killing pagination here. Is that intentional?

archolewa:
After talking to @aklish off line, it looks like we're not (hopefully) breaking pagination, the pagination is just being moved into the collectors that are being built and merged.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
aklish: I think it makes sense to pull out the non-anonymous class since it is named.  But I think the anonymous class makes more sense to leave inside the function.

Pulling out the anonymous class means it has to be named and it also pollutes the list of nouns/classes in the source tree (increasing the mental model of that package).

I think if someone else wants this functionality, it can be refactored at that time.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
aklish: It is actually handled in the newCollector function.  That function constructs a collector with `lastEntry` as an argument.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
aklish: One more point about this, the anonymous class is actually using closure to take arguments from the parent context.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
cdeszaq: Yeah, I looked closely at this too, and the code we have here is almost exactly the same logic that the TopN collector is doing internally. The difference is that, by using a manager, we're actually parallelizing the work across more than 1 thread and the reducing down, which we were not doing before, so things _should_ be faster (and more concretely bounded).

The only worry I have is that having more disjoint thread pools (ie. Lucene will use it's own for this parallelism) makes it harder to tune things, since you can have more contention for processors more easily, but if we see that, we can deal with it.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/DataServlet.java
```diff
@@ -85,6 +74,15 @@
...
 import javax.ws.rs.core.UriInfo;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+
+import static com.yahoo.bard.webservice.web.handlers.workflow.DruidWorkflow.REQUEST_WORKFLOW_TIMER;
+import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
+import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+import static javax.ws.rs.core.Response.Status.REQUEST_TIMEOUT;
```
archolewa: Looks like your import configuration doesn't agree with ours. The imports should be static, then `com.yahoo` then `com` then `org` then other, then `java`, then `javax`.

archolewa:
A small nitpick: It'd be nice if we could add a bit of documentation to the `SearchProvider` interface that explains that the `getDimensionsPaged` method may throw the `TimeoutException`.  For one, it'll help people implementing future `SearchProviders` know that timing out is a thing they can do.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProviderManager.java
```diff
@@ -7,11 +7,10 @@
 import com.yahoo.bard.webservice.util.Utils;
 import com.yahoo.bard.webservice.web.util.PaginationParameters;
 
+import javax.validation.constraints.NotNull;
 import java.util.HashMap;
 import java.util.Map;
 
-import javax.validation.constraints.NotNull;
-
```
archolewa: Should probably undo this change as well.

archolewa:
👍 Once the changes to the import orders that I pointed out have been addressed.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/DataServlet.java
```diff
@@ -448,6 +446,9 @@ public void getData(
         } catch (NoMatchFoundException e) {
             LOG.info("Exception processing request", e);
             asyncResponse.resume(RequestHandlerUtils.makeErrorResponse(INTERNAL_SERVER_ERROR, e, writer));
+        } catch (TimeoutException e) {
+            LOG.info("Exception processing request", e);
+            asyncResponse.resume(RequestHandlerUtils.makeErrorResponse(REQUEST_TIMEOUT, e, writer));
```
aklish: I think 408 is actually what we want here.  A 408:
```The server did not receive a complete request message within the time that it was prepared to wait```

PR #190, opened 2017-03-06
Add PermissiveAvailability & PermissiveConcretePhysicalTable

QubitPi:


garyluoex:
Need `CHANGELOG.md`, also add a test to test `PermissiveConcretePhysicalTable` correctly creates `PermissiveAvailability` and that the `getAvailableIntervals` and `getAllAvailableIntervals` uses the `PermissiveAvailability` appropriately.

fili-core/src/main/java/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTable.java
```diff
@@ -0,0 +1,35 @@
...
+     * @param columns  The columns for this table
+     * @param logicalToPhysicalColumnNames  Mappings from logical to physical names
+     * @param availability the instance of Availability specifying the availabilities of table columns
+     */
+    protected PermissiveConcretePhysicalTable(
+            @NotNull TableName name,
+            @NotNull ZonedTimeGrain timeGrain,
+            @NotNull Set<Column> columns,
+            @NotNull Map<String, String> logicalToPhysicalColumnNames,
+            @NotNull Availability availability
```
cdeszaq: This is the exact same constructor as exists in `ConcretePhysicalTable`, and there are otherwise no methods in this class that are not directly from the parent. What added value is this class giving us?

As a minimum, I would expect that the constructor here would require a `PermissiveAvailability`, rather than _any_ `Availability`, otherwise there's no assertion that this `PhysicalTable` is going to do anything "permissive".

fili-core/src/main/java/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTable.java
```diff
@@ -0,0 +1,35 @@
...
+import com.yahoo.bard.webservice.data.config.names.TableName;
+import com.yahoo.bard.webservice.data.time.ZonedTimeGrain;
+import com.yahoo.bard.webservice.table.availability.Availability;
+
+import javax.validation.constraints.NotNull;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * An instance of permissive Physical table that is backed by a single fact table.
```
cdeszaq: Unless a class is set up as a singleton, it isn't really an _instance_. It may be an _implementation_, which is perhaps what you meant?

Since this is a specialization of a `ConcretePhysicalTable`, the JavaDoc on this class should talk about what is different about this implementation. _Why_ does this class exist, what is it for, what concepts does it add on top of what `ConcretePhysicalTable` already does.

To put it a different way, merely adding (essentially) the class name to the javadoc of the parent class doesn't really make this a valuable / useful comment. 😄

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -0,0 +1,75 @@
...
+
+import javax.validation.constraints.NotNull;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * An availability which doesn't force immutability on its contents.
```
cdeszaq: Is it really the case that "permissive" means "mutable"? My understanding was that "permissive" in this context has more to do with something related to how intervals from multiple columns were counted, to determine if all columns were available or not. It's entirely possible I'm misunderstanding something, but if mutability is really the question here, I think we should rename this class to better reflect that.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -0,0 +1,75 @@
...
+            @NotNull DataSourceMetadataService metadataService
+    ) {
+        this.name = tableName;
+        this.columns = columns;
+        this.metadataService = metadataService;
+    }
+
+    @Override
+    public Set<TableName> getDataSourceNames() {
+        return Collections.singleton(name);
```
cdeszaq: Rather than make this wrapping every time `getDataSourceNames` is called, we should do it only once in the constructor and store the collection.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,80 @@
+// Copyright 2016 Yahoo Inc.
```
cdeszaq: update

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,80 @@
...
+
+        column1 = new Column('column1')
+        column2 = new Column('column2')
+
+        interval1 = new Interval('2010-01-01/2020-12-31')
+        interval2 = new Interval('2017-01-01/2020-12-31')
+
+        permissiveAvailability = new PermissiveAvailability(
+                TableName.of('table'),
+                [column1, column2].toSet(),
```
cdeszaq: In Groovy, it's generally more idiomatic to use the coercion operator: `[col1, col2] as Set`

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,80 @@
...
+                ])
+        )
+    }
+
+    def "getDataSourceNames returns the correct table name"() {
+        expect:
+        permissiveAvailability.dataSourceNames == Collections.singleton(TableName.of('table'))
+    }
+
+    def "getAllAvailability returns the correct availabilities for each columns configured to the table"() {
```
cdeszaq: "correct" in test names is generally not as useful as writing out what it means to _be correct_.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,80 @@
...
+        expect:
+        permissiveAvailability.getAllAvailableIntervals() == [
+                (column1): [interval1].toSet(),
+                (column2): [interval2].toSet()
```
cdeszaq: again, using `as Set` is generally better

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,80 @@
...
+        ] as LinkedHashMap
+    }
+
+    def "getAvailableIntervals returns the intersection of the requested column available intervals"() {
+        given:
+        DataSourceConstraint dataSourceConstraint = Mock(DataSourceConstraint)
+        dataSourceConstraint.getAllColumnNames() >> ['column1', 'column2']
+
+        expect:
+        permissiveAvailability.getAvailableIntervals(dataSourceConstraint) == new SimplifiedIntervalList([interval1])
```
cdeszaq: This test seems to be wrong. `interval1` is _not_ the _intersection_ of the intervals for the two columns, it's the union.

Also, additional test cases for _many_ more overlap / non-overlap cases would be good to include here. In particular:
- full overlap (`start/end, start/end`)
- 0 overlap (`-10/-1, 0/10`)
- 0 overlap abutting (`-10/0, 0/10`)
- partial front overlap (`0/10, 5/15`)
- partial back overlap (`0/10, -5/5`)
- full front overlap (`0/10, 5/10`)
- full back overlap (`0/10, 0/5`)

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    def "getAvailableInterval returns empty interval if given column not configured"() {
```
cdeszaq: not configured _where_? It's unclear because the setup clearly configures it _somewhere_.

cdeszaq: Also need to run this same check for `getAllAvailableIntervals`, since it's behavior should be similar.

Or rather, it _shouldn't_ include columns in the underlying metadata that this availability isn't configured for.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,80 @@
...
+        given:
+        DataSourceConstraint dataSourceConstraint = Mock(DataSourceConstraint)
+        dataSourceConstraint.getAllColumnNames() >> ['ignored']
+
+        expect:
+        permissiveAvailability.getAvailableIntervals(dataSourceConstraint) == new SimplifiedIntervalList()
+    }
+
+    def "getAvailableInterval returns empty interval if given empty column request"() {
```
cdeszaq: Is this the correct behavior? Should this throw an exception instead? It's not clear to me what's right, since "all data is available if I don't need data from any columns" seems like it _might_ be a reasonable response in this case.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
QubitPi: Right, `PermissiveAvailability` in this case should return the union. This is how I understand "Concrete"(intersection) and "Permissive"(union) availavilities.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
QubitPi: This `ignored` interval is not associated with any column in setup. I changed `ignored` to `notConfiguredColumn`

QubitPi: I think this check is done in the "getAllAvailability returns the union of all availabilities for each columns configured to the table" test?

fili-core/src/main/java/com/yahoo/bard/webservice/table/ConcretePhysicalTable.java
```diff
@@ -42,6 +43,31 @@ public ConcretePhysicalTable(
...
+     * @param availability the instance of Availability specifying the availabilities of table columns
+     */
+    protected ConcretePhysicalTable(
+            @NotNull TableName name,
+            @NotNull ZonedTimeGrain timeGrain,
+            @NotNull Set<Column> columns,
+            @NotNull Map<String, String> logicalToPhysicalColumnNames,
+            @NotNull Availability availability
+    ) {
+        super(
```
garyluoex: Since we have this constructor now, can you change the `super` in the other constructor to `this`? Thanks.

fili-core/src/main/java/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTable.java
```diff
@@ -0,0 +1,40 @@
...
+public class PermissiveConcretePhysicalTable extends ConcretePhysicalTable {
+    /**
+     * Create a permissive concrete physical table.
+     *
+     * @param name  Name of the physical table as TableName
+     * @param timeGrain  time grain of the table
+     * @param columns  The columns for this table
+     * @param logicalToPhysicalColumnNames  Mappings from logical to physical names
+     * @param permissiveAvailability the instance of Permissive Availability which makes sure that
+     *                               the table is backed by permissive availability
```
garyluoex: Get rid of the indention on the second line here, use intellij's auto format will help.

fili-core/src/main/java/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTable.java
```diff
@@ -0,0 +1,40 @@
...
+     */
+    protected PermissiveConcretePhysicalTable(
```
garyluoex: This should be public and does not take a `permissiveAvailability` as its argument, it should construct the availability and pass it to `super`.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * An availability which allows multiple disconnected intervals(union) on its contents.
```
garyluoex: Instead of saying multiple disconnected intervals, just say missing intervals might be more appropriate.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+public class PermissiveAvailability implements Availability {
+
+    private final TableName tableName;
```
garyluoex: This is not needed if we are going to store a `Set<TableName> dataSourceNames`

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -0,0 +1,84 @@
...
+     */
+    public PermissiveAvailability(
+            TableName tableName,
+            @NotNull Set<Column> columns,
+            @NotNull DataSourceMetadataService metadataService
+    ) {
+        this.tableName = tableName;
+        this.columns = columns;
+        this.metadataService = metadataService;
+        dataSourceNames = Collections.singleton(tableName);
```
garyluoex: Add a `this.` to be consistent with everything else so that it looks nicer.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param tableName The name of the data source
+     * @param columns A set of columns associated with the data source
+     * @param metadataService A service containing the data source segment data
+     */
+    public PermissiveAvailability(
+            TableName tableName,
```
garyluoex: Add a `@NotNull` here so that we won't be asking `DataSourceMetadataService with` a `null` table.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -0,0 +1,84 @@
...
+                        )
+                );
+    }
+
+    @Override
+    public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constraints) {
+        Set<String> allColumnNames = constraints.getAllColumnNames();
+
+        if (allColumnNames.isEmpty()) {
+            throw new IllegalArgumentException("DataSourceConstraint needs to have at least 1 constraint column");
```
garyluoex: Follow the behavior if things are empty in `ConcreteAvailability`, we do not want different behavior for each Availability.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+
+        Map<Column, Set<Interval>> x = getAllAvailableIntervals();
```
garyluoex: `x` is not acceptable.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -0,0 +1,84 @@
...
+        dataSourceNames = Collections.singleton(tableName);
+    }
+
+    @Override
+    public Set<TableName> getDataSourceNames() {
+        return dataSourceNames;
+    }
+
+    @Override
+    public Map<Column, Set<Interval>> getAllAvailableIntervals() {
```
garyluoex: I don't you have to `Override` this method again here, since we can use the super class version.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -0,0 +1,84 @@
...
+            @NotNull DataSourceMetadataService metadataService
+    ) {
+        this.tableName = tableName;
+        this.columns = columns;
+        this.metadataService = metadataService;
+        dataSourceNames = Collections.singleton(tableName);
+    }
+
+    @Override
+    public Set<TableName> getDataSourceNames() {
```
garyluoex: Move this function logic to `ConcreteAvailability` along with the variable
```java
private final Set<TableName> dataSourceNames;
```
and remove the method from here.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/AvailabilityTestingUtils.groovy
```diff
@@ -62,4 +64,21 @@ class AvailabilityTestingUtils extends Specification {
...
+     *
+     * @param map a map of Column in String representation to its all available intervals in String representations
+     *
+     * @return a map of Column object to its all available intervals
+     */
+    static def constructColumnToIntervalMap(Map<String, Set<String>> map) {
+        return map.entrySet().stream()
+                .collect(
+                    Collectors.toMap(
+                            {Map.Entry it -> new Column(it.key)},
```
garyluoex: can we try remove the type cast and directly use `it`? since intellij complains anyways, if it compile and runs might as well remove the type cast.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,117 @@
...
+import java.util.stream.Collectors
+
+class PermissiveAvailabilitySpec extends Specification  {
+
+    PermissiveAvailability permissiveAvailability
+    TableName tableName
+    Column column1
+    Column column2
+    Interval interval1
+    Interval interval2
```
garyluoex: Grouping the ones with the same type into a single line make things look cleaner

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
garyluoex: The current logic is `if you don't ask for anything, you don't get anything`, consider the case where the user asks for a column not on the table in which is filtered out in the availability (although this should not be possible).

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,117 @@
...
+
+    def setup() {
+        tableName = TableName.of('table')
+
+        column1 = new Column('column1')
+        column2 = new Column('column2')
+
+        interval1 = new Interval('2010-01-01/2020-12-31')
+        interval2 = new Interval('2017-01-01/2020-12-31')
```
garyluoex: Move the setup things that are not used by more than one test into the single test that uses it.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
QubitPi: Could you explain a bit why we do this?

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
cdeszaq: Actually, I generally _hate_ lumping things onto one line. It may just be my personal preference, but it makes it harder to see how many things there are, and harder to see what changes if variable names need to change. It also makes it more difficult to initialize variables when they are defined.

It's not a big deal, but I definitely would rather have more lines, with each definition on it's own line, rather than not.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
QubitPi: Right, I forgot to remove this debug line

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
cdeszaq: And for reference, here's the relevant style guide indication for this: https://google.github.io/styleguide/javaguide.html#s4.8.2-variable-declarations

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/AvailabilityTestingUtils.groovy
cdeszaq: It's not a cast in this context, it's a type indication, but otherwise you are correct, Groovy's dynamic typing means that it's not needed here (unless we wanted it for clarity)

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
garyluoex: Follow the behavior if things are empty in `ConcreteAvailability`, we do not want different behavior for each Availability.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -53,7 +51,11 @@ public PermissiveAvailability(
 
     @Override
     public Map<Column, Set<Interval>> getAllAvailableIntervals() {
-        Map<String, Set<Interval>> allAvailableIntervals = metadataService.getAvailableIntervalsByTable(tableName);
+        Map<String, Set<Interval>> allAvailableIntervals = metadataService.getAvailableIntervalsByTable(
+                dataSourceNames.stream()
+                        .findFirst()
+                        .orElseThrow(NullPointerException::new)
```
garyluoex: This is not necessary since the constructor only allows a single name to be passed in, also throwing a null pointer exception with out any additional information is not recommended.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -68,17 +70,12 @@ public PermissiveAvailability(
     public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constraints) {
         Set<String> allColumnNames = constraints.getAllColumnNames();
 
-        if (allColumnNames.isEmpty()) {
```
garyluoex: Refer to `ConcreteAvailability` the behavior of `getAvailableIntervals` behavior, it should filter out to only the columns that is configured in this table. Should add a test to ensure this behavior.

QubitPi:
@garyluoex @cdeszaq More comments welcome :)

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTableSpec.groovy
```diff
@@ -0,0 +1,100 @@
...
+        constraints.getAllColumnNames() >> allColumnNames
+
+        expect:
+        permissiveConcretePhysicalTable.getAvailableIntervals(constraints).asList() == new SimplifiedIntervalList(expected).asList()
+
+        where:
+        allColumnNames           | expected
+        ["dimension"]            | [interval1] as Set
+        ["metric1"]              | [interval2] as Set
+        ["metric2"]              | [interval3] as Set
```
cdeszaq: Can we give these things names that more clearly indicate what sort of state they are holding or should be used for? `interval1` and `metric1` don't tell us anything about if the test is correct or not, and the reader has to connect the data to the variables, understand how it's working under the covers, and map from the test to reality all in their head.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTableSpec.groovy
```diff
@@ -0,0 +1,100 @@
...
+                        (dimensionColumn): [interval1] as Set,
+                        (metricColumn1): [interval2] as Set,
+                        (metricColumn2): [interval3] as Set,
+                        (new Column('ignored')): [new Interval('2010-01-01/2500-12-31')] as Set
+                ])
+        )
+    }
+
+    @Unroll
+    def "Physical table getAvailableIntervals returns #expected for column #allColumnNames"() {
```
cdeszaq: It would be better to have the test name indicate _why_ or under what conditions the behavior being tested is true.

For example "The spanning intervals are returned even when individual columns are not fully covering the interval" might be a good name because it's clear what the behavior is, and under what conditions.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTableSpec.groovy
```diff
@@ -0,0 +1,100 @@
...
+        DataSourceConstraint constraints = Mock(DataSourceConstraint)
+        constraints.getAllColumnNames() >> allColumnNames
+
+        expect:
+        permissiveConcretePhysicalTable.getAvailableIntervals(constraints).asList() == new SimplifiedIntervalList(expected).asList()
+
+        where:
+        allColumnNames           | expected
+        ["dimension"]            | [interval1] as Set
+        ["metric1"]              | [interval2] as Set
```
cdeszaq: Could we add some tests that would return more than a single interval?

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/AvailabilityTestingUtils.groovy
```diff
@@ -2,18 +2,20 @@
...
 
+import com.sun.corba.se.impl.encoding.OSFCodeSetRegistry
 import com.yahoo.bard.webservice.application.JerseyTestBinder
 import com.yahoo.bard.webservice.data.dimension.DimensionColumn
 import com.yahoo.bard.webservice.data.metric.MetricColumn
 import com.yahoo.bard.webservice.metadata.TestDataSourceMetadataService
 import com.yahoo.bard.webservice.table.Column
 import com.yahoo.bard.webservice.table.ConcretePhysicalTable
 import com.yahoo.bard.webservice.table.PhysicalTableDictionary
-
```
cdeszaq: Keep

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/AvailabilityTestingUtils.groovy
```diff
@@ -2,18 +2,20 @@
...
+import io.netty.util.internal.chmv8.ConcurrentHashMapV8
```
cdeszaq: Unused?

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/AvailabilityTestingUtils.groovy
```diff
@@ -2,18 +2,20 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.table.availability
 
+import com.sun.corba.se.impl.encoding.OSFCodeSetRegistry
```
cdeszaq: Unused?

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/AvailabilityTestingUtils.groovy
```diff
@@ -2,18 +2,20 @@
...
 import com.yahoo.bard.webservice.table.Column
 import com.yahoo.bard.webservice.table.ConcretePhysicalTable
 import com.yahoo.bard.webservice.table.PhysicalTableDictionary
-
+import io.netty.util.internal.chmv8.ConcurrentHashMapV8
 import org.joda.time.Interval
 
 import spock.lang.Specification
 
+import java.util.function.Function
```
cdeszaq: Unused

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/AvailabilityTestingUtils.groovy
```diff
@@ -62,4 +64,21 @@ class AvailabilityTestingUtils extends Specification {
...
     }
+
+    /**
+     * Construct a map of Column object to its all available intervals
+     *
+     * @param map a map of Column in String representation to its all available intervals in String representations
+     *
+     * @return a map of Column object to its all available intervals
+     */
+    static def constructColumnToIntervalMap(Map<String, Set<String>> map) {
```
cdeszaq: Rather than `def`, we should use proper types wherever possible. `def` should only be used where we want to leverage the dynamic typing capabilities Groovy provides.

cdeszaq: Can we give this parameter a better name? Parameter names that are nothing more than their types are not very helpful

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/AvailabilityTestingUtils.groovy
```diff
@@ -62,4 +64,21 @@ class AvailabilityTestingUtils extends Specification {
...
+        return map.entrySet().stream()
+                .collect(
+                    Collectors.toMap(
+                            {it -> new Column(it.key)},
+                            {(it.value).each {new Interval(it)}}
+                    )
```
cdeszaq: This entire method body can be simplified using Groovy's collection extensions directly:

```groovy
map.collectEntries{key, value -> [ (new Column(key)): value.collect{new Interval(it)} ]}
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,145 @@
...
+    def "getAvailableInterval returns empty interval if given column not configured"() {
+        given:
+        DataSourceConstraint dataSourceConstraint = Mock(DataSourceConstraint)
+        dataSourceConstraint.getAllColumnNames() >> ['unConfiguredColumn']
+
+        expect:
+        permissiveAvailability.getAvailableIntervals(dataSourceConstraint) == new SimplifiedIntervalList()
+    }
+
+    def "getAvailableInterval returns empty interval if given empty column request"() {
```
cdeszaq: s/getAvailableInterval/getAvailableIntervals

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,145 @@
...
+        DataSourceConstraint dataSourceConstraint = Mock(DataSourceConstraint)
+        dataSourceConstraint.getAllColumnNames() >> ['column1', 'column2', 'unConfiguredColumn']
+
+        expect:
+        permissiveAvailability.getAvailableIntervals(dataSourceConstraint) == new SimplifiedIntervalList(
+                [interval1, interval2]
+        )
+    }
+
+    def "getAvailableInterval returns empty interval if given column not configured"() {
```
cdeszaq: s/getAvailableInterval/getAvailableIntervals

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
cdeszaq: It's not an empty _interval_, it's an empty _list_ that gets returned. Those are two very different things.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
cdeszaq: It's not an empty _interval_, it's an empty _list_ that gets returned. Those are two very different things.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,145 @@
...
+        '2017-01-01/2017-02-01' | '2018-01-01/2018-02-01' | ['2017-01-01/2017-02-01', '2018-01-01/2018-02-01'] // 0 overlap (-10/-1, 0/10)
+        '2017-01-01/2017-02-01' | '2017-02-01/2017-03-01' | ['2017-01-01/2017-03-01']                          // 0 overlap abutting (-10/0, 0/10)
+        '2017-01-01/2017-02-01' | '2017-01-15/2017-03-01' | ['2017-01-01/2017-03-01']                          // partial front overlap (0/10, 5/15)
+        '2017-01-01/2017-02-01' | '2016-10-01/2017-01-15' | ['2016-10-01/2017-02-01']                          // partial back overlap (0/10, -5/5)
+        '2017-01-01/2017-02-01' | '2017-01-15/2017-02-01' | ['2017-01-01/2017-02-01']                          // full front overlap (0/10, 5/10)
+        '2017-01-01/2017-02-01' | '2017-01-01/2017-01-15' | ['2017-01-01/2017-02-01']                          // full back overlap (0/10, 0/5)
+        '2017-01-01/2017-02-01' | '2017-01-15/2017-01-25' | ['2017-01-01/2017-02-01']                          // fully contain (0/10, 3/9)
+    }
+
+    def "getAvailableInterval returns only configured column"() {
```
cdeszaq: s/getAvailableInterval/getAvailableIntervals

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,145 @@
...
+        '2017-01-01/2017-02-01' | '2017-01-01/2017-02-01' | ['2017-01-01/2017-02-01']                          // full overlap (start/end, start/end)
+        '2017-01-01/2017-02-01' | '2018-01-01/2018-02-01' | ['2017-01-01/2017-02-01', '2018-01-01/2018-02-01'] // 0 overlap (-10/-1, 0/10)
+        '2017-01-01/2017-02-01' | '2017-02-01/2017-03-01' | ['2017-01-01/2017-03-01']                          // 0 overlap abutting (-10/0, 0/10)
+        '2017-01-01/2017-02-01' | '2017-01-15/2017-03-01' | ['2017-01-01/2017-03-01']                          // partial front overlap (0/10, 5/15)
+        '2017-01-01/2017-02-01' | '2016-10-01/2017-01-15' | ['2016-10-01/2017-02-01']                          // partial back overlap (0/10, -5/5)
+        '2017-01-01/2017-02-01' | '2017-01-15/2017-02-01' | ['2017-01-01/2017-02-01']                          // full front overlap (0/10, 5/10)
+        '2017-01-01/2017-02-01' | '2017-01-01/2017-01-15' | ['2017-01-01/2017-02-01']                          // full back overlap (0/10, 0/5)
+        '2017-01-01/2017-02-01' | '2017-01-15/2017-01-25' | ['2017-01-01/2017-02-01']                          // fully contain (0/10, 3/9)
```
cdeszaq: It would cool to have these comments as a string variable and have that expanded into the test name...

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,145 @@
...
+        )
+
+        DataSourceConstraint dataSourceConstraint = Mock(DataSourceConstraint)
+        dataSourceConstraint.getAllColumnNames() >> ['column1', 'column2']
+
+        expect:
+        permissiveAvailability.getAvailableIntervals(dataSourceConstraint) == new SimplifiedIntervalList(
+                expected.stream()
+                        .map{it -> new Interval(it)}
+                        .collect(Collectors.toSet())
```
cdeszaq: Can be simplified using groovy: `expected.collect{new Interval(it)} as Set`

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,145 @@
...
+                new TestDataSourceMetadataService()
+        )
+    }
+
+    def "getDataSourceNames returns the correct table name"() {
+        expect:
+        permissiveAvailability.dataSourceNames == Collections.singleton(TableName.of('table'))
+    }
+
+    def "getAllAvailability returns all availabilities for each columns configured to the table"() {
```
cdeszaq: method name here in the description doesn't seem to match the method under test

cdeszaq: We should also have a test like this for only returning availabilities for columns that are configured on the Availability, even if they are in the underlying DataSourceMetadataService.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,145 @@
...
+import com.yahoo.bard.webservice.util.SimplifiedIntervalList
+
+import org.joda.time.Interval
+
+import spock.lang.Specification
+import spock.lang.Unroll
+
+import java.util.stream.Collectors
+
+class PermissiveAvailabilitySpec extends Specification  {
```
cdeszaq: If `PermissiveAvailability` extends from `ConcreteAvailability`, then this Spec only needs to worry about the specific differences in behavior from it's base class, which will likely simplify this Spec significantly.

That said, if the comments I've left here can make the Spec for the parent better, then they should be rolled into the Parent spec.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTableSpec.groovy
```diff
@@ -0,0 +1,100 @@
...
+
+    @Shared TableName tableName
+
+    @Shared Set<Column> columns
+
+    @Shared Dimension dimension
+    @Shared DimensionColumn dimensionColumn
+    @Shared MetricColumn metricColumn1, metricColumn2
+
+    @Shared Interval interval1, interval2, interval3
```
QubitPi: We need to access their states across tests

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTableSpec.groovy
```diff
@@ -0,0 +1,100 @@
...
+
+    def setupSpec() {
```
QubitPi: `setup` fails all tests. I guess we have to keep `@shared` and `setupSpec` for tests to pass

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
QubitPi: @cdeszaq Agree. How do I do that?

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
QubitPi: @cdeszaq This test is removed because `PermissiveAvailability` extends `ConcreteAvailability`, do we still need to add the test?

QubitPi: I think the new test is the same as `getAvailableInterval returns only configured column`

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
QubitPi: OMG. I'll never see this

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -41,13 +44,35 @@ public ConcreteAvailability(
             @NotNull DataSourceMetadataService metadataService
     ) {
         this.name = tableName;
-        this.columns = columns;
         this.metadataService = metadataService;
+        this.names = Collections.singleton(tableName);
+        this.columns = columns;
```
cdeszaq: Should turn this into an `ImmutableSet` so that callers can't modify it if they get it via `getColumns`

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -41,13 +44,35 @@ public ConcreteAvailability(
...
+        this.columnNames = columns.stream()
+                .map(Column::getName)
+                .collect(Collectors.toSet());
```
cdeszaq: Should also be made into an `ImmutableSet`, though it matters a _little_ less, since it's never returned directly. Better to be safe though.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
cdeszaq: One way to do it would be like this:

```groovy

Bar mockBar = Mock(Bar)

@Unroll
def "The #thing always does #actionCount #action whenever it is poked and #reason"() {
    given: "A configured thing and a mock"
    thing."set$initialParam"(initialState)

    when: "It gets poked"
    thing.poke()

    then: "It has taken the action the correct number of times"
    actionCount * mockBar."action"()

    where:
    thing     | actionCount | action | initialParam | initialState | reason
    new Dog() | 1           | "jump" | "Barking"    | true         | "it is barking"
}
```

This example is a bit fancy, and shows off much more than how to get a string to unroll as part of the test name, but you can see that the `reason` `where` variable is _only_ used as an unrolled parameter in the test name.

In a nutshell, you can put a `#` in front of any `where` parameter in the test name when using `@Unroll`, and the value of that variable will be swapped into the test name when Spock generates the tests (which is on for each iteration in the `where` block.

cdeszaq: Some of the other stuff that's in that example shows off how Groovy will let you use strings as method names :-)

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTableSpec.groovy
cdeszaq: Ahh, right, it's because `@Shared` is needed to be able to reference variables in the `where` block. I consider this a Spock bug :(

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
cdeszaq: More specifically, for this test, of you make the `where` block like this:

```groovy
where:
firstInterval           | secondInterval          | expected        | reason
'2017-01-01/2017-02-01' | '2017-01-01/2017-02-01' | ['2017-01-01/2017-02-01'] | "full overlap (start/end, start/end)"
'2017-01-01/2017-02-01' | '2018-01-01/2018-02-01' | ['2017-01-01/2017-02-01', '2018-01-01/2018-02-01'] | "0 overlap (-10/-1, 0/10)"
'2017-01-01/2017-02-01' | '2017-02-01/2017-03-01' | ['2017-01-01/2017-03-01'] | "0 overlap, abutting (-10/0, 0/10)"
'2017-01-01/2017-02-01' | '2017-01-15/2017-03-01' | ['2017-01-01/2017-03-01'] | "partial front overlap (0/10, 5/15)"
'2017-01-01/2017-02-01' | '2016-10-01/2017-01-15' | ['2016-10-01/2017-02-01'] | "partial back overlap (0/10, -5/5)"
'2017-01-01/2017-02-01' | '2017-01-15/2017-02-01' | ['2017-01-01/2017-02-01'] | "full front overlap (0/10, 5/10)"
'2017-01-01/2017-02-01' | '2017-01-01/2017-01-15' | ['2017-01-01/2017-02-01'] | "full back overlap (0/10, 0/5)"
'2017-01-01/2017-02-01' | '2017-01-15/2017-01-25' | ['2017-01-01/2017-02-01'] | "fully contain (0/10, 3/9)"
```

Then you can add it to the test name, like this:

```groovy
getAvailableIntervals returns the union of requested columns when available intervals have #reason
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTableSpec.groovy
```diff
@@ -0,0 +1,85 @@
...
+                        (disjointIntervalColumn)     : [disjointInterval] as Set,
+                        (leftAbuttingIntervalColumn) : [leftAbuttingInterval] as Set,
+                        (rightAbuttingIntervalColumn): [rightAbuttingInterval] as Set,
+                        (new Column('ignored'))      : [new Interval('2010-01-01/2500-12-31')] as Set
+                ])
+        )
+    }
+
+    @Unroll
+    def "The spanning intervals are returned even when individual columns are not fully covering the interval"() {
```
cdeszaq: `@Unroll` tests should have _some_ sort of `where` expansion, to differentiate the test iterations and indicate what is being tested in each iteration.

QubitPi: I though it makes more sense to remove `@Unroll` :)

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
QubitPi: Brilliant! Thanks! I'll take a note.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTableSpec.groovy
cdeszaq: If we have a data-driven test (ie. a test with a `where` block), we should _always_ use `@Unroll`. The reason is that `@Unroll` makes a test (as seen by the JUnit Test Runner) for each "row" in a `where` table. Without that, all of the iterations in the `where` block are run as a single test, so a failure with the 1st row marks the entire test as a failure, instead of allowing _each_ iteration to fail on it's own, which gives much more information.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -0,0 +1,54 @@
...
+     * com.yahoo.bard.webservice.table.availability.ConcreteAvailability#getAvailableIntervals(DataSourceConstraint)};
+     * Instead of returning the intersection of all available intervals, this method returns the union of them.
+     *
+     * @param constraints  Data constraint containing columns and api filters
+     *
+     * @return the union of all available intervals
+     */
+    @Override
+    public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constraints) {
+        return getIntervalsForColumnNames(getColumns().stream().map(Column::getName).collect(Collectors.toSet()))
```
cdeszaq: Rather than recompute this, we should pull this collection from the parent class: https://github.com/yahoo/fili/pull/190/files#diff-6923730d44e29354ccfcc8d1ca622d8eR33

We may need to add a `protected` getter for it to the parent, but that's where we should use it from.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,117 @@
...
+
+    def "getAvailableIntervals returns empty list of intervals if given column not configured"() {
+        given:
+        DataSourceConstraint dataSourceConstraint = Mock(DataSourceConstraint)
+        dataSourceConstraint.getAllColumnNames() >> ['unConfiguredColumn']
+
+        expect:
+        permissiveAvailability.getAvailableIntervals(dataSourceConstraint) == new SimplifiedIntervalList()
+    }
+
```
michael-mclawhorn: These three tests have incorrect results.  The reason it's not obvious is because you haven't initialized the metadata service they use with actual intervals.  They should be returning the full set of intervals supplied by the columns of the permissive table.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,102 @@
...
+        dataSourceConstraint.getAllColumnNames() >> ['unConfiguredColumn']
+
+        expect:
+        permissiveAvailability.getAvailableIntervals(dataSourceConstraint) == new SimplifiedIntervalList(
+                [interval1, interval2]
+        )
+    }
+
+    def "getAvailableIntervals returns empty list of intervals if given empty column request"() {
+        given:
```
michael-mclawhorn: You changed the test behavior but not the test name.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -0,0 +1,102 @@
...
+        '2017-01-01/2017-02-01' | '2017-02-01/2017-03-01' | ['2017-01-01/2017-03-01']                          | "0 overlap abutting (-10/0, 0/10)"
+        '2017-01-01/2017-02-01' | '2017-01-15/2017-03-01' | ['2017-01-01/2017-03-01']                          | "partial front overlap (0/10, 5/15)"
+        '2017-01-01/2017-02-01' | '2016-10-01/2017-01-15' | ['2016-10-01/2017-02-01']                          | "partial back overlap (0/10, -5/5)"
+        '2017-01-01/2017-02-01' | '2017-01-15/2017-02-01' | ['2017-01-01/2017-02-01']                          | "full front overlap (0/10, 5/10)"
+        '2017-01-01/2017-02-01' | '2017-01-01/2017-01-15' | ['2017-01-01/2017-02-01']                          | "full back overlap (0/10, 0/5)"
+        '2017-01-01/2017-02-01' | '2017-01-15/2017-01-25' | ['2017-01-01/2017-02-01']                          | "fully contain (0/10, 3/9)"
+    }
+
+    def "getAvailableIntervals returns empty list of intervals if given column not configured"() {
+        given:
```
michael-mclawhorn: You changed the test behavior but not the test name.

michael-mclawhorn:
Once you've got these last changes in, please rebase onto the updated UDCore3 and resolve any conflicts.

cdeszaq:
I'm pulling my approval back, since this needs a rebase onto the underlying PR, but once this is cleanly rebased, I'm happy to give this a quick glance and fast re-approval.

fili-core/src/main/java/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTable.java
```diff
@@ -0,0 +1,50 @@
...
+public class PermissiveConcretePhysicalTable extends ConcretePhysicalTable {
+    /**
+     * Create a permissive concrete physical table.
+     *
+     * @param name  Name of the physical table as TableName
+     * @param timeGrain  time grain of the table
+     * @param columns  The columns for this table
+     * @param logicalToPhysicalColumnNames  Mappings from logical to physical names
+     * @param dataSourceMetadataService data source metadata service containing availability data for the table
+     */
```
michael-mclawhorn: Spacing:  two spaces after param name.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -105,4 +105,16 @@ public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constra
...
+        return metadataService;
+    }
+
+    public TableName getName() {
+        return name;
+    }
+
+    public Set<String> getCachedColumnNames() {
+        return cachedColumnNames;
+    }
```
michael-mclawhorn: Let's drop the 'cached' prefix on the member variables in this class and in the getter.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -105,4 +105,16 @@ public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constra
...
+
+    public DataSourceMetadataService getMetadataService() {
+        return metadataService;
+    }
+
+    public TableName getName() {
+        return name;
+    }
+
+    public Set<String> getCachedColumnNames() {
```
cdeszaq: Probably should make this `protected`, since it's intended as a helper method more than anything else, right?

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/AvailabilityTestingUtils.groovy
```diff
@@ -62,4 +62,18 @@ class AvailabilityTestingUtils extends Specification {
...
+     * Construct a map of Column object to its all available intervals
+     *
+     * @param map a map of Column in String representation to its all available intervals in String representations
+     *
+     * @return a map of Column object to its all available intervals
+     */
+    static Map<Column, Set<Interval>> constructColumnToIntervalMap(Map<String, Set<String>> columnToIntervalInString) {
+        return columnToIntervalInString.collectEntries{
+            key,
+            value -> [ (new Column(key)): value.collect{new Interval(it)} ]
```
cdeszaq: Typically, format for Groovy closures is to put the closure params on the opening line, and then wrap after that:

```groovy
[:].collectEntries {key, value ->
    [(key): value]
}
```

cdeszaq:
Closing in favor of #200, which is the rebased version of this.

PR #191, opened 2017-03-08
Config slurper master

kevinhinterlong:
The config slurper allows fili to be set up in front of any instance of druid and probe druid to build metrics, dimensions, and tables.


fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericBinder.java
```diff
@@ -0,0 +1,49 @@
...
+import com.yahoo.wiki.webservice.data.config.table.GenericTableLoader;
+
+import java.util.LinkedHashSet;
+
+/**
+ * Created by kevin on 3/7/2017.
+ */
+public class GenericBinder {
+    private static GenericBinder genericBinder;
+    private DruidWebService druidWebService;
```
kevinhinterlong: This class has no longer has unused private fields

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/metric/GenericMetricLoader.java
```diff
@@ -0,0 +1,98 @@
...
+ * such as the LongSumMaker for performing the longSum aggregation, and the divisionMaker, which performs division
+ * of two other metrics.
+ */
+public class GenericMetricLoader implements MetricLoader {
+
+    public static final int BYTES_PER_KILOBYTE = 1024;
+    public static final int DEFAULT_KILOBYTES_PER_SKETCH = 16;
+    public static final int DEFAULT_SKETCH_SIZE_IN_BYTES = DEFAULT_KILOBYTES_PER_SKETCH * BYTES_PER_KILOBYTE;
+    private static final Logger LOG = LoggerFactory.getLogger(GenericMetricLoader.class);
+    final int sketchSize;
```
kevinhinterlong: This is copied directly from WikiMetricLoader so same issue would apply there

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/dimension/GenericDimensions.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Hold all the dimension configurations for the sample Bard instance.
+ */
+public class GenericDimensions {
+
+    private final SystemConfig systemConfig = SystemConfigProvider.getInstance();
+
+    private final String defaultDimensionBackendKey = systemConfig.getPackageVariableName("dimension_backend");
```
kevinhinterlong: Also copied directly from WikiDimensions

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/metric/MetricNameGenerator.java
```diff
@@ -0,0 +1,36 @@
...
+        defaultTimeGrain = t;
+    }
+
+    public static DruidMetricName getDruidMetric(String name) {
+        return new DruidMetricName(name);
+    }
+
+    public static FiliMetricName getFiliMetricName(String name) {
+        if (defaultTimeGrain == null) {
+            throw new RuntimeException("Default time grain must be specified if no time grains are given");
```
cdeszaq: This is generally an OK warning to ignore

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/WikiMain.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+    /**
+     * Run the Wikipedia application.
+     *
+     * @param args  command line arguments
+     * @throws Exception if the server fails to start or crashes
+     */
+    public static void main(String[] args) throws Exception {
```
michael-mclawhorn: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/ConfigLoader.java
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.wiki.webservice.data.config.auto;
+
+import java.util.List;
+
+/**
+ * Created by kevin on 3/3/2017.
+ */
+public interface ConfigLoader {
+    public List<? extends DruidConfig> getTableNames();
```
michael-mclawhorn: 👍

kevinhinterlong: Would you prefer that these get removed?

michael-mclawhorn: Ok, in general comments on classes are a good idea, and necessary on interfaces.

However, I'm not sure I understand the connection between this interface and it's name.  Further, there's already a class ConfigurationLoader, which I think this interface could easily be confused with.

Also, this interface appears to be a single method function.  It's possible rather than have this be an explicit interface, you could instead make anything that is currently implementing it instead to implement:

Supplier<List<? extends DruidConfig>>

https://docs.oracle.com/javase/8/docs/api/java/util/function/Supplier.html

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/StaticWikiConfigLoader.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Created by kevin on 3/7/2017.
+ */
+public class StaticWikiConfigLoader implements ConfigLoader {
+
```
michael-mclawhorn: Implements Supplier<List<DataSourceConfiguration>> 
(method name would change to get() )

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/dimension/GenericDimensionField.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * DimensionField enum.
+ */
+public enum GenericDimensionField implements DimensionField {
+    ID("Dimension ID"),
+    DESC("Dimension Description");
+
+    private String description;
+    private String camelName;
```
michael-mclawhorn: Is there any value in having description here?  I'd get rid of it.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/metric/DruidMetricName.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.Locale;
+
+/**
+ * Created by kevin on 3/7/2017.
+ */
+public class DruidMetricName implements FieldName {
+    private String lowerCaseName;
+
+    public DruidMetricName(String lowerCaseName) {
+        this.lowerCaseName = lowerCaseName.toLowerCase(Locale.ENGLISH);
```
michael-mclawhorn: clearly the incoming parameter should not be called lowerCaseName, since that's not being enforced or expected.  

Also, constructor javadoc indicated, with parameter description.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/metric/FiliMetricName.java
```diff
@@ -0,0 +1,41 @@
...
+import com.yahoo.bard.webservice.util.EnumUtils;
+
+import java.util.List;
+import java.util.Locale;
+
+/**
+ * Created by kevin on 3/7/2017.
+ */
+public class FiliMetricName implements ApiMetricName {
+    private String apiName;
```
michael-mclawhorn: Should probably be FiliApiMetricName since it extends ApiMetricName

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/metric/FiliMetricName.java
```diff
@@ -0,0 +1,41 @@
...
+    private List<TimeGrain> satisfyingGrains;
+
+    public FiliMetricName(String lowerCaseName, List<TimeGrain> timeGrains) {
+        this.apiName = EnumUtils.camelCase(lowerCaseName);
```
michael-mclawhorn: Javadoc.
Is it REALLY a lowercase name?
Should this be plural time grains?  This config currently only allows a single default grain.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/dimension/GenericDimensions.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private final Set<DimensionConfig> dimensionConfigs;
+    private final LinkedHashMap<String, DimensionConfig> wikiApiDimensionNameToConfig;
+
+    /**
+     * Construct the dimension configurations.
+     */
+    public GenericDimensions(ConfigLoader configLoader) {
+        DruidConfig config = configLoader.getTableNames().get(0);
+        this.dimensionConfigs = Collections.unmodifiableSet(
```
michael-mclawhorn: why are we taking something that supplies several DruidConfigs and then only taking the first one?
If we're only loading one, let's only pass in one.  If we're taking several, you should be looping over all the dimension configurations.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/DruidNavigator.java
```diff
@@ -0,0 +1,136 @@
...
+            LOG.info("Loaded table " + table.getName());
+        }, url);
+    }
+
+    private void loadMetrics(final TableConfig table, final JsonNode rootNode) {
+        JsonNode metricsArray = rootNode.get("metrics");
+        String[] metrics = metricsArray.asText().split(",");
+        for (String m : metrics) {
+            table.addMetric(m);
+        }
```
michael-mclawhorn: avoid overabbreviation, 'm' may as well be 'metric'

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/DruidNavigator.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    private void loadDimensions(final TableConfig tableName, final JsonNode rootNode) {
+        JsonNode dimensionsArray = rootNode.get("dimensions");
+        String[] dimensions = dimensionsArray.asText().split(",");
+        for (String d : dimensions) {
+            tableName.addDimension(d);
```
michael-mclawhorn: over-abbreviated variable 'd'

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/StaticWikiConfigLoader.java
michael-mclawhorn: Class comment

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/TableConfig.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Created by kevin on 3/3/2017.
+ */
+public class TableConfig implements DruidConfig {
+    private String tableName;
+    private List<String> metrics;
+    private List<String> dimensions;
+    private List<TimeGrain> timeGrains;
+
```
michael-mclawhorn: These probably deserve to be final.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/dimension/GenericDimensionField.java
```diff
@@ -0,0 +1,42 @@
...
+import com.yahoo.bard.webservice.data.dimension.DimensionField;
+import com.yahoo.bard.webservice.util.EnumUtils;
+
+/**
+ * DimensionField enum.
+ */
+public enum GenericDimensionField implements DimensionField {
+    ID("Dimension ID"),
+    DESC("Dimension Description");
+
```
michael-mclawhorn: I'm not sure 'DESC' is good for anything on dimensions loaded directly from druid.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/metric/DruidMetricName.java
```diff
@@ -0,0 +1,28 @@
...
+import com.yahoo.bard.webservice.data.time.TimeGrain;
+
+import java.util.List;
+import java.util.Locale;
+
+/**
+ * Created by kevin on 3/7/2017.
+ */
+public class DruidMetricName implements FieldName {
+    private String lowerCaseName;
```
michael-mclawhorn: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/metric/FiliMetricName.java
michael-mclawhorn: This may be better named FiliApiMetricName because it extends ApiMetricName.
(I'm experiencing deja vu, I thought I already made this comment elsewhere)

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/table/GenericTableLoader.java
```diff
@@ -0,0 +1,171 @@
...
+        this.configLoader = configLoader;
+        configureSample(genericDimensions);
+    }
+
+    /**
+     * Set up the tables for this table loader.
+     *
+     * @param genericDimensions  The dimensions to load into test tables.
+     */
+    private void configureSample(GenericDimensions genericDimensions) {
```
michael-mclawhorn: codacy apparently gets confused by the foreach syntax.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/table/GenericTableLoader.java
```diff
@@ -0,0 +1,171 @@
...
+                    Utils.asLinkedHashSet(
+                            getAllGranularities(druidConfig)
+                    )
+            );
+        }
+
+    }
+
+    //TODO check AllGranularity.INSTANCE
+    private Granularity[] getAllGranularities(final DruidConfig druidConfig) {
```
michael-mclawhorn: final in parameter lists should be removed. (Just a style thing)

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/table/GenericTableLoader.java
```diff
@@ -0,0 +1,171 @@
...
+            new HashMap<>();
+    // Set up the metrics
+    private final Map<String, Set<FieldName>> druidMetricNames =
+            new HashMap<>();
+    private final Map<String, Set<ApiMetricName>> apiMetricNames =
+            new HashMap<>();
+    // Set up the table definitions
+    private final Map<String, Set<PhysicalTableDefinition>> tableDefinitions =
+            new HashMap<>();
+    private ConfigLoader configLoader;
```
michael-mclawhorn: These variable names end up being confusing below because it's not clear that they are maps by table name.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericBinder.java
```diff
@@ -0,0 +1,49 @@
...
+import com.yahoo.wiki.webservice.data.config.metric.GenericMetricLoader;
+import com.yahoo.wiki.webservice.data.config.table.GenericTableLoader;
+
+import java.util.LinkedHashSet;
+
+/**
+ * Created by kevin on 3/7/2017.
+ */
+public class GenericBinder {
+    private static GenericBinder genericBinder;
```
michael-mclawhorn: I'm not sure this should be called Binder in that it does not implement org.glassfish.hk2.utilities.Binder which is the sense of Binder used in the AbstractBinderFactory.

michael-mclawhorn: Also, class javadoc.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericBinder.java
```diff
@@ -0,0 +1,49 @@
...
+    private DruidWebService druidWebService;
+    private ConfigLoader configLoader;
+
+    private GenericBinder() {
+        configLoader = new StaticWikiConfigLoader();
+        //druidWebService = null; //TODO how to initialize with actual DruidWebService
```
michael-mclawhorn: I may be missing some context here.  Why is this using the hard-coding StaticWikiConfigLoader?

michael-mclawhorn:
Some general thoughts.

The set of abstractions here is a little loose, and I think doing class level javadoc comments will help flush that out.

Your interface DruidConfig seems sane, and making a supplier for it available to each of your loaders also seems sane.  I'm not sure your use of GenericDimensions to arbitrarily build and distribute Dimensions around to unconnected places is well justified.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericBinder.java
kevinhinterlong: Should I just remove GenericBinder? It seemed appropriate at the time but looking back it doesn't seem very necessary

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericBinder.java
kevinhinterlong: This was before I found out how to get a reference to a DruidWebService from GenericBinderFactory

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/table/GenericTableLoader.java
kevinhinterlong: How about `tableNameTo` { tableDefinition, apiMetricNames, ...}?

fili-core/src/main/java/com/yahoo/bard/webservice/application/DruidDimensionsLoader.java
```diff
@@ -84,6 +84,54 @@ public DruidDimensionsLoader(
...
+     * DruidDimensionsLoader fetches data from Druid and adds it to the dimension cache.
+     * @param druidWebService  The druid webservice to query.
+     * @param dimensions  The dimensions to use.
+     * @param dataSources  The datasources to use.
+     */
+    public DruidDimensionsLoader(
+            DruidWebService druidWebService,
+            List<List<Dimension>> dimensions,
+            List<DataSource> dataSources
+    ) {
```
cdeszaq: This constructor is called from `GenericBinderFactory::buildDruidDimensionsLoader`, right? It looks like it's doing something different with the dimensions (ie. different than what the `buildDimensionList` method does)

fili-core/src/main/java/com/yahoo/bard/webservice/application/DruidDimensionsLoader.java
```diff
@@ -84,6 +84,54 @@ public DruidDimensionsLoader(
...
+                buildDataSourcesList(physicalTableDictionary)
+        );
+    }
+
+    /**
+     * Builds a list of datasources from the physical tables.
+     * @param physicalTableDictionary  The physical tables to load from.
+     * @return the list of datasources.
+     */
+    public static List<DataSource> buildDataSourcesList(PhysicalTableDictionary physicalTableDictionary) {
```
cdeszaq: 👍 It's probably better to have another constructor that takes the dimensions and the physical table dictionary and converts the dict into the datasources, instead of having the convertToDatasources method getting called as part of the call to some other constructor outside this class.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericMain.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+
+    /**
+     * Run a generic setup which mirrors all information from druid into fili configuration.
+     *
+     * @param args  Command line arguments.
+     * @throws Exception if the server fails to start or crashes.
+     */
+    public static void main(String[] args) throws Exception {
```
cdeszaq: `main` methods are intended as the entry point into a program. As such, they can throw pretty much any exception, and the result will usually be that the application dies. We should leave this as-is, since it's the common pattern for `main` methods.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/DruidNavigator.java
```diff
@@ -0,0 +1,190 @@
...
+    @Override
+    public List<? extends DataSourceConfiguration> get() {
+        if (tableConfigurations.isEmpty()) {
+            loadAllDatasources();
+            LOG.info("Loading all datasources");
+            try {
+                //TODO Fix this so it waits for druid's responses
+                // instead of waiting 1 second for everything to be configured
+                // in order to wait for druid's response, DruidWebService will have to return a future
+                Thread.sleep(1000);
```
cdeszaq: Good call! I always forget about `CountDownLatch`! It's definitely the right choice here.

That said, the other option (which I think might be the route to take) would be to make the async client return a `Future` so that consumers (like this one) can switch to a blocking behavior if they wish.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericBinderFactory.java
```diff
@@ -0,0 +1,87 @@
...
+     */
+    public GenericBinderFactory() {
+        DruidWebService druidWebService = buildMetadataDruidWebService(getMappers().getMapper());
+        configLoader = new DruidNavigator(druidWebService);
+        configLoader.get();
+    }
+
+    /**
+     * Provides a way for {@link GenericMain} to access the dimensions.
+     * TODO: This is a bad solution to getting the dimensions from GenericMain
```
cdeszaq: If we're using the DruidDimensionLoader to load the dimension values from Druid, we _shouldn't_ need to mark the dimensions healthy ourselves.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericBinderFactory.java
```diff
@@ -0,0 +1,87 @@
...
+     * @return the dimensions found from druid.
+     */
+    public static Set<DimensionConfig> getDimensions() {
+        return dimensions;
+    }
+
+    @Override
+    protected Set<DimensionConfig> getDimensionConfigurations() {
+        //NOTE: This is guaranteed to be called before getTableLoader()
+        genericDimensions = new GenericDimensions(configLoader);
```
cdeszaq: Why not make this call in the constructor? That should let `genericDimensions` be `final` then, which is always nice. It also means _you_ are more in control of the ordering, which is _also_ nice.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericMain.java
```diff
@@ -0,0 +1,120 @@
...
+     * {
+     *     "name": "d",
+     *     "lastUpdated": "2016-01-01"
+     * }
+     *
+     * @param port  The port through which we access the webservice
+     *
+     * @throws IOException If something goes terribly wrong when building the JSON or sending it
+     */
+    private static void markDimensionCacheHealthy(int port) throws IOException {
```
cdeszaq: Now that we're using the DruidDimensionLoader, we shouldn't need this method at all I think.

And if we yank this method, the `LOG` isn't used anymore either.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/DataSourceConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+     * Gets the names of all the dimensions for the current datasource.
+     * @return a list of names of dimensions for the current datasource.
+     */
+    List<String> getDimensions();
+
+    /**
+     * Gets the list of {@link TimeGrain} which are valid for use in queries.
+     * @return a list of {@link TimeGrain} for the current datasource.
+     */
+    List<TimeGrain> getValidTimeGrains();
```
cdeszaq: I feel like this method is trying to be a little bit too smart... If this class is modeling a Druid DataSource (and it sure feels like that's what it's doing), then it should model just a single time grain, since a Druid data source _should_ have just a single grain. 

The other smarts that this method is wrapping should be someone else's responsibilities.

In theory, a Datasource in Druid _could_ have more than 1 grain at different times, but for now we only want to support one. It would be worth calling out this restriction somewhere as well (possibly in the README as well as the JavaDoc on this class)

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/DruidNavigator.java
```diff
@@ -0,0 +1,190 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Supplier;
+
+/**
+ * Searches for all datasources and their dimensions, metrics, and timegrains from druid.
+ */
+public class DruidNavigator implements Supplier<List<? extends DataSourceConfiguration>> {
+    private static final Logger LOG = LoggerFactory.getLogger(DruidNavigator.class);
+    private static final String DATASOURCES = "/datasources/";
```
cdeszaq: We should give this constant a better (ie. more descriptive) name. Names are best if they indicate what the thing _means_, rather than what it _is_.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericBinderFactory.java
```diff
@@ -0,0 +1,87 @@
...
+    ) {
+        List<List<Dimension>> dimensionsList = genericDimensions.getAllDimensionConfigurations().stream()
+                .map(DimensionConfig::getApiName)
+                .map(dimensionDictionary::findByApiName)
+                .map(Collections::singletonList)
+                .collect(Collectors.toList());
+        return new DruidDimensionsLoader(
+                webService,
+                dimensionsList,
+                DruidDimensionsLoader.buildDataSourcesList(physicalTableDictionary)
```
garyluoex: I tried to trace it, all of them source from `GenericDimensions`, i.e. `ConfigurationLoader` gets `TypeAwareDimensionLoader` which gets dimension configurations from `getDimensionConfigurations` which is getting the configs from `GenericDimensions` which is what we use to construct `dimensionsList` here.

fili-core/src/main/java/com/yahoo/bard/webservice/application/DruidDimensionsLoader.java
garyluoex: It should be the same, let me know otherwise.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericBinderFactory.java
garyluoex: We should change it in the constructor so that if the `DRUID_DIM_LOADER_DIMENSIONS` is empty list, then we directly use everything in the dimensionDictionary, otherwise, use the `DRUID_DIM_LOADER_DIMENSIONS` list. This way we can just have one constructor without the static helper methods in `DruidDimensionLoader` class right?

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/DruidNavigator.java
kevinhinterlong: Is there any way to set a timeout with this?

kevinhinterlong: Nevermind just started writing it out and there is 👍

fili-generic-example/README.md
```diff
@@ -0,0 +1,83 @@
+Fili Generic Loader Application
+==================================
+
+This is application is meant to be able to run and connect to any instance of Druid to show the basic metrics and dimensions.
```
kevinhinterlong: Is there a standard way fili accepts command line parameters other places in the code? Or some library you use for parsing them?
Just want to make sure before I write something specific for this

fili-generic-example/README.md
```diff
@@ -0,0 +1,83 @@
...
+1. Follow the [Druid instructions](http://druid.io/docs/latest/tutorials/quickstart.html) to bring up a local Druid 
+   cluster on your Unix based machine.
+   
+2. Clone this repository to your computer.
+    ```bash
+    git clone git@github.com:yahoo/fili.git
+    ```
+3. Use Maven to install and launch the Fili Generic example:
+
+Simliar to the [Wikipedia example](../fili-wikipedia-example)
```
kevinhinterlong: 👍

fili-generic-example/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+- List dimensions:  
+
+      GET http://localhost:9998/v1/dimensions
+
+- List metrics:
+  
+      GET http://localhost:9998/v1/metrics/
+
+### Specific to Wikipedia Example
```
kevinhinterlong: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericBinderFactory.java
kevinhinterlong: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericBinderFactory.java
kevinhinterlong: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericMain.java
kevinhinterlong: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/dimension/GenericDimensionField.java
```diff
@@ -0,0 +1,40 @@
+// Copyright 2016 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.wiki.webservice.data.config.dimension;
+
+import com.yahoo.bard.webservice.data.dimension.DimensionField;
+
+/**
+ * DimensionField enum.
+ */
+public enum GenericDimensionField implements DimensionField {
```
kevinhinterlong: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/dimension/GenericDimensionConfig.java
```diff
@@ -0,0 +1,95 @@
...
+import com.yahoo.bard.webservice.data.dimension.SearchProvider;
+
+import java.util.LinkedHashSet;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * Holds dimensions configuration details.
+ */
+public class GenericDimensionConfig implements DimensionConfig {
```
kevinhinterlong: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/dimension/GenericDimensions.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+/**
+ * Hold all the dimension configurations for a generic druid configuration.
+ */
+public class GenericDimensions {
```
kevinhinterlong: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericMain.java
kevinhinterlong: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericBinderFactory.java
```diff
@@ -0,0 +1,87 @@
...
+        return new GenericTableLoader(configLoader, genericDimensions, getDataSourceMetadataService());
+    }
+
+    @Override
+    protected DruidDimensionsLoader buildDruidDimensionsLoader(
+            DruidWebService webService,
+            PhysicalTableDictionary physicalTableDictionary,
+            DimensionDictionary dimensionDictionary
+    ) {
+        List<List<Dimension>> dimensionsList = genericDimensions.getAllDimensionConfigurations().stream()
```
kevinhinterlong: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericBinderFactory.java
```diff
@@ -0,0 +1,87 @@
...
+    public static Set<DimensionConfig> getDimensions() {
+        return dimensions;
+    }
+
+    @Override
+    protected Set<DimensionConfig> getDimensionConfigurations() {
+        //NOTE: This is guaranteed to be called before getTableLoader()
+        genericDimensions = new GenericDimensions(configLoader);
+        dimensions = genericDimensions.getAllDimensionConfigurations();
+        return genericDimensions.getAllDimensionConfigurations();
```
kevinhinterlong: 👍

fili-generic-example/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-checkstyle-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>exec-maven-plugin</artifactId>
+                <version>1.1</version>
+                <executions>
+                    <execution><goals><goal>java</goal></goals></execution>
```
kevinhinterlong: 👍

fili-generic-example/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                        <configuration>
+                            <target>
+                                <exec executable="/bin/sh">
+                                    <env key="BUILD_NUMBER" value="${project.build}" />
+                                    <arg value="-c" />
+                                    <arg
+                                        value="echo ${project.version}.${BUILD_NUMBER:=0} | sed -e s/-SNAPSHOT//g &gt; target/buildVersion" />
```
kevinhinterlong: 👍

fili-generic-example/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+    <parent>
+        <groupId>com.yahoo.fili</groupId>
+        <artifactId>fili-parent-pom</artifactId>
+        <version>0.8-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>fili-generic-example</artifactId>
+    <packaging>war</packaging>
+    <name>Fili: Generic Example</name>
+    <description>Fili Wikipedia Example serves as a standalone sample implementation using core Fili services to stand
```
kevinhinterlong: Was there anywhere specific in the code you remember this? I know the package still says wiki

fili-core/src/main/java/com/yahoo/bard/webservice/application/DruidDimensionsLoader.java
```diff
@@ -84,6 +84,54 @@ public DruidDimensionsLoader(
...
+                buildDimensionList(dimensionDictionary),
+                buildDataSourcesList(physicalTableDictionary)
+        );
+    }
+
+    /**
+     * Builds a list of datasources from the physical tables.
+     * @param physicalTableDictionary  The physical tables to load from.
+     * @return the list of datasources.
+     */
```
kevinhinterlong: Is this the case for `@throws` as well?

fili-generic-example/README.md
```diff
@@ -0,0 +1,83 @@
+Fili Generic Loader Application
+==================================
+
+This is application is meant to be able to run and connect to any instance of Druid to show the basic metrics and dimensions.
+
+## Setup and Launching
```
kevinhinterlong: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericMain.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    /**
+     * Run a generic setup which mirrors all information from druid into fili configuration.
+     *
+     * @param args  Command line arguments.
+     * @throws Exception if the server fails to start or crashes.
+     */
+    public static void main(String[] args) throws Exception {
+        int port = 9998;
```
kevinhinterlong: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericMain.java
kevinhinterlong: 👍

fili-generic-example/README.md
kevinhinterlong: Nevermind 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/DataSourceConfiguration.java
kevinhinterlong: 👍

fili-core/src/main/java/com/yahoo/bard/webservice/application/DruidDimensionsLoader.java
cdeszaq: Generally, it's:

1. Blank line between description and `@` clauses
2. Blank line between `@param` section and the remaining `@` clauses.
3. Optionally a blank line after the `@return` clause (if there is one, and if there are additional clauses)
    - And the `@return` clause should be the next one after the `@param` ones

The only ones that really matter are the 1st and 2nd ones. The 3rd one is optional (ie. if you think it makes it more readable / visible)

fili-generic-example/pom.xml
cdeszaq: The package is the most likely culprit. But you could do a `grep` through the directory holding the `fili-generic-example` and find all instances of `[Ww]iki` pretty easily.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericBinderFactory.java
cdeszaq: Making this change would be a difference in how the `DruidDimensionLoader` has worked, but it _should_ be OK to do (I think), but we'll want to check what the defaults have been and what it has been doing from a behavior standpoint. 

I think the best approach (and we've started to move this way) is to have 2 constructors:

1. That takes the list of dimensions (ie. no dependencies on SystemConfig at all)
2. One that loads the dimensions from the configuration, resolves them, and then calls to the other constructor that takes the list.

The advantages of this are multiple, but here are a few:

1. Easier to test

    If we have to go through SystemConfig, it's harder to test because we have to set up the dimensions via both being in the dictionary, as well as set and revert a config parameter. If we can just pass in the Dimensions to load, it's much simpler

2. More reusable

    If we have a different source for what dimensions to load (ie. the ones that we're pulling in from Druid), then that use-case can just call the "primary" constructor directly and pass in the Dimensions desired.

If I'm remembering correctly (it's been a long weekend), this PR pushes us in the direction I was just talking about, though perhaps it doesn't go quite all the way there?

fili-generic-example/README.md
cdeszaq: 😄 To answer this question, the SystemConfig component _should_ pick up these parameters as something like environment variables, so we shouldn't need to do anything special. If I'm wrong, we should loop @michael-mclawhorn in, since he's most familiar with the SystemConfig system.

fili-core/src/main/java/com/yahoo/bard/webservice/application/DruidDimensionsLoader.java
```diff
@@ -84,6 +84,54 @@ public DruidDimensionsLoader(
...
+                .collect(Collectors.toList());
+    }
+
+    /**
+     * DruidDimensionsLoader fetches data from Druid and adds it to the dimension cache.
+     * @param druidWebService  The druid webservice to query.
+     * @param dimensions  The dimensions to use.
+     * @param dataSources  The datasources to use.
+     */
+    public DruidDimensionsLoader(
```
kevinhinterlong: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericBinderFactory.java
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * Builds dimensions, metrics, and tables for all datasources found from druid.
+ */
+public class GenericBinderFactory extends AbstractBinderFactory {
+    private final Supplier<List<? extends DataSourceConfiguration>> configLoader;
+    private GenericDimensions genericDimensions;
+    private static Set<DimensionConfig> dimensions;
+
+    /**
+     * Constructs a GenericBinderFactory which starts configuring with druid.
```
kevinhinterlong: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/application/GenericBinderFactory.java
kevinhinterlong: 👍

fili-core/src/main/java/com/yahoo/bard/webservice/application/DruidDimensionsLoader.java
kevinhinterlong: I think I've come up with an okay solution to the constructor problem

fili-core/src/main/java/com/yahoo/bard/webservice/application/DruidDimensionsLoader.java
```diff
@@ -84,6 +84,54 @@ public DruidDimensionsLoader(
...
+                .map(TableDataSource::new)
+                .collect(Collectors.toList());
+    }
+
+    /**
+     * Builds a list of dimensions from the dimension dictionary.
+     * @param dimensionDictionary  The dimension dictionary to load from.
+     * @return the list of dimensions.
+     */
+    public static List<List<Dimension>> buildDimensionList(DimensionDictionary dimensionDictionary) {
```
kevinhinterlong: 👍

fili-core/src/main/java/com/yahoo/bard/webservice/application/DruidDimensionsLoader.java
kevinhinterlong: 👍

fili-core/src/main/java/com/yahoo/bard/webservice/application/DruidDimensionsLoader.java
```diff
@@ -84,6 +84,54 @@ public DruidDimensionsLoader(
...
+        // A DruidSearchQuery requires a list of dimensions, which we would have to explicitly create at serialization
+        // time if `dimensions` were a flat list instead of a list of singleton lists.
+        return dimensionStr.stream()
+                .map(dimensionDictionary::findByApiName)
+                .map(Collections::singletonList)
+                .collect(Collectors.toList());
+    }
+
+    /**
+     * DruidDimensionsLoader fetches data from Druid and adds it to the dimension cache.
```
kevinhinterlong: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/DruidNavigator.java
```diff
@@ -0,0 +1,177 @@
...
+            }
+        }, COORDINATOR_TABLES_PATH);
+    }
+
+    /**
+     * Load a specific table with all reported metrics, dimensions, and timegrains.
+     *
+     * @param table The TableConfig to be loaded with queries against druid.
+     */
+    private void loadTable(TableConfig table) {
```
kevinhinterlong: These should have javadoc comments showing example json being parsed

fili-generic-example/pom.xml
```diff
@@ -0,0 +1,146 @@
...
+        <maven>3.0</maven>
+    </prerequisites>
+
+    <parent>
+        <groupId>com.yahoo.fili</groupId>
+        <artifactId>fili-parent-pom</artifactId>
+        <version>0.8-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>fili-generic-example</artifactId>
```
kevinhinterlong: Should I change the the name `fili-generic-example`?

fili-generic-example/pom.xml
```diff
@@ -0,0 +1,146 @@
...
+                <version>1.1</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>java</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <mainClass>com.yahoo.wiki.webservice.application.GenericMain</mainClass>
```
kevinhinterlong: Refactor `.wiki`

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/DruidNavigator.java
```diff
@@ -0,0 +1,214 @@
...
+     *      ]
+     * }
+     *
+     * @param table The TableConfig to be loaded with queries against druid.
+     */
+    private void loadTable(TableConfig table) {
+        String url = COORDINATOR_TABLES_PATH + table.getName() + "/?full";
+        queryDruid(rootNode -> {
+            JsonNode segments = rootNode.get("segments").get(0);
+            loadMetrics(table, segments);
```
michael-mclawhorn: Make a note in class comments that the schema of the first segment is for configuration.

michael-mclawhorn: NPE will be thrown if the segment list is empty.  Perhaps we should catch and rethrow with a better error message.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/metric/GenericMetricLoader.java
```diff
@@ -0,0 +1,85 @@
...
+        this.configLoader = configLoader;
+    }
+
+    /**
+     * (Re)Initialize the metric makers with the given metric dictionary.
+     *
+     * @param metricDictionary  Metric dictionary to use for generating the metric makers.
+     */
+    protected void buildMetricMakers(MetricDictionary metricDictionary) {
+        // Create the various metric makers
```
michael-mclawhorn: Not that it's a big issue, but why is this implemented this way rather than having an initialize-once pattern?

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/table/GenericTableLoader.java
```diff
@@ -0,0 +1,163 @@
...
+                            metricName,
+                            dataSourceConfiguration.getValidTimeGrain()
+                    ))
+                    .collect(Collectors.toSet());
+
+            validGrains = getGranularities(dataSourceConfiguration);
+        });
+
+
+    }
```
michael-mclawhorn: extra vertical whitespace

kevinhinterlong: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/DruidNavigator.java
kevinhinterlong: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/metric/GenericMetricLoader.java
kevinhinterlong: Did you want something like this?
```java
if (doubleSumMaker == null) {
    doubleSumMaker = new DoubleSumMaker(metricDictionary);
}
```

PR #193, opened 2017-03-10
Implement MetricUnionAvailability & MetricUnionCompositeTable

QubitPi:
**TODO**
- [x] Add tests to `MetricUnionCompositeTable` once @cdeszaq @michael-mclawhorn @garyluoex are OK with its implementation.
- [x] Update ChangeLog

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
@@ -7,6 +7,9 @@ import com.yahoo.bard.webservice.data.time.ZonedTimeGrain;
 import com.yahoo.bard.webservice.table.availability.MetricUnionAvailability;
 import com.yahoo.bard.webservice.table.resolver.GranularityComparator;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -16,14 +19,53 @@ import javax.validation.constraints.NotNull;
 
 /**
  * An implementation of <tt>BasePhysicalTable</tt> backed by metric union availability.
+ * <p>
+ * The composite table joins physical tables puts metric columns of different tables together so that we can
+ * query different metric columns from different tables at the same time.
+ * <p>
+ * For example, two tables of the following
+ *
+ * table1:
+ * +---------+---------+---------+
+ * | metric1 | metric2 | metric3 |
+ * +---------+---------+---------+
+ * |         |         |         |
+ * |         |         |         |
+ * +---------+---------+---------+
+ *
+ * table2
+ * +---------+---------+---------+
+ * | metric4 | metric5 | metric5 |
+ * +---------+---------+---------+
+ * |         |         |         |
+ * |         |         |         |
+ * +---------+---------+---------+
+ *
+ * are put together into a table
+ *
+ * +---------+---------+---------+---------+---------+---------+
+ * | metric1 | metric2 | metric3 | metric4 | metric5 | metric5 |
+ * +---------+---------+---------+---------+---------+---------+
+ * |         |         |         |         |         |         |
+ * |         |         |         |         |         |         |
+ * +---------+---------+---------+---------+---------+---------+
+ *
+ * and this joined table is backed by the <tt>MetricUnionAvailability</tt>
+ *
  */
 public class MetricUnionCompositeTable extends BasePhysicalTable {
+    private static final Logger LOG = LoggerFactory.getLogger(MetricUnionCompositeTable.class);
+    private static final GranularityComparator GRANULARITY_COMPARATOR = new GranularityComparator();
+
     /**
      * Constructor.
      *
-     * @param name  Name of the physical table as TableName, also used as fact table name
+     * @param name  Name that represents set of fact table names joined together
      * @param columns  The columns for this table
-     * @param physicalTables  A set of <tt>PhysicalTable</tt>s
+     * @param physicalTables  A set of <tt>PhysicalTable</tt>s whose same metric schema is to be joined together. The
+     * tables will be used to construct MetricUnionAvailability, as well as to compute common/coarsest time grain among
+     * them. The <tt>PhysicalTable</tt>s needs to have mutually satisfying time grains in order to calculate the
+     * common/coarsest time grain.
      * @param logicalToPhysicalColumnNames  Mappings from logical to physical names
      */
     public MetricUnionCompositeTable(
@@ -45,44 +87,40 @@ public class MetricUnionCompositeTable extends BasePhysicalTable {
      * Returns the coarsest <tt>ZonedTimeGrain</tt> that satisfies all tables.
      * <p>
      * If the set of <tt>PhysicalTables</tt>'s is empty or the coarsest <tt>ZonedTimeGrain</tt> is not
-     * compatible with any of the <tt>PhysicalTables</tt>s, throw <tt>IllegalArgumentException</tt>.
+     * compatible with all of the <tt>PhysicalTables</tt>s, throw <tt>IllegalArgumentException</tt>.
      *
      * @param physicalTables  A set of <tt>PhysicalTable</tt>s among which the coarsest <tt>ZonedTimeGrain</tt>
      * is to be returned.
      *
      * @return the coarsest <tt>ZonedTimeGrain</tt> among a set of <tt>PhysicalTables</tt>s
+     * @throws IllegalArgumentException when no PhysicalTable is provided or there is not mutually satisfying grain
+     * among the table's time grain
      */
-    private static ZonedTimeGrain getCoarsestTimeGrain(Set<PhysicalTable> physicalTables) {
-        if (physicalTables.isEmpty()) {
-            throw new IllegalArgumentException("At least 1 physical table needs to be provided");
-        }
-
-        GranularityComparator granularityComparator = new GranularityComparator();
-
-        // sort tables by <tt>ZonedTimeGrain</tt> in increasing order
-        List<PhysicalTable> sortedTables = physicalTables.stream()
-                .sorted((table1, table2) -> granularityComparator.compare(table1, table2))
+    private static ZonedTimeGrain getCoarsestTimeGrain(Set<PhysicalTable> physicalTables)
+            throws IllegalArgumentException {
+        // sort <tt>ZonedTimeGrain</tt>s in decreasing order
+        List<ZonedTimeGrain> sortedTimeGrains = physicalTables.stream()
+                .sorted(GRANULARITY_COMPARATOR)
+                .map(PhysicalTable::getSchema)
+                .map(PhysicalTableSchema::getTimeGrain)
                 .collect(Collectors.toList());
 
         // check to see if all <tt>ZonedTimeGrain</tt>'s is compatible with the coarsest <tt>ZonedTimeGrain</tt>
-        ZonedTimeGrain coarsestTimeGrain = sortedTables.get(0).getSchema().getTimeGrain();
-        List<PhysicalTable> incompatibles = sortedTables.stream()
-                .filter(table -> !table.getSchema().getTimeGrain().satisfiedBy(coarsestTimeGrain))
-                .collect(Collectors.toList());
-        if (!incompatibles.isEmpty()) {
-            throw new IllegalArgumentException(
-                    String.format(
-                            "The following ZonedTimeGrains are not compatible with the coarsest ZoneTimeGrain({}) - {}",
-                            coarsestTimeGrain,
-                            incompatibles.stream()
-                                    .collect(
-                                            Collectors.toMap(
-                                                    table -> table.getSchema().getTimeGrain(),
-                                                    table -> table.getTableName().asName()
-                                            )
-                                    )
-                    )
+        ZonedTimeGrain coarsestTimeGrain = sortedTimeGrains.stream()
+                .findFirst()
+                .orElseThrow(() -> {
+                        String message = "At least 1 physical table needs to be provided";
+                        LOG.error(message);
+                        return new IllegalArgumentException(message);
+                });
+
+        if (sortedTimeGrains.stream().anyMatch(grain -> grain.satisfiedBy(coarsestTimeGrain))) {
+            String message = String.format(
+                    "There is no mutually satisfying grain among: %s",
+                    sortedTimeGrains
             );
+            LOG.error(message);
+            throw new IllegalArgumentException(message);
         }
 
         return coarsestTimeGrain;
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
@@ -11,7 +11,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 
 import javax.validation.constraints.NotNull;
@@ -78,47 +77,38 @@ public class MetricUnionCompositeTable extends BasePhysicalTable {
     ) {
         super(
                 name,
-                verifyCoarsestTimeGrain(IntervalUtils.getCoarsestTimeGrain(physicalTables), physicalTables),
+                IntervalUtils.getCoarsestTimeGrain(physicalTables).orElseThrow(() -> {
+                    String message = "At least 1 physical table needs to be provided";
+                    LOG.error(message);
+                    return new IllegalArgumentException(message);
+                }),
                 columns,
                 logicalToPhysicalColumnNames,
                 new MetricUnionAvailability(physicalTables, columns)
         );
+        verifyGrainSatisfiesAllTables(IntervalUtils.getCoarsestTimeGrain(physicalTables).get(), physicalTables);
     }
 
     /**
-     * Verifies and returns the coarsest <tt>ZonedTimeGrain</tt> if that satisfies all tables.
+     * Verifies that the coarsest <tt>ZonedTimeGrain</tt> satisfies all tables.
      *
-     * @param coarsestGrain  The coarsest <tt>ZonedTimeGrain</tt> to be verified
+     * @param coarsestTimeGrain  The coarsest <tt>ZonedTimeGrain</tt> to be verified
      * @param physicalTables  A set of <tt>PhysicalTable</tt>s whose <tt>ZonedTimeGrain</tt>s are checked to make sure
      * they all satisfies with the given coarsest <tt>ZonedTimeGrain</tt>
      *
-     * @return the coarsest <tt>ZonedTimeGrain</tt> among a set of <tt>PhysicalTables</tt>s
-     * @throws IllegalArgumentException when no PhysicalTable is provided or there is not mutually satisfying grain
-     * among the table's time grain
+     * @throws IllegalArgumentException when there is no mutually satisfying grain among the table's time grains
      */
-    private static ZonedTimeGrain verifyCoarsestTimeGrain(
-            Optional<ZonedTimeGrain> coarsestGrain,
+    private static void verifyGrainSatisfiesAllTables(
+            ZonedTimeGrain coarsestTimeGrain,
             Set<PhysicalTable> physicalTables
     ) throws IllegalArgumentException {
-        ZonedTimeGrain coarsestTimeGrain = coarsestGrain
-                .orElseThrow(() -> {
-                        String message = "At least 1 physical table needs to be provided";
-                        LOG.error(message);
-                        return new IllegalArgumentException(message);
-                });
-
-        if (physicalTables.stream()
+        if (!physicalTables.stream()
                 .map(PhysicalTable::getSchema)
                 .map(PhysicalTableSchema::getTimeGrain)
-                .anyMatch(grain -> grain.satisfiedBy(coarsestTimeGrain))) {
-            String message = String.format(
-                    "There is no mutually satisfying grain among: %s",
-                    physicalTables
-            );
+                .allMatch(grain -> grain.satisfiedBy(coarsestTimeGrain))) {
+            String message = String.format("There is no mutually satisfying grain among: %s", physicalTables);
             LOG.error(message);
             throw new IllegalArgumentException(message);
         }
-
-        return coarsestTimeGrain;
     }
 }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
@@ -20,7 +20,7 @@ import javax.validation.constraints.NotNull;
 /**
  * An implementation of <tt>BasePhysicalTable</tt> backed by metric union availability.
  * <p>
- * The composite table joins physical tables puts metric columns of different tables together so that we can
+ * The composite table puts metric columns of different tables together so that we can
  * query different metric columns from different tables at the same time.
  * <p>
  * For example, two tables of the following
@@ -55,7 +55,7 @@ import javax.validation.constraints.NotNull;
  */
 public class MetricUnionCompositeTable extends BasePhysicalTable {
     private static final Logger LOG = LoggerFactory.getLogger(MetricUnionCompositeTable.class);
-    private static final GranularityComparator GRANULARITY_COMPARATOR = new GranularityComparator();
+    private static final GranularityComparator GRANULARITY_COMPARATOR = GranularityComparator.getInstance();
 
     /**
      * Constructor.
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
@@ -5,15 +5,14 @@ package com.yahoo.bard.webservice.table;
 import com.yahoo.bard.webservice.data.config.names.TableName;
 import com.yahoo.bard.webservice.data.time.ZonedTimeGrain;
 import com.yahoo.bard.webservice.table.availability.MetricUnionAvailability;
-import com.yahoo.bard.webservice.table.resolver.GranularityComparator;
+import com.yahoo.bard.webservice.util.IntervalUtils;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 import javax.validation.constraints.NotNull;
 
@@ -59,7 +58,6 @@ import javax.validation.constraints.NotNull;
  */
 public class MetricUnionCompositeTable extends BasePhysicalTable {
     private static final Logger LOG = LoggerFactory.getLogger(MetricUnionCompositeTable.class);
-    private static final GranularityComparator GRANULARITY_COMPARATOR = GranularityComparator.getInstance();
 
     /**
      * Constructor.
@@ -80,7 +78,7 @@ public class MetricUnionCompositeTable extends BasePhysicalTable {
     ) {
         super(
                 name,
-                getCoarsestTimeGrain(physicalTables),
+                verifyCoarsestTimeGrain(IntervalUtils.getCoarsestTimeGrain(physicalTables), physicalTables),
                 columns,
                 logicalToPhysicalColumnNames,
                 new MetricUnionAvailability(physicalTables, columns)
@@ -88,39 +86,34 @@ public class MetricUnionCompositeTable extends BasePhysicalTable {
     }
 
     /**
-     * Returns the coarsest <tt>ZonedTimeGrain</tt> that satisfies all tables.
-     * <p>
-     * If the set of <tt>PhysicalTables</tt>'s is empty throw <tt>IllegalArgumentException</tt>.
+     * Verifies and returns the coarsest <tt>ZonedTimeGrain</tt> if that satisfies all tables.
      *
-     * @param physicalTables  A set of <tt>PhysicalTable</tt>s among which the coarsest <tt>ZonedTimeGrain</tt>
-     * is to be returned.
+     * @param coarsestGrain  The coarsest <tt>ZonedTimeGrain</tt> to be verified
+     * @param physicalTables  A set of <tt>PhysicalTable</tt>s whose <tt>ZonedTimeGrain</tt>s are checked to make sure
+     * they all satisfies with the given coarsest <tt>ZonedTimeGrain</tt>
      *
      * @return the coarsest <tt>ZonedTimeGrain</tt> among a set of <tt>PhysicalTables</tt>s
      * @throws IllegalArgumentException when no PhysicalTable is provided or there is not mutually satisfying grain
      * among the table's time grain
      */
-    private static ZonedTimeGrain getCoarsestTimeGrain(Set<PhysicalTable> physicalTables)
-            throws IllegalArgumentException {
-        // sort <tt>ZonedTimeGrain</tt>s in decreasing order
-        List<ZonedTimeGrain> sortedTimeGrains = physicalTables.stream()
-                .sorted(GRANULARITY_COMPARATOR)
-                .map(PhysicalTable::getSchema)
-                .map(PhysicalTableSchema::getTimeGrain)
-                .collect(Collectors.toList());
-
-        // check to see if all <tt>ZonedTimeGrain</tt>'s is compatible with the coarsest <tt>ZonedTimeGrain</tt>
-        ZonedTimeGrain coarsestTimeGrain = sortedTimeGrains.stream()
-                .findFirst()
+    private static ZonedTimeGrain verifyCoarsestTimeGrain(
+            Optional<ZonedTimeGrain> coarsestGrain,
+            Set<PhysicalTable> physicalTables
+    ) throws IllegalArgumentException {
+        ZonedTimeGrain coarsestTimeGrain = coarsestGrain
                 .orElseThrow(() -> {
                         String message = "At least 1 physical table needs to be provided";
                         LOG.error(message);
                         return new IllegalArgumentException(message);
                 });
 
-        if (sortedTimeGrains.stream().anyMatch(grain -> grain.satisfiedBy(coarsestTimeGrain))) {
+        if (physicalTables.stream()
+                .map(PhysicalTable::getSchema)
+                .map(PhysicalTableSchema::getTimeGrain)
+                .anyMatch(grain -> grain.satisfiedBy(coarsestTimeGrain))) {
             String message = String.format(
                     "There is no mutually satisfying grain among: %s",
-                    sortedTimeGrains
+                    physicalTables
             );
             LOG.error(message);
             throw new IllegalArgumentException(message);
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
@@ -90,8 +90,7 @@ public class MetricUnionCompositeTable extends BasePhysicalTable {
     /**
      * Returns the coarsest <tt>ZonedTimeGrain</tt> that satisfies all tables.
      * <p>
-     * If the set of <tt>PhysicalTables</tt>'s is empty or the coarsest <tt>ZonedTimeGrain</tt> is not
-     * compatible with all of the <tt>PhysicalTables</tt>s, throw <tt>IllegalArgumentException</tt>.
+     * If the set of <tt>PhysicalTables</tt>'s is empty throw <tt>IllegalArgumentException</tt>.
      *
      * @param physicalTables  A set of <tt>PhysicalTable</tt>s among which the coarsest <tt>ZonedTimeGrain</tt>
      * is to be returned.
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
@@ -24,7 +24,8 @@ import javax.validation.constraints.NotNull;
  * query different metric columns from different tables at the same time.
  * <p>
  * For example, two tables of the following
- *
+ * <pre>
+ * {@code
  * table1:
  * +---------+---------+---------+
  * | metric1 | metric2 | metric3 |
@@ -40,16 +41,19 @@ import javax.validation.constraints.NotNull;
  * |         |         |         |
  * |         |         |         |
  * +---------+---------+---------+
- *
+ * }
+ * </pre>
  * are put together into a table
- *
+ * <pre>
+ * {@code
  * +---------+---------+---------+---------+---------+---------+
  * | metric1 | metric2 | metric3 | metric4 | metric5 | metric5 |
  * +---------+---------+---------+---------+---------+---------+
  * |         |         |         |         |         |         |
  * |         |         |         |         |         |         |
  * +---------+---------+---------+---------+---------+---------+
- *
+ * }
+ * </pre>
  * and this joined table is backed by the <tt>MetricUnionAvailability</tt>
  *
  */
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
@@ -86,7 +86,7 @@ public class MetricUnionCompositeTable extends BasePhysicalTable {
                 logicalToPhysicalColumnNames,
                 new MetricUnionAvailability(physicalTables, columns)
         );
-        verifyGrainSatisfiesAllTables(IntervalUtils.getCoarsestTimeGrain(physicalTables).get(), physicalTables);
+        verifyGrainSatisfiesAllTables(getSchema().getTimeGrain(), physicalTables);
     }
 
     /**
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
@@ -1,0 +1,90 @@
+// Copyright 2017 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.table;
+
+import com.yahoo.bard.webservice.data.config.names.TableName;
+import com.yahoo.bard.webservice.data.time.ZonedTimeGrain;
+import com.yahoo.bard.webservice.table.availability.MetricUnionAvailability;
+import com.yahoo.bard.webservice.table.resolver.GranularityComparator;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * An implementation of <tt>BasePhysicalTable</tt> backed by metric union availability.
+ */
+public class MetricUnionCompositeTable extends BasePhysicalTable {
+    /**
+     * Constructor.
+     *
+     * @param name  Name of the physical table as TableName, also used as fact table name
+     * @param columns  The columns for this table
+     * @param physicalTables  A set of <tt>PhysicalTable</tt>'s
+     * @param logicalToPhysicalColumnNames  Mappings from logical to physical names
+     */
+    public MetricUnionCompositeTable(
+            @NotNull TableName name,
+            @NotNull Set<Column> columns,
+            @NotNull Set<PhysicalTable> physicalTables,
+            @NotNull Map<String, String> logicalToPhysicalColumnNames
+    ) {
+        super(
+                name,
+                getCoarsestTimeGrain(physicalTables),
+                columns,
+                logicalToPhysicalColumnNames,
+                new MetricUnionAvailability(physicalTables, columns)
+        );
+    }
+
+    /**
+     * Returns the coarsest <tt>ZonedTimeGrain</tt> among a set of <tt>PhysicalTables</tt>'s.
+     * <p>
+     * If the set of <tt>PhysicalTables</tt>'s is empty or the coarsest <tt>ZonedTimeGrain</tt> is not
+     * compatible with any of the <tt>PhysicalTables</tt>'s, throw <tt>IllegalArgumentException</tt>.
+     *
+     * @param physicalTables  A set of <tt>PhysicalTable</tt>'s among which the coarsest <tt>ZonedTimeGrain</tt>
+     * is to be returned.
+     *
+     * @return the coarsest <tt>ZonedTimeGrain</tt> among a set of <tt>PhysicalTables</tt>'s
+     */
+    private static ZonedTimeGrain getCoarsestTimeGrain(Set<PhysicalTable> physicalTables) {
+        if (physicalTables.isEmpty()) {
+            throw new IllegalArgumentException("At least 1 physical table needs to be provided");
+        }
+
+        GranularityComparator granularityComparator = new GranularityComparator();
+
+        // sort tables by <tt>ZonedTimeGrain</tt> in increasing order
+        List<PhysicalTable> sortedTables = physicalTables.stream()
+                .sorted((table1, table2) -> granularityComparator.compare(table1, table2))
+                .collect(Collectors.toList());
+
+        // check to see if all <tt>ZonedTimeGrain</tt>'s is compatible with the coarsest <tt>ZonedTimeGrain</tt>
+        ZonedTimeGrain coarsestTimeGrain = sortedTables.get(0).getSchema().getTimeGrain();
+        List<PhysicalTable> incompatibles = sortedTables.stream()
+                .filter(table -> !table.getSchema().getTimeGrain().satisfiedBy(coarsestTimeGrain))
+                .collect(Collectors.toList());
+        if (!incompatibles.isEmpty()) {
+            throw new IllegalArgumentException(
+                    String.format(
+                            "The following ZonedTimeGrains are not compatible with the coarsest ZoneTimeGrain({}) - {}",
+                            coarsestTimeGrain,
+                            incompatibles.stream()
+                                    .collect(
+                                            Collectors.toMap(
+                                                    table -> table.getSchema().getTimeGrain(),
+                                                    table -> table.getTableName().asName()
+                                            )
+                                    )
+                    )
+            );
+        }
+
+        return coarsestTimeGrain;
+    }
+}
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
@@ -23,7 +23,7 @@ public class MetricUnionCompositeTable extends BasePhysicalTable {
      *
      * @param name  Name of the physical table as TableName, also used as fact table name
      * @param columns  The columns for this table
-     * @param physicalTables  A set of <tt>PhysicalTable</tt>'s
+     * @param physicalTables  A set of <tt>PhysicalTable</tt>s
      * @param logicalToPhysicalColumnNames  Mappings from logical to physical names
      */
     public MetricUnionCompositeTable(
@@ -42,15 +42,15 @@ public class MetricUnionCompositeTable extends BasePhysicalTable {
     }
 
     /**
-     * Returns the coarsest <tt>ZonedTimeGrain</tt> among a set of <tt>PhysicalTables</tt>'s.
+     * Returns the coarsest <tt>ZonedTimeGrain</tt> that satisfies all tables.
      * <p>
      * If the set of <tt>PhysicalTables</tt>'s is empty or the coarsest <tt>ZonedTimeGrain</tt> is not
-     * compatible with any of the <tt>PhysicalTables</tt>'s, throw <tt>IllegalArgumentException</tt>.
+     * compatible with any of the <tt>PhysicalTables</tt>s, throw <tt>IllegalArgumentException</tt>.
      *
-     * @param physicalTables  A set of <tt>PhysicalTable</tt>'s among which the coarsest <tt>ZonedTimeGrain</tt>
+     * @param physicalTables  A set of <tt>PhysicalTable</tt>s among which the coarsest <tt>ZonedTimeGrain</tt>
      * is to be returned.
      *
-     * @return the coarsest <tt>ZonedTimeGrain</tt> among a set of <tt>PhysicalTables</tt>'s
+     * @return the coarsest <tt>ZonedTimeGrain</tt> among a set of <tt>PhysicalTables</tt>s
      */
     private static ZonedTimeGrain getCoarsestTimeGrain(Set<PhysicalTable> physicalTables) {
         if (physicalTables.isEmpty()) {
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
@@ -78,7 +78,11 @@ public class MetricUnionCompositeTable extends BasePhysicalTable {
         super(
                 name,
                 IntervalUtils.getCoarsestTimeGrain(physicalTables).orElseThrow(() -> {
-                    String message = "At least 1 physical table needs to be provided";
+                    String message = String.format(
+                            "At least 1 physical table needs to be provided in order to calculate " +
+                                    "coarsest time grain for %s",
+                            name.asName()
+                    );
                     LOG.error(message);
                     return new IllegalArgumentException(message);
                 }),
@@ -86,7 +90,7 @@ public class MetricUnionCompositeTable extends BasePhysicalTable {
                 logicalToPhysicalColumnNames,
                 new MetricUnionAvailability(physicalTables, columns)
         );
-        verifyGrainSatisfiesAllTables(getSchema().getTimeGrain(), physicalTables);
+        verifyGrainSatisfiesAllTables(getSchema().getTimeGrain(), physicalTables, name);
     }
 
     /**
@@ -95,18 +99,25 @@ public class MetricUnionCompositeTable extends BasePhysicalTable {
      * @param coarsestTimeGrain  The coarsest <tt>ZonedTimeGrain</tt> to be verified
      * @param physicalTables  A set of <tt>PhysicalTable</tt>s whose <tt>ZonedTimeGrain</tt>s are checked to make sure
      * they all satisfies with the given coarsest <tt>ZonedTimeGrain</tt>
+     * @param name  Name of the current <tt>MetricUnionCompositeTable</tt> that represents set of fact table names
+     * joined together
      *
      * @throws IllegalArgumentException when there is no mutually satisfying grain among the table's time grains
      */
     private static void verifyGrainSatisfiesAllTables(
             ZonedTimeGrain coarsestTimeGrain,
-            Set<PhysicalTable> physicalTables
+            Set<PhysicalTable> physicalTables,
+            TableName name
     ) throws IllegalArgumentException {
         if (!physicalTables.stream()
                 .map(PhysicalTable::getSchema)
                 .map(PhysicalTableSchema::getTimeGrain)
                 .allMatch(grain -> grain.satisfiedBy(coarsestTimeGrain))) {
-            String message = String.format("There is no mutually satisfying grain among: %s", physicalTables);
+            String message = String.format("There is no mutually satisfying grain among: %s for current table " +
+                    "%s",
+                    physicalTables,
+                    name.asName()
+            );
             LOG.error(message);
             throw new IllegalArgumentException(message);
         }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
@@ -109,4 +109,9 @@ public class ConcreteAvailability implements Availability {
     protected Set<String> getColumnNames() {
         return columnNames;
     }
+
+    @Override
+    public String toString() {
+        return "concrete availability";
+    }
 }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
@@ -106,12 +106,31 @@ public class ConcreteAvailability implements Availability {
         return metadataService.getAvailableIntervalsByTable(name);
     }
 
+    /**
+     * Returns the configured columns in their String representation.
+     *
+     * @return the configured columns in their String representation
+     */
     protected Set<String> getColumnNames() {
         return columnNames;
     }
 
+    /**
+     * Returns the name of the table and data source associated with this Availability.
+     *
+     * @return the name of the table and data source associated with this Availability
+     */
+    protected TableName getName() {
+        return name;
+    }
+
     @Override
     public String toString() {
-        return "concrete availability";
+        return String.format("ConcreteAvailability with table name = %s and Configured columns = [%s]",
+                name.asName(),
+                columns.stream()
+                        .map(Column::getName)
+                        .collect(Collectors.joining(", "))
+        );
     }
 }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
@@ -89,9 +89,7 @@ public class MetricUnionAvailability implements Availability {
                                 availability ->
                                         Sets.intersection(
                                                 Utils.getSubsetByType(
-                                                        availability
-                                                                .getAllAvailableIntervals()
-                                                                .keySet(),
+                                                        availability.getAllAvailableIntervals().keySet(),
                                                         MetricColumn.class
                                                 ),
                                                 metricColumns
@@ -133,7 +131,7 @@ public class MetricUnionAvailability implements Availability {
      * Intervals of the same metric column are associated with the same metric column key. Overlapping intervals under
      * the same metric column key are collapsed into single interval.
      *
-     * @return a map of metric column to all of its available intervals in union
+     * @return a map of column to all of its available intervals in union
      */
     @Override
     public Map<Column, List<Interval>> getAllAvailableIntervals() {
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
@@ -1,0 +1,226 @@
+// Copyright 2017 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.table.availability;
+
+import com.yahoo.bard.webservice.data.config.names.TableName;
+import com.yahoo.bard.webservice.data.metric.MetricColumn;
+import com.yahoo.bard.webservice.table.Column;
+import com.yahoo.bard.webservice.table.PhysicalTable;
+import com.yahoo.bard.webservice.table.resolver.DataSourceConstraint;
+import com.yahoo.bard.webservice.util.IntervalUtils;
+import com.yahoo.bard.webservice.util.SimplifiedIntervalList;
+import com.yahoo.bard.webservice.util.Utils;
+
+import com.google.common.collect.Sets;
+
+import org.joda.time.Interval;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.AbstractMap;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * An implementation of availability which joins physical tables with the same metric schema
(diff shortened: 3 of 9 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,125 @@
...
+                .flatMap(Set::stream)
+                .collect(
+                        Collectors.groupingBy(
+                                Function.identity(),
+                                Collectors.counting()
+                        )
+                )
+                .entrySet()
+                .stream()
+                .anyMatch(entry -> entry.getValue() > 1)) {
```
garyluoex: Indention is a bit weird, otherwise good.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,125 @@
...
+    Map<Availability, Set<String>> metricAvailability;
+
+    /**
+     * Constructor.
+     *
+     * @param metricAvailability A mapping from a table to its contents availability
+     * @param columns The columns associated with all tables
+     */
+    public MetricUnionAvailability(
+            @NotNull Map<Table, Set<String>> metricAvailability,
```
garyluoex: Change `Table` to `PhysicalTable`

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,125 @@
...
+                )
+                .entrySet()
+                .stream()
+                .anyMatch(entry -> entry.getValue() > 1)) {
+            throw new IllegalArgumentException("Metric columns are not unique.");
+        }
+
+        this.columns = columns;
+        this.dataSourceNames = metricAvailability.keySet().stream()
+                .map(table -> TableName.of(table.getName()))
```
garyluoex: Use `getTableName` method on `PhysicalTable`

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,125 @@
...
+                .collect(Collectors.toSet());
+        this.metricAvailability = metricAvailability.entrySet().stream()
+                .collect(
+                        Collectors.toMap(
+                                entry -> PhysicalTable.class.cast(entry.getKey()).getAvailability(),
```
garyluoex: Should not need to cast

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @Override
+    public Set<TableName> getDataSourceNames() {
+        return dataSourceNames;
+    }
+
+    @Override
+    public Map<Column, Set<Interval>> getAllAvailableIntervals() {
+        return metricAvailability.entrySet().stream()
```
garyluoex: I think there might and will be repeated columns, we need to union those availabilities together I think using group by.

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/DataSourceConstraint.java
```diff
@@ -80,4 +104,40 @@ public DataSourceConstraint(DataApiRequest dataApiRequest, DruidAggregationQuery
...
+     * @param other the user provided set of columns
+     *
+     * @return The new <tt>DataSourceConstraint</tt> instance with a new subset of table columns
+     */
+    public DataSourceConstraint withIntersectedColumnNames(Set<String> other) {
+        return new DataSourceConstraint(
+                intersectionBetween(getRequestDimensions(), other),
+                intersectionBetween(getFilterDimensions(), other),
+                intersectionBetween(getMetricDimensions(), other),
+                metricNames.stream()
```
garyluoex: Just need the `metricNames` to intersect `other` not the dimensions here, change the name to include `metricIntersection`, leave the dimensions the same.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,125 @@
...
+     */
+    private Map<Availability, DataSourceConstraint> constructSubConstraint(DataSourceConstraint constraint) {
+        return metricAvailability.entrySet().stream()
+                .collect(
+                        Collectors.toMap(
+                                entry -> entry.getKey(),
+                                entry -> constraint.withIntersectedColumnNames(entry.getValue())
+                        )
+                ).entrySet().stream()
+                        .filter(entry -> !entry.getValue().getAllColumnNames().isEmpty())
```
garyluoex: check if the metric is empty, not including dimensions.

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/DataSourceConstraint.java
```diff
@@ -80,4 +104,40 @@ public DataSourceConstraint(DataApiRequest dataApiRequest, DruidAggregationQuery
...
+
+    /**
+     * Return intersection between two sets.
+     *
+     * @param first a set of <tt>Dimension</tt> objects
+     * @param second a set of <tt>Dimension</tt> objects in <tt>String</tt> representation
+     *
+     * @return The intersection, containing <tt>Dimension</tt> objects, between the two sets
+     */
+    private static Set<Dimension> intersectionBetween(Set<Dimension> first, Set<String> second) {
```
garyluoex: Can remove, no need.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,203 @@
+// Copyright 2016 Yahoo Inc.
```
garyluoex: 17

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
cdeszaq: Also, this is a strange way to cast. Generally, casts in Java are done like:

```java
entry -> (PhysicalTable) entry.getKey()
```

QubitPi:
The refactor breaks down methods into smaller helper methods so they are readable and can be tested better

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,357 @@
...
+
+    Set<Column> columns
+    Column column1
+    Column column2
+
+    DataSourceConstraint dataSourceConstraint
+
+    class TestMetricUnionAvailability extends MetricUnionAvailability {
+
+        List<TestAvailability> testAvailabilities
```
QubitPi: Does this need to be `List<Availability>` instead?

garyluoex: Can you briefly explain why do we need a `TestAvailability`?

QubitPi: Sure, it's explained in the next commit.

garyluoex: I mean... explain why do we need `TestAvailability` to exist? Explain briefly in this comment.

QubitPi: Since `Availability` is an interface. An implementation of it, `TestAvailability`, can return specified value from `getAvailableIntervals`. We need `Availability.getAvailableIntervals` to return something in test

QubitPi: I could've mock `Availability` and then `Availability.getAvailableIntervals >> ...`, but that's not allowed in `TestMetricUnionAvailability`

garyluoex: You should be able to test `MetricUnionAvailability` directly using mock without having to create another `TestMetricUnionAvailability`. I will review and figure out how to do this later today, in the mean time try if this is possible

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,190 @@
...
+                                Map.Entry::getKey,
+                                Map.Entry::getValue
+                        )
+                );
+    }
+
+    /**
+     * Construct a mapping from a type of <tt>Availability</tt> to <tt>DataSourceConstraint</tt> of that availability.
+     *
+     * @param constraint data constraint containing columns and api filters
```
cdeszaq: How is this parameter used? What does it mean?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,190 @@
...
+                        Collectors.toMap(
+                                Map.Entry::getKey,
+                                Map.Entry::getValue
+                        )
+                );
+    }
+
+    /**
+     * Construct a mapping from a type of <tt>Availability</tt> to <tt>DataSourceConstraint</tt> of that availability
+     * and filter out mappings that maps to <tt>DataSourceConstraint</tt> with empty set of metric names.
```
cdeszaq: Which things are are getting filtered out? It's not clear what things might have empty metrics, nor under what conditions this might happen.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,190 @@
...
+     *
+     * @param constraint data constraint containing columns and api filters
```
cdeszaq: What is this param for?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,190 @@
...
+     *
+     * @return A mapping from <tt>Availability</tt> to <tt>DataSourceConstraint</tt> with non-empty metric names
+     */
+    private Map<Availability, DataSourceConstraint> constructSubConstraint(DataSourceConstraint constraint) {
+        return getAvailabilityToConstraintMapping(constraint).entrySet().stream()
+                .filter(entry -> !entry.getValue().getMetricNames().isEmpty())
```
cdeszaq: We seem to be filtering these maps based on values (or keys) a whole lot. Perhaps we can combine some of these streams into a single stream, rather than going to an intermediate collection every time?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,190 @@
...
+     * <p>
+     * The return value is a map of duplicate <tt>MetricColumn</tt> to all <tt>Availability</tt>'s that contains
+     * this <tt>MetricColumn</tt>
+     *
+     * @param availabilitySetMap  A map from <tt>Availability</tt> to set of <tt>MetricColumn</tt> contained in that
+     * <tt>Availability</tt>
+     *
+     * @return duplicate values of <tt>MetricColumn</tt>'s
+     */
+    private static Map<MetricColumn, Set<Availability>> getDuplicateValue(
```
cdeszaq: What do we really care about here? Do we care about the availabilities that the metric columns are duplicated in, or do we only care _which_ metric columns are duplicated? If we care about both, how are we using that information?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,190 @@
...
+                        metricColumn,
+                        Sets.union(
+                                metricColumnSetMap.getOrDefault(metricColumn, new HashSet<>()),
+                                Sets.newHashSet(availability)
+                        )
+                );
+            }
+        }
+
+        return metricColumnSetMap.entrySet().stream()
```
cdeszaq: More comments in here in general would greatly aid understanding

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,190 @@
...
+    @Override
+    public Map<Column, List<Interval>> getAllAvailableIntervals() {
+        return metricAvailability.keySet().stream()
+                .map(availability -> availability.getAllAvailableIntervals().entrySet())
+                .flatMap(Set::stream)
+                .filter(entry -> metricColumns.contains(entry.getKey()))
+                .collect(
+                        Collectors.toMap(
+                                entry -> entry.getKey(),
+                                entry -> entry.getValue(),
```
cdeszaq: Can be method references

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,190 @@
...
+
+    @Override
+    public Set<TableName> getDataSourceNames() {
+        return dataSourceNames;
+    }
+
+    @Override
+    public Map<Column, List<Interval>> getAllAvailableIntervals() {
+        return metricAvailability.keySet().stream()
+                .map(availability -> availability.getAllAvailableIntervals().entrySet())
```
cdeszaq: Can be decomposed to 2 map calls: 

```java
.map(Availability::getAllAvailableIntervals)
.map(Map::entrySet)
```

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,190 @@
...
+                .filter(entry -> metricColumns.contains(entry.getKey()))
+                .collect(
+                        Collectors.toMap(
+                                entry -> entry.getKey(),
+                                entry -> entry.getValue(),
+                                (value1, value2) -> {
+                                    List<Interval> combined = new ArrayList<>();
+                                    combined.addAll(value1);
+                                    combined.addAll(value2);
+                                    return combined;
```
cdeszaq: I _think_ this body can be simpler, not needing the extra list construction:

```java
value1.addAll(value2);
return value1;
```

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,190 @@
...
+/**
+ * An implementation of availability which which joins physical tables with the same dimension schema
+ * but non-intersecting metric schemas. Metrics on the <tt>MetricUnionAvailability</tt> are sourced from
+ * exactly one of the participating tables.
+ */
+public class MetricUnionAvailability implements Availability {
+
+    private final Set<TableName> dataSourceNames;
+    private final Set<MetricColumn> metricColumns;
+    private final Map<Availability, Set<MetricColumn>> metricAvailability;
```
cdeszaq: Can we give thses a name that indicates what these things _mean_ rather than what these things _are_? For example, when `metricColumns` shows up in the code of this class, it's often unclear what this collection is being used for, and a better name would help that.

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/DataSourceConstraint.java
```diff
@@ -54,6 +55,39 @@ public DataSourceConstraint(DataApiRequest dataApiRequest, DruidAggregationQuery
...
+     * @param requestDimensions  Dimensions contained in request
+     * @param filterDimensions  Filtered dimensions
+     * @param metricDimensions  Metric related dimensions
+     * @param metricNames  Names of metrics
+     * @param allDimensions  Set of all dimension objects
+     * @param allDimensionNames  Set of all dimension names
+     * @param allColumnNames  Set of all column names
+     * @param apiFilters  Map of dimension to its set of API filters
+     */
+    public DataSourceConstraint(
```
cdeszaq: If this is only supposed to be used within this class, it should be protected, not public.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,190 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * An implementation of availability which which joins physical tables with the same dimension schema
+ * but non-intersecting metric schemas. Metrics on the <tt>MetricUnionAvailability</tt> are sourced from
+ * exactly one of the participating tables.
```
michael-mclawhorn: 'physical tables' here is shorthand for the availabilities of physical tables, I suppose.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,190 @@
...
+
+    private final Set<TableName> dataSourceNames;
+    private final Set<MetricColumn> metricColumns;
+    private final Map<Availability, Set<MetricColumn>> metricAvailability;
+
+    /**
+     * Constructor.
+     *
+     * @param physicalTables  A set of <tt>PhysicalTable</tt>'s
+     * @param columns  The columns associated with all tables
```
michael-mclawhorn: Perhaps we should describe this as 'The set of columns supplied by this availability'

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,190 @@
...
+        if (!duplicates.isEmpty()) {
+            throw new IllegalArgumentException(
+                    String.format(
+                            "Metric columns are not unique - {}",
+                            duplicates
+                    )
+            );
+        }
+
+        dataSourceNames = physicalTables.stream()
```
michael-mclawhorn: Good catch.
This is collecting physical table dictionary names and not datasource names.
We need to call 'getDataSourceNames' on the the underlying availabilities.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
michael-mclawhorn: It was my suggestion to log the duplicate column information.  If we are logging the source of the conflicts, the logging above needs to be enhanced.

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/DataSourceConstraint.java
michael-mclawhorn: What?

cdeszaq: @michael-mclawhorn This constructor is a copy-constructor that doesn't go through the same defensive copies that the other constructor does. It's only purpose in life is to support the `with*` method that this PR added, and should not be called from outside this class.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
cdeszaq: I've no problem with logging more stuff, but we should make sure that we're doing it at the right point, and keeping things as encapsulated and clear as we can. This class in general seems to be passing around a lot of Maps with signatures that are not particularly clear, so I want to make sure that we're letting the code be simple and easy to understand.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
cdeszaq: I would vote for using the longer but more descriptive description, to remove confusion and ambiguity. 5 months from now, when we forget how this is architected, we should be able to read this and understand what is going on.

cdeszaq:
The branch this had been based on got merged and deleted, which auto-closed this PR. I've re-set it's base to master and reopened it.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,226 @@
...
+                                Map.Entry::getKey,
+                                Map.Entry::getValue
+                        )
+                );
+    }
+
+    /**
+     * Given a <tt>DataSourceConstraint</tt> - DSC1, construct a mapping from a type of <tt>Availability</tt>, A, to its
+     * <tt>DataSourceConstraint</tt>, DSC2. DSC2 is constructed as the intersection of metric columns between DSC1 and
+     * A's available metric columns. There are cases in which the intersection is empty; this method filters out
```
michael-mclawhorn: The first sentence of javadoc needs to be newline terminated   It should probably be more concise, summarizing the rest of this explanation. 
Also, 'construct a mapping from a type of `Availability` should probably be `construct a map from each availability in this MetricUnion` ' or something like that.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,226 @@
...
+     * mapping entries that maps to <tt>DataSourceConstraint</tt> with empty set of metric names.
+     *
+     * @param constraint data constraint whose contained metric columns will be intersected with availabilities'
+     * metric columns
+     *
```
michael-mclawhorn: spacing, two spaces after param name. 
Use an article on param description, e.g.: 
@param constraint 'the data constraint' or 'a data constraint'.

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/DataSourceConstraint.java
```diff
@@ -85,4 +118,31 @@ public DataSourceConstraint(DataApiRequest dataApiRequest, DruidAggregationQuery
...
+
+
+    /**
+     * Create a new <tt>DataSourceConstraint</tt> instance with a new subset of metric names.
+     * <p>
+     * The new set of metric names will be an intersection between old metric names and
+     * a user provided set of metric names
+     *
+     * @param metricColumns  The user provided set of metric names
+     *
```
michael-mclawhorn: I don't think there is a 'user' in this context.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/resolver/DataSourceConstraintSpec.groovy
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.table.resolver
+
+import com.yahoo.bard.webservice.data.metric.MetricColumn
+import spock.lang.Specification
+
+class DataSourceConstraintSpec extends Specification {
+
+    def "test withIntersectedColumnNames"() {
+        given:
```
michael-mclawhorn: Use @Unroll and use #arguments in test definition to explain the parameters of each test case:
e.g.
#metricNames intersected with #other produces #newMetricNames

michael-mclawhorn: ps @cdeszaq hates the use of 'Test' in test names.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,189 @@
...
+        availability2 = Mock(Availability)
+
+        metricColumn1 = new MetricColumn('metric1')
+        metricColumn2 = new MetricColumn('metric2')
+
+        interval1 = new Interval('2018-01-01/2018-02-01')
+        interval2 = new Interval('2018-02-01/2018-03-01')
+        interval3 = new Interval('2019-01-01/2019-02-01')
+        interval4 = new Interval('2018-11-01/2018-12-01')
+
```
michael-mclawhorn: I'm not sure what to make of these intervals and the order of the intervals listed here.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,226 @@
...
+     * @return A mapping from <tt>Availability</tt> to <tt>DataSourceConstraint</tt> with non-empty metric names
+     */
+    private Map<Availability, DataSourceConstraint> constructSubConstraint(DataSourceConstraint constraint) {
+        return availabilityToAvailableColumns.entrySet().stream()
```
michael-mclawhorn: This variable should be plural

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,189 @@
...
+        expect:
+        metricUnionAvailability.getAllAvailableIntervals() == [
+                (metricColumn1): [interval1, interval4] as List,
+                (metricColumn2): [interval2] as List,
+        ]
+    }
+
+    def "test getAvailableIntervals"() {
+        given:
```
michael-mclawhorn: This test tests only a path with no overlap.  Would it make sense to also test when there is an overlap?

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,189 @@
...
+
+        physicalTable1.getAvailability() >> availability1
+        physicalTable2.getAvailability() >> availability2
+
+        physicalTable1.getSchema() >> schema1
+        physicalTable2.getSchema() >> schema2
+    }
+
+    def "dataSourceNames aggregates from availabilities not physical tables in MetricUnionAvailability constructor"() {
+        given:
```
michael-mclawhorn: I'm not sure I understand from the name what this test is testing.

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,90 @@
...
+            throw new IllegalArgumentException("At least 1 physical table needs to be provided");
+        }
+
+        GranularityComparator granularityComparator = new GranularityComparator();
+
+        // sort tables by <tt>ZonedTimeGrain</tt> in increasing order
+        List<PhysicalTable> sortedTables = physicalTables.stream()
+                .sorted((table1, table2) -> granularityComparator.compare(table1, table2))
+                .collect(Collectors.toList());
+
```
QubitPi: It does. Actually that was my original implementation. It's better until when I was trying to format exception message: if you keep table, user will be able to see which table has incompatible `ZonedTimeGrains`, so that they can understand error better.

But I can change it if you still prefer doing that :) I don't have strong preference on which

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,226 @@
...
+                        )
+                );
+
+        // validate metric uniqueness such that
+        // each table's underlying datasource schema don't have repeated metric column
+        Map<MetricColumn, Set<Availability>> duplicates = getDuplicateValue(availabilityToAvailableColumns);
+        if (!duplicates.isEmpty()) {
+            String message = String.format(
+                    "While constructing MetricUnionAvailability, Metric columns are not unique - {}",
+                    duplicates
```
QubitPi: Agree. I've added `toString` to all availabilities

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
QubitPi: Did you mean you were confused about how they are used?

michael-mclawhorn: Sort of.  They aren't ordered.  And they don't overlap at all.  So they feel kind of arbitrary.  Is there some meaning that they have relative to each other?

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
michael-mclawhorn: You don't save the sortedTables.get(0) value though, you simply pass through it to filter.
The sorted tables you stream over later doesn't use the fact that it's sorter, so you can just replace it with physicalTables.

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,90 @@
...
+     *
+     * @param physicalTables  A set of <tt>PhysicalTable</tt>s among which the coarsest <tt>ZonedTimeGrain</tt>
+     * is to be returned.
+     *
+     * @return the coarsest <tt>ZonedTimeGrain</tt> among a set of <tt>PhysicalTables</tt>s
+     */
+    private static ZonedTimeGrain getCoarsestTimeGrain(Set<PhysicalTable> physicalTables) {
+        if (physicalTables.isEmpty()) {
+            throw new IllegalArgumentException("At least 1 physical table needs to be provided");
+        }
```
cdeszaq: This check can actually be bundled into the retrieval of the coarsest grain down below. Not only does it make things more concise, but it also moves the error to the place where it matters and removes the potentially unsafe list access:

```java
.findFirst()
.orElseThrow(() -> {
        String message = "At least 1 physical table needs to be provided";
        LOG.error(message);
        return new IllegalArgumentException(message);
});
```

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,90 @@
...
+     * compatible with any of the <tt>PhysicalTables</tt>s, throw <tt>IllegalArgumentException</tt>.
+     *
+     * @param physicalTables  A set of <tt>PhysicalTable</tt>s among which the coarsest <tt>ZonedTimeGrain</tt>
+     * is to be returned.
+     *
+     * @return the coarsest <tt>ZonedTimeGrain</tt> among a set of <tt>PhysicalTables</tt>s
+     */
+    private static ZonedTimeGrain getCoarsestTimeGrain(Set<PhysicalTable> physicalTables) {
+        if (physicalTables.isEmpty()) {
+            throw new IllegalArgumentException("At least 1 physical table needs to be provided");
```
cdeszaq: Should log an ERROR here in case the exception doesn't bubble as expected.

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,90 @@
...
+                .sorted((table1, table2) -> granularityComparator.compare(table1, table2))
+                .collect(Collectors.toList());
+
+        // check to see if all <tt>ZonedTimeGrain</tt>'s is compatible with the coarsest <tt>ZonedTimeGrain</tt>
+        ZonedTimeGrain coarsestTimeGrain = sortedTables.get(0).getSchema().getTimeGrain();
+        List<PhysicalTable> incompatibles = sortedTables.stream()
+                .filter(table -> !table.getSchema().getTimeGrain().satisfiedBy(coarsestTimeGrain))
+                .collect(Collectors.toList());
+        if (!incompatibles.isEmpty()) {
+            throw new IllegalArgumentException(
```
cdeszaq: Should log an ERROR here in case the exception doesn't bubble as expected.

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,90 @@
...
+                    String.format(
+                            "The following ZonedTimeGrains are not compatible with the coarsest ZoneTimeGrain({}) - {}",
```
cdeszaq: Since it's not clear if there's an error with the coarsest grain, or with any of the other tables, we should use a more general message:

```java
        "There is no mutually satisfying grain for the physical tables: %s",
        sortedTables
);
LOG.error(message);
throw new IllegalStateException(message);
```

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,90 @@
...
+
+        // sort tables by <tt>ZonedTimeGrain</tt> in increasing order
+        List<PhysicalTable> sortedTables = physicalTables.stream()
+                .sorted((table1, table2) -> granularityComparator.compare(table1, table2))
+                .collect(Collectors.toList());
+
+        // check to see if all <tt>ZonedTimeGrain</tt>'s is compatible with the coarsest <tt>ZonedTimeGrain</tt>
+        ZonedTimeGrain coarsestTimeGrain = sortedTables.get(0).getSchema().getTimeGrain();
+        List<PhysicalTable> incompatibles = sortedTables.stream()
+                .filter(table -> !table.getSchema().getTimeGrain().satisfiedBy(coarsestTimeGrain))
```
cdeszaq: Rather than embedding this call string into the filter, it's generally better to pivot it to be declarative via `map` calls if possible. Once that's done, I think this helper method may be useful:

```java
private static Stream<ZonedTimeGrain> getSortedZonedTimeGrainStream(List<PhysicalTable> sortedTables) {
    return sortedTables.stream()
            .map(PhysicalTable::getSchema)
            .map(PhysicalTableSchema::getTimeGrain);
}
```

Especially if coupled with a more generic condition, rather than accumulating things for which we're not sure of the true error:

```java
if (getSortedZonedTimeGrainStream(sortedTables).anyMatch(ztg -> !ztg.satisfiedBy(coarsestTimeGrain))) {
```

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,90 @@
...
+     */
+    public MetricUnionCompositeTable(
+            @NotNull TableName name,
+            @NotNull Set<Column> columns,
+            @NotNull Set<PhysicalTable> physicalTables,
+            @NotNull Map<String, String> logicalToPhysicalColumnNames
+    ) {
+        super(
+                name,
+                getCoarsestTimeGrain(physicalTables),
```
michael-mclawhorn: The messaging might indicate that the problem is that there may or may not be a solution grain that satisfies all the tables, but since proving that there is not such a solution is non trivial without consulting all possible grains, that we're only trying to solve using grains on the tables being supplied.

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,90 @@
...
+                logicalToPhysicalColumnNames,
+                new MetricUnionAvailability(physicalTables, columns)
+        );
+    }
+
+    /**
+     * Returns the coarsest <tt>ZonedTimeGrain</tt> that satisfies all tables.
+     * <p>
+     * If the set of <tt>PhysicalTables</tt>'s is empty or the coarsest <tt>ZonedTimeGrain</tt> is not
+     * compatible with any of the <tt>PhysicalTables</tt>s, throw <tt>IllegalArgumentException</tt>.
```
michael-mclawhorn: Maybe instead of 'is not compatible with any' 'is compatible with all' would be clearer.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -109,4 +109,9 @@ public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constra
     protected Set<String> getColumnNames() {
         return columnNames;
     }
+
+    @Override
+    public String toString() {
+        return "concrete availability";
```
QubitPi: Sure! Good idea.

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,90 @@
...
+     * is to be returned.
+     *
+     * @return the coarsest <tt>ZonedTimeGrain</tt> among a set of <tt>PhysicalTables</tt>s
+     */
+    private static ZonedTimeGrain getCoarsestTimeGrain(Set<PhysicalTable> physicalTables) {
+        if (physicalTables.isEmpty()) {
+            throw new IllegalArgumentException("At least 1 physical table needs to be provided");
+        }
+
+        GranularityComparator granularityComparator = new GranularityComparator();
```
QubitPi: @cdeszaq A temp static instance is there in 1MetricUnionCompositeTable` to address your comment. Issue https://github.com/yahoo/fili/issues/224 has been created. I can pick it up once it's approved to be implemented.

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,90 @@
...
+
+        // sort tables by <tt>ZonedTimeGrain</tt> in increasing order
+        List<PhysicalTable> sortedTables = physicalTables.stream()
+                .sorted((table1, table2) -> granularityComparator.compare(table1, table2))
```
QubitPi: @cdeszaq Beautiful!

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
QubitPi: @cdeszaq I like this and really want to do it this way. [A compiler bug](http://stackoverflow.com/questions/39076077/throw-runtimeexception-inside-stream-with-optional-orelsethrow), however, stops me from doing this :( Let's talk about this offline

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
cdeszaq: Ok, that makes sense :-)

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
cdeszaq: It compiles for me, so perhaps just a JDK update is all you need?

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
cdeszaq: Feel free to pick up anything that's already called out as an issue whenever. If you think more details are needed on it before you can pick an issue up, please do ask (on the issue) for more clarification / thoughts

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,234 @@
...
+            );
+            LOG.error(message);
+            throw new IllegalArgumentException(message);
+        }
+
+        dataSourceNames = availabilitiesToAvailableColumns.entrySet().stream()
+                .map(Map.Entry::getKey)
+                .map(Availability::getDataSourceNames)
+                .flatMap(Set::stream)
+                .collect(Collectors.toSet());
```
QubitPi: @cdeszaq Beautiful design!

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,124 @@
...
+                });
+
+        if (physicalTables.stream()
+                .map(PhysicalTable::getSchema)
+                .map(PhysicalTableSchema::getTimeGrain)
+                .anyMatch(grain -> grain.satisfiedBy(coarsestTimeGrain))) {
+            String message = String.format(
+                    "There is no mutually satisfying grain among: %s",
+                    physicalTables
+            );
```
cdeszaq: Doesn't need to wrap

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,124 @@
...
+                .orElseThrow(() -> {
+                        String message = "At least 1 physical table needs to be provided";
+                        LOG.error(message);
+                        return new IllegalArgumentException(message);
+                });
+
+        if (physicalTables.stream()
+                .map(PhysicalTable::getSchema)
+                .map(PhysicalTableSchema::getTimeGrain)
+                .anyMatch(grain -> grain.satisfiedBy(coarsestTimeGrain))) {
```
cdeszaq: Shouldn't this be `allMatch`?

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Verifies and returns the coarsest <tt>ZonedTimeGrain</tt> if that satisfies all tables.
+     *
+     * @param coarsestGrain  The coarsest <tt>ZonedTimeGrain</tt> to be verified
+     * @param physicalTables  A set of <tt>PhysicalTable</tt>s whose <tt>ZonedTimeGrain</tt>s are checked to make sure
+     * they all satisfies with the given coarsest <tt>ZonedTimeGrain</tt>
+     *
+     * @return the coarsest <tt>ZonedTimeGrain</tt> among a set of <tt>PhysicalTables</tt>s
+     * @throws IllegalArgumentException when no PhysicalTable is provided or there is not mutually satisfying grain
+     * among the table's time grain
```
cdeszaq: The `mutually satisfying` part of this isn't valid any more, is it?

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * Verifies and returns the coarsest <tt>ZonedTimeGrain</tt> if that satisfies all tables.
+     *
+     * @param coarsestGrain  The coarsest <tt>ZonedTimeGrain</tt> to be verified
+     * @param physicalTables  A set of <tt>PhysicalTable</tt>s whose <tt>ZonedTimeGrain</tt>s are checked to make sure
+     * they all satisfies with the given coarsest <tt>ZonedTimeGrain</tt>
+     *
+     * @return the coarsest <tt>ZonedTimeGrain</tt> among a set of <tt>PhysicalTables</tt>s
```
cdeszaq: This isn't correct any more, is it?

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,124 @@
...
+     */
+    public MetricUnionCompositeTable(
+            @NotNull TableName name,
+            @NotNull Set<Column> columns,
+            @NotNull Set<PhysicalTable> physicalTables,
+            @NotNull Map<String, String> logicalToPhysicalColumnNames
+    ) {
+        super(
+                name,
+                verifyCoarsestTimeGrain(IntervalUtils.getCoarsestTimeGrain(physicalTables), physicalTables),
```
cdeszaq: Rather than make this `verify*` method return something, it would be better to just pass the coarsest grain up the constructor chain, and then call the `verify` method after `super(...)` has come back. `verifyCoarsestTimeGrain` can then do just what it's name says: Verify.

Also, I would consider making the name a bit more indicative of _what verification_ the method is doing. (eg. `verifyGrainSatisfiesAllTables`)

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,124 @@
...
+    private static ZonedTimeGrain verifyCoarsestTimeGrain(
+            Optional<ZonedTimeGrain> coarsestGrain,
+            Set<PhysicalTable> physicalTables
+    ) throws IllegalArgumentException {
+        ZonedTimeGrain coarsestTimeGrain = coarsestGrain
+                .orElseThrow(() -> {
+                        String message = "At least 1 physical table needs to be provided";
+                        LOG.error(message);
+                        return new IllegalArgumentException(message);
+                });
```
cdeszaq: Move this check up onto the call to `IntervalUtils.getCoarsestTimeGrain(physicalTables)` in the param list to the `super(...)` call.

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,124 @@
...
+     * @param coarsestGrain  The coarsest <tt>ZonedTimeGrain</tt> to be verified
+     * @param physicalTables  A set of <tt>PhysicalTable</tt>s whose <tt>ZonedTimeGrain</tt>s are checked to make sure
+     * they all satisfies with the given coarsest <tt>ZonedTimeGrain</tt>
+     *
+     * @return the coarsest <tt>ZonedTimeGrain</tt> among a set of <tt>PhysicalTables</tt>s
+     * @throws IllegalArgumentException when no PhysicalTable is provided or there is not mutually satisfying grain
+     * among the table's time grain
+     */
+    private static ZonedTimeGrain verifyCoarsestTimeGrain(
+            Optional<ZonedTimeGrain> coarsestGrain,
```
cdeszaq: In general, taking in an `Optional<>` is often a code smell that this method has too many responsibilities. Optionals are generally great for return types, but rarely good for input types.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,233 @@
...
+    Set<PhysicalTable> physicalTables
+
+    MetricUnionAvailability metricUnionAvailability
+
+    def setup() {
+        availability1 = Mock(Availability)
+        availability2 = Mock(Availability)
+
+        availability1.getDataSourceNames() >> Sets.newHashSet(TableName.of('source1'))
+        availability2.getDataSourceNames() >> Sets.newHashSet(TableName.of('source2'))
```
cdeszaq: List literals + coercion is a little cleaner in Groovy:

```groovy
[TableName.of('source1')] as Set
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,233 @@
...
+        physicalTables = [physicalTable1, physicalTable2] as Set
+    }
+
+    def "Metric columns are initialized by fetching columns from availabilities, not from physical tables"() {
+        given:
+        availability1.getAllAvailableIntervals() >> [
+                (metricColumn1): Collections.emptyList()
+        ]
+        availability2.getAllAvailableIntervals() >> [
+                (metricColumn2): Collections.emptyList()
```
cdeszaq: Groovy's list literals are a bit cleaner here: `[]`

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,233 @@
...
+
+        availabilitiesToAvailableColumns.containsKey(availability2)
+        Set<MetricColumn> availableColumns2 = availabilitiesToAvailableColumns.get(availability2)
+        availableColumns2.contains(metricColumn2)
+        !availableColumns2.contains(tableColumn1)
+        !availableColumns2.contains(tableColumn2)
+    }
+
+    @Unroll
+    def "checkDuplicateValue returns duplicates of metric column names from 2 physical tables in the case of #caseDescription"() {
```
cdeszaq: This description doesn't seem to make sense, since for many of the cases we're testing, there are no duplicates. Perhaps there's some more info from the `where` block that needs to be surfaced in the test name?

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,233 @@
...
+            actual.containsKey(duplicate)
+            actual.get(duplicate).size() == entry.value
+        }
+
+        where:
+        metricColumnNames1     | metricColumnNames2 | expected       | caseDescription
+        []                     | []                 | [:]            | 'both physical tables having empty metric column names'
+        ['metric1']            | []                 | [:]            | 'one empty table and one non-empty table'
+        ['metric1']            | ['metric1']        | ['metric1': 2] | 'both physical tables having the same metric column names'
+        ['metric1', 'metric2'] | ['metric1']        | ['metric1': 2] | 'two tables having intersections'
```
cdeszaq: It would be nice to add a case where there is more than 1 metric that has a duplicate, just to check it.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,233 @@
...
+                        (availability2): metricColumnNames2.collect{it -> new MetricColumn(it)} as Set
+                ]
+        )
+
+        then:
+        actual.size() == expected.size()
+        for (Map.Entry<String, Integer> entry : expected.entrySet()) {
+            MetricColumn duplicate = new MetricColumn(entry.key)
+            actual.containsKey(duplicate)
+            actual.get(duplicate).size() == entry.value
```
cdeszaq: When you're inside a `for` loop in a `then` block, you need to add `assert` to any tests you want it to check. Only top-level expressions are evaluated as tests by spock, so it doesn't see statements that are inside a conditional block like a `for` loop body.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,233 @@
...
+        ['metric1']            | ['metric2']        | [:]            | 'two tables having no intersections'
+    }
+
+    def "constructor throws IllegalArgumentException when 2 availabilities have the same metric column"() {
+        given:
+        availability1.getAllAvailableIntervals() >> [
+                (metricColumn1): Collections.emptyList()
+        ]
+        availability2.getAllAvailableIntervals() >> [
+                (metricColumn1): Collections.emptyList()
```
cdeszaq: Use list literal

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,233 @@
...
+        ]
+
+        when:
+        metricUnionAvailability = new MetricUnionAvailability(physicalTables, Collections.singleton(metricColumn1))
```
cdeszaq: List literal

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,233 @@
...
+        exception.message.startsWith("While constructing MetricUnionAvailability, Metric columns are not unique")
+    }
+
+
+    def "getDataSourceNames returns sources from availabilities not from physical tables"() {
+        given:
+        availability1.getAllAvailableIntervals() >> Collections.emptyMap()
+        availability2.getAllAvailableIntervals() >> Collections.emptyMap()
+
+        metricUnionAvailability = new MetricUnionAvailability(physicalTables, Collections.emptySet())
```
cdeszaq: Map literal, coerced list literal

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,233 @@
...
+
+        PhysicalTable physicalTable3 = Mock(PhysicalTable)
+        physicalTable3.getAvailability() >> metricUnionAvailability
+        MetricUnionAvailability outerMetricUnionAvailability = new MetricUnionAvailability(
+                Collections.singleton(physicalTable3),
+                Collections.emptySet()
```
cdeszaq: List literals

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,233 @@
...
+        )
+
+        expect:
+        outerMetricUnionAvailability.getDataSourceNames() == [TableName.of('source1'), TableName.of('source2')] as Set
+    }
+
+    def "getAllAvailableIntervals returns the combined intervals of configured metric columns of all availabilities"() {
```
cdeszaq: This test needs to cover dimensions as well

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,233 @@
...
+        given:
+        availability1.getAllAvailableIntervals() >> [
+                (metricColumn1): [new Interval('2018-01-01/2018-02-01')]
+        ]
+        availability2.getAllAvailableIntervals() >> [
+                (metricColumn2): [new Interval('2019-01-01/2019-02-01')]
+        ]
+
+        metricUnionAvailability = new MetricUnionAvailability(physicalTables, Collections.singleton(metricColumn1))
```
cdeszaq: Literal

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,233 @@
...
+
+        DataSourceConstraint dataSourceConstraint = new DataSourceConstraint(
+                Collections.emptySet(),
+                Collections.emptySet(),
+                Collections.emptySet(),
+                ['metric1', 'metric2'] as Set,
+                Collections.emptySet(),
+                Collections.emptySet(),
+                Collections.emptySet(),
+                Collections.emptyMap()
```
cdeszaq: More literal usage opportunities in this test (many places) as well.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,233 @@
...
+
+        metricUnionAvailability = new MetricUnionAvailability(physicalTables, Collections.singleton(metricColumn1))
+
+        expect:
+        metricUnionAvailability.getAllAvailableIntervals() == [
+                (metricColumn1): [new Interval('2018-01-01/2018-02-01')]
+        ]
+
+    }
+    def "getAvailableIntervals returns the intersection of requested columns when available intervals have #reason"() {
```
cdeszaq: Should have blank space and `@Unroll`

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/resolver/DataSourceConstraintSpec.groovy
```diff
@@ -0,0 +1,35 @@
...
+        DataSourceConstraint newDataSourceConstraint = new DataSourceConstraint(
+                Collections.emptySet(),
+                Collections.emptySet(),
+                Collections.emptySet(),
+                metricNames as Set,
+                Collections.emptySet(),
+                Collections.emptySet(),
+                Collections.emptySet(),
+                Collections.emptyMap()
```
cdeszaq: More literals

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,124 @@
...
+ * +---------+---------+---------+---------+---------+---------+
+ * |         |         |         |         |         |         |
+ * |         |         |         |         |         |         |
+ * +---------+---------+---------+---------+---------+---------+
+ * }
+ * </pre>
+ * and this joined table is backed by the <tt>MetricUnionAvailability</tt>
+ *
+ */
+public class MetricUnionCompositeTable extends BasePhysicalTable {
```
cdeszaq: Would be good to get a test for this class (ie. the constructor validity)

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,233 @@
...
+
+        availability1.getAvailableIntervals(_ as DataSourceConstraint) >> new SimplifiedIntervalList(
+                availableIntervals1.collect{it -> new Interval(it)} as Set
+        )
+        availability2.getAvailableIntervals(_ as DataSourceConstraint) >> new SimplifiedIntervalList(
+                availableIntervals2.collect{it -> new Interval(it)} as Set
+        )
+
+        schema1.getColumns(MetricColumn.class) >> Sets.newHashSet(metricColumn1)
+        schema2.getColumns(MetricColumn.class) >> Sets.newHashSet(metricColumn2)
```
cdeszaq: It would be good to include some dimensions in here to make sure it handles dimensions the way it should as well.

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/DataSourceConstraint.java
```diff
@@ -85,4 +118,32 @@ public DataSourceConstraint(DataApiRequest dataApiRequest, DruidAggregationQuery
...
+     * @return the new <tt>DataSourceConstraint</tt> instance with a new subset of metric names
+     */
+    public DataSourceConstraint withMetricIntersection(Set<MetricColumn> metricColumns) {
+        return new DataSourceConstraint(
+                requestDimensions,
+                filterDimensions,
+                metricDimensions,
+                metricColumns.stream()
+                        .filter(metricColumn -> metricNames.contains(metricColumn.getName()))
+                        .map(MetricColumn::getName)
```
QubitPi: Beautiful

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,258 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(MetricUnionAvailability.class);
+
+    private final Set<TableName> dataSourceNames;
+    private final Set<MetricColumn> metricColumns;
+    private final Map<Availability, Set<MetricColumn>> availabilitiesToAvailableColumns;
+
+    /**
+     * Constructor.
+     *
+     * @param physicalTables  A set of <tt>PhysicalTable</tt>s containing overlapping metric schema
```
QubitPi: Right

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,258 @@
...
+        // i.e. metricColumns
+        availabilitiesToAvailableColumns = physicalTables.stream()
+                .collect(
+                        Collectors.toMap(
+                                PhysicalTable::getAvailability,
+                                physicalTable ->
+                                        Sets.intersection(
+                                                Utils.getSubsetByType(
+                                                        physicalTable
+                                                                .getAvailability()
```
QubitPi: Beautiful!

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,258 @@
...
+        return dataSourceNames;
+    }
+
+    /**
+     * Combines and returns intervals of all availabilities' metric columns.
+     * <p>
+     * Intervals of the same metric column are associated with the same metric column key. Overlapping intervals under
+     * the same metric column key are collapsed into single interval.
+     *
+     * @return a map of metric column to all of its available intervals in union
```
QubitPi: @cdeszaq I think @garyluoex and @michael-mclawhorn have decided to have this method return only metric columns.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,258 @@
...
+     */
+    @Override
+    public Map<Column, List<Interval>> getAllAvailableIntervals() {
+        // get all availabilities
+        Set<Availability> availabilities = availabilitiesToAvailableColumns.keySet();
+
+        // take available interval maps from all availabilities and merge the maps together
+        return Stream.of(availabilities)
```
QubitPi: @cdeszaq I think so too. I remember you suggested splitting it to make it more readable. Maybe I interpreted it wrong.

cdeszaq: :) It's possible. Though I may have been referring to splitting some other aspect of what this was doing previously. I often tend to pack more into 1 chained expression, rather than breaking things up, much to others' chagrin I'm sure 😉

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
QubitPi: @cdeszaq This might be a groovy bug. If I use `[TableName.of('source1')] as Set` I get `NullPointerException: Cannot invoke method rightShift() on null object`

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
QubitPi: @cdeszaq Hmm...why do you think there are not duplicates? I thinks at least this one tests the case of duplicates:
`['metric1', 'metric2'] | ['metric1']        | ['metric1': 2] | 'two tables having intersections'`

This says metric1 is duplicate

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
cdeszaq: It sounds like something else is amiss. The `rightShift()` operation is being called on `availabilityX.getDataSourceNames()`, not on the set.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
cdeszaq: Some of them are cases where there are duplicates, but some of them are not. eg. `[]` + `[]` doesn't have any duplicates, yet the test description doesn't indicate that for that instance of the test.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
QubitPi: Nice catch!

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
QubitPi: Yes

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
QubitPi: @cdeszaq Hmm...I don't see that. Could you explain?

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
QubitPi: Thanks

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
QubitPi: Yes

fili-core/src/main/java/com/yahoo/bard/webservice/table/MetricUnionCompositeTable.java
```diff
@@ -0,0 +1,114 @@
...
+                IntervalUtils.getCoarsestTimeGrain(physicalTables).orElseThrow(() -> {
+                    String message = "At least 1 physical table needs to be provided";
+                    LOG.error(message);
+                    return new IllegalArgumentException(message);
+                }),
+                columns,
+                logicalToPhysicalColumnNames,
+                new MetricUnionAvailability(physicalTables, columns)
+        );
+        verifyGrainSatisfiesAllTables(IntervalUtils.getCoarsestTimeGrain(physicalTables).get(), physicalTables);
```
QubitPi: Beautiful!

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,252 @@
...
+ * {@code
+ * +---------------+---------------+---------------+---------------+---------------+
+ * | metricColumn1 | metricColumn2 | metricColumn3 | metricColumn4 | metricColumn5 |
+ * +---------------+---------------+---------------+---------------+---------------+
+ * | [1/10]        | [20/30]       | [5/15]        | [25/50]       | [90/100]      |
+ * +---------------+---------------+---------------+---------------+---------------+
+ * }
+ * </pre>
+ * Note that the joining availabilities must have completely different set of metric columns.
+ */
```
michael-mclawhorn: I would reword this and put it higher.  I would say that metric columns available on one availability must not exist on any other availabilities.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,252 @@
...
+     * For example, when a map of {availability1: [metricCol1, metricCol2], availability2: [metricCol1]} is passed to
+     * this method, the method returns {metricCol1: [availability1, availability2]}
+     *
+     * @param availabilityToAvailableColumns  A map from <tt>Availability</tt> to set of <tt>MetricColumn</tt>
+     * contained in that <tt>Availability</tt>
+     *
+     * @return duplicate values of <tt>MetricColumn</tt>s
+     */
+    private static Map<MetricColumn, Set<Availability>> getDuplicateValue(
+            Map<Availability, Set<MetricColumn>> availabilityToAvailableColumns
```
michael-mclawhorn: The should probably be getDuplicateValues (plural)

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -0,0 +1,252 @@
...
+                        )
+                );
+    }
+
+    @Override
+    public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constraints) {
+        return new SimplifiedIntervalList(
+                constructSubConstraint(constraints).entrySet().stream()
+                        .map(entry -> entry.getKey().getAvailableIntervals(entry.getValue()))
+                        .map(i -> (Set<Interval>) new HashSet<>(i))
```
garyluoex: what is `i`? Does method reference `HashSet::new` work?

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
garyluoex: I think it turns out that if you wrap the right part with a parenthesis, things will work.

PR #194, opened 2017-03-15
Add groups shields

cdeszaq:
We have a Users and a Developers Google Group. This PR adds shields to the README so people can find them!

This PR also removes shields that had been added previously, but were either not fully hooked up, or have stopped working. Also reduces the cache time on the Code Quality badge from some huge time down to 6 hours, so that we get faster feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Fili 
 ====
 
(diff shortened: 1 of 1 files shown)


michael-mclawhorn:
👍

PR #195, opened 2017-03-16
Implement PartitionAvailability & PartitionCompositeTable

QubitPi:


QubitPi:
@cdeszaq @garyluoex @michael-mclawhorn 

I haven't tested `getAvailableIntervals` because I'm not sure its implementation is correct. The current implementation of `getAvailableIntervals` takes a `DataSourceConstraint` and pass that constraint to all `Availabilities.getAvailableIntervals(DataSourceConstraint)`. My concern is should we pass a sub-constraint that is specific to an `Availability`? By sub-constraint I mean something similar used in [`MetricUnionAvailability`](https://github.com/ContributedOpenSourceProjects/fili/blob/600dd4939ac7aa8cabc5cb2050d7d198f971a438/fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java#L141)

garyluoex:
No, we do not need to since we should only call `getAvailableInterval` on the table in concern while skipping other tables that we do not care and all the tables should have the same schema.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,110 @@
...
+    public Set<TableName> getDataSourceNames() {
+        return sourceTables.stream()
+                .map(PhysicalTable::getTableName)
+                .collect(Collectors.toSet());
+    }
+
+    @Override
+    public Map<Column, List<Interval>> getAllAvailableIntervals() {
+        return getColumnToIntervalsMapping().entrySet().stream()
+                .filter(entry -> columns.contains(entry.getKey()))
```
cdeszaq: I feels like we might be able to use a single stream, rather than dumping to an intermediate collection, applying this filter before we start to aggregate intervals.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * An implementation of {@link com.yahoo.bard.webservice.table.availability.Availability}
+ * that unions different tables in druid that have the same columns together,
+ * in a way that the same column in two different table contains different values.
+ * Therefore if only values in one of the table is requested we only need to consider that table's availability.
+ */
+public class PartitionAvailability implements Availability {
+
+    private final Set<PhysicalTable> sourceTables;
```
QubitPi: @cdeszaq Did you mean ` private final Set<Availability> sourceAvailabilities;`?

cdeszaq: I think so?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,110 @@
...
+     * a {@link com.yahoo.bard.webservice.table.resolver.DataSourceConstraint} to a set of
+     * {@link com.yahoo.bard.webservice.table.PhysicalTable}'s
+     */
+    public PartitionAvailability(
+            @NotNull Set<PhysicalTable> sourceTables,
+            @NotNull Set<Column> columns,
+            @NotNull Function<DataSourceConstraint, Set<PhysicalTable>> partitionFunction
+    ) {
+        this.sourceTables = sourceTables;
+        this.columns = columns;
```
QubitPi: @cdeszaq I'm not sure if I understand "defensively" copy

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,110 @@
...
+        this.partitionFunction = partitionFunction;
+    }
+
+    @Override
+    public Set<TableName> getDataSourceNames() {
+        return sourceTables.stream()
+                .map(PhysicalTable::getTableName)
```
QubitPi: @cdeszaq https://github.com/yahoo/fili/issues/223

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
QubitPi: @cdeszaq Wow, beautiful design!

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,110 @@
...
+
+                unionedColumnToIntervalMapping.put(columnKey, union);
+            }
+        }
+
+        return unionedColumnToIntervalMapping;
+    }
+
+    @Override
+    public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constraints) {
```
QubitPi: Agree. I have refactored `getAllAvailableIntervals()`

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,86 @@
...
+    public PartitionAvailability(
+            @NotNull Set<Availability> sourceAvailabilities,
+            @NotNull Set<Column> columns,
+            @NotNull Function<DataSourceConstraint, Set<Availability>> partitionFunction
+    ) {
+        this.sourceAvailabilities = sourceAvailabilities;
+        this.columns = new HashSet<>();
+        columns.forEach(column -> {
+            this.columns.add(new Column(column.getName()));
+        });
```
cdeszaq: Why not something simpler like `this.columns = new HashSet<>(columns);`?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,86 @@
...
+public class PartitionAvailability implements Availability {
+
+    private final Set<Availability> sourceAvailabilities;
+    private final Set<Column> columns;
+    private final Function<DataSourceConstraint, Set<Availability>> partitionFunction;
+
+    /**
+     * Constructor.
+     *
+     * @param sourceAvailabilities  The set of availabilities that have the same columns
```
cdeszaq: We should go into more detail on the requirement here about same columns

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,86 @@
...
+     * @param columns  The set of configured columns
+     * @param partitionFunction  A function that transform a DataSourceConstraint to a set of
+     * Availabilities
```
cdeszaq: More detail needed on what the expectations are of the partition function. (ie. what is it partitioning on, how is it partitioning, etc.)

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,86 @@
...
+
+        this.partitionFunction = partitionFunction;
+    }
+
+    @Override
+    public Set<TableName> getDataSourceNames() {
+        return sourceAvailabilities.stream()
+                .map(Availability::getDataSourceNames)
+                .flatMap(Set::stream)
+                .collect(Collectors.toSet());
```
cdeszaq: We should do this once, in the constructor

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Override
+    public Map<Column, List<Interval>> getAllAvailableIntervals() {
+        return sourceAvailabilities.stream()
+                .map(Availability::getAllAvailableIntervals)
+                .map(Map::entrySet)
+                .flatMap(Set::stream)
+                .filter(entry -> columns.contains(entry.getKey()) && !entry.getValue().isEmpty())
```
cdeszaq: Why are we not keeping empty interval lists here? @michael-mclawhorn could you help explain why? (Also, it's probably a good idea to add comments in here _explaining_ why it makes sense to do so.)

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PartitionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,159 @@
...
+    def setup() {
+        availability1 = Mock(Availability)
+        availability2 = Mock(Availability)
+    }
+
+    @Unroll
+    def "getDataSourceNames returns #expected from #dataSourceNames1 and #dataSourceNames2"() {
+        given:
+        availability1.getDataSourceNames() >> Sets.newHashSet(dataSourceNames1.collect{it -> TableName.of(it)})
+        availability2.getDataSourceNames() >> Sets.newHashSet(dataSourceNames2.collect{it -> TableName.of(it)})
```
cdeszaq: Can be grooiver: `dataSourceNames.collect {TableName.of(it)} as Set`

- No need to wrap in a HashSet, can just coerce
- No need to declare `it` as a closure param, just use it if it's the only param

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PartitionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,159 @@
...
+    def "getAllAvailableIntervals returns merged intervals grouped by columns when two availabilities have #caseDescription"() {
+        given:
+        column1 = new Column('col1')
+        column2 = new Column(column2Name)
+
+        availability1.getAllAvailableIntervals() >> [
+                (column1): intervallist1.collect{it -> new Interval(it)}
+        ]
+        availability2.getAllAvailableIntervals() >> [
+                (column2): intervallist2.collect{it -> new Interval(it)}
```
cdeszaq: No need to wrap this if you don't have the necessary `it -> ` in the closures: `[(column2): intervallist2.collect {new Interval(it)}]`

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PartitionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,159 @@
...
+        given:
+        column1 = new Column('col1')
+        column2 = new Column('col2')
+
+        availability1.getAllAvailableIntervals() >> [
+                (column1): ['2018-01-01/2018-02-01']
+        ]
+        availability2.getAllAvailableIntervals() >> [
+                (column2): ['2018-01-01/2018-02-01']
+        ]
```
cdeszaq: No need to wrap these maps, just inline them since there's only 1 entry

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PartitionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,159 @@
...
+        given:
+        availability1.getAvailableIntervals(_ as DataSourceConstraint) >> new SimplifiedIntervalList(
+                availableIntervals1.collect{it -> new Interval(it)} as Set
+        )
+        availability2.getAvailableIntervals(_ as DataSourceConstraint) >> new SimplifiedIntervalList(
+                availableIntervals2.collect{it -> new Interval(it)} as Set
+        )
+
+        Function<DataSourceConstraint, Set<Availability>> partitionFunction = Mock(Function.class)
+        partitionFunction.apply(_ as DataSourceConstraint) >> Sets.newHashSet(availability1, availability2)
```
cdeszaq: Groovier: `([availability1, availability2] as Set)

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PartitionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,159 @@
...
+    def "getAvailableIntervals returns intervals in intersection when intervals have #reason"() {
+        given:
+        availability1.getAvailableIntervals(_ as DataSourceConstraint) >> new SimplifiedIntervalList(
+                availableIntervals1.collect{it -> new Interval(it)} as Set
+        )
+        availability2.getAvailableIntervals(_ as DataSourceConstraint) >> new SimplifiedIntervalList(
+                availableIntervals2.collect{it -> new Interval(it)} as Set
+        )
+
+        Function<DataSourceConstraint, Set<Availability>> partitionFunction = Mock(Function.class)
```
cdeszaq: No need for `.class` in groovy. Class names _are_ the class instance.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PartitionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,159 @@
...
+
+        when:
+        Map<Column, List<Interval>> actual = partitionAvailability.getAllAvailableIntervals()
+
+        then:
+        actual.containsKey(column1)
+        !actual.containsKey(column2)
+    }
+
+    def "getAvailableIntervals returns intervals in intersection when intervals have #reason"() {
```
cdeszaq: Missing `@Unroll`?

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PartitionAvailabilitySpec.groovy
```diff
@@ -0,0 +1,159 @@
...
+        partitionFunction.apply(_ as DataSourceConstraint) >> Sets.newHashSet(availability1, availability2)
+
+        partitionAvailability = new PartitionAvailability(
+                [availability1, availability2] as Set,
+                Collections.emptySet(),
+                partitionFunction
+        )
+
+        expect:
+        partitionAvailability.getAvailableIntervals(Mock(DataSourceConstraint)) == new SimplifiedIntervalList(
```
cdeszaq: Do we have a test that verifies the behavior of the partition stuff? In particular, I would expect a test that has many availabilities, and a partition function that maps to a subset, and the resulting available intervals includes only the subset of available intervals, not the full set of available intervals.

This test _looks_ like it might be doing it, but I don't think it is...

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,86 @@
...
+                                (value1, value2) -> SimplifiedIntervalList.simplifyIntervals(value1, value2)
+                        )
+                );
+    }
+
+    @Override
+    public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constraints) {
+        return new SimplifiedIntervalList(
+                partitionFunction.apply(constraints).stream()
+                        .map(availability -> availability.getAvailableIntervals(constraints))
```
cdeszaq: Is the MetricUnionAvailability just a specialized (ie. fixed) form of the PartitionAvailability? @garyluoex or @michael-mclawhorn, what do you think?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
QubitPi: @cdeszaq That's just my pervious personal preference. Will keep empty interval list in the next commit

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * An implementation of Availability that unions same columns from Druid tables together, in a way that the same column
+ * in 2 different availabilities contains different values. Therefore if only values in one of the availabilities is
+ * requested we only need to consider that table's availability.
+ * <p>
+ * Note that this is for union across availabilities with the same dimension columns who's values are non-overlapping.
+ * For example, two availabilities of the following
+ * <pre>
+ * {@code
+ * +-------------------------+--------------------------+
```
michael-mclawhorn: This documentation is totally wrong.

cdeszaq: Wrong how? What's wrong about it? Feedback pointing in the right direction would be nice.

michael-mclawhorn: Ok. I'm confused.  I'm pretty sure when I was looking at this code I was reading the metric union documentation in here.  I think perhaps there was a gap between my local copy and the current commit.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,137 @@
...
+ * </pre>
+ * are joined into a partition availability below
+ * <pre>
+ * {@code
+ * +-------------------------+--------------------------+
+ * | column1(availability1)  |  column1(availability2)  |
+ * +-------------------------+--------------------------+
+ * | [2017-01-01/2017-02-01] |  [2017-03-01/2017-04-01] |
+ * +-------------------------+--------------------------+
+ * }
```
michael-mclawhorn: This documentation highlights my confusion.  Partition doesn't use column information at all.  It simply uses the availability of child tables without considering columns.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param sourceAvailabilities  The set of availabilities that have the same columns whose values are
+     * non-overlapping
+     * @param columns  The set of configured columns
+     * @param partitionFunction  A function that transform a DataSourceConstraint to a set of
+     * Availabilities. The function is composed of the following works functions/transformations
+     * <pre>
```
michael-mclawhorn: Technically Partition may assume common columns, but it doesn't require it.  It would be better to emphasize that the availability is the result of considering the availability of the child tables, full stop.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
michael-mclawhorn: Actually, no.  
Metric union contains inner-join semantics.
Partition applies more of an unconstrained outer join behavior.

garyluoex:
This PR is reopened here and merged #244

PR #197, opened 2017-03-20
Fix to prevent NPE when attempting to bind segment file columns not c…

michael-mclawhorn:
…onfigured in the dimension dictionary.

(Columns in segment files don't all need to be configured, unused columns should be skipped)
Also, warn if NO columns are resolved.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ImmutableAvailability.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ImmutableAvailability.java
@@ -22,7 +22,9 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
+import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 /**
@@ -31,6 +33,9 @@ import java.util.stream.Collectors;
 public class ImmutableAvailability implements Availability {
 
     private static final Logger LOG = LoggerFactory.getLogger(ImmutableAvailability.class);
+
+    public static final String FILTERED_DIMENSION_FORMAT =
+            "Found unknown dimension name '{}' when building availability for table: '{}'.";
 
     private final TableName name;
     private final Map<Column, List<Interval>> columnIntervals;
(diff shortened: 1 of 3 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ImmutableAvailability.java
```diff
@@ -97,11 +101,15 @@ public ImmutableAvailability(
                 entry -> new SimplifiedIntervalList(entry.getValue());
 
         Map<Column, List<Interval>> map = dimensionIntervals.entrySet().stream()
+                .filter(entry -> dimensionDictionary.findByApiName(entry.getKey()) != null)
```
archolewa: Would it make sense to have a debug level log statement here that logs the fact that we're filtering out a dimension value? I could see people getting confused if they are working on exposing a dimension across multiple services, and they accidentally skipped a step.

archolewa:
👍

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ImmutableAvailability.java
```diff
@@ -97,11 +101,15 @@ public ImmutableAvailability(
...
                 .collect(Collectors.toMap(dimensionKeyMapper, valueMapper));
         map.putAll(
                 metricIntervals.entrySet().stream()
                         .collect(Collectors.toMap(metricKeyMapper, valueMapper))
         );
+        if (map.size() == 0) {
```
cdeszaq: Definitely should use `isEmpty` here

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ImmutableAvailability.java
```diff
@@ -97,11 +101,15 @@ public ImmutableAvailability(
...
+            LOG.warn("No legal availability columns discovered when building availability instance.");
```
cdeszaq: We should include some info about this availability so that we have a sense of _what_ failed. The `tableName` is probably a good candidate.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ImmutableAvailability.java
cdeszaq: I would put it at `trace` rather than debug, since it's expected to happen during a migration, but I agree, it might not be a bad thing to indicate that something is being filtered out. I really wish I understood grouped streams better, because "split on include/exclude, log the exclude" seems to be a very useful pattern.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/ImmutableAvailabilitySpec.groovy
```diff
@@ -0,0 +1,58 @@
...
+    Set<Dimension> dimensions = [dimension1, dimension2] as Set
+    DimensionDictionary dictionary = new DimensionDictionary(dimensions)
+
+    def "Empty segment data does not create an error"() {
+        setup:
+        Map<String, List<Interval>> segmentDimensionData = [:]
+        Map<String, List<Interval>> segmentMetricData = [:]
+
+        expect: "segment availability is passed to be bound to dimension and metric columns"
+        (ImmutableAvailability.buildAvailabilityMap(segmentDimensionData, segmentMetricData, dictionary)).size() == 0
```
cdeszaq: Extra parens here are not needed

michael-mclawhorn: 👍

PR #198, opened 2017-03-20
Restored query based schema injection override capability removed as …

michael-mclawhorn:
…a side effect of immutable schemas.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidResponseParser.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidResponseParser.java
@@ -9,6 +9,8 @@ import com.yahoo.bard.webservice.data.dimension.DimensionRow;
 import com.yahoo.bard.webservice.data.metric.MetricColumn;
 import com.yahoo.bard.webservice.druid.model.DefaultQueryType;
 import com.yahoo.bard.webservice.druid.model.QueryType;
+import com.yahoo.bard.webservice.druid.model.query.DruidAggregationQuery;
+import com.yahoo.bard.webservice.table.Column;
 
 import com.fasterxml.jackson.databind.JsonNode;
 
@@ -21,6 +23,7 @@ import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Stream;
 
 import javax.inject.Singleton;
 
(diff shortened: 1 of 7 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidResponseParser.java
```diff
@@ -305,4 +308,16 @@ private Object getNodeValue(JsonNode node) {
                 node.isNull() ? null :
                 node;
     }
+
+    /**
+     * Produce the schema-defining columns for a given druid query.
+     * Pass through to druid query to allow for possible behavior customization on injected DruidResponseParsers.
```
cdeszaq: This is an implementation detail, right? Shouldn't this go _inside_ the method, since it doesn't really impact it's contract?

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/DruidAggregationQuery.java
```diff
@@ -96,4 +101,22 @@
...
+     * @return A stream of columns based on the signature of the Druid Query.
+     */
+    default Stream<Column> buildSchemaColumns() {
+        return Stream.of(
+                getDimensions().stream()
+                        .map(DimensionColumn::new),
+                getAggregations().stream()
+                        .map(Aggregation::getName)
+                        .map(MetricColumn::new),
+                getPostAggregations().stream()
```
cdeszaq: Instead of doing streams over both of these, consider adding a `getMetricFieldNames()` method that gives back aggregations _and_ post-aggregations (since they both implement `metricField`.

It would make this method a little bit simpler, and provide a useful, reusable hook for other things to get a unified view of aggs and post-aggs.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/DruidResponseParserSpec.groovy
```diff
@@ -404,7 +407,19 @@ class DruidResponseParserSpec extends Specification {
...
+        setup:
+        Stream<Column> columnStream = Mock(Stream)
+        DruidAggregationQuery query = Mock(DruidAggregationQuery)
+        1 * query.buildSchemaColumns() >> columnStream
 
+        when:
+        def result = responseParser.buildSchemaColumns(query)
+
+        then:
+        result == columnStream
```
cdeszaq: Just make this an `expect`, since it's just "grab the output, and check it", without much other "check the state of the world" aspects to it.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/DruidAggregationQuery.java
michael-mclawhorn: This?
```

    default Stream<String> getMetricFieldNames() {
        return Stream.concat(
                getAggregations().stream()
                    .map(Aggregation::getName),
                getPostAggregations().stream()
                        .map(PostAggregation::getName)
        );
    }
    /**
     * Produce the schema-defining columns for a given druid query.
     *
     * @return A stream of columns based on the signature of the Druid Query.
     */
    default Stream<Column> buildSchemaColumns() {
        return Stream.of(
                getDimensions().stream()
                        .map(DimensionColumn::new),
                getMetricFieldNames().map(MetricColumn::new)
        ).flatMap(Function.identity());
    }
```

PR #199, opened 2017-03-23
Added 'isValidFor' that takes a Logical Metric to the ApiMetricName i…

michael-mclawhorn:
…nterface.  Move LogicalTableSchema to use that to filter metrics instead of the granularity only one.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/names/ApiMetricName.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/names/ApiMetricName.java
@@ -36,7 +36,7 @@ public interface ApiMetricName extends FieldName {
      * This version only takes the time grain into account.  The default implementation simply sets a required minimum
      * grain for the metric.
      * <p>
-     * An example of this is a DailyAverage metric than doesn't make sense to query by the HOUR grain, but does at WEEK.
+     * An example of this is a DailyAverage metric that doesn't make sense to query by the HOUR grain, but does at WEEK.
      *
      * @param granularity  TimeGrain to determine validity for
      *
@@ -56,14 +56,14 @@ public interface ApiMetricName extends FieldName {
      * LogicalTable instances at configuration time.  This version allows the logical metric itself to be used for
      * filtering.
      * <p>
-     * An example of this is a DailyAverage metric than doesn't make sense to query by the HOUR grain, but does at WEEK.
+     * An example of this is a DailyAverage metric that doesn't make sense to query by the HOUR grain, but does at WEEK.
      *
-     * @param logicalMetric  The metric whose validity is being tested.
      * @param granularity  TimeGrain to determine validity for
+     * @param logicalMetric  The metric whose validity is being tested.
      *
      * @return True if the ApiMetricName is valid for the time grain
      */
-    default boolean isValidFor(LogicalMetric logicalMetric, Granularity granularity) {
+    default boolean isValidFor(Granularity granularity, LogicalMetric logicalMetric) {
         return isValidFor(granularity);
     }
 
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/names/ApiMetricName.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/names/ApiMetricName.java
@@ -2,6 +2,7 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.data.config.names;
 
+import com.yahoo.bard.webservice.data.metric.LogicalMetric;
 import com.yahoo.bard.webservice.data.time.TimeGrain;
 import com.yahoo.bard.webservice.druid.model.query.AllGranularity;
 import com.yahoo.bard.webservice.druid.model.query.Granularity;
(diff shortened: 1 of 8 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/config/names/ApiMetricName.java
```diff
@@ -45,6 +51,23 @@ default boolean isValidFor(Granularity granularity) {
     }
 
     /**
+     * Determine if this API Metric Name is valid for the given time grain.
+     * This capability is provided as a convenience for configurers to filter metrics from a TableGroup onto a set of
+     * LogicalTable instances at configuration time.  This version allows the logical metric itself to be used for
+     * filtering.
+     * <p>
+     * An example of this is a DailyAverage metric than doesn't make sense to query by the HOUR grain, but does at WEEK.
```
cdeszaq: s/than/that

fili-core/src/main/java/com/yahoo/bard/webservice/table/LogicalTableSchema.java
```diff
@@ -45,10 +49,30 @@ public LogicalTableSchema(TableGroup tableGroup, Granularity granularity, Metric
         return Stream.concat(
                 tableGroup.getDimensions().stream()
                 .map(DimensionColumn::new),
```
cdeszaq: I think `.map(...)` should be indented 2 levels here.

fili-core/src/main/java/com/yahoo/bard/webservice/table/LogicalTableSchema.java
```diff
@@ -45,10 +49,30 @@ public LogicalTableSchema(TableGroup tableGroup, Granularity granularity, Metric
...
+     * Buid a stream of metric columns by filtering the table group's collection of ApiMetricNames.
+     * Metric names are resolved and then filtered considering the table granularity and the metric referenced.
+     *
+     * @param apiMetricNames  The names of the apiMetrics being bound and filtered
+     * @param granularity  The grain used to filter those metric names
+     * @param metricDictionary  The dictionary to resolve the logical metric instances from
+     *
+     * @return A stream of metric columns, filtered for compatibility with the grain.
+     */
+    private static Stream<MetricColumn> buildMetricColumns(
```
cdeszaq: Should this return `Stream<MetricColumn>` or `Stream<LogicalMetricColumn>`? If the former, why the type relaxation?

fili-core/src/main/java/com/yahoo/bard/webservice/table/LogicalTableSchema.java
```diff
@@ -45,10 +49,30 @@ public LogicalTableSchema(TableGroup tableGroup, Granularity granularity, Metric
...
+            Collection<ApiMetricName> apiMetricNames,
+            Granularity granularity,
+            MetricDictionary metricDictionary
+    ) {
+        return apiMetricNames.stream()
+                .collect(Collectors.toMap(Function.identity(), entry -> metricDictionary.get(entry.getApiName())))
+                .entrySet().stream()
+                .filter(entry -> entry.getKey().isValidFor(entry.getValue(), granularity))
+                .map(Map.Entry::getValue)
```
cdeszaq: Can this be simplified? I don't _think_ we need to build a `Map<ApiMetricName, LogicalMetric>`, then filter by the keys, only to then grab the values. I _think_ we could just filter out the APIMetricNames, and then convert the remaining ones to LogicalMetrics. I've not tested this code change though.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/names/ApiMetricName.java
```diff
@@ -45,6 +51,23 @@ default boolean isValidFor(Granularity granularity) {
...
+     *
+     * @param logicalMetric  The metric whose validity is being tested.
+     * @param granularity  TimeGrain to determine validity for
+     *
+     * @return True if the ApiMetricName is valid for the time grain
+     */
+    default boolean isValidFor(LogicalMetric logicalMetric, Granularity granularity) {
```
cdeszaq: Can we flip the order of these parameters? It's a little odd to have an overload that drops parameters from the front, rather than the back.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/LogicalTableSchemaSpec.groovy
```diff
@@ -0,0 +1,39 @@
...
+        LogicalMetric logicalMetric2 = Mock(LogicalMetric)
+
+        ApiMetricName apiName1 = Mock(ApiMetricName) {
+            getApiName() >> metricName1
+        }
+        ApiMetricName apiName2 = Mock(ApiMetricName) {
+            getApiName() >> metricName2
+        }
+        1 * apiName1.isValidFor(logicalMetric1, _) >> true
+        1 * apiName2.isValidFor(logicalMetric2, _) >> false
```
cdeszaq: This is a small thing, but could we give these better names that "1" and "2"? Perhaps something like "validMetric" and "invalidMetric" so that it was easier to connect the dots about what's being tested, and why these mock values make sense?

fili-core/src/main/java/com/yahoo/bard/webservice/table/LogicalTableSchema.java
michael-mclawhorn:
```
        return apiMetricNames.stream()
                .filter(name -> name.isValidFor(metricDictionary.get(name), granularity))
                .map(metricDictionary::get)
                .map(LogicalMetricColumn::new);
```

PR #200, opened 2017-03-27
Add implementations of PermissiveAvailability & PermissiveConcretePhy…

michael-mclawhorn:
…sicalTable

Add a hook for getting dimension dictionary in application binder factory

Rebase of PR #190

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,17 @@ pull request if there was one.
 Current
 -------
 ### Added:
-- [CompositePhsyicalTable Core Components Refactor](https://github.com/yahoo/fili/pull/179)
+- [CompositePhysicalTable Core Components Refactor](https://github.com/yahoo/fili/pull/179)
     * Added `ConcretePhysicalTable` and `ConcreteAvailability` to model table in druid datasource and its availabillity in the new table availability structure
     * Added class variable for `DataSourceMetadataService` and `ConfigurationLoader` into `AbstractBinderFactory` for application to access
+
+- [PermissiveAvailability and PermissiveConcretePhysicalTable](https://github.com/yahoo/fili/pull/190)
+    * Added `PermissiveConcretePhysicalTable` and `PermissiveAvailability` to model table in druid datasource and its availability in the new table availability structure.
+    `PermissiveConcretePhysicalTable` and `PermissiveAvailability` extends `ConcretePhysicalTable` and `ConcreteAvailability`, respectively.
+    `PermissiveAvailability` differs from `ConcreteAvailability` in the way of returning available intervals: `ConcreteAvailability` returns
+    the available intervals constraint by `DataSourceConstraint` and provides them in intersection. `PermissiveAvailability`, however, returns
+    them without constraint from `DataSourceConstraint` and provides them in union. `PermissiveConcretePhysicalTable` is different from `ConcretePhysicalTable`
+    in that the former is backed by `PermissiveAvailability` while the latter is backed by `ConcreteAvailability`.
 
 - [Refactor DatasourceMetaDataService to fit composite table needs](https://github.com/yahoo/fili/pull/173)
     * `DataSourceMetadataService` also stores interval data from segment data as intervals by column name map and provides method `getAvailableIntervalsByTable` to retrieve it
@@ -52,14 +60,14 @@ Current
 ### Changed:
 
 - [CompositePhsyicalTable Core Components Refactor](https://github.com/yahoo/fili/pull/179)
-    * `TableLoader` now takes an additional constructor argument `DataSourceMetadataService` for creating tables
+    * `TableLoader` now takes an additional constructor argument `DataSourceMetadataService` for creating tables     
     * `findMissingRequestTimeGrainIntervals` method in `PartialDataHandler` now takes `DataSourceConstraint`
-
+ 
 - [Restored flexibility about columns for query from DruidResponseParser](https://github.com/yahoo/fili/pull/198)
     * Immutable schemas prevented custom query types from changing `ResultSetSchema` columns.
     * Columns are now sourced from `DruidResponseParser` and default implemented on `DruidAggregationQuery`
 
-- [Refactor DatasourceMetadataService to fit composite table needs](https://github.com/yahoo/fili/pull/173)
+- [Refactor DatasourceMetaDataService to fit composite table needs](https://github.com/yahoo/fili/pull/173)
     * `BasePhysicalTable` now stores table name as the `TableName` instead of `String`
     * `SegmentInfo` now stores dimension and metrics from segment data for constructing column to available interval map
 
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
@@ -178,7 +178,6 @@ public abstract class AbstractBinderFactory implements BinderFactory {
 
     private final TaskScheduler loaderScheduler = new TaskScheduler(LOADER_SCHEDULER_THREAD_POOL_SIZE);
 
-
     /**
      * Constructor.
      */
@@ -231,9 +230,6 @@ public abstract class AbstractBinderFactory implements BinderFactory {
 
                 //Initialize the metrics filter helper
                 FieldConverterSupplier.metricsFilterSetBuilder = initializeMetricsFilterSetBuilder();
-
-                // Build the datasource metadata service containing the data segments
-                bind(getDataSourceMetadataService()).to(DataSourceMetadataService.class);
 
                 // Build the datasource metadata service containing the data segments
                 bind(getDataSourceMetadataService()).to(DataSourceMetadataService.class);
@@ -315,8 +311,6 @@ public abstract class AbstractBinderFactory implements BinderFactory {
                 // Call post-binding hook to allow for additional binding
                 afterBinding(this);
             }
-
-
         };
     }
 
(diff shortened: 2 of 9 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -234,6 +235,9 @@ protected void configure() {
                 // Build the datasource metadata service containing the data segments
                 bind(getDataSourceMetadataService()).to(DataSourceMetadataService.class);
 
+                // Build the datasource metadata service containing the data segments
+                bind(getDataSourceMetadataService()).to(DataSourceMetadataService.class);
```
cdeszaq: Duplicate

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -105,4 +105,8 @@ public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constra
     protected Map<String, List<Interval>> getLatestAvailableIntervalsByTable() {
         return metadataService.getAvailableIntervalsByTable(name);
     }
+
+    public Set<String> getColumnNames() {
```
cdeszaq: `protected`?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -0,0 +1,65 @@
...
+     * missing intervals and returning unions of available intervals
+     *
+     * @return the union of all available intervals
+     */
+    @Override
+    public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint ignoredConstraints) {
+        Map<String, List<Interval>> allAvailableIntervals = getLatestAvailableIntervalsByTable();
+        return new SimplifiedIntervalList(
+                getColumnNames().stream()
+                        .map(columnName -> allAvailableIntervals.getOrDefault(columnName, Collections.emptyList()))
```
garyluoex: I believe this filtering process is not needed since all the things returned from `allAvailableIntervals` are filtered already. Simply pass `allAvailbleIntervals.values()` into `SimplifiedIntervalList` should be sufficient.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/AvailabilityTestingUtils.groovy
```diff
@@ -62,4 +62,18 @@ class AvailabilityTestingUtils extends Specification {
...
+     * Construct a map of Column object to its all available intervals
+     *
+     * @param map a map of Column in String representation to its all available intervals in String representations
+     *
+     * @return a map of Column object to its all available intervals
+     */
+    static Map<Column, Set<Interval>> constructColumnToIntervalMap(Map<String, Set<String>> columnToIntervalInString) {
+        return columnToIntervalInString.collectEntries{
+            key,
+            value -> [ (new Column(key)): value.collect{new Interval(it)} ]
```
garyluoex: This method should take some kind of `schema`, we lose the ability to distinguish `DimensionColumn` and `MetricColumn`. This `Column` constructor should only be used in tests.

CHANGELOG.md
```diff
@@ -8,10 +8,18 @@ pull request if there was one.
...
 -------
 ### Added:
-- [CompositePhsyicalTable Core Components Refactor](https://github.com/yahoo/fili/pull/179)
+- [CompositePhysicalTable Core Components Refactor](https://github.com/yahoo/fili/pull/179)
     * Added `ConcretePhysicalTable` and `ConcreteAvailability` to model table in druid datasource and its availabillity in the new table availability structure
     * Added class variable for `DataSourceMetadataService` and `ConfigurationLoader` into `AbstractBinderFactory` for application to access
 
+- [PermissiveAvailability and PermissiveConcretePhysicalTable](https://github.com/yahoo/fili/pull/190)
+    * Added `PermissiveConcretePhysicalTable` and `PermissiveAvailability` to model table in druid datasource and its availability in the new table availability structure.
+    `PermissiveConcretePhysicalTable` and `PermissiveAvailability` extends `ConcretePhysicalTable` and `ConcreteAvailability`, respectively.
```
garyluoex: these things are too detailed, remove it, leave it in the PR description or javadoc, don't put it in CHANNGELOG.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
garyluoex: I think this method is a bit dull to be here, remove it makes less code to maintain without losing basically anything.

cdeszaq: The loss is that `PermissiveAvailability` wouldn't be able to get at the column names, which it needs it. So, I think making it `protected` is the best bet for it, rather than dropping it.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
cdeszaq: Really? I don't see where it's filtered already. `getLatestAvailableIntervalsByTable()` is just `metadataService.getAvailableIntervalsByTable(name)`, which doesn't do any filtering I don't think. Am I missing something?

garyluoex: `allAvailableIntervals` filters it.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/AvailabilityTestingUtils.groovy
cdeszaq: Actually, I don't think this method is even used. Perhaps we can just drop it entirely? (IntelliJ doesn't think so, anyways)

CHANGELOG.md
cdeszaq: It's OK to have too much detail in the changelog. It's easier to trim it out than to add it in, so it's better to err on the side of too much.

I tend to go back and clean up the changelogs when we go to cut the stable version anyways, so I wouldn't worry about this.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
garyluoex: ok got it

garyluoex:
Ops, accidentally clicked the approve.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
cdeszaq: `ConcreteAvailability::getAllAvilableIntervals` does, but `ConcreteAvailability::getLatestAvailableIntervalsByTable` does not. Since the local variable `allAvailableInterbals` is the latter, not the former, it's not filtered.

That said (and perhaps this is what you mean), calling `getAllAvailableIntervals()` here _would_ allow the filter to not be applied again, but that's not `Map<String, List<Interval>>`, it's `Map<Column, List<Interval>>` so some unwrapping is needed, as well as the other downside of moving from collection to collection, rather than just a single stream.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -0,0 +1,65 @@
...
+     * @param ignoredConstraints  Data constraint containing columns and api filters. Constrains are ignored, because
+     * <tt>PermissiveAvailability</tt> returns as much available intervals as possible by, for example, allowing
+     * missing intervals and returning unions of available intervals
+     *
+     * @return the union of all available intervals
+     */
+    @Override
+    public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint ignoredConstraints) {
+        Map<String, List<Interval>> allAvailableIntervals = getLatestAvailableIntervalsByTable();
+        return new SimplifiedIntervalList(
```
cdeszaq: This extra constructor actually isn't needed, because the collector on the stream will give you an SIL.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
garyluoex: ok

cdeszaq: The other alternative, after looking at it more, would be this:

```java
return getAllAvailableIntervals().values().stream()
        .flatMap(List::stream)
        .collect(SimplifiedIntervalList.getCollector());
```

The downside is that the alternative goes through a temporary map, rather than just operating on the stream, and would end up making more SIL objects / transforms than the impl in the PR.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -102,7 +102,11 @@ public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constra
      *
      * @return map of column name to list of avialable intervals
      */
-    protected Map<String, List<Interval>> getLatestAvailableIntervalsByTable() {
+    protected Map<String, List<Interval>> getAvailableIntervalsByTable() {
         return metadataService.getAvailableIntervalsByTable(name);
     }
+
+    public Set<String> getColumnNames() {
```
cdeszaq: Protected?

PR #202, opened 2017-03-28
Restore non-default query type support in TestDruidWebService

michael-mclawhorn:
Base TableDataSource serialization on ConcretePhysicalTable fact name


cdeszaq:
👍

PR #205, opened 2017-03-28
Expose the durations stored in the RequestLog.

archolewa:
--Some customers may have their own custom logging format that differs
significantly from how the `BardLoggingFilter` logs the `RequestLog`.
However, they may still want access to the many timings that the
`RequestLog` tracks.

The ideal solution is to rearchitect the
`RequestLog` to make it easy to replace with a custom logging scheme
and/or make it truly a-la carte. However, as a stopgap, simply
providing a means of getting at the durations is sufficient.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
@@ -127,7 +127,6 @@ public class RequestLog {
         MDC.put(ID_KEY, logId);
     }
 
-
     /**
      * Get the aggregate durations for this request.
      *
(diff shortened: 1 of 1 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
```diff
@@ -127,22 +127,43 @@ private void init() {
...
      *
      * @return the map containing all the recorded times per phase in milliseconds
      */
-    private Map<String, Long> getDurations() {
+    private Map<String, Long> durations() {
         return times.values().stream().collect(Collectors.toMap(TimedPhase::getName, TimedPhase::getDuration));
     }
 
+
+    /**
```
michael-mclawhorn: whitespace: vertical

fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java
```diff
@@ -127,22 +127,43 @@ private void init() {
         MDC.put(ID_KEY, logId);
     }
 
+
+    /**
```
michael-mclawhorn: Whitespace: vertical

PR #206, opened 2017-03-29
Implement MetricUnionAvailability and MetricUnionCompositeTable

QubitPi:
Same as forke-based https://github.com/yahoo/fili/pull/193, which cannot show build status.


QubitPi:
Closed. Same PR as https://github.com/yahoo/fili/pull/193

PR #207, opened 2017-03-29
Refactor PhysicalTableDefinition and Update BaseTableLoader

garyluoex:
* `PhysicalTableDefinition` is now an abstract class, construct using `ConcretePhysicalTableDefinition` instead
* `PhysicalTableDefinition` now requires a `build` methods to be implemented that builds a physical table
* `BaseTableLoader` now constructs physical tables by calling `build` method on `PhysicalTableDefinition`s in `buildPhysicalTablesWithDependency`
* `buildDimensionSpanningTableGroup` method in `BaseTableLoader` now uses `loadPhysicalTablesWithDependency` instead of removed `loadPhysicalTables`
* `buildDimensionSpanningTableGroup` method in `BaseTableLoader` now does not take druid metric as arguments, instead `PhysicalTableDefinition` does

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
@@ -177,7 +177,7 @@ public abstract class BaseTableLoader implements TableLoader {
      * <p>
      * note: current physical table not loaded into physical table dictionary, only dependencies will
      *
-     * @param currentTableName  Iterator for the mutable definition
+     * @param currentTableName  Table name of the table being built
      * @param availableTableDefinitions  A map of table name to table definition that are available globally
      * @param dictionaries  Contains both dimension and physical table dictionary for building and dependency resolution
      *
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
@@ -26,7 +26,6 @@ import org.slf4j.LoggerFactory;
 import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -71,11 +70,11 @@ public abstract class BaseTableLoader implements TableLoader {
     public abstract void loadTableDictionary(ResourceDictionaries dictionaries);
 
     /**
-     * Builds a table group.
+     * Builds a table group that derive its dimensions by taking the union of all the underlying physical dimensions.
      * <p>
      * Builds and loads the physical tables from table definitions for the current table group.
      *
-     * @param tableNames  Set of table name of tables belonging to this table group
+     * @param currentTableGroupTableNames  Set of table name of tables belonging to this table group
      * @param tableDefinitions  A list of config objects for building physical tables and its dependent physical tables
      * @param dictionaries  The container for all the data dictionaries
      * @param apiMetrics  The set of metric names surfaced to the api
@@ -83,27 +82,33 @@ public abstract class BaseTableLoader implements TableLoader {
      * @return A table group binding all the tables for a logical table view together.
      */
     public TableGroup buildDimensionSpanningTableGroup(
-            Set<TableName> tableNames,
+            Set<TableName> currentTableGroupTableNames,
             Set<PhysicalTableDefinition> tableDefinitions,
             ResourceDictionaries dictionaries,
             Set<ApiMetricName> apiMetrics
     ) {
-        Map<String, PhysicalTableDefinition> availableTableDefinitions =
-                buildPhysicalTableDefinitionDictionary(tableDefinitions);
+        Map<String, PhysicalTableDefinition> availableTableDefinitions = buildPhysicalTableDefinitionDictionary(
+                tableDefinitions
+        );
 
         PhysicalTableDictionary physicalTableDictionary = dictionaries.getPhysicalDictionary();
 
         // Get the physical table from physical table dictionary, if not exist, build it and put it in dictionary
-        LinkedHashSet<PhysicalTable> physicalTables = tableNames.stream()
+        LinkedHashSet<PhysicalTable> physicalTables = currentTableGroupTableNames.stream()
+                .map(TableName::asName)
                 .map(
                         tableName -> physicalTableDictionary.computeIfAbsent(
-                                tableName.asName(),
-                                key -> buildPhysicalTablesWithDependency(key, availableTableDefinitions, dictionaries)
+                                tableName,
+                                tableNameKey -> buildPhysicalTableWithDependency(
+                                        tableNameKey,
+                                        availableTableDefinitions,
+                                        dictionaries
+                                )
                         )
                 ).collect(Collectors.toCollection(LinkedHashSet::new));
 
 
-        //Derive the logical dimensions by taking the union of all the physical dimensions
+        // Derive the dimensions by taking the union of all the physical dimensions
         Set<Dimension> dimensions = physicalTables.stream()
                 .map(PhysicalTable::getSchema)
                 .map(schema -> schema.getColumns(DimensionColumn.class))
@@ -167,7 +172,10 @@ public abstract class BaseTableLoader implements TableLoader {
     }
 
     /**
-     * Iterate through given definitions and builds all corresponding physical table with satisfied dependency.
+     * Build and return the current physical table given its table name and definition, if dependency exists, build its
+     * dependencies and load the dependencies into physical table dictionary.
+     * <p>
+     * note: current physical table not loaded into physical table dictionary, only dependencies will
      *
      * @param currentTableName  Iterator for the mutable definition
      * @param availableTableDefinitions  A map of table name to table definition that are available globally
@@ -175,7 +183,7 @@ public abstract class BaseTableLoader implements TableLoader {
      *
      * @return the current physical table built from table definitions
      */
-    protected PhysicalTable buildPhysicalTablesWithDependency(
+    protected PhysicalTable buildPhysicalTableWithDependency(
             String currentTableName,
             Map<String, PhysicalTableDefinition> availableTableDefinitions,
             ResourceDictionaries dictionaries
@@ -197,36 +205,19 @@ public abstract class BaseTableLoader implements TableLoader {
                             "circular dependency", currentTableName);
 
             LOG.error(message);
-            throw new RuntimeException(message);
+            throw new IllegalStateException(message);
         }
 
         // If dependent table not in physical table dictionary, try to recursively build it and put it in the dictionary
         currentTableDefinition.getDependentTableNames()
                 .forEach(tableName -> physicalTableDictionary.computeIfAbsent(
                         tableName.asName(),
-                        name -> buildPhysicalTablesWithDependency(
-                                name,
-                                availableTableDefinitions,
-                                dictionaries
-                        )
+                        name -> buildPhysicalTableWithDependency(name, availableTableDefinitions, dictionaries)
                 ));
 
+        LOG.debug("Table Loader Building Physical Table {}");
         // Build the current physical table using physical table dictionary to resolve dependency
-        Optional<PhysicalTable> optionalPhysicalTable = currentTableDefinition.build(
-                dictionaries,
-                getDataSourceMetadataService()
-        );
-
-        // If the table definition correctly builds a physical table
-        if (optionalPhysicalTable.isPresent()) {
-            return optionalPhysicalTable.get();
-        } else {
-            String message = String.format(
-                    "Unable to build physical table: %s, might be table definition error", currentTableName);
-
-            LOG.error(message);
-            throw new RuntimeException(message);
-        }
+        return currentTableDefinition.build(dictionaries, getDataSourceMetadataService());
     }
 
     /**
@@ -253,6 +244,31 @@ public abstract class BaseTableLoader implements TableLoader {
     }
 
     /**
+     * Load a new physical table into the dictionary and return the loaded physical table.
+     *
+     * @param definition  A config object for the physical table
+     * @param metricNames  The Set of metric names on the table
+     * @param dictionaries  The resource dictionaries for reading and storing resource data
+     *
+     * @return The physical table created
+     *
+     * @deprecated use buildPhysicalTableWithDependency instead, which supports building table with dependencies
+     */
+    @Deprecated
+    protected PhysicalTable loadPhysicalTable(
+            PhysicalTableDefinition definition,
+            Set<FieldName> metricNames,
+            ResourceDictionaries dictionaries
+    ) {
+        LOG.debug(
+                "Building table {} with deprecated loadPhysicalTable method, use buildPhysicalTableWithDependency " +
+                        "instead",
+                definition.getName().asName()
+        );
+        return definition.build(dictionaries, getDataSourceMetadataService());
+    }
+
+    /**
      * Builds a table group.
      * <p>
      * Builds and loads the physical tables for the physical table definitions as well.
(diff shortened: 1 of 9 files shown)


fili-core/src/test/java/com/yahoo/bard/webservice/data/config/table/TestPhysicalTableDefinitionUtils.java
```diff
@@ -43,9 +44,11 @@
      *
      * @return the hourly table definitions
      */
-    public static LinkedHashSet<PhysicalTableDefinition> buildHourlyTableDefinitions(TestDimensions testDimensions) {
+    public static Set<PhysicalTableDefinition> buildHourlyTableDefinitions(
```
garyluoex: Changed to `Set` to avoid codacy annoying warnings.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
```diff
@@ -159,6 +155,113 @@ public void loadLogicalTableWithGranularities(
...
+
+        // Remove the current definition from map so that we don't try to circle back to it if circular dependency exist
+        PhysicalTableDefinition currentTableDefinition = mutableTableDefinitionMap.remove(currentTableName);
+
+        // If the dependent table definition is currently being build or missing
+        if (Objects.isNull(currentTableDefinition)) {
+            LOG.error("Unable to resolve physical table dependency for physical table: " + currentTableName.asName());
+            throw new RuntimeException("Unable to resolve physical table dependency for physical table: " +
+                    currentTableName
+                    .asName());
```
garyluoex: Move up a line

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
```diff
@@ -159,6 +155,113 @@ public void loadLogicalTableWithGranularities(
...
+    protected LinkedHashSet<PhysicalTable> loadPhysicalTablesWithDependency(
+            Set<PhysicalTableDefinition> definitions,
+            ResourceDictionaries dictionaries
+    ) {
+        LinkedHashSet<PhysicalTable> loadedTables = new LinkedHashSet<>();
+
+        Map<TableName, PhysicalTableDefinition> mutableCopyDefinitionMap =
+                buildPhysicalTableDefinitionDictionary(definitions);
+
+        definitions.forEach(definition -> buildPhysicalTablesWithDependency(
```
cdeszaq: This construct feels odd... It's like we're doing a recursion, but modifying a structure and filling up a bucket as we go. I think things would be much cleaner if we used a more traditional DepthFirstBuilding approach, where the return value of the method returns a LinkedHashSet of all of the tables it built, or perhaps just the 1 table it wanted to build.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/PhysicalTableDefinition.java
```diff
@@ -1,93 +1,144 @@
...
      */
-    @Deprecated
-    public PhysicalTableDefinition(
-            TableName name,
-            ZonelessTimeGrain grain,
-            Iterable<? extends DimensionConfig> dimensions
-    ) {
-        this(name, grain.buildZonedTimeGrain(DateTimeZone.UTC), dimensions);
-    }
+    public abstract Optional<PhysicalTable> build(
```
garyluoex: Not much now, but it feels right that it should return an `Optional` when you build something. Maybe some fili application will extend some new dimension and dimension definition that will fail to build for some reason. Let me know if you want me to remove the `Optional` or not.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
```diff
@@ -77,24 +76,21 @@ protected BaseTableLoader(DateTimeZone defaultTimeZone, DataSourceMetadataServic
...
         // Load a physical table for each of the table definitions
-        LinkedHashSet<PhysicalTable> physicalTables = new LinkedHashSet<>();
-        for (PhysicalTableDefinition def : tableDefinitions) {
-            PhysicalTable table = loadPhysicalTable(def, druidMetrics, dictionaries);
-            physicalTables.add(table);
-        }
+        LinkedHashSet<PhysicalTable> physicalTables = loadPhysicalTablesWithDependency(
+                tableDefinitions,
+                dictionaries
+        );
```
michael-mclawhorn: Per discussion yesterday, for this to work we need to have both the list of table names or table definitions which define the table group and ALSO have the total set of table definitions for building dependent tables.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/PhysicalTableDefinition.java
```diff
@@ -1,93 +1,144 @@
...
-                                            }
-                                            return physicalName;
-                                        }
-                                )
-                        )
-        );
+        this.timeGrain = timeGrain;
+        this.metricNames = ImmutableSet.copyOf(metricNames);
+        this.dimensionConfigs = ImmutableSet.copyOf(dimensionConfigs);
+        this.logicalToPhysicalNames = Collections.unmodifiableMap(buildLogicalToPhysicalNames(dimensionConfigs));
```
cdeszaq: If we really want this class to be primarily a holder for data needed to build a physical table, we need a constructor that takes this `logicalToPhysicalNames` mapping in as a parameter, rather than just assuming we can build it from the dimension configs.

The reason for this stems from the reason for the mapping in the first place: Dimensions do not have the same physical name on all physical tables. Because of that, you'll either need to have a complete set of DimensionConfig objects for every table, or you need to be able to pass the map into the PTDefinition directly.

(Note that this concern likely ripples down to the constructors of the subclasses as well.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/PhysicalTableDefinition.java
```diff
@@ -1,93 +1,144 @@
...
     /**
-     * Define a physical table using a zoneless time grain, defaulting it to UTC.
+     * Get the set of physical tables required to build the current physical table.
+     *
+     * @return a set of physical table names
+     */
+    public abstract Set<TableName> getDependentTableNames();
+
+    /**
+     * Given a set of metric names and a data source metadata service, build the corresponding physical table.
```
cdeszaq: I don't think this javadoc is correct?

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/PhysicalTableDefinition.java
cdeszaq: I left a comment elsewhere with deeper thoughts. tldr: Yes, let's remove the `Optional` wrapper.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/PhysicalTableDefinition.java
```diff
@@ -1,93 +1,144 @@
...
+        return dimensionConfigs.stream()
+                .collect(
+                        Collectors.toMap(
+                                DimensionConfig::getApiName,
+                                config -> {
+                                    String physicalName = config.getPhysicalName();
+                                    if (physicalName == null) {
+                                        throw new RuntimeException("No physical name found for dimension: "
+                                                + config.getApiName());
+                                    }
```
cdeszaq: I'm not sure this behavior is correct, but I'm also not sure what the correct behavior _is_. 

Per the docs on `DimensionConfig::getPhysicalName`, this value: `This field (if set) is used as the only physical name`, which means that it is OK not to have a value for this field. It also means that if it's _not_ set, the expectation is that there is more than 1 physical name for this dimension, and that information is being set on the PhysicalTables through the mapping directly.

That said, if this method is being made available as a default / helper (ie. if no mapping was provided), which is the expected use of this I think, then there _should_ be a single physical name present for each dimension.

The other concern I have is that by throwing an exception, we're forcing users to specify the physical name of the dimension. While I think it's highly likely that most Fili instances will have different API names and physical names for dimensions, would it ease onboarding (ie. reduce required configuration) if we instead had this method log a warning, and then use the API name as the default physical name?

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/table/BaseTableLoaderSpec.groovy
```diff
@@ -8,95 +8,269 @@ import com.yahoo.bard.webservice.data.config.ResourceDictionaries
...
+                return Collections.emptySet()
+            }
+            return Collections.singleton(TableName.of(dependentTableName));
+        }
+
+        @Override
+        Optional<PhysicalTable> build(
+                ResourceDictionaries dictionaries,
+                DataSourceMetadataService metadataService
+        ) {
```
cdeszaq: Can unwrap

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/table/BaseTableLoaderSpec.groovy
```diff
@@ -8,95 +8,269 @@ import com.yahoo.bard.webservice.data.config.ResourceDictionaries
...
     Collection<Dimension> dims
+    Set<TableName> tableNames
+    Set<PhysicalTableDefinition> allDependentDefinitions
+
+    PhysicalTableDefinition definition1
+    PhysicalTableDefinition definition2
+    PhysicalTableDefinition definition3
+    PhysicalTableDefinition definition4
+    PhysicalTableDefinition definition5
+    PhysicalTableDefinition definition6
```
cdeszaq: Can we give these more descriptive names? Seeing it in code, `definition4` means nothing, but `selfDependentDefinition` is much more meaningful, for example.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/table/BaseTableLoaderSpec.groovy
```diff
@@ -8,95 +8,269 @@ import com.yahoo.bard.webservice.data.config.ResourceDictionaries
...
     def setup() {
-        loader = new SimpleBaseTableLoader(new TestDataSourceMetadataService())
+        loader = new SimpleBaseTableLoader(Mock(DataSourceMetadataService))
         dicts = new ResourceDictionaries()
         apiNames = TestApiMetricName.getByLogicalTable(SHAPES)
         metricNames = TestDruidMetricName.getByLogicalTable(SHAPES)
-        physDefs = TestPhysicalTableDefinitionUtils.buildShapeTableDefinitions(new TestDimensions())
+        physDefs = TestPhysicalTableDefinitionUtils.buildShapeTableDefinitions(new TestDimensions(), metricNames)
         dimNames = TestApiDimensionName.getByLogicalTable(SHAPES)
+        tableNames = physDefs.stream().map({it -> it.getName()}).collect(Collectors.toSet())
```
cdeszaq: Since we're in groovy, it's generally more idiomatic to use groovy's built-in collection processing, closures, and property access:

```groovy
physDefs.collect {it.name} as Set
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/table/BaseTableLoaderSpec.groovy
```diff
@@ -8,95 +8,269 @@ import com.yahoo.bard.webservice.data.config.ResourceDictionaries
...
+                dicts,
+                apiNames
+        )
 
         then:
-        dicts.physicalDictionary.size() == 2
-        dicts.physicalDictionary.containsValue(t1)
-        dicts.physicalDictionary.containsValue(t2)
+        dicts.physicalDictionary.size() == 6
+        group.physicalTables.stream().map({it.getTableName()}).collect(Collectors.toSet()) as Set == tableNames
```
cdeszaq: could be groovier: `group.physicalTables.collect {it.tableName} as Set`

<sub>Note: Applies elsewhere in PR</sub>

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/table/BaseTableLoaderSpec.groovy
```diff
@@ -8,95 +8,269 @@ import com.yahoo.bard.webservice.data.config.ResourceDictionaries
...
     }
 
-    def "load duplicate physical tables results in sharing definitions"() {
-        setup:
-        List<PhysicalTableDefinition> allDefs = physDefs as List
+    def "loading physical tables with dependency loads all satisfied dependency physical tables"() {
+        given:
+        Set<TableName> currentTableNames = [definition1.getName(), definition2.getName(), definition3.getName()]
```
cdeszaq: property access is groovier: `definition1.name, definition2.name, definition3.name`

<sub>Note: Applies elsewhere in PR</sub>

garyluoex: So we are not testing the behavior of `getName()` anymore?

cdeszaq: property access in groovy goes through the getter/setter methods. If you want direct field access, it's a different operator (something like `definition1@.name` I think). Groovy realized that the more common case is getter/setter access, rather than field access, and so they made accessing JavaBean-style data easier.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
```diff
@@ -74,27 +73,35 @@ protected BaseTableLoader(DateTimeZone defaultTimeZone, DataSourceMetadataServic
...
+     * @param tableDefinitions  A list of config objects for building physical tables and its dependent physical tables
      * @param dictionaries  The container for all the data dictionaries
+     * @param apiMetrics  The set of metric names surfaced to the api
      *
      * @return A table group binding all the tables for a logical table view together.
      */
     public TableGroup buildDimensionSpanningTableGroup(
-            Set<ApiMetricName> apiMetrics,
-            Set<FieldName> druidMetrics,
+            Set<TableName> tableNames,
```
garyluoex: `tableDefinitions` is a super set of `tableNames`

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
```diff
@@ -159,6 +167,138 @@ public void loadLogicalTableWithGranularities(
     }
 
     /**
+     * Iterate through given definitions and builds all corresponding physical table with satisfied dependency.
+     *
+     * @param currentTableName  Iterator for the mutable definition
+     * @param availableTableDefinitions  A map of table name to table definition that are available globally
```
garyluoex: `availableTableDefinitions` simply means the table definitions that is available for you to use, it may hold definitions that is already built, so better not use untouched or unbuilt

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
```diff
@@ -159,6 +167,138 @@ public void loadLogicalTableWithGranularities(
...
+     * @param apiMetrics  The set of metric names surfaced to the api
+     * @param druidMetrics  Names of druid datasource metric columns
+     * @param tableDefinitions  A list of config objects for physical tables
+     * @param dictionaries  The container for all the data dictionaries
+     *
+     * @return A table group binding all the tables for a logical table view together.
+     *
+     * @deprecated does not load table with external dependency, use the other buildDimensionSpanningTableGroup instead
+     */
+    public TableGroup buildDimensionSpanningTableGroup(
```
garyluoex: To deprecate the old method, instead of simply removing it without deprecation.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
```diff
@@ -159,6 +167,138 @@ public void loadLogicalTableWithGranularities(
...
+            Set<ApiMetricName> apiMetrics,
+            Set<FieldName> druidMetrics,
```
garyluoex: To deprecate the old method, instead of simply removing it without deprecation.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
cdeszaq: I see. Perhaps we could rename `tableNames` to something like `tablesInTableGroup` or something that indicates their purpose more clearly?

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
cdeszaq: ahh, ok, so it's only "unbuilt" for the _current_ table group.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
cdeszaq: So, rather than duplicating the code, can we just have the deprecated one call the non-deprecated one?

Also, if this is to be deprecated, it should have the `@Deprecated` annotation on the method so that the compiler will flag it.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/PhysicalTableDefinition.java
cdeszaq: Perhaps not a warning, but log _something_ so that it's not surprising that we're filling in the blanks. Info or Debug levels would be fine by me, as long as the method indicates it's going to do this.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
```diff
@@ -159,6 +167,138 @@ public void loadLogicalTableWithGranularities(
     }
 
     /**
+     * Iterate through given definitions and builds all corresponding physical table with satisfied dependency.
+     *
+     * @param currentTableName  Iterator for the mutable definition
```
cdeszaq: Param still seems to have an odd description?

Issue #59, opened 2016-10-07
Migrate builds to Gradle


cdeszaq:
As nice as Maven is, it's been a challenge to get it working the way we'd like and making it do more seems to be more painful than painless. So, to ease our pain (and hopefully get faster builds), we want to move to Grade.

archolewa:
💯 
💯 
💯 
💯 
💯 
💯

dejan2609:
@cdeszaq would love to work on this. Am I allowed to share some ideas / send PR (if external contributions are allowed for this specific ticket) ?

michael-mclawhorn:
Right now I'm looking for some Fili implementor to pilot using gradle on their fili-derived project to figure out how difficult or easy it is to adapt the build process to it.  Our chief gradle evangelist from the Core Team(@cdeszaq) moved and took a new job and has been fairly idle since then.

cdeszaq:
By all means feel free to take a stab at it / provide input. Our Gradle efforts are very nascent so anyone with experience (or desire to learn) is welcome.

As Mike said, I've been more checked out than originally planned, but I am watching from afar more than it may seem ;-)

I will say that one of the biggest hurdles will be to help the whole Fili community get up to speed with Gradle and be comfortable with it. So with that in mind, lots of comments and even links to documentation so that others can follow along, and to explain why a choice makes sense, would be a good idea.

dejan2609:
@michael-mclawhorn I'll try to provide non-biased Maven vs. Gradle comparison (followed by some examples / pitfalls / suggestions).

## Maven (devil's) advocacy combined with just a few Gradle pitfalls 
### (main idea is to show that Maven is not all that bad _**or**_ that Gradle is not a silver bullet solution):
- [Maven's Inflexibility Is Its Best Feature](https://timboudreau.com/blog/maven/read)
- http://www.wakaleo.com/blog/what-has-maven-ever-done-for-us
- `Gradle doesn't separate declarative bits from code, so builds descend rapidly into complex mess. Maven is boring but much more reliable` https://twitter.com/jodastephen/status/888364803416367104
- old dog can learn some new tricks indeed: see my PR here #461 (maven wrapper related)
- Gradle tends to incubate some features for a way too long: 
-- Gradle **1.4** (released on ***Jan 28, 2013***) https://docs.gradle.org/1.4/release-notes.html#the-new-‘maven-publish’-plugin
[...]

PR #208, opened 2017-03-29
[WIP] Assert1test config

cdeszaq:
Having more than 1 `testApplicationConfig` on the classpath when running tests wasn't alerted, and made debugging something hard recently. There should only ever be 1 `testApplicationConfig` at a time, so this change asserts that to be true and blows up if it isn't.

This also cleans up some of the logging and information the config loading process spits out as it does it's thing, especially in error scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -163,6 +163,11 @@ Current
 - [Make `TemplateDruidQuery::getMetricField` get the first field instead of any field](https://github.com/yahoo/fili/pull/210)
     * Previously, order was by luck, now it's by the contract of `findFirst`
 
+- [Clean up config loading and add more logs and checks]()
+    * Assert no more than 1 `testApplicationConfig` on classpath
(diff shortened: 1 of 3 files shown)


cdeszaq:
While this passes with `mvn verify`, it _doesn't_ pass with `mvn test` due to picking up more than one `testApplicationConfig`. This needs to be solved before we can merge this.

QubitPi:
Is this branch now stale @cdeszaq ?

cdeszaq:
Yes, it's stale. I think we should come back to this at some point, but not right now. I'll close this PR shortly (leaving the branch), and move it into a techdebt ticket to pay off soon. (Probably when we move to Gradle from Maven)

QubitPi:
Whut? Seriously? Why do we move from Maven to Gradle?

cdeszaq:
@QubitPi See [this issue](https://github.com/yahoo/fili/issues/59)

garyluoex:
Not needed

PR #209, opened 2017-03-29
Deprecation unusingness

cdeszaq:
This PR removes many instances where we were using deprecated APIs and such in Fili code. These are generally in implementations, not interfaces, so there's no real impact.

There's also some cleanups (groovy cleanups and other simplifications) in here as well, but broken out as separate commits.

### DIFF
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/query/WeightEvaluationQuerySpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/query/WeightEvaluationQuerySpec.groovy
@@ -214,7 +214,7 @@ class WeightEvaluationQuerySpec extends Specification {
 
         and: "The weight-check query doesn't have a sort at either level"
         !new WeightEvaluationQuery(groupByQuery, 1).limitSpec.columns
-        !new WeightEvaluationQuery(groupByQuery, 1).innermostQuery.limitSpec.schema.columns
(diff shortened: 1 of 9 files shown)


fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/query/WeightEvaluationQuerySpec.groovy
```diff
@@ -214,7 +214,7 @@ class WeightEvaluationQuerySpec extends Specification {
 
         and: "The weight-check query doesn't have a sort at either level"
         !new WeightEvaluationQuery(groupByQuery, 1).limitSpec.columns
-        !new WeightEvaluationQuery(groupByQuery, 1).innermostQuery.limitSpec.columns
+        !new WeightEvaluationQuery(groupByQuery, 1).innermostQuery.limitSpec.schema.columns
```
michael-mclawhorn: This is an error, I'm pretty sure

cdeszaq: Yeah, I'm pretty sure you're right, since `limitSpec` doesn't have a schema 😉 Let me give this PR a little sprucing.

cdeszaq: This has been spruced and builds correctly now.

michael-mclawhorn:
👍 looks fine

QubitPi:
👍

PR #210, opened 2017-03-29
Un-luck TemplateDruidQuery::getMetricField

cdeszaq:
`findAny` will use the 1st one it sees in a parallel world, which _may_ not be in the order the streams gave the results. `findFirst` ensures order is followed, meaning that we'll get a PostAgg if there is one, otherwise we might not have.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,9 @@ Current
 
 ### Changed:
 
+- [Make `TemplateDruidQuery::getMetricField` get the first field instead of any field](https://github.com/yahoo/fili/pull/210)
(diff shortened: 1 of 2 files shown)


michael-mclawhorn:
👍

archolewa:
👍

CHANGELOG.md
```diff
@@ -59,6 +59,9 @@ Current
 
 ### Changed:
 
+- [Make `TemplateDruidQuery::getMetricField` get the first field instead of any field](https://github.com/yahoo/fili/pull/203)
```
archolewa: The PR number is wrong.

PR #211, opened 2017-03-29
Deprecate unused Column constructors

cdeszaq:
Some of the column constructors are or almost are unused. This deprecates those so we can clean them up soon.


michael-mclawhorn:
👍

QubitPi:
👍

PR #213, opened 2017-03-29
Fix pull request builds

kevinhinterlong:
Hopefully this fixes pull request builds

[travis-ci pull requests security restrictions](https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions)


travis/pull_request_setup.bash
```diff
@@ -0,0 +1 @@
+#!/bin/bash
```
archolewa: Is this file supposed to be blank? If so, why we do we even need it?

kevinhinterlong: I'm not sure if there's any set up required for pull requests (if there is then it'll be added there). If nothing needs to be done then it can be removed.
@cdeszaq ?

cdeszaq: At this point, I don't think there's anything special needed for PR builds, so we can probably drop the file

PR #214, opened 2017-03-30
Make druid client follow redirects by default

cdeszaq:
Not following redirects is a pain if hosts need to move and leave a redirect behind to point to their new location, or if there is some sort of redirection-based load balancing. The Druid clients now will follow redirects (max count defaults to 5)


archolewa:
👍

Issue #224, opened 2017-04-03
Allow GranularityComparator to return static instance


QubitPi:
![screen shot 2017-04-03 at 9 54 11 am](https://cloud.githubusercontent.com/assets/16126939/24615310/88353e64-1853-11e7-9eab-4eb8951c9d79.png)

Implementation of PR https://github.com/yahoo/fili/pull/193 suggests an possible improvement on `GranularityComparator`: Put the static instance on the `GranularityComparator` class itself, so that everywhere in the system that wants it could just call `GranularityComparator.getInstance()`

QubitPi:
https://github.com/yahoo/fili/pull/225

PR #225, opened 2017-04-03
Allow GranularityComparator to return static instance

ghost:
https://github.com/yahoo/fili/issues/224

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/GranularityComparator.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/GranularityComparator.java
@@ -27,7 +27,7 @@ public class GranularityComparator implements Comparator<PhysicalTable> {
      *
      * @return a new GranularityComparator instance
      */
-    public static synchronized GranularityComparator getInstance() {
+    public static GranularityComparator getInstance() {
         return COMPARATOR;
     }
 
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/GranularityComparator.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/GranularityComparator.java
@@ -17,6 +17,15 @@ public class GranularityComparator implements Comparator<PhysicalTable> {
     private static final Logger LOG = LoggerFactory.getLogger(GranularityComparator.class);
 
     /**
+     * Factory method for creating new GranularityComparator instance.
+     *
+     * @return a new GranularityComparator instance
+     */
+    public static GranularityComparator getInstance() {
+        return new GranularityComparator();
+    }
+
+    /**
      * Compare two physical tables identifying which one has fewer time buckets.
      *
      * @param table1 The first table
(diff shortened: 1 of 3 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/GranularityComparator.java
```diff
@@ -17,6 +17,15 @@
     private static final Logger LOG = LoggerFactory.getLogger(GranularityComparator.class);
 
     /**
+     * Factory method for creating new GranularityComparator instance.
+     *
+     * @return a new GranularityComparator instance
+     */
+    public static GranularityComparator getInstance() {
+        return new GranularityComparator();
```
cdeszaq: This class should create an instance statically, cache it, and then this method can return the cached instance every time. We explicitly _don't_ want to be making new instances all the time because there's no point (ie. no difference between the instances)

cdeszaq (changes requested):
- Changelog needed
- This PR should go through and update the places where we're calling this constructor, and switch them to use the new `getInstance()` method
- The is perhaps more controversial, but this class should have it's constructor converted to be protected, forcing people to go through the `getInstance()` method instead, to require code reuse

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/GranularityComparator.java
```diff
@@ -15,6 +15,24 @@
 public class GranularityComparator implements Comparator<PhysicalTable> {
 
     private static final Logger LOG = LoggerFactory.getLogger(GranularityComparator.class);
+    private static GranularityComparator granularityComparator;
+
+    /**
+     * Disable constructor to force using getInstance() method.
+     */
+    private GranularityComparator() { }
```
cdeszaq: Make this `protected` in case someone wants to subclass.

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/GranularityComparator.java
```diff
@@ -15,6 +15,24 @@
 public class GranularityComparator implements Comparator<PhysicalTable> {
 
     private static final Logger LOG = LoggerFactory.getLogger(GranularityComparator.class);
+    private static GranularityComparator granularityComparator;
```
cdeszaq: Make this a `private static final` and call the constructor to assign it statically.

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/GranularityComparator.java
```diff
@@ -15,6 +15,24 @@
...
+     */
+    private GranularityComparator() { }
+
+    /**
+     * Factory method for creating new GranularityComparator instance.
+     *
+     * @return a new GranularityComparator instance
+     */
+    public static synchronized GranularityComparator getInstance() {
+        if (granularityComparator == null) {
```
cdeszaq: Remove this condition. We should just instantiate the instance variable once when the class is loaded, rather than incur the overhead of the check every time.

cdeszaq (changes requested):

cdeszaq (approved):

michael-mclawhorn (approved):

PR #226, opened 2017-04-04
Fix to use physical name instead of logical name to retrieve available interval

garyluoex:
* Added `PhysicalDataSourceConstraint` class to capture physical names of columns for retrieving availble interval
* `getAllAvailbleIntervals` in `ConcreteAvailability` no longer filters out unconfigured columns, instead table's `getAllAvailbleIntervals` does
* `getAvailbleIntervals` in `Availbality` now takes `PhysicalDataSourceConstraint` instead of `DataSourceConstraint`
* `Availbility` no longer takes a set of columns on the table, only table needs to know
* `getAllAvailbleIntervals` in `Availbility` now returns a map of column physical name to interval list instead of column to interval list
* `TestDataSourceMetadataService` now takes map from string to list of intervals instead of column to list of intervals for constructor

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
@@ -6,6 +6,7 @@ import com.yahoo.bard.webservice.data.config.names.TableName;
 import com.yahoo.bard.webservice.data.time.ZonedTimeGrain;
 import com.yahoo.bard.webservice.table.availability.Availability;
 import com.yahoo.bard.webservice.table.resolver.DataSourceConstraint;
+import com.yahoo.bard.webservice.table.resolver.PhysicalDataSourceConstraint;
 import com.yahoo.bard.webservice.util.IntervalUtils;
 import com.yahoo.bard.webservice.util.SimplifiedIntervalList;
 
@@ -16,6 +17,8 @@ import org.slf4j.LoggerFactory;
 
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import javax.validation.constraints.NotNull;
 
@@ -79,12 +82,21 @@ public abstract class BasePhysicalTable implements PhysicalTable {
 
     @Override
     public Map<Column, List<Interval>> getAllAvailableIntervals() {
-        return getAvailability().getAllAvailableIntervals();
+        Map<String, List<Interval>> availableIntervals = getAvailability().getAllAvailableIntervals();
+
+        return getSchema().getColumns().stream()
+                .collect(Collectors.toMap(
+                        Function.identity(),
+                        column -> availableIntervals.getOrDefault(
+                                getSchema().getPhysicalColumnName(column.getName()),
+                                new SimplifiedIntervalList()
+                        )
+                ));
     }
 
     @Override
     public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constraint) {
-        return getAvailability().getAvailableIntervals(constraint);
+        return getAvailability().getAvailableIntervals(new PhysicalDataSourceConstraint(constraint, getSchema()));
     }
 
     @Override
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
@@ -17,6 +17,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -96,6 +97,22 @@ public abstract class BasePhysicalTable implements PhysicalTable {
 
     @Override
     public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constraint) {
+
+        Set<String> tableColumnNames = getSchema().getColumnNames();
+        Set<String> invalidColumnNames = constraint.getAllColumnNames().stream()
+                .filter(name -> !tableColumnNames.contains(name))
+                .collect(Collectors.toSet());
+
+        if (!constraint.getAllColumnNames().stream().allMatch(tableColumnNames::contains)) {
+            String message = String.format(
+                    "Received invalid request requesting for columns: %s that is not available in this table: %s",
+                    invalidColumnNames.stream().collect(Collectors.joining(",")),
+                    getName()
+            );
+            LOG.error(message);
+            throw new RuntimeException(message);
+        }
+
         return getAvailability().getAvailableIntervals(new PhysicalDataSourceConstraint(constraint, getSchema()));
     }
 
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
@@ -99,16 +99,14 @@ public abstract class BasePhysicalTable implements PhysicalTable {
     public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constraint) {
 
         Set<String> tableColumnNames = getSchema().getColumnNames();
-        Set<String> invalidColumnNames = constraint.getAllColumnNames().stream()
-                .filter(name -> !tableColumnNames.contains(name))
-                .collect(Collectors.toSet());
 
+        // Validate that the requested columns are answerable by the current table
         if (!constraint.getAllColumnNames().stream().allMatch(tableColumnNames::contains)) {
             String message = String.format(
                     "Received invalid request requesting for columns: %s that is not available in this table: %s",
-                    invalidColumnNames.stream().collect(Collectors.joining(",")),
-                    getName()
-            );
+                    constraint.getAllColumnNames().stream()
+                            .filter(name -> !tableColumnNames.contains(name))
+                            .collect(Collectors.joining(",")), getName());
             LOG.error(message);
             throw new RuntimeException(message);
         }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTable.java
@@ -44,7 +44,7 @@ public class PermissiveConcretePhysicalTable extends ConcretePhysicalTable {
                 timeGrain,
                 columns,
                 logicalToPhysicalColumnNames,
-                new PermissiveAvailability(name, columns, dataSourceMetadataService)
+                new PermissiveAvailability(name, dataSourceMetadataService)
         );
     }
 }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTable.java
@@ -6,9 +6,6 @@ import com.yahoo.bard.webservice.data.config.names.TableName;
 import com.yahoo.bard.webservice.data.time.ZonedTimeGrain;
 import com.yahoo.bard.webservice.metadata.DataSourceMetadataService;
 import com.yahoo.bard.webservice.table.availability.PermissiveAvailability;
-import com.yahoo.bard.webservice.table.resolver.DataSourceConstraint;
-import com.yahoo.bard.webservice.table.resolver.PhysicalDataSourceConstraint;
-import com.yahoo.bard.webservice.util.SimplifiedIntervalList;
 
 import java.util.Map;
 import java.util.Set;
@@ -49,13 +46,5 @@ public class PermissiveConcretePhysicalTable extends ConcretePhysicalTable {
                 logicalToPhysicalColumnNames,
                 new PermissiveAvailability(name, dataSourceMetadataService)
         );
-    }
-
-    @Override
-    public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constraint) {
-        return getAvailability().getAvailableIntervals(PhysicalDataSourceConstraint.buildWithSchemaUnion(
-                constraint,
-                getSchema()
-        ));
     }
 }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTable.java
@@ -6,6 +6,9 @@ import com.yahoo.bard.webservice.data.config.names.TableName;
 import com.yahoo.bard.webservice.data.time.ZonedTimeGrain;
 import com.yahoo.bard.webservice.metadata.DataSourceMetadataService;
 import com.yahoo.bard.webservice.table.availability.PermissiveAvailability;
+import com.yahoo.bard.webservice.table.resolver.DataSourceConstraint;
+import com.yahoo.bard.webservice.table.resolver.PhysicalDataSourceConstraint;
+import com.yahoo.bard.webservice.util.SimplifiedIntervalList;
 
 import java.util.Map;
 import java.util.Set;
@@ -46,5 +49,13 @@ public class PermissiveConcretePhysicalTable extends ConcretePhysicalTable {
                 logicalToPhysicalColumnNames,
                 new PermissiveAvailability(name, dataSourceMetadataService)
         );
+    }
+
+    @Override
+    public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constraint) {
+        return getAvailability().getAvailableIntervals(PhysicalDataSourceConstraint.buildWithSchemaUnion(
+                constraint,
+                getSchema()
+        ));
     }
 }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
@@ -4,12 +4,9 @@ package com.yahoo.bard.webservice.table.availability;
 
 import com.yahoo.bard.webservice.data.config.names.TableName;
 import com.yahoo.bard.webservice.metadata.DataSourceMetadataService;
-import com.yahoo.bard.webservice.table.Column;
-import com.yahoo.bard.webservice.table.resolver.DataSourceConstraint;
+import com.yahoo.bard.webservice.table.resolver.PhysicalDataSourceConstraint;
 import com.yahoo.bard.webservice.util.IntervalUtils;
 import com.yahoo.bard.webservice.util.SimplifiedIntervalList;
-
-import com.google.common.collect.ImmutableSet;
 
 import org.joda.time.Interval;
 
@@ -18,8 +15,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.function.Function;
-import java.util.stream.Collectors;
 
 import javax.validation.constraints.NotNull;
 
@@ -29,29 +24,23 @@ import javax.validation.constraints.NotNull;
 public class ConcreteAvailability implements Availability {
 
     private final TableName name;
-    private final Set<Column> columns;
     private final DataSourceMetadataService metadataService;
 
-    private final Set<String> columnNames;
     private final Set<TableName> dataSourceNames;
 
     /**
      * Constructor.
      *
      * @param tableName  The name of the table and data source associated with this Availability
-     * @param columns  The columns associated with the table and availability
      * @param metadataService  A service containing the datasource segment data
      */
     public ConcreteAvailability(
             @NotNull TableName tableName,
-            @NotNull Set<Column> columns,
             @NotNull DataSourceMetadataService metadataService
     ) {
         this.name = tableName;
-        this.columns = ImmutableSet.copyOf(columns);
         this.metadataService = metadataService;
 
-        this.columnNames = columns.stream().map(Column::getName).collect(Collectors.toSet());
         this.dataSourceNames = Collections.singleton(name);
     }
 
@@ -61,59 +50,30 @@ public class ConcreteAvailability implements Availability {
     }
 
     @Override
-    public Map<Column, List<Interval>> getAllAvailableIntervals() {
-
-        Map<String, List<Interval>> allAvailableIntervals = getAvailableIntervalsByTable();
-        return columns.stream()
-                .collect(
-                        Collectors.toMap(
-                                Function.identity(),
-                                column -> new SimplifiedIntervalList(
-                                        allAvailableIntervals.getOrDefault(column.getName(), Collections.emptyList())
-                                )
-                        )
-                );
+    public Map<String, List<Interval>> getAllAvailableIntervals() {
+        return metadataService.getAvailableIntervalsByTable(name);
     }
 
     @Override
-    public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constraint) {
+    public SimplifiedIntervalList getAvailableIntervals(PhysicalDataSourceConstraint constraint) {
 
-        Set<String> requestColumns = constraint.getAllColumnNames().stream()
-                .filter(columnNames::contains)
-                .collect(Collectors.toSet());
+        Set<String> requestColumns = constraint.getAllColumnPhysicalNames();
 
         if (requestColumns.isEmpty()) {
             return new SimplifiedIntervalList();
         }
 
-        Map<String, List<Interval>> allAvailableIntervals = getAvailableIntervalsByTable();
+        Map<String, List<Interval>> allAvailableIntervals = getAllAvailableIntervals();
 
         // Need to ensure requestColumns is not empty in order to prevent returning null by reduce operation
         return new SimplifiedIntervalList(
                 requestColumns.stream()
-                        .map(columnName -> allAvailableIntervals.getOrDefault(columnName, Collections.emptyList()))
+                        .map(physicalName -> allAvailableIntervals.getOrDefault(physicalName, Collections.emptyList()))
                         .map(intervals -> (Collection<Interval>) intervals)
                         .reduce(null, IntervalUtils::getOverlappingSubintervals)
         );
     }
 
-    /**
-     * Retrieves the most up to date column to available interval map from data source metadata service.
-     *
-     * @return map of column name to list of avialable intervals
-     */
-    protected Map<String, List<Interval>> getAvailableIntervalsByTable() {
-        return metadataService.getAvailableIntervalsByTable(name);
-    }
-
-    /**
-     * Returns the configured columns in their String representation.
-     *
-     * @return the configured columns in their String representation
-     */
-    protected Set<String> getColumnNames() {
-        return columnNames;
-    }
 
     /**
      * Returns the name of the table and data source associated with this Availability.
@@ -126,11 +86,8 @@ public class ConcreteAvailability implements Availability {
 
     @Override
     public String toString() {
-        return String.format("ConcreteAvailability with table name = %s and Configured columns = [%s]",
-                name.asName(),
-                columns.stream()
-                        .map(Column::getName)
-                        .collect(Collectors.joining(", "))
+        return String.format("ConcreteAvailability for table: %s",
+                name.asName()
         );
     }
 }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
@@ -68,7 +68,10 @@ public class ConcreteAvailability implements Availability {
         // Need to ensure requestColumns is not empty in order to prevent returning null by reduce operation
         return new SimplifiedIntervalList(
                 requestColumns.stream()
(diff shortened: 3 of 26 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
```diff
@@ -79,12 +82,18 @@ public DateTime getTableAlignment() {
 
     @Override
     public Map<Column, List<Interval>> getAllAvailableIntervals() {
-        return getAvailability().getAllAvailableIntervals();
+        Map<String, List<Interval>> availableIntervals = getAvailability().getAllAvailableIntervals();
+
+        return getSchema().getColumns().stream()
+                .collect(Collectors.toMap(
+                        Function.identity(),
+                        column -> availableIntervals.get(getSchema().getPhysicalColumnName(column.getName()))
```
cdeszaq: Should this be `getOrDefault`?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -106,7 +104,7 @@ public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constra
         return metadataService.getAvailableIntervalsByTable(name);
     }
 
-    protected Set<String> getColumnNames() {
-        return columnNames;
+    protected Set<String> getColumnPhysicalNames() {
```
cdeszaq: Should we rename this `physicalColumnNames`? `columnPhysicalNames` feels a bit backwards, for some reason.

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/DataSourceConstraint.java
```diff
@@ -54,6 +54,22 @@ public DataSourceConstraint(DataApiRequest dataApiRequest, DruidAggregationQuery
         ).collect(Collectors.toSet()));
     }
 
+    /**
+     * Copy Constructor.
+     *
+     * @param dataSourceConstraint  The data source constarint to copy from
+     */
+    public DataSourceConstraint(DataSourceConstraint dataSourceConstraint) {
```
cdeszaq: Should be `protected`, not `public`

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/PhysicalDataSourceConstraint.java
```diff
@@ -0,0 +1,40 @@
...
+                .map(physicalTableSchema::getPhysicalColumnName)
+                .collect(Collectors.toSet());
+    }
+
+    /**
+     * Getter for the all column names as physical names.
+     *
+     * @return the physical name of all the columns
+     */
+    public Set<String> getAllColumnPhysicalNames() {
```
cdeszaq: Should we rename this `physicalColumnNames`? `columnPhysicalNames` feels a bit backwards, for some reason.

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/PhysicalDataSourceConstraint.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private final Set<String> allColumnPhysicalNames;
+
+    /**
+     * Constructor.
+     *
+     * @param dataSourceConstraint  Data soruce contraint containing all the column names as logical names
+     * @param physicalTableSchema  A map from logical column name to physical column names
+     */
+    public PhysicalDataSourceConstraint(DataSourceConstraint dataSourceConstraint, PhysicalTableSchema physicalTableSchema) {
```
cdeszaq: Rather than depending on the `PhysicalTableSchema`, we should just take the one mapping function we care about:

```java
Function<String, String> logicalToPhysicalNameMapper
```

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/PhysicalDataSourceConstraint.java
garyluoex: We don't have `PhysicalColumn`s, we do have `Column` physical name.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -26,36 +25,36 @@
...
      * @return the union of all available intervals
      */
     @Override
-    public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint ignoredConstraints) {
-        Map<String, List<Interval>> allAvailableIntervals = getAvailableIntervalsByTable();
-        return getColumnNames().stream()
+    public SimplifiedIntervalList getAvailableIntervals(PhysicalDataSourceConstraint constraint) {
+        Map<String, List<Interval>> allAvailableIntervals = getAllAvailableIntervals();
+
+        return constraint.getAllColumnPhysicalNames().stream()
```
michael-mclawhorn: This is a change in behavior and I don't think it is correct.

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/PhysicalDataSourceConstraint.java
```diff
@@ -0,0 +1,48 @@
...
+    public PhysicalDataSourceConstraint(
+            DataSourceConstraint dataSourceConstraint,
+            PhysicalTableSchema physicalTableSchema
+    ) {
+        super(dataSourceConstraint);
+
+        Set<String> schemaColumnNames = physicalTableSchema.getColumns().stream()
+                .map(Column::getName)
+                .collect(Collectors.toSet());
+
```
michael-mclawhorn: This should just be a get method on physical table schema

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/PhysicalDataSourceConstraint.java
cdeszaq: I see it more as having the "physical" "column names", rather than "physical column" "names", but I suppose it's not a big difference either way 😄

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/PhysicalDataSourceConstraint.java
garyluoex: We also need all the columns from it, might as well pass the entire schema.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
garyluoex: Same reason as we don't have physicalColumn names

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -26,36 +25,37 @@
...
      * This is different from its parent's
      * {@link
-     * com.yahoo.bard.webservice.table.availability.ConcreteAvailability#getAvailableIntervals(DataSourceConstraint)};
+     * com.yahoo.bard.webservice.table.availability.ConcreteAvailability#getAvailableIntervals(
+     * PhysicalDataSourceConstraint
+     * )};
      * Instead of returning the intersection of all available intervals, this method returns the union of them.
      *
-     * @param ignoredConstraints  Data constraint containing columns and api filters. Constrains are ignored, because
+     * @param constraint  Data constraint containing columns and api filters. Constrains are ignored, because
```
michael-mclawhorn: Why was this changed?  I'm pretty sure that the original 'ignored' was intentional.  
Also, spelling 'Constraints'

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
```diff
@@ -49,15 +50,14 @@ class PermissiveAvailabilitySpec extends Specification  {
```
michael-mclawhorn: This test is incorrect to design.  Create a 'long_interval' column and confirm that even when not selecte that the getAvailableIntervals()l is always the long_interval or the union of the long_interval with any non overlapping intervals.

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/DataSourceConstraint.java
```diff
@@ -54,6 +54,22 @@ public DataSourceConstraint(DataApiRequest dataApiRequest, DruidAggregationQuery
...
+    protected DataSourceConstraint(DataSourceConstraint dataSourceConstraint) {
+        this.requestDimensions = dataSourceConstraint.getRequestDimensions();
+        this.filterDimensions = dataSourceConstraint.getFilterDimensions();
+        this.metricDimensions = dataSourceConstraint.getMetricDimensions();
+        this.metricNames = dataSourceConstraint.getMetricNames();
+        this.apiFilters = dataSourceConstraint.getApiFilters();
+        this.allDimensions = dataSourceConstraint.getAllDimensions();
+        this.allDimensionNames = this.getAllDimensionNames();
+        this.allColumnNames = this.getAllColumnNames();
+    }
```
QubitPi: Since all private instance variables are `final`, should we use defensive copy? @cdeszaq @michael-mclawhorn

QubitPi: Since all private instance variables are final, should we use defensive copy? @cdeszaq @michael-mclawhorn

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/PhysicalDataSourceConstraint.java
```diff
@@ -0,0 +1,82 @@
...
+    private final Set<String> allColumnPhysicalNames;
+
+    /**
+     * Constructor.
+     *
+     * @param dataSourceConstraint  Data source constraint containing all the column names as logical names
+     * @param physicalTableSchema  A map from logical column name to physical column names
+     */
+    public PhysicalDataSourceConstraint(
+            DataSourceConstraint dataSourceConstraint,
```
QubitPi: Are we missing `@NotNull`?

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/PhysicalDataSourceConstraint.java
```diff
@@ -0,0 +1,82 @@
...
+            PhysicalTableSchema physicalTableSchema
```
QubitPi: Are we missing `@NotNull`?

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/QueryBuildingTestingResources.groovy
```diff
@@ -233,21 +233,21 @@ public class QueryBuildingTestingResources {
...
 
-        Map<Column, Set<Interval>> availabilityMap1 = [:]
-        Map<Column, Set<Interval>> availabilityMap2 = [:]
+        Map<String, Set<Interval>> availabilityMap1 = [:]
+        Map<String, Set<Interval>> availabilityMap2 = [:]
 
         [d1, d2, m1, m2, m3].each {
-            availabilityMap1.put(toColumn(it), [interval1].toSet())
-            availabilityMap2.put(toColumn(it), [interval2].toSet())
+            availabilityMap1.put(toColumn(it).getName(), [interval1].toSet())
```
QubitPi: I guess `[interval1] as Set` is more readable?

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/ConcreteAvailabilitySpec.groovy
```diff
@@ -16,35 +15,34 @@ import spock.lang.Specification
...
         concreteAvailability = new ConcreteAvailability(
                 TableName.of('table'),
-                [column1, column2, column3] as Set,
                 new TestDataSourceMetadataService([
-                        (column1): [interval1] as Set,
-                        (column2): [interval2] as Set,
-                        (new Column('ignored')): [new Interval('2010-01-01/2500-12-31')] as Set
+                        (columnPhysicalName1): [interval1] as Set,
+                        (columnPhysicalName2): [interval2] as Set,
+                        'hidden_column'     : [interval1] as Set
```
QubitPi: small - colon allignment

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/ConcreteAvailabilitySpec.groovy
```diff
@@ -16,35 +15,34 @@ import spock.lang.Specification
...
-    def "getAllAvailability returns the correct availabilities for each columns configured to the table"() {
+    def "getAllAvailability returns all availabilities for all column in datasource metadata service"() {
         expect:
         concreteAvailability.getAllAvailableIntervals() == [
-                (column1): [interval1],
-                (column2): [interval2],
-                (column3): [],
+                (columnPhysicalName1): [interval1],
+                (columnPhysicalName2): [interval2],
+                'hidden_column'     : [interval1],
```
QubitPi: small - colon allignment

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
garyluoex: I did a change here that makes it the same I think, [here](https://github.com/yahoo/fili/pull/226/files#diff-2cf930e01f4320995434024e1678779cR56)

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/DataSourceConstraint.java
cdeszaq: In general, `protected` and `private` constructors like this are relying on the fact that the external interfaces are ensuring the collections like this are immutable, so there's usually no need to defensively copy collections for those sorts of copy-constructors.

tldr: In this case, it's OK to use the item directly, not need to defensively copy

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
cdeszaq: What? I'm not sure I understand the problem here...

fili-core/src/main/java/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTable.java
```diff
@@ -44,7 +47,15 @@ public PermissiveConcretePhysicalTable(
...
+                new PermissiveAvailability(name, dataSourceMetadataService)
         );
     }
+
+    @Override
+    public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constraint) {
+        return getAvailability().getAvailableIntervals(PhysicalDataSourceConstraint.buildWithSchemaUnion(
+                constraint,
+                getSchema()
+        ));
```
cdeszaq: This should wrap differently:

```java
return getAvailability().getAvailableIntervals(
        PhysicalDataSourceConstraint.buildWithSchemaUnion(constraint, getSchema())
);
```

In general, the outer-most param list is where the wrapping should occur, rather than the inner-most, in a case like this where there are nested calls.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -61,52 +50,27 @@ public ConcreteAvailability(
...
         }
 
-        Map<String, List<Interval>> allAvailableIntervals = getAvailableIntervalsByTable();
+        Map<String, List<Interval>> allAvailableIntervals = getAllAvailableIntervals();
 
         // Need to ensure requestColumns is not empty in order to prevent returning null by reduce operation
         return new SimplifiedIntervalList(
                 requestColumns.stream()
-                        .map(columnName -> allAvailableIntervals.getOrDefault(columnName, Collections.emptyList()))
+                        .map(physicalName -> allAvailableIntervals.getOrDefault(physicalName, Collections.emptyList()))
```
cdeszaq: Does it make sense to use `new SimplifiedIntervalList()` instead of `Collections.emptyList()` for the default value here? It would bring it in line with what `getAllAvailableIntervals()` is doing in the `BasePhysicalTable`.

It may also allow for some optimizations or smarts with the rest of this stream, possibly letting us drop the wrapping construction of `new SimplifiedIntervalList` that is ultimately getting returned here if we can build one from this stream directly...

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
cdeszaq: While I think @garyluoex is right, in that it _should_ work as long as `PhysicalDataSourceConstraint::buildWithSchemaUnion` does the right thing. That said, I think the JavaDoc on _this_ method needs significant upgrades to reflect the new assumptions this method has, in that the unioning needs to have already been done to the inbound `constraint` object.

And with this new formulation, there is significant overlap between this method and the method in `ConcreteAvailability`, even though they do it a little differently. Here are the relevant portions:

```java
// ConcreteAvailability
.map(physicalName -> allAvailableIntervals.getOrDefault(physicalName, Collections.emptyList()))
.map(intervals -> (Collection<Interval>) intervals)
.reduce(null, IntervalUtils::getOverlappingSubintervals)

// PermissiveAvailability
.map(columnName -> allAvailableIntervals.getOrDefault(columnName, Collections.emptyList()))
.flatMap(List::stream)
.collect(SimplifiedIntervalList.getCollector());
```

If that core can be lifted into a common protected helper in `ConcreteAvailability`, that would be nice.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/PermissiveAvailabilitySpec.groovy
michael-mclawhorn: Basically a column that doesn't participate at all should still overwhelm everything else if it's big in time.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
garyluoex: I would change things to `Collections.emptyList` since it is shorter and looks less expensive than creating a `new SimplifiedIntervalList`, let me know otherwise.

fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
```diff
@@ -79,12 +82,21 @@ public DateTime getTableAlignment() {
 
     @Override
```
cdeszaq: Still needs to be addressed I think?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
cdeszaq: I think this should use a SIL as it's default. The biggest reasons is that it allows the following `map` and `reduce` calls to use SIL-aware methods which can be faster, since an SIL has known ordering and compactness qualities that a `List` does not. It also allows the SIL produced by that reduction to be returned directly, rather than having to convert a `List` into an SIL.

So, while `Collections.emptyList()` may seem shorter and faster than `new SimplifiedIntervalList()`, the net result may actually be that using SIL everywhere is faster overall.

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/PhysicalDataSourceConstraint.java
```diff
@@ -0,0 +1,82 @@
...
+    ) {
+        super(dataSourceConstraint);
+
+        Set<String> schemaColumnNames = physicalTableSchema.getColumnNames();
+
+        this.allColumnPhysicalNames = dataSourceConstraint.getAllColumnNames().stream()
+                .filter(schemaColumnNames::contains)
+                .map(physicalTableSchema::getPhysicalColumnName)
+                .collect(Collectors.toSet());
```
cdeszaq: Still needs to be addressed

fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
garyluoex: added comment on the interface

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -65,51 +64,37 @@
...
     public MetricUnionAvailability(@NotNull Set<PhysicalTable> physicalTables, @NotNull Set<Column> columns) {
-        metricColumns = Utils.getSubsetByType(columns, MetricColumn.class);
+        metricNames = Utils.getSubsetByType(columns, MetricColumn.class).stream()
+                .map(MetricColumn::getName)
+                .collect(Collectors.toSet());
 
-        // get a map from availability to its available metric columns intersected with configured metric columns
-        // i.e. metricColumns
-        availabilitiesToAvailableColumns = physicalTables.stream()
+        // map each metric to its corresponding availability object
```
cdeszaq: This comment change is losing the fact that this is intersected with the configured (ie. passed in) metric columns.

This comment change is also losing the fact that this is building `Map<Availability, Set<metric column>>`, rather than `Map<metric column, Availability>`, which seems to be what the comment is saying

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -165,59 +162,26 @@ public String toString() {
...
-        // {metricCol1: [availability1, availability2]}
-        return metricColumnSetMap.entrySet().stream()
-                .filter(entry -> entry.getValue().size() > 1)
-                .collect(
-                        Collectors.toMap(
-                                Map.Entry::getKey,
-                                Map.Entry::getValue
-                        )
-                );
+        return availabilityToMetricNames.entrySet().stream()
```
cdeszaq: Could just use `.values()` and skip the extra `map` operation

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -26,46 +19,44 @@
...
     }
 
     @Override
     public String toString() {
-        return String.format("PermissiveAvailability with table name = %s and Configured columns = [%s]",
-                getName().asName(),
-                getColumnNames()
+        return String.format("PermissiveAvailability with table name = %s",
+                getName().asName()
         );
```
cdeszaq: Doesn't need to wrap?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PermissiveAvailability.java
```diff
@@ -26,46 +19,44 @@
...
      */
     @Override
-    public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint ignoredConstraints) {
-        Map<String, List<Interval>> allAvailableIntervals = getAvailableIntervalsByTable();
-        return getColumnNames().stream()
-                .map(columnName -> allAvailableIntervals.getOrDefault(columnName, Collections.emptyList()))
-                .flatMap(List::stream)
-                .collect(SimplifiedIntervalList.getCollector());
+    public SimplifiedIntervalList getAvailableIntervals(PhysicalDataSourceConstraint ignoredConstraint) {
+
```
cdeszaq: Can remove blank line

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/PhysicalDataSourceConstraint.java
```diff
@@ -0,0 +1,127 @@
...
+     * @param filterDimensions  Filtered dimensions
+     * @param metricDimensions  Metric related dimensions
+     * @param metricNames  Names of metrics
+     * @param allDimensions  Set of all dimension objects
+     * @param allDimensionNames  Set of all dimension names
+     * @param allColumnNames  Set of all column names
+     * @param apiFilters  Map of dimension to its set of API filters
+     * @param allColumnPhysicalNames  Set of all column physical names
+     */
+    protected PhysicalDataSourceConstraint(
```
cdeszaq: Why do we have this constructor? It looks like it does exactly the same thing as the private constructor just above, except unpacks the parameters. Can we remove this one?

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy
```diff
@@ -54,79 +53,74 @@ class MetricUnionAvailabilitySpec extends Specification {
...
-        Set<MetricColumn> availableColumns2 = availabilitiesToAvailableColumns.get(availability2)
-        availableColumns2.contains(metricColumn2)
+        Set<String> availableColumns2 = availabilitiesToAvailableColumns.get(availability2)
+        availableColumns2.contains(metric2)
         !availableColumns2.contains(tableColumn1)
         !availableColumns2.contains(tableColumn2)
     }
 
     @Unroll
     def "checkDuplicateValue returns duplicate metric column names or empty, if no duplicates, from 2 physical tables in the case of #caseDescription"() {
```
cdeszaq: Test name needs to be updated

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/PhysicalDataSourceConstraint.java
```diff
@@ -41,7 +38,7 @@ public PhysicalDataSourceConstraint(
     }
 
     /**
-     * Constructor, use with care, beware of caching behavior.
+     * Constructor, use with care, beware of inconsistent caching behavior.
```
cdeszaq: I think I'm still confused. What inconsistencies are possible?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
cdeszaq: Still needs addressing?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -148,9 +147,15 @@ public MetricUnionAvailability(@NotNull Set<PhysicalTable> physicalTables, @NotN
     }
 
     @Override
-    public SimplifiedIntervalList getAvailableIntervals(PhysicalDataSourceConstraint constraints) {
+    public SimplifiedIntervalList getAvailableIntervals(PhysicalDataSourceConstraint constraint) {
+
+        // If there are columns requested that are not configured on this table
+        if (constraint.getAllColumnPhysicalNames().size() != constraint.getAllColumnNames().size()) {
```
cdeszaq: This will likely always be true, due to the defaulting behavior of how mapping from `logicalToPhysicalName` works. We probably need to look at the union of the columns in the subconstraints to see if any of the requested columns are not present in the subconstraints.

garyluoex: There is also a filtering step during the mapping.

fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
```diff
@@ -96,6 +97,22 @@ public DateTime getTableAlignment() {
...
+
+        Set<String> tableColumnNames = getSchema().getColumnNames();
+        Set<String> invalidColumnNames = constraint.getAllColumnNames().stream()
+                .filter(name -> !tableColumnNames.contains(name))
+                .collect(Collectors.toSet());
+
+        if (!constraint.getAllColumnNames().stream().allMatch(tableColumnNames::contains)) {
+            String message = String.format(
+                    "Received invalid request requesting for columns: %s that is not available in this table: %s",
+                    invalidColumnNames.stream().collect(Collectors.joining(",")),
```
cdeszaq: Rather than computing `invalidColumnNames` every time, let's replace the formatting stream here with the computation for the invalid names:

```java
constraint.getAllColumnNames().stream().filter(name -> !tableColumnNames.contains(name)).collect(Collectors.joining(","))
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/ConcretePhysicalTableSpec.groovy
```diff
@@ -86,7 +86,20 @@ class ConcretePhysicalTableSpec extends Specification {
         dimensionColumn             | intervalSet1
         metricColumn1               | intervalSet2
         metricColumn2               | intervalSet3
-        new Column("MissingName")   | [] as Set
+    }
+
+    @Unroll
```
cdeszaq: No need for `@Unroll` here, since there's no `where` block and no templated test name.

cdeszaq:
Just those few trivial things, and this will merge.

PR #229, opened 2017-04-05
Add 'satisfies' as complementary relationship to 'satisfiedBy'

michael-mclawhorn:


fili-core/src/test/groovy/com/yahoo/bard/webservice/data/time/TimeGrainSpec.groovy
```diff
@@ -2,6 +2,9 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.data.time
 
+import static com.yahoo.bard.webservice.data.time.DefaultTimeGrain.*
```
cdeszaq: no wildcard imports

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/query/AllGranularitySpec.groovy
```diff
@@ -26,6 +28,19 @@ class AllGranularitySpec extends Specification {
...
+    def "All granularity #relationship  #grain"() {
+
+        expect:
+        allGranularity.satisfiedBy(grain)
+        grain.satisfies(allGranularity)
+
+        where:
+
+        grain << ((DefaultTimeGrain.values() as List) + allGranularity)
+        relationship = " is satisfied by "
```
cdeszaq: Is this supposed to be the same every time? If so, should we just move it up into the description directly rather than being a param?

PR #230, opened 2017-04-05
Implement method for finding coarsest ZonedTimeGrain

ghost:


fili-core/src/test/groovy/com/yahoo/bard/webservice/util/IntervalUtilsSpec.groovy
```diff
@@ -326,6 +328,32 @@ class IntervalUtilsSpec extends Specification {
         DAY   | ["2012-02-03/2017-04-04"] | ["2012-02-03/2012-05-04", "2017-02-03/2017-04-04"]
     }
 
+    @Unroll
+    def "getCoarsestTimeGrain returns duration of #expected among #duration1, #duration2, #duration3"() {
```
cdeszaq: Why are we testing / asserting behavior about durations when we are generally dealing with `ZonedTimeGrain`?

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/IntervalUtilsSpec.groovy
```diff
@@ -326,6 +328,32 @@ class IntervalUtilsSpec extends Specification {
...
+        zonedTimeGrain3.getEstimatedDuration() >> new Duration(duration3)
+
+        expect:
+        IntervalUtils.getCoarsestTimeGrain([zonedTimeGrain1, zonedTimeGrain2, zonedTimeGrain3]).getEstimatedDuration().equals(new Duration(expected))
+
+        where:
+        duration1 | duration2 | duration3 | expected
+        1         | 2         | 3         | 1
+        1         | 1         | 3         | 1
+        3         | 3         | 3         | 3
```
cdeszaq: Any reason we're using mocked grains instead of using real grains and giving them a timezone?

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/IntervalUtilsSpec.groovy
```diff
@@ -326,6 +328,32 @@ class IntervalUtilsSpec extends Specification {
...
+        zonedTimeGrain2.getEstimatedDuration() >> new Duration(duration2)
+        zonedTimeGrain3.getEstimatedDuration() >> new Duration(duration3)
+
+        expect:
+        IntervalUtils.getCoarsestTimeGrain([zonedTimeGrain1, zonedTimeGrain2, zonedTimeGrain3]).getEstimatedDuration().equals(new Duration(expected))
+
+        where:
+        duration1 | duration2 | duration3 | expected
+        1         | 2         | 3         | 1
+        1         | 1         | 3         | 1
```
cdeszaq: We should make sure to test what happens when we have the same grain but different time zones.

fili-core/src/main/java/com/yahoo/bard/webservice/util/IntervalUtils.java
```diff
@@ -209,4 +211,20 @@ public static long getTotalDuration(Collection<Interval> intervals) {
                 .map(Interval::getStart)
                 .reduce(Utils::getMinValue);
     }
+
+    /**
+     * Returns the coarsest ZonedTimeGrain among a set of ZonedTimeGrains.
+     * <p>
+     * If the set of ZonedTimeGrains is empty, return null.
```
cdeszaq: Why are we returning `null` here? Do we really expect `null` to be a valid return value? If we do, we should make that explicit in the return type by using `Optional` and _force_ callers to deal with the fact that it could come back as null.

cdeszaq (changes requested):

fili-core/src/main/java/com/yahoo/bard/webservice/util/IntervalUtils.java
```diff
@@ -209,4 +211,20 @@ public static long getTotalDuration(Collection<Interval> intervals) {
...
+     *
+     * @param zonedTimeGrains  A set of ZonedTimeGrains among which the coarsest ZonedTimeGrain is to be returned.
+     *
+     * @return the coarsest ZonedTimeGrain among a set of ZonedTimeGrains
+     */
+    public static ZonedTimeGrain getCoarsestTimeGrain(Collection<ZonedTimeGrain> zonedTimeGrains) {
+        return zonedTimeGrains.stream()
+                .sorted(Comparator.comparing(ZonedTimeGrain::getEstimatedDuration))
```
QubitPi: @cdeszaq Because `GranularityComparator` compares by taking `PhysicalTable` not `ZonedTimeGrain`. I can write a comparator for `ZonedTimeGrain` if that's you would like to do.

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/IntervalUtilsSpec.groovy
QubitPi: @cdeszaq By digging into the source code, I found that comparing `ZonedTimeGrain` is the same thing as comparing duration. Would you like to change the behavior anyway?

cdeszaq: Yes. It's better to keep the tests focused on the external interface and the behavior, and to try to not duplicate what the code under test is doing when testing it.

fili-core/src/main/java/com/yahoo/bard/webservice/util/IntervalUtils.java
cdeszaq: Good point. That's unfortunate. If the intent of this helper method is to make it easy to select the coarsest grain from a collection of PhysicalTables, perhaps we should convert this method to do that, and then use the GranularityComparator?

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/IntervalUtilsSpec.groovy
```diff
@@ -326,6 +331,46 @@ class IntervalUtilsSpec extends Specification {
...
+        ).get() == new ZonedTimeGrain(expectedTimeGrain, DateTimeZone.forID(expectedTimeZone))
+
+        where:
+        timeGrain1 | timeGrain2 | timeGrain3 | timeZone1         | timeZone2             | timeZone3         | expectedTimeGrain | expectedTimeZone  | description
+        DAY        | DAY        | DAY        | 'America/Chicago' | 'America/Los_Angeles' | 'America/Phoenix' | DAY               | 'America/Chicago' | 'same grain but different time zones'
+        MINUTE     | HOUR       | DAY        | 'America/Phoenix' | 'America/Phoenix'     | 'America/Phoenix' | DAY               | 'America/Phoenix' | 'same time zone but different grans'
+        MINUTE     | HOUR       | DAY        | 'America/Chicago' | 'America/Los_Angeles' | 'America/Phoenix' | DAY               | 'America/Phoenix' | 'different grains and different time zones'
+    }
+
+    def "getCoarsestTimeGrain returns null on empty input time grain collections"() {
```
cdeszaq: Should be `returns empty`, not `null`, right?

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/IntervalUtilsSpec.groovy
```diff
@@ -326,6 +331,46 @@ class IntervalUtilsSpec extends Specification {
...
+        expect:
+        IntervalUtils.getCoarsestTimeGrain(Collections.emptyList()) == Optional.empty()
```
cdeszaq: Just use groovy's list literal for an empty list: `[]`

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/IntervalUtilsSpec.groovy
```diff
@@ -326,6 +331,46 @@ class IntervalUtilsSpec extends Specification {
...
+                        physicalTable2,
+                        physicalTable3
+                ]
+        ).get() == new ZonedTimeGrain(expectedTimeGrain, DateTimeZone.forID(expectedTimeZone))
+
+        where:
+        timeGrain1 | timeGrain2 | timeGrain3 | timeZone1         | timeZone2             | timeZone3         | expectedTimeGrain | expectedTimeZone  | description
+        DAY        | DAY        | DAY        | 'America/Chicago' | 'America/Los_Angeles' | 'America/Phoenix' | DAY               | 'America/Chicago' | 'same grain but different time zones'
+        MINUTE     | HOUR       | DAY        | 'America/Phoenix' | 'America/Phoenix'     | 'America/Phoenix' | DAY               | 'America/Phoenix' | 'same time zone but different grans'
+        MINUTE     | HOUR       | DAY        | 'America/Chicago' | 'America/Los_Angeles' | 'America/Phoenix' | DAY               | 'America/Phoenix' | 'different grains and different time zones'
```
cdeszaq: Could we add some more cases that use week, month, quarter, and year? Some of those (ie. week) are odd in terms of how they line up, so that would be good to double-check.

cdeszaq (approved):
Just 3 very small things. I'm good with this once those are looked at.

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/IntervalUtilsSpec.groovy
QubitPi: Yes

garyluoex (approved):
Good

PR #231, opened 2017-04-06
Should also setConnectTimeout when using setReadTimeout

wcekan:
Extends #160


archolewa:
Should add an entry to the changelog. Other than that, 👍

cdeszaq:
Meh, merging. No harm in getting a small tweak in quickly.

PR #232, opened 2017-04-07
Improve Lucene search speed with internal cache

wcekan:
We found some Lucene searches could be much too slow to be repeated often.  The cache speeds our searches significantly.

The cache will always be cleared on writes.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
@@ -33,9 +33,11 @@ import org.apache.lucene.index.Term;
 import org.apache.lucene.search.BooleanClause;
 import org.apache.lucene.search.BooleanQuery;
 import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.LRUQueryCache;
 import org.apache.lucene.search.MatchAllDocsQuery;
 import org.apache.lucene.search.PrefixQuery;
 import org.apache.lucene.search.Query;
+import org.apache.lucene.search.QueryCachingPolicy;
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.TermQuery;
 import org.apache.lucene.search.TimeLimitingCollector;
@@ -76,6 +78,21 @@ public class LuceneSearchProvider implements SearchProvider {
     public static final int LUCENE_SEARCH_TIMEOUT_MS = SYSTEM_CONFIG.getIntProperty(
             SYSTEM_CONFIG.getPackageVariableName("lucene_search_timeout_ms"), 600000
     );
+
+    public static final LRUQueryCache LUCENE_QUERY_CACHE;
+    static {
+        // Lucene 5.3.0 disables query caching.  Create cache for use here.
+        // min of 32MB or 5% of the heap size.  Same default as Lucene 5.4.0.
+        final long maxRamBytesUsed = SYSTEM_CONFIG.getLongProperty(
+            SYSTEM_CONFIG.getPackageVariableName("lucene_cache_size_bytes"),
+            Math.min(1L << 25, Runtime.getRuntime().maxMemory() / 20));
+
+        // default max cached queries based on memory used
+        final int maxCachedQueries = SYSTEM_CONFIG.getIntProperty(
+            SYSTEM_CONFIG.getPackageVariableName("lucene_cache_queries"),
+            (int) Math.max(1000L, Math.min(1000000L, maxRamBytesUsed / 1000)));
+        LUCENE_QUERY_CACHE = new LRUQueryCache(maxCachedQueries, maxRamBytesUsed);
+    }
 
     /**
      * The maximum number of results per page.
@@ -155,6 +172,8 @@ public class LuceneSearchProvider implements SearchProvider {
 
             // Open a new IndexSearcher on a new DirectoryReader
             luceneIndexSearcher = new IndexSearcher(DirectoryReader.open(luceneDirectory));
+            luceneIndexSearcher.setQueryCache(LUCENE_QUERY_CACHE);
+            luceneIndexSearcher.setQueryCachingPolicy(QueryCachingPolicy.ALWAYS_CACHE);
         } catch (IOException reopenException) {
             // If there is no index file, this is expected. On the 1st time through, write an empty index and try again
             if (firstTimeThrough) {
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/cache/HashDataCache.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/cache/HashDataCache.java
@@ -10,6 +10,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.Serializable;
+import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.AbstractMap;
@@ -76,7 +77,7 @@ public class HashDataCache<T extends Serializable> implements DataCache<T> {
     public String hash(String key) {
         try {
             MessageDigest md = MessageDigest.getInstance(algorithm);
-            byte[] keyBytes = key.getBytes("UTF-8");
+            byte[] keyBytes = key.getBytes(StandardCharsets.UTF_8);
             md.update(keyBytes, 0, keyBytes.length);
             byte[] binaryhash = md.digest();
             return Base64.encode(binaryhash);
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/cache/MemTupleDataCache.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/cache/MemTupleDataCache.java
@@ -9,6 +9,7 @@ import net.spy.memcached.MemcachedClient;
 
 import java.io.IOException;
 import java.io.Serializable;
+import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.Base64;
@@ -73,7 +74,7 @@ public class MemTupleDataCache<V extends Serializable>
     protected String hash(String key) {
         try {
             MessageDigest gen = HASH_GENERATOR.get();
-            byte[] keyBytes = key.getBytes("UTF-8");
+            byte[] keyBytes = key.getBytes(StandardCharsets.UTF_8);
             gen.update(keyBytes, 0, keyBytes.length);
             byte[] binaryhash = gen.digest();
 


fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -624,11 +648,22 @@ private Query getFilterQuery(Set<ApiFilter> filters) {
...
                 Collections.unmodifiableList(filteredDimRows.stream().collect(Collectors.toList())),
                 paginationParameters,
                 documentCount
         );
+        if (LUCENE_CACHE_SIZE > 0) {
+            try {
+                SinglePagePagination<DimensionRow> clone = SerializationUtils.clone(page);
+                synchronized (queryCache) {
+                    queryCache.put(new Pair<>(query, paginationParameters), clone);
+                }
```
will-lauer: Synchronizing on the cache is evil (although necessary in some scenarios) and is going to impact your overall throughput. Maybe we should be looking at a ConcurrentLinkedHashMap implementation for this (like the one available at https://github.com/ben-manes/caffeine)

wcekan: Actually JRE 1.8 has vastly improved synchronization handling.  Also, since this lock is only on put after long call to Lucene and of very brief duration, it should be very unlikely to cause contention.

However, if you know of a concurrent linked hash map that performs the required LRU mechanism, I would happily switch.

wcekan:
Will close for now.  We may need to simply tune Lucene query caches better.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -597,6 +591,11 @@ private Query getFilterQuery(Set<ApiFilter> filters) {
 
         lock.readLock().lock();
         try {
+            // return from cache if available
+            SinglePagePagination<DimensionRow> page = queryCache.get(new Pair<>(query, paginationParameters));
```
will-lauer: This bothers me. You have an operation that _looks_ like a read inside a readers lock, but since this cache is actually mutating internal state as it handles the LRU magic, it may actually need a writers lock to avoid corrupting itself.

cdeszaq:
@wcekan how often were you seeing searches being repeated? It shouldn't be more than once per request I think. We'd talked in the past about putting a cache in front of Lucene and possibly kicking it over to be async, but hadn't seen enough need to tackle that yet. Any more info you might have would be wonderful.

wcekan:
@cdeszaq I changed to pull in the Lucene LRU query cache.  Lucene `5.3.0` disables caching while `5.4.0` enables it by default.  We are currently seeing a single Fili request repeat the same Lucene query for each filter mapper and again for the transform mapper.

archolewa:
Rather than adding the cache directly to the `LuceneSearchProvider`, I think we should instead create a `CachingSearchProvider` that wraps an arbitrary `SearchProvider`, and only delegates to said `SearchProvider` when there is a cache miss.

ben-manes:
Per @will-lauer's comment,

[https://github.com/ben-manes/concurrentlinkedhashmap](https://github.com/ben-manes/concurrentlinkedhashmap) would give you a small, concurrent, JDK6 LRU cache. That would be much faster than `LRUQueryCache` which is guarded by a lock, if and only if you foresee contention due to multiple readers.

Caffeine would provide a concurrent, JDK8 cache with a near-optimal hit rate. Since search is frequency-biased there is typically a [large gap](https://github.com/ben-manes/caffeine/wiki/Efficiency#search) between LRU and optimal, so you might get a significant boost.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -77,6 +79,21 @@
...
+            SYSTEM_CONFIG.getPackageVariableName("lucene_cache_size_bytes"),
+            Math.min(1L << 25, Runtime.getRuntime().maxMemory() / 20));
+
+        // default max cached queries based on memory used
+        final int maxCachedQueries = SYSTEM_CONFIG.getIntProperty(
+            SYSTEM_CONFIG.getPackageVariableName("lucene_cache_queries"),
+            (int) Math.max(1000L, Math.min(1000000L, maxRamBytesUsed / 1000)));
+        LUCENE_QUERY_CACHE = new LRUQueryCache(maxCachedQueries, maxRamBytesUsed);
+    }
+
```
michael-mclawhorn: Probably the cache minimum and/or heap size percentage should be tunable via configuration.
Where did these numbers come from?

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -77,6 +79,21 @@
             SYSTEM_CONFIG.getPackageVariableName("lucene_search_timeout_ms"), 600000
     );
 
+    public static final LRUQueryCache LUCENE_QUERY_CACHE;
+    static {
+        // Lucene 5.3.0 disables query caching.  Create cache for use here.
+        // min of 32MB or 5% of the heap size.  Same default as Lucene 5.4.0.
+        final long maxRamBytesUsed = SYSTEM_CONFIG.getLongProperty(
+            SYSTEM_CONFIG.getPackageVariableName("lucene_cache_size_bytes"),
+            Math.min(1L << 25, Runtime.getRuntime().maxMemory() / 20));
```
QubitPi: It's better if closing parenthesis goes to the next line.

```
final long maxRamBytesUsed = SYSTEM_CONFIG.getLongProperty(
        SYSTEM_CONFIG.getPackageVariableName("lucene_cache_size_bytes"),
        Math.min(1L << 25, Runtime.getRuntime().maxMemory() / 20)
);
```

@cdeszaq Does fili refer 8 space indent on the next-line-continuation?

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -77,6 +79,21 @@
...
+
+        // default max cached queries based on memory used
+        final int maxCachedQueries = SYSTEM_CONFIG.getIntProperty(
+            SYSTEM_CONFIG.getPackageVariableName("lucene_cache_queries"),
+            (int) Math.max(1000L, Math.min(1000000L, maxRamBytesUsed / 1000)));
```
QubitPi: It's better if closing parenthesis goes to the next line.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
cdeszaq: Line wraps should have 2 levels of indent, yes.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
cdeszaq: They now appear to use the config if set, and otherwise default, so that aspect has been addressed. But I'm still eager to hear why these are the defaults?

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
wcekan: My IDE flat out refuses :(

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
wcekan: @cdeszaq @michael-mclawhorn 
https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java#L121

wcekan: I see no explanation of why these were used.

cdeszaq: 😄 "Their code uses this" is sufficient for me.

wcekan: Perhaps you consider instead upgrading to a more current Lucene.

cdeszaq: We have in fact updated to the most recent Lucene: https://github.com/yahoo/fili/blob/master/pom.xml#L81

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
wcekan: I see you publish only a IntelliJ code style.  Have you considered adding Eclipse as well?

https://github.com/yahoo/fili/blob/master/Fili-Project-intellij-code-style.xml

cdeszaq: I'd love to include an Eclipse checkstyle, but I'm not sure you'll be able to get it to match what IntelliJ has fully. I don't think Eclipse supports chop-wrapping of param lists. The only reason it's not included yet is that most of the active committers don't use Eclipse 😄 

So, feel free to submit one, and perhaps @will-lauer can take a look at it (since I think he uses Eclipse), and we'll get it added.

<sub>It's also been brought up that the existing IntelliJ config is a bit out of date vs. the Checkstyle config, so we need work on that too, unfortunately 😢 </sub>

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
wcekan: In that case this whole PR is obsolete.

cdeszaq: Let's leave it outstanding for a little bit longer. It would be good to confirm that the most recent version of Fili (the version with the updated Lucene dependency) fixes you guys. If we can confirm that, I'm cool closing this PR.

If you don't want / can't take the most recent dev version of Fili, I _think_ you should be able to bump your local Lucene dependency, since the Fili PR to bump it didn't require any code changes.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
will-lauer: If you are going to start getting into eclipse vs intellij formatting and requiring a specific formatting, I suggest you abandon using the IDE to enforce the formatting and instead add a maven plugin to do the formatting for you. The build would fix the formatting, requiring that people build before submitting their PR. That would move all formatting related stuff to a central location and remove any IDE related arguments.

cdeszaq: Ultimately, checkstyle and PR reviews are the arbiters of truth, and the IDE configs are really only a guide to help people get up and running with a sane starting point. If there are some quick wins we can get to handle unambiguous common cases, I'm all for automating wherever we can to make everyone's life easier. Unfortunately, not everything has a clear "doing it this way makes it easier to read and understand", which is the ultimate goal.

I'm not familiar with the Maven plugins that do auto-formatting, but if someone wants to take a stab at hooking one up and starting with some simple rules (eg. indent levels and spaces instead of tabs) to keep changes generally in line with the current codebase, I would be happy to pull that in.

cdeszaq:
With the upgrade to the Lucene version recently merging to master, this PR's effective changes are activated by default, since the newer version of Lucene activates it's cache by default. Given that, I'm closing this.

PR #234, opened 2017-04-12
Stops using `IndexSearcher::count` in Lucene.

archolewa:
--It looks like `IndexSearcher::count` fires off an entire query, and
then counts the number of documents that were hit. Currently, we are
using it to determine the total number of dimension rows that satisfy a
given query. However, this is unnecessary and wasteful, because the
`TopDocs` object returned by the query for the actual data _also_
contains information on the total number of documents hit by the query.


michael-mclawhorn:
I would put a CHANGELOG entry around this even though it's a small internal change.

archolewa:
@michael-mclawhorn Is the test I added insufficient somehow? In our case, we only use the `documentCount` to populate the pagination metadata, and I verify that the pagination metadata is correct.

cdeszaq:
I think the test this PR adds is a good addition. I think Mike's point is more that we're making an internal code change which should result in a performance boost, but we're not really checking the internal implementation anywhere.

Personally, I'm 100% OK with _not_ testing this particular difference:

1. It doesn't (shouldn't) change the correctness of the result
2. It's going to be difficult to test, since it's _very_ intimate with the class itself (and even that particular method)
3. I don't think we get much value from testing it

PR #235, opened 2017-04-12
Update dependency versions

QubitPi:


pom.xml
```diff
@@ -72,12 +72,12 @@
...
-        <version.slf4j>1.7.21</version.slf4j>
+        <version.slf4j>1.7.25</version.slf4j>
         <version.servlet>3.1.0</version.servlet>
         <version.jersey>2.24</version.jersey>
-        <version.hk2>2.5.0-b05</version.hk2>
-        <version.metrics>3.1.2</version.metrics>
-        <version.logback>1.1.7</version.logback>
+        <version.hk2>2.5.0-b36</version.hk2>
+        <version.metrics>3.2.2</version.metrics>
+        <version.logback>1.2.3</version.logback>
```
cdeszaq: What changes in LogBack are we picking up when going from `v1.1.7` to `v1.2.3`?

Also, do we know what's changing in hk2 between versions `b05` and `b36`?

PR #242, opened 2017-04-14
Implement BaseCompositePhysicalTable

QubitPi:
screen shot 2017-04-14 at 10 38 14 am

PR https://github.com/yahoo/fili/pull/195 reveals a need to implement a common class between `PartitionCompositeTable` and `MetricUnionCompositeTable`.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseCompositePhysicalTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseCompositePhysicalTable.java
@@ -5,7 +5,6 @@ package com.yahoo.bard.webservice.table;
 import com.yahoo.bard.webservice.data.config.names.TableName;
 import com.yahoo.bard.webservice.data.time.ZonedTimeGrain;
 import com.yahoo.bard.webservice.table.availability.Availability;
-import com.yahoo.bard.webservice.util.IntervalUtils;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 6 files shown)


michael-mclawhorn:
The changes seem fine.  Rick and I discussed at some point whether or not we should take out the 'we'll calculate the grain for you' from the composite tables and replace it with only passing in an acceptable grain.

fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseCompositePhysicalTable.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Set;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * An implementation of BasePhysicalTable that contains multiple tables.
+ */
+public class BaseCompositePhysicalTable extends BasePhysicalTable {
+
+    private static final Logger LOG = LoggerFactory.getLogger(MetricUnionCompositeTable.class);
```
cdeszaq: Wrong logger

fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseCompositePhysicalTable.java
```diff
@@ -0,0 +1,90 @@
...
+    public BaseCompositePhysicalTable(
+            @NotNull TableName name,
+            @NotNull Set<Column> columns,
+            @NotNull Set<PhysicalTable> physicalTables,
+            @NotNull Map<String, String> logicalToPhysicalColumnNames,
+            @NotNull Availability availability
+    ) {
+        super(
+                name,
+                IntervalUtils.getCoarsestTimeGrain(physicalTables).orElseThrow(() -> {
```
cdeszaq: As commented before, just take in the grain, don't compute it. A param will need to be added, and the javadoc for `physicalTables` likely needs to be updated as well.

fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseCompositePhysicalTable.java
```diff
@@ -0,0 +1,90 @@
...
+     */
+    private static void verifyGrainSatisfiesAllTables(
+            ZonedTimeGrain coarsestTimeGrain,
+            Set<PhysicalTable> physicalTables,
+            TableName name
+    ) throws IllegalArgumentException {
+        if (!physicalTables.stream()
+                .map(PhysicalTable::getSchema)
+                .map(PhysicalTableSchema::getTimeGrain)
+                .allMatch(grain -> grain.satisfiedBy(coarsestTimeGrain))) {
```
cdeszaq: Can switch to a method reference for the new `satisfies` method on grain: `coarsestTimeGrain::satisfies`

fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseCompositePhysicalTable.java
```diff
@@ -0,0 +1,90 @@
...
+            String message = String.format("There is no mutually satisfying grain among: %s for current table " +
+                            "%s",
```
cdeszaq: When wrapping, chop-wrap all parameters. That also means that the string doesn't need to be split either:

```java
        "There is no mutually satisfying grain among: %s for current table %s",
        physicalTables,
        name.asName()
);
```

PR #244, opened 2017-04-18
Pr/195

michael-mclawhorn:
Local copy of #195 
Rebased, pushed to github mail
Updated javadoc
Added test for partition layering

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ Current
 -------
 ### Added:
 
+- [Add partition availability and table](https://github.com/yahoo/fili/pull/244)
+
 - [Add constructor to specify DruidDimensionLoader dimensions directly](https://github.com/yahoo/fili/pull/191)
 
 - [Add `IntervalUtils::getTimeGrain` to determine the grain given an Interval](https://github.com/yahoo/fili/pull/191)
@@ -91,6 +93,10 @@ Current
 - [Support timeouts for lucene search provider](https://github.com/yahoo/fili/pull/183)
 
 ### Changed:
+
+- [Update availabilities for PartitionAvailability](https://github.com/yahoo/fili/pull/244)
+    * Created `BaseCompositeAvailability` for common features
+    * Refactored `DataSourceMetadataService` methods to use SimplifiedIntervaList to standardize intersections
 
 - [Queries to Druid Web Service now return a Future](https://github.com/yahoo/fili/pull/254)
     * Queries now return a `Future<Response>` in addition to having method callbacks.
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseCompositePhysicalTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseCompositePhysicalTable.java
@@ -47,33 +47,31 @@ public abstract class BaseCompositePhysicalTable extends BasePhysicalTable {
                 logicalToPhysicalColumnNames,
                 availability
         );
-        verifyGrainSatisfiesAllTables(getSchema().getTimeGrain(), physicalTables, name);
+        verifyGrainSatisfiesAllSourceTables(getSchema().getTimeGrain(), physicalTables);
     }
 
     /**
      * Verifies that the coarsest ZonedTimeGrain satisfies all tables.
      *
-     * @param coarsestTimeGrain  The coarsest ZonedTimeGrain to be verified
+     * @param timeGrain  The coarsest ZonedTimeGrain being validated
      * @param physicalTables  A set of PhysicalTables whose ZonedTimeGrains are checked to make sure
      * they all satisfies with the given coarsest ZonedTimeGrain
-     * @param name  Name of the current MetricUnionCompositeTable that represents set of fact table names
-     * joined together
      *
      * @throws IllegalArgumentException when there is no mutually satisfying grain among the table's time grains
      */
-    private static void verifyGrainSatisfiesAllTables(
-            ZonedTimeGrain coarsestTimeGrain,
-            Set<PhysicalTable> physicalTables,
-            TableName name
+    private void verifyGrainSatisfiesAllSourceTables(
+            ZonedTimeGrain timeGrain,
+            Set<PhysicalTable> physicalTables
     ) throws IllegalArgumentException {
         if (!physicalTables.stream()
                 .map(PhysicalTable::getSchema)
                 .map(PhysicalTableSchema::getTimeGrain)
-                .allMatch(coarsestTimeGrain::satisfies)) {
+                .allMatch(timeGrain::satisfies)) {
             String message = String.format(
-                    "There is no mutually satisfying grain among: %s for current table %s",
-                    physicalTables,
-                    name.asName()
+                    "There is no mutually satisfying grain among: %s for composite table %s",
+                    physicalTables.stream()
+                            .map(Table::getName),
+                    getTableName()
             );
             LOG.error(message);
             throw new IllegalArgumentException(message);
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseCompositePhysicalTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseCompositePhysicalTable.java
@@ -11,6 +11,8 @@ import org.slf4j.LoggerFactory;
 
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
 import javax.validation.constraints.NotNull;
 
@@ -63,14 +65,21 @@ public abstract class BaseCompositePhysicalTable extends BasePhysicalTable {
             ZonedTimeGrain timeGrain,
             Set<PhysicalTable> physicalTables
     ) throws IllegalArgumentException {
-        if (!physicalTables.stream()
-                .map(PhysicalTable::getSchema)
-                .map(PhysicalTableSchema::getTimeGrain)
-                .allMatch(timeGrain::satisfies)) {
+        Predicate<PhysicalTable> tableDoesNotSatisfyFilter = (physicalTable) -> ! physicalTable.getSchema()
+                .getTimeGrain()
+                .satisfies(timeGrain);
+
+        Set<String> unsatisfied = physicalTables.stream()
+                .filter(tableDoesNotSatisfyFilter)
+                .map(PhysicalTable::getTableName)
+                .map(TableName::asName)
+                .collect(Collectors.toSet());
+
+        if (! unsatisfied.isEmpty()) {
             String message = String.format(
-                    "There is no mutually satisfying grain among: %s for composite table %s",
-                    physicalTables.stream().map(Table::getName),
-                    getTableName()
+                    "Time grain: '%s' cannot be satisfied by source table(s) %s",
+                    timeGrain,
+                    unsatisfied
             );
             LOG.error(message);
             throw new IllegalArgumentException(message);
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseCompositePhysicalTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseCompositePhysicalTable.java
@@ -69,8 +69,7 @@ public abstract class BaseCompositePhysicalTable extends BasePhysicalTable {
                 .allMatch(timeGrain::satisfies)) {
             String message = String.format(
                     "There is no mutually satisfying grain among: %s for composite table %s",
-                    physicalTables.stream()
-                            .map(Table::getName),
+                    physicalTables.stream().map(Table::getName),
                     getTableName()
             );
             LOG.error(message);
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/table/PartitionCompositeTable.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/table/PartitionCompositeTable.java
@@ -1,0 +1,59 @@
+// Copyright 2017 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.table;
+
+import com.yahoo.bard.webservice.data.config.names.TableName;
+import com.yahoo.bard.webservice.data.time.ZonedTimeGrain;
+import com.yahoo.bard.webservice.table.availability.Availability;
+import com.yahoo.bard.webservice.table.availability.PartitionAvailability;
+import com.yahoo.bard.webservice.table.resolver.DataSourceFilter;
+
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * An implementation of <tt>BaseCompositePhysicalTable</tt> backed by partition availability.
+ */
+public class PartitionCompositeTable extends BaseCompositePhysicalTable {
+
+    /**
+     * Constructor.
+     *
+     * @param name  Name of the physical table as TableName, also used as fact table name
+     * @param timeGrain  The time grain of the table. The time grain has to satisfy all grains of the tables
+     * @param columns  The columns for this table
+     * @param logicalToPhysicalColumnNames  Mappings from logical to physical names
+     * @param availabilityFilters  A map of table to filters which apply to those tables
+     */
+    public PartitionCompositeTable(
+            @NotNull TableName name,
+            @NotNull ZonedTimeGrain timeGrain,
+            @NotNull Set<Column> columns,
+            @NotNull Map<String, String> logicalToPhysicalColumnNames,
+            @NotNull Map<PhysicalTable, DataSourceFilter> availabilityFilters
+    ) {
+        super(
+                name,
+                timeGrain,
+                columns,
+                availabilityFilters.keySet(),
+                logicalToPhysicalColumnNames,
+                buildAvailability(availabilityFilters)
+        );
+    }
+
+    /**
+     * Build a PartitionAvailability based on a map of table based filters.
+     *
+     * @param dataSourceFilterMap  A map of part tables to filters
+     *
+     * @return  The availability describing the partition
+     */
+    private static Availability buildAvailability(final Map<PhysicalTable, DataSourceFilter> dataSourceFilterMap) {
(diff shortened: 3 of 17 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -21,35 +21,38 @@
 import javax.validation.constraints.NotNull;
 
 /**
- * An implementation of Availability that unions same columns from Druid tables together, in a way that the same column
- * in 2 different availabilities contains different values. Therefore if only values in one of the availabilities is
- * requested we only need to consider that table's availability.
+ * An implementation of Availability which merges distinct record rows from multiple availabilities, using the
+ * intersection of the availability of the child availabilities, taking into consideration only children whose rows
+ * are expected to participate.
+ * The default assumption is that the schemas of all tables will be the same, and if they are not, outer join
```
cdeszaq: If this is supposed to be a paragraph break, it should have a `<p>` on a line all by itself between the paragraphs. Otherwise it renders as 1 big paragraph.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/BaseCompositeAvailability.java
```diff
@@ -0,0 +1,60 @@
...
+ * A class implementing common capabilities for availabilities backed by a collection of other availabilities.
+ */
+public abstract class BaseCompositeAvailability implements Availability {
+
+    /**
+     * Return a stream of all the dependent availabilities.
+     *
+     * @return A stream of availabilities
+     */
+    protected abstract Stream<Availability> getAllDependentAvailabilities();
```
cdeszaq: Why make children implement this if it's supposed to give back all of them? Why not have this base class _hold_ the children, and provide them when asked?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/BaseCompositeAvailability.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Override
+    public Set<TableName> getDataSourceNames() {
+        return getAllDependentAvailabilities()
```
cdeszaq: If this class takes the collection of dependent availabilities, this calculation can be done in the constructor once, rather than each time this method is called.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/BaseCompositeAvailability.java
```diff
@@ -0,0 +1,60 @@
...
+                        )
+                );
+    }
+
+    /**
+     * Retrieve all available intervals for all columns across all the underlying datasources.
+     * <p>
+     * Available intervals for the same columns are unioned into a <tt>SimplifiedIntervalList</tt>
+     *
+     * @return a map of column to all of its available intervals in union
```
cdeszaq: Shouldn't mention column

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,158 @@
...
+     * Retrieve all available intervals for all child availabilities.
+     * <p>
+     * Available intervals for the same datasource columns are unioned into a <tt>SimplifiedIntervalList</tt>
+     *
+     * @return a map of column to all of its available intervals in union
+     */
+    @Override
+    public Map<String, SimplifiedIntervalList> getAllAvailableIntervals() {
+        // get all availabilities take available interval maps from all availabilities and merge the maps together
+        return availabilityFilters.keySet().stream()
```
cdeszaq: Other than getting this from the keyset rather than a different list, this sure _looks_ like it's exactly the same implementation as the parent class. Can we dedupe?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/MetricUnionAvailability.java
```diff
@@ -95,15 +93,7 @@ public MetricUnionAvailability(@NotNull Set<PhysicalTable> physicalTables, @NotN
...
-        dataSourceNames = availabilitiesToMetricNames.keySet().stream()
-                .map(Availability::getDataSourceNames)
-                .flatMap(Set::stream)
-                .collect(
-                        Collectors.collectingAndThen(
-                                Collectors.toSet(),
-                                Collections::unmodifiableSet
-                        )
-                );
+        dataSourceNames = super.getDataSourceNames();
```
cdeszaq: In order for this to work, doesn't the `super` constructor need to be called?

fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseCompositePhysicalTable.java
```diff
@@ -47,33 +47,31 @@ public BaseCompositePhysicalTable(
...
                 .map(PhysicalTableSchema::getTimeGrain)
-                .allMatch(coarsestTimeGrain::satisfies)) {
+                .allMatch(timeGrain::satisfies)) {
             String message = String.format(
-                    "There is no mutually satisfying grain among: %s for current table %s",
-                    physicalTables,
-                    name.asName()
+                    "There is no mutually satisfying grain among: %s for composite table %s",
+                    physicalTables.stream()
+                            .map(Table::getName),
```
cdeszaq: I think this needs a `collect` to do what you're expecting, and since it's so simple, I might just put it all on 1 line.

fili-core/src/main/java/com/yahoo/bard/webservice/table/PartitionCompositeTable.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * Build a PartitionAvailability based on a map of table based filters.
+     *
+     * @param dataSourceFilterMap  A map of part tables to filters
+     *
+     * @return  The availability describing the partition
+     */
+    private static Availability buildAvailability(final Map<PhysicalTable, DataSourceFilter> dataSourceFilterMap) {
```
cdeszaq: `final` not needed

/NotABlocker

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -89,8 +78,7 @@ protected TableName getName() {
 
     @Override
     public String toString() {
-        return String.format("ConcreteAvailability for table: %s",
-                name.asName()
+        return String.format("ConcreteAvailability for table: %s", name.asName()
         );
```
cdeszaq: can unwrap

/NotABlocker

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/BaseCompositeAvailability.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * A class implementing common capabilities for availabilities backed by a collection of other availabilities.
+ */
+public abstract class BaseCompositeAvailability implements Availability {
+
+    /**
+     * Return a stream of all the availabilities which this table composites from.
```
cdeszaq: s/which this table composites from/from which this availability composites

/Pedant

<sub>There is 1 useful bit in here: switch from `table` to `availability`</sub>

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/BaseCompositeAvailability.java
```diff
@@ -0,0 +1,60 @@
...
+     *
+     * @return A stream of availabilities
+     */
+    protected abstract Stream<Availability> getAllSourceAvailabilities();
```
cdeszaq: Why make children implement this if it's supposed to give back all of them? Why not have this base class hold the children, and provide them when asked?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/BaseCompositeAvailability.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Override
+    public Set<TableName> getDataSourceNames() {
+        return getAllSourceAvailabilities()
```
cdeszaq: If this class takes the collection of dependent availabilities, this calculation can be done in the constructor once, rather than each time this method is called.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param availabilityFilters  A map of availabilities to filter functions that determine which requests they
+     * participate in
+     */
+    public PartitionAvailability(
+            @NotNull Map<Availability, DataSourceFilter> availabilityFilters
+    ) {
```
cdeszaq: can unwrap?

/NotABlocker

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,134 @@
...
+        this.availabilityFilters = availabilityFilters;
+        this.dataSourceNames = super.getDataSourceNames();
+    }
+
+    @Override
+    protected Stream<Availability> getAllSourceAvailabilities() {
+        return availabilityFilters.keySet().stream();
```
cdeszaq: Do this in the constructor?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @Override
+    public Set<TableName> getDataSourceNames() {
+        return dataSourceNames;
+    }
+
+    /**
+     * Retrieve the union of all available intervals for all source availabilities.
+     *
+     * @return a map of column to all of its available intervals in union
```
cdeszaq: mention of column in here is wrong, correct?

Also, if this is just a copy of the parent method's doc, perhaps we don't need it here? (I didn't check)

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,158 @@
...
+     *
+     * 3. Map<PartitionKey, Availability>
+     * }
+     * </pre>
+     */
+    public PartitionAvailability(
+            @NotNull Map<Availability, DataSourceFilter> availabilityFilters
+    ) {
+        this.availabilityFilters = availabilityFilters;
+        this.dataSourceNames = super.getDataSourceNames();
```
michael-mclawhorn: Technically, it does not need to because of the dependency on getAllSourceAvailabilities(), but I'm happy to make this change since performance is a real factor on these classes.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    @Override
+    protected Stream<Availability> getAllDependentAvailabilities() {
+        return availabilityFilters.keySet().stream();
+    }
+
+    @Override
+    public Set<TableName> getDataSourceNames() {
```
michael-mclawhorn: Yes.  In fact if we build the bound version, there will never be a DSC taking method on the interface.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java
michael-mclawhorn: Yes

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/BaseCompositeAvailability.java
```diff
@@ -16,30 +16,39 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param availabilityStream  A potentially ordered stream of availabilities which supply this composite view
+     */
+    public BaseCompositeAvailability(Stream<Availability> availabilityStream) {
+        sourceAvailabilities = StreamUtils.toUnmodifiableSet(availabilityStream);
+        dataSourcesNames = StreamUtils.toUnmodifiableSet(sourceAvailabilities.stream()
+                .map(Availability::getDataSourceNames).flatMap(Set::stream)
```
cdeszaq: This wrapping confused me. Can we move the `sourceAvailabilities.stream()` down a line? The way I read it initially, it looked like we were trying to assign a `Stream` to `dataSourceNames` and forgetting to `collect`.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/BaseCompositeAvailability.java
cdeszaq: Still missed it in the return description?

CHANGELOG.md
```diff
@@ -98,6 +98,9 @@ Current
     * Created `BaseCompositeAvailability` for common features
     * Refactored `DataSourceMetadataService` methods to use SimplifiedIntervaList to standardize intersections
 
+- [Queries to Druid Web Service now return a Future](https://github.com/yahoo/fili/pull/254)
+    * Queries now return a `Future<Response>` in addition to having method callbacks.
```
cdeszaq: Where is this coming from? Isn't this already in master?

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/BaseCompositePhysicalTableSpec.groovy
```diff
@@ -32,7 +32,14 @@ class BaseCompositePhysicalTableSpec extends Specification {
...
         when:
-        MetricUnionCompositeTable.verifyGrainSatisfiesAllTables(coarsestTimeGrain, [physicalTable1, physicalTable2] as Set, TableName.of("table1"))
+        new BaseCompositePhysicalTable(
+                TableName.of("test"),
+                coarsestTimeGrain,
+                Collections.emptySet(),
+                [physicalTable1, physicalTable2] as Set,
+                [:],
+                Mock(Availability)
+                ) {}
```
garyluoex: What is the curly bracket used for here?

fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseCompositePhysicalTable.java
```diff
@@ -47,33 +47,30 @@ public BaseCompositePhysicalTable(
...
                 .map(PhysicalTable::getSchema)
                 .map(PhysicalTableSchema::getTimeGrain)
-                .allMatch(coarsestTimeGrain::satisfies)) {
+                .allMatch(timeGrain::satisfies)) {
             String message = String.format(
-                    "There is no mutually satisfying grain among: %s for current table %s",
-                    physicalTables,
-                    name.asName()
+                    "There is no mutually satisfying grain among: %s for composite table %s",
+                    physicalTables.stream().map(Table::getName),
```
cdeszaq: Good catch. It likely needs to be collected into something (like a list), though if it works as-is, that would be sweet. If it's a collection, most Java collections extend from `AbstractCollection`, who's `toString` method will give a string like `[foo,bar,baz]`

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/BaseCompositeAvailability.java
```diff
@@ -0,0 +1,70 @@
...
+
+    private final Set<Availability> sourceAvailabilities;
+    private final Set<TableName> dataSourcesNames;
+
+    /**
+     * Constructor.
+     *
+     * @param availabilityStream  A potentially ordered stream of availabilities which supply this composite view
+     */
+    protected BaseCompositeAvailability(Stream<Availability> availabilityStream) {
```
cdeszaq: The general principal is to have interfaces be only as restrictive as required. If all this thing needs is, essentially, an iterator, and even then, if order doesn't really matter, a stream is a perfect fit here. Essentially, `Stream` is the better / more event-based version of `Collection`.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/BaseCompositeAvailability.java
```diff
@@ -0,0 +1,70 @@
...
+                        .map(Availability::getDataSourceNames).flatMap(Set::stream)
+        );
+    }
+
+    /**
+     * Return a stream of all the availabilities which this availability composites from.
+     *
+     * @return A stream of availabilities
+     */
+    protected Stream<Availability> getAllSourceAvailabilities() {
```
cdeszaq: There's no "store" going on here, simply passing of information. If we're expecting the caller to convert any collection we give them into a stream for processing anyways, why not give them the stream right away, rather than collecting into a collection, and then streaming out again after that?

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/BaseCompositePhysicalTableSpec.groovy
cdeszaq: It's the body of the anonymous subclass.

garyluoex:
ok

fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseCompositePhysicalTable.java
michael-mclawhorn: Yeah, this message was not good.  Proposed new implementation:
Time grain: 'MockTimeGrain' does not satisfy: source table(s) [notTable]

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/BaseCompositeAvailability.java
michael-mclawhorn: I don't especially like Streams for public APIs, but since this is a protected constructor on an abstract class I thought we could be more tightly tuned.

Issue #216, opened 2017-03-30
DruidWebService should return a Future for druid queries


kevinhinterlong:
The AsyncDruidWebServiceImpl gives a callback for when it finishes, but it can't be forced to finish with the current interface methods 
```java
void postDruidQuery(...);
void getJsonObject(...);
```
so maybe refactor it to something along the lines of 
```java
Future<Response> postDruidQuery(...);
Future<Response> getJsonObject(...);
```

cdeszaq:
Fixed.

Issue #239, opened 2017-04-13
Druid errors are not easy to differentiate from Fili errors


cdeszaq:
It would be good if the error responses could indicate that an error was from Druid (or the underlying fact store). Right now, the only way to determine it is to know that Druid error messages are JSON, and so the error message in the Fili response starts with `{` in those cases, and error messages from Fili errors generally don't start with `{`.

One way to solve this would be to capture errors from Druid and wrap them in our own exception that prefixes the Druid exception message with some sort of flag indicating it came from the underlying fact store (or just indicate Druid)

PR #254, opened 2017-04-21
Druid web service queries return a future

kevinhinterlong:
Addresses #216

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
@@ -9,7 +9,6 @@ import static com.yahoo.bard.webservice.web.handlers.workflow.DruidWorkflow.RESP
 import com.yahoo.bard.webservice.application.MetricRegistryFactory;
 import com.yahoo.bard.webservice.druid.client.DruidServiceConfig;
 import com.yahoo.bard.webservice.druid.client.DruidWebService;
-import com.yahoo.bard.webservice.druid.client.FailedFutureResponse;
 import com.yahoo.bard.webservice.druid.client.FailureCallback;
 import com.yahoo.bard.webservice.druid.client.HttpErrorCallback;
 import com.yahoo.bard.webservice.druid.client.SuccessCallback;
@@ -17,6 +16,7 @@ import com.yahoo.bard.webservice.druid.model.query.DruidQuery;
 import com.yahoo.bard.webservice.druid.model.query.WeightEvaluationQuery;
 import com.yahoo.bard.webservice.logging.RequestLog;
 import com.yahoo.bard.webservice.logging.blocks.DruidResponse;
+import com.yahoo.bard.webservice.util.FailedFuture;
 import com.yahoo.bard.webservice.web.handlers.RequestContext;
 
 import com.codahale.metrics.Meter;
@@ -181,6 +181,8 @@ public class AsyncDruidWebServiceImpl implements DruidWebService {
      * @param requestBuilder  The bound request builder for the request to be sent.
      * @param timerName  The name that distinguishes this request as part of a druid query or segment metadata request
      * @param outstanding  The counter that keeps track of the outstanding (in flight) requests for the top level query
+     *
+     * @return a future response for the query being sent
      */
     protected Future<Response> sendRequest(
             final SuccessCallback success,
@@ -268,7 +270,7 @@ public class AsyncDruidWebServiceImpl implements DruidWebService {
             }
             LOG.error("druid {} http request failed: ", serviceConfig.getNameAndUrl(), t);
             failure.invoke(t);
-            responseFuture = new FailedFutureResponse("Failed to reach druid");
+            responseFuture = new FailedFuture<>("Failed to reach druid");
         }
         return responseFuture;
     }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
@@ -41,6 +41,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Supplier;
 
@@ -180,7 +181,7 @@ public class AsyncDruidWebServiceImpl implements DruidWebService {
      * @param timerName  The name that distinguishes this request as part of a druid query or segment metadata request
      * @param outstanding  The counter that keeps track of the outstanding (in flight) requests for the top level query
      */
-    protected void sendRequest(
+    protected Future<Response> sendRequest(
             final SuccessCallback success,
             final HttpErrorCallback error,
             final FailureCallback failure,
@@ -190,9 +191,9 @@ public class AsyncDruidWebServiceImpl implements DruidWebService {
     ) {
         RequestLog.startTiming(timerName);
         final RequestLog logCtx = RequestLog.dump();
-
+        Future<Response> responseFuture = null;
         try {
-            requestBuilder.execute(
+            responseFuture = requestBuilder.execute(
                 new AsyncCompletionHandler<Response>() {
                     @Override
                     public Response onCompleted(Response response) {
@@ -267,10 +268,11 @@ public class AsyncDruidWebServiceImpl implements DruidWebService {
             LOG.error("druid {} http request failed: ", serviceConfig.getNameAndUrl(), t);
             failure.invoke(t);
         }
+        return responseFuture;
     }
 
     @Override
-    public void getJsonObject(
+    public Future<Response> getJsonObject(
             SuccessCallback success,
             HttpErrorCallback error,
             FailureCallback failure,
@@ -281,7 +283,7 @@ public class AsyncDruidWebServiceImpl implements DruidWebService {
         BoundRequestBuilder requestBuilder = webClient.prepareGet(url);
         headersToAppend.get().forEach(requestBuilder::addHeader);
 
-        sendRequest(
+        return sendRequest(
                 success,
                 error,
                 failure,
@@ -292,7 +294,7 @@ public class AsyncDruidWebServiceImpl implements DruidWebService {
     }
 
     @Override
-    public void postDruidQuery(
+    public Future<Response> postDruidQuery(
             RequestContext context,
             SuccessCallback success,
             HttpErrorCallback error,
@@ -333,7 +335,7 @@ public class AsyncDruidWebServiceImpl implements DruidWebService {
         headersToAppend.get().forEach(requestBuilder::addHeader);
 
         LOG.debug("druid json request: {}", entityBody);
-        sendRequest(
+        return sendRequest(
                 success,
                 error,
                 failure,
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
@@ -9,6 +9,7 @@ import static com.yahoo.bard.webservice.web.handlers.workflow.DruidWorkflow.RESP
 import com.yahoo.bard.webservice.application.MetricRegistryFactory;
 import com.yahoo.bard.webservice.druid.client.DruidServiceConfig;
 import com.yahoo.bard.webservice.druid.client.DruidWebService;
+import com.yahoo.bard.webservice.druid.client.FailedFutureResponse;
 import com.yahoo.bard.webservice.druid.client.FailureCallback;
 import com.yahoo.bard.webservice.druid.client.HttpErrorCallback;
 import com.yahoo.bard.webservice.druid.client.SuccessCallback;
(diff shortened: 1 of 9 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -266,11 +270,13 @@ public void onThrowable(Throwable t) {
             }
             LOG.error("druid {} http request failed: ", serviceConfig.getNameAndUrl(), t);
             failure.invoke(t);
+            responseFuture = new FailedFuture<>("Failed to reach druid");
         }
+        return responseFuture;
```
cdeszaq: Rather than capture into a temporary variable and then return, just return the `new FailedFuture` and from `requestBuilder.execute` directly. There's minimal use in the local variable.

fili-core/src/main/java/com/yahoo/bard/webservice/util/FailedFuture.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    public boolean isDone() {
+        return true;
+    }
+
+    @Override
+    public E get() throws InterruptedException, ExecutionException {
+        throw new ExecutionException(new RuntimeException(reason));
```
cdeszaq: Rather than make a new `RuntimeException`, instead take an exception in the constructor and use that here as the cause for the `ExecutionException`

fili-core/src/main/java/com/yahoo/bard/webservice/util/FailedFuture.java
```diff
@@ -0,0 +1,52 @@
...
+     * Construct a future which is already known to have failed.
+     *
+     * @param reason  The reason the future could not be completed.
+     */
+    public FailedFuture(String reason) {
+        this.reason = reason;
+    }
+
+    @Override
+    public boolean cancel(final boolean mayInterruptIfRunning) {
```
cdeszaq: no need for `final` on this param

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/DruidNavigator.java
```diff
@@ -54,14 +53,13 @@ public DruidNavigator(DruidWebService druidWebService) {
...
     public List<? extends DataSourceConfiguration> get() {
         if (tableConfigurations.isEmpty()) {
-            loadAllDatasources();
+            Future<Response> responseFuture = loadAllDatasources();
             LOG.info("Loading all datasources");
             try {
-                //TODO Druid should ideally return a future, but this accomplishes the same result.
-                countDownLatch.await(5, TimeUnit.SECONDS);
-            } catch (InterruptedException e) {
+                responseFuture.get();
```
cdeszaq: Add a comment in here that indicates we're only calling `get` so that we wait for the response to come back and be processed before continuing execution.

kevinhinterlong: 👍

fili-core/src/main/java/com/yahoo/bard/webservice/util/FailedFuture.java
kevinhinterlong: 👍

fili-core/src/main/java/com/yahoo/bard/webservice/util/FailedFuture.java
kevinhinterlong: 👍

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
kevinhinterlong: 👍

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/DruidNavigator.java
cdeszaq: Rather than 👍 comments, feel free to use the "reaction" capability on comments to show your reactions to things 😉 It generates a little less email spam.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/handlers/AsyncWebServiceRequestHandlerSpec.groovy
```diff
@@ -42,6 +42,7 @@ class AsyncWebServiceRequestHandlerSpec extends Specification {
         1 * dws.postDruidQuery(rc, _, null, null, groupByQuery) >> { a0, a1, a2, a3, a4 ->
             // Save the success callback
             sc = a1
+            return null
```
archolewa: This should be unnecessary.

cdeszaq: @archolewa this method (`postDruidQuery`) needs to return _something_, but it should be able to return just about anything, since nothing in this test is looking at the return value. Given that, `sc = a1` _should_ have some sort of return value and should be OK. _If_ that's not true, then having `return null` here is fine I think.

fili-core/src/test/java/com/yahoo/bard/webservice/models/druid/client/impl/TestDruidWebService.java
```diff
@@ -151,6 +153,8 @@ public void postDruidQuery(
         } catch (IOException e) {
             failure.invoke(e);
         }
+
+        return ConcurrentUtils.constantFuture(null);
```
cdeszaq: I think this `return` should be lifted up and happen inside the `catch` (2 lines up) wrapping the `IOException`, as well as happening on each of the branches inside that `try` block as well.

fili-core/src/test/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplWrapper.java
```diff
@@ -59,6 +62,7 @@ protected void sendRequest(
             final AtomicLong outstanding
     ) {
         this.request = requestBuilder.build();
+        return ConcurrentUtils.constantFuture(null); //no request to be sent
```
cdeszaq: Should call out what this is doing in the JavaDoc on this method since it's A-typical.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -267,10 +270,11 @@ public void onThrowable(Throwable t) {
             LOG.error("druid {} http request failed: ", serviceConfig.getNameAndUrl(), t);
             failure.invoke(t);
         }
+        return new FailedFuture<>(new RuntimeException("Failed to reach druid"));
```
cdeszaq: This should happen inside the `catch` block, and the exception it's constructed with should make sure it wraps the exception that the `catch` is catching. (It's OK to wrap that exception and give it an extra message if you wish, but the causal chain needs to be preserved.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/handlers/AsyncWebServiceRequestHandlerSpec.groovy
kevinhinterlong: So are you suggesting `return sc = a1`? It was failing with just `sc = a1`

cdeszaq: If just `sc = a1` was failing, that means that it wasn't returning whatever it needed to (possibly because assignment doesn't have a return value). How was it failing with only `sc = a1`?

kevinhinterlong: It looks like it was trying to force something to be a future but was unable to cast it.
```java
GroovyCastException: Cannot cast object 'com.yahoo.bard.webservice.web.handlers.AsyncWebServiceRequestHandler$1@33e01298' with class 'com.yahoo.bard.webservice.web.handlers.AsyncWebServiceRequestHandler$1' to class 'java.util.concurrent.Future'
```

cdeszaq: Ahh, right, so it was looking for something that it could call a `Future`. Whatever was being returned with just `sc = a1` wasn't castable, but `null` is. So, returning a `Mock(Future<Response>)` is probably the best thing there. (a bit better than `null` at least)

fili-core/src/test/java/com/yahoo/bard/webservice/models/druid/client/impl/TestDruidWebService.java
```diff
@@ -111,7 +113,7 @@ public void postDruidQuery(
         // Invoke failure callback if we have a throwable to give it
         if (throwable != null) {
             failure.invoke(throwable);
-            return;
+            return ConcurrentUtils.constantFuture(null);
```
cdeszaq: Still needs addressing

fili-core/src/test/java/com/yahoo/bard/webservice/models/druid/client/impl/TestDruidWebService.java
```diff
@@ -238,5 +246,6 @@ public void getJsonObject(
         } catch (IOException e) {
             failure.invoke(e);
         }
+        return ConcurrentUtils.constantFuture(null);
```
cdeszaq: Should get lifted up into the `catch` I think?

fili-core/src/test/java/com/yahoo/bard/webservice/models/druid/client/impl/TestDruidWebService.java
kevinhinterlong: This is actually changed, somehow github didn't catch it. [Link](https://github.com/yahoo/fili/pull/254/commits/d4054a96c200ed06ea7121ecc355b200b50114dc#diff-dcbc81d29aa4af827abee6b6fe2819e1R118)

fili-core/src/test/java/com/yahoo/bard/webservice/models/druid/client/impl/TestDruidWebService.java
```diff
@@ -86,10 +89,11 @@ public TestDruidWebService() {
     /**
      * If TestDruidWebService#throwable is set, invokes the failure callback.  Otherwise invokes success or failure
      * dependent on whether TestDruidWebService#statusCode equals 200.
+     * Note that since this doesn't send requests to druid all the responses will be null or {@link FailedFuture}.
      */
     @Override
     @SuppressWarnings("checkstyle:cyclomaticcomplexity")
-    public void postDruidQuery(
+    public Future<org.asynchttpclient.Response> postDruidQuery(
```
garyluoex: Why do we have to specify the full class name here?

cdeszaq: Good catch. We should just import that if we can

cdeszaq: Actually, it's because this class already has a `Response` object imported (ie. `javax.ws.rs.core.Response`). Because of that, one of them has to be fully qualified

garyluoex: instead of importing `javax.ws.rs.core.Response`, we can import `javax.ws.rs.core.Response.Status`?

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/DruidNavigator.java
```diff
@@ -217,5 +208,13 @@ private void queryDruid(SuccessCallback successCallback, String url) {
                 },
                 url
         );
+
+        try {
+            //calling get so we wait until responses are loaded before returning and processing continues
+            responseFuture.get();
+        } catch (InterruptedException | ExecutionException e) {
```
garyluoex: How about `TimeoutException`?

cdeszaq: What do we gain by catching that exception as well? Is it already covered by either `InterruptedException` or `ExecutionException`?

garyluoex: All three `TimeoutException`, `InterruptedException` and `ExecutionException` extends exception, I am just wondering why we are catching the latter two while not catching the first one.

cdeszaq: Good point. It probably makes sense to capture it. And it's likely part of the answer for https://github.com/yahoo/fili/issues/239

kevinhinterlong: The plain `.get()` method of `Future` doesn't throw a `TimeoutException`. Do you think I should specify a timeout and then catch it?

garyluoex: I believe `.get()` on `ListenableFuture` which is returned from `execute()` on `BoundRequestBuilder` does.

garyluoex: I see, you are right. But we should probably use the `get` with timeout version? @cdeszaq

cdeszaq: Yes, it wouldn't hurt to put a timeout there, just in case the system takes a long time for some reason. A timeout of 30 seconds should be sufficient, but could make it configurable if you really wanted.

In general, timeouts and bounded buffers are good for keeping a system resilient 😉

PR #258, opened 2017-04-26
Implement MetricUnionCompositeTableDefinition

QubitPi:


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ Current
 -------
 ### Added:
 
+- [Add implement-MetricUnionCompositeTableDefinition](https://github.com/yahoo/fili/pull/258)
```
garyluoex: add implement?

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/MetricUnionCompositeTableDefinition.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Override
+    public Set<TableName> getDependentTableNames() {
+        return Collections.emptySet();
+    }
+
+    @Override
+    public PhysicalTable build(ResourceDictionaries dictionaries, DataSourceMetadataService metadataService) {
+        return new ConcretePhysicalTable(
```
garyluoex: .....??

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/MetricUnionCompositeTableDefinition.java
```diff
@@ -0,0 +1,75 @@
...
+            Set<? extends DimensionConfig> dimensionConfigs,
+            Map<String, String> logicalToPhysicalNames
+
+    ) {
+        super(name, timeGrain, metricNames, dimensionConfigs, logicalToPhysicalNames);
+    }
+
+    @Override
+    public Set<TableName> getDependentTableNames() {
+        return Collections.emptySet();
```
garyluoex: This should not be empty since we are trying to build a Composite Table.

garyluoex:
@QubitPi  I will update the design on the Jira ticket for this metric union composite table definition, I will let you know when I am finish writing it

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/MetricUnionCompositeTableDefinition.java
```diff
@@ -0,0 +1,100 @@
...
+     * Returns set of PhysicalTables from ResourceDictionaries.
+     *
+     * @param resourceDictionaries  The ResourceDictionaries from which the tables are to be retrieved
+     *
+     * @return set of PhysicalTables from the ResourceDictionaries
+     */
+    private Set<PhysicalTable> getPhysicalTables(ResourceDictionaries resourceDictionaries) {
+        PhysicalTableDictionary physicalTableDictionary = resourceDictionaries.getPhysicalDictionary();
+        return dependentTableNames.stream()
+                .map(name -> physicalTableDictionary.getOrDefault(name, null))
```
garyluoex: getOrDefault null is same as get

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/MetricUnionCompositeTableDefinition.java
```diff
@@ -0,0 +1,100 @@
...
+                .filter(Objects::nonNull)
```
garyluoex: If there are nulls, we should probably at least log a warning.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/MetricUnionCompositeTableDefinition.java
```diff
@@ -93,7 +99,12 @@ public PhysicalTable build(ResourceDictionaries dictionaries, DataSourceMetadata
     private Set<PhysicalTable> getPhysicalTables(ResourceDictionaries resourceDictionaries) {
         PhysicalTableDictionary physicalTableDictionary = resourceDictionaries.getPhysicalDictionary();
         return dependentTableNames.stream()
-                .map(name -> physicalTableDictionary.getOrDefault(name, null))
+                .peek(name -> {
```
garyluoex: Instead of doing this to every element once, maybe just check if the resulting table collection is the same size as the dependent table names? I guess its a preference thing, but it feels like size checking is a bit simpler. Also the error message should be a bit more verbose saying that one of the dependent table is missing when building this metric union composite table.

garyluoex: Also fix the unused import thing

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/MetricUnionCompositeTableDefinition.java
```diff
@@ -0,0 +1,111 @@
...
+     * @param dimensionConfigs  The dimension configurations
+     * @param dependentTableNames  The set of dependent table names on the table
+     * @param logicalToPhysicalNames  A map from logical column names to physical column names
+     */
+    public MetricUnionCompositeTableDefinition(
+            TableName name,
+            ZonedTimeGrain timeGrain,
+            Set<FieldName> metricNames,
+            Set<? extends DimensionConfig> dimensionConfigs,
+            Set<TableName> dependentTableNames,
```
cdeszaq: Let's reorder these two params so that they match the order in the other constructor.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/MetricUnionCompositeTableDefinition.java
cdeszaq: I'd leave this check in here. It only happens at table load time, and it's better to warn about any that are found missing, rather than not I think.

cdeszaq: Also note that the log message will indicate that it's the `MetricUnionCompositeTableDefinition` class that's emitting the log, so including that in the message isn't strictly needed (though it doesn't hurt)

garyluoex: At least mention that the table not found in physical table dictionary {dependent table name} is a table that we need to build the current table {current table name}.

PR #260, opened 2017-04-27
Add test for MetricUnionCompositeTableDefinition

QubitPi:


fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/MetricUnionCompositeTableDefinition.java
```diff
@@ -98,13 +98,14 @@ public PhysicalTable build(ResourceDictionaries dictionaries, DataSourceMetadata
...
         PhysicalTableDictionary physicalTableDictionary = resourceDictionaries.getPhysicalDictionary();
         return dependentTableNames.stream()
+                .map(TableName::asName)
                 .peek(name -> {
                     if (physicalTableDictionary.get(name) == null) {
                         LOG.warn(
                                 "{} is needed to build {}, but it's not found in ResourceDictionaries in" +
                                         "MetricUnionCompositeTableDefinition",
                                 name,
-                                getName()
```
michael-mclawhorn: Why is this a warn and not an error?

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/MetricUnionCompositeTableDefinitionSpec.groovy
```diff
@@ -0,0 +1,40 @@
...
+        PhysicalTable physicalTable = Mock(PhysicalTable)
+        ResourceDictionaries resourceDictionaries = Mock(ResourceDictionaries)
+        resourceDictionaries.getPhysicalDictionary() >> [(existing.asName()) : physicalTable]
+
+        when:
+        Set<PhysicalTable> physicalTables = metricUnionCompositeTableDefinition.getPhysicalTables(resourceDictionaries)
+
+        then:
+        physicalTables.size() == 1
+        physicalTables.contains(physicalTable)
```
cdeszaq: Can simplify this to an `expect` test that just compares the call to `getPhysicalTables` to `[physicalTable] as Set`. But this is not critical

QubitPi: @cdeszaq Beautiful!

PR #262, opened 2017-04-28
Constrained availability

michael-mclawhorn:
Create a ConstrainedTable that applies DataSourceConstraints to a Table and caches and closes over that constraint.

Switch to using the ConstrainedTable in the post-query space.

Remove getAvailability from the general TableSpace and make a ConfigPhysicalTable subinterface to expose availability.

This feature is necessary to allow `DataSourceConstraint` filtered datasources on query serialization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ Current
 -------
 ### Added:
 
+- [Constrained Table Support for Table Serialization](https://github.com/yahoo/fili/pull/262/files)
+    * Add ConstrainedTable which closes over a physical table and an availability, caching all availability merges.
+    * Add PartialDataHandler method to use `ConstrainedTable`
+    * Add SimplifiedIntervalLit.empty() to produce empty SILs
+
 - [Prepare For Partial Data V2](https://github.com/yahoo/fili/pull/264)
     * Add new query context for druid's uncovered interval feature
     * Add a configurable property named "druid_uncovered_interval_limit"
@@ -105,10 +110,18 @@ Current
 
 ### Changed:
 
+- [Constrained Table Support for Table Serialization](https://github.com/yahoo/fili/pull/262/files)
+    * Switched `PartialDataRequestHandler` to use the table from the query rather than the `PhysicalTableDictionary`
+    * `DruidQueryBuilder` uses constrained tables to dynamically pick between Union and Table DataSource implementations.
+    * `PartialDataHandler` has multiple different entrypoints now depending on pre or post constraint conditions.
+    * `getAvailability` moved to a `ConfigTable` interface and all configured Tables to that interface
+    * DataSource implementations bind to `ConstrainedTable` and only ConstrainedTable is used after table selection
+    * `PhysicalTable.getAllAvailableIntervals` explicitly rather than implicitly uses `SimplifiedIntervalList`
+    * Bound and default versions of getAvailableIntervals and getAllAvailableIntervals added to PhysicalTable interface    
+    
 - [Druid filters are now lazy.](https://github.com/yahoo/fili/pull/269)
     - The Druid filter is built when requested, NOT at DatApiRequest construction. This will
         make it easier to write performant `DataApiRequest` mappers.
-
 
 - [Reduce log level of failure to store a result in the asynchronous job store](https://github.com/yahoo/fili/pull/266)
     * Customers who aren't using the asynchronous infrastructure shouldn't be seeing spurious warnings about a failure
@@ -273,6 +286,9 @@ Current
     * Converted to 404 when error was cause by not finding a path element
 
 ### Deprecated:
+
+- [Constrained Table Support for Table Serialization](https://github.com/yahoo/fili/pull/262/files)
+    * Deprecated static empty instance of SimplifiedIntervalList.NO_INTERVALS
 
 - [Support for Lucene 5 indexes](https://github.com/yahoo/fili/pull/265)
     * Added lucene-backward-codecs.jar as a dependency to restore support for indexes built on earlier instances.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Current
     * Add PartialDataHandler method to use `ConstrainedTable`
     * Add SimplifiedIntervalLit.empty() to produce empty SILs
 
+- [Moved UnionDataSource to support only single tables](https://github.com/yahoo/fili/pull/262/files)
+    * `DataSource` now supports getDataSource() operation
+
 - [Prepare For Partial Data V2](https://github.com/yahoo/fili/pull/264)
     * Add new query context for druid's uncovered interval feature
     * Add a configurable property named "druid_uncovered_interval_limit"
@@ -120,6 +123,11 @@ Current
     * Bound and default versions of getAvailableIntervals and getAllAvailableIntervals added to PhysicalTable interface
     * Package-private optimize tests in `DruidQueryBuilder` moved to protected
     * Immutable `NoVolatileIntervalsFunction` class made final
+    
+- [Moved UnionDataSource to support only single tables](https://github.com/yahoo/fili/pull/262/files)
+    * `UnionDataSource` now accepts only single tables instead of sets of tables.
+    * `DataSource` now supports getDataSource() operation
+    * `IntervalUtils.collectBucketedIntervalsNotInIntervalList` moved to `PartialDataHandler`
     
     
 - [Druid filters are now lazy.](https://github.com/yahoo/fili/pull/269)
@@ -290,6 +298,9 @@ Current
 
 ### Deprecated:
 
+- [Moved UnionDataSource to support only single tables](https://github.com/yahoo/fili/pull/262/files)
+    * `DataSource` deprecates getDataSources()
+
 - [Constrained Table Support for Table Serialization](https://github.com/yahoo/fili/pull/262/files)
     * Deprecated static empty instance of SimplifiedIntervalList.NO_INTERVALS
 
@@ -353,6 +364,9 @@ Current
 - [Refactor Physical Table Definition and Update Table Loader](https://github.com/yahoo/fili/pull/207)
     * Removed deprecated `PhysicalTableDefinition` constructor that takes an `ZonlessTimeGrain`, use `ZonedTimeGrain` instead
     * Removed `buildPhysicalTable` in `BaseTableLoader`, building table logic is pushed into `PhysicalTableDefinition`
+
+- [Moved UnionDataSource to support only single tables](https://github.com/yahoo/fili/pull/262/files)
+    * `DataSource` no longer accepts Set<Table> constructor
 
 - [CompositePhsyicalTable Core Components Refactor](https://github.com/yahoo/fili/pull/179)
     * Removed deprecated method `findMissingRequestTimeGrainIntervals` from `PartialDataHandler`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Current
     * Add PartialDataHandler method to use `ConstrainedTable`
     * Add SimplifiedIntervalLit.empty() to produce empty SILs
 
+- [Testing: ClassScannerSpec now supports 'discoverable' depenencies ](https://github.com/yahoo/fili/pull/262/files)
+    * Creating 'supplyDependencies' method on a class's spec allows definitions of dependencies for dynamic equality testing
+
 - [Moved UnionDataSource to support only single tables](https://github.com/yahoo/fili/pull/262/files)
     * `DataSource` now supports getDataSource() operation
 
@@ -298,11 +301,12 @@ Current
 
 ### Deprecated:
 
-- [Moved UnionDataSource to support only single tables](https://github.com/yahoo/fili/pull/262/files)
-    * `DataSource` deprecates getDataSources()
-
 - [Constrained Table Support for Table Serialization](https://github.com/yahoo/fili/pull/262/files)
     * Deprecated static empty instance of SimplifiedIntervalList.NO_INTERVALS
+    * PartialDataRequestHandler constructor using `PhysicalTableDictionary`
+
+- [Moved UnionDataSource to support only single tables](https://github.com/yahoo/fili/pull/262/files)
+    * `DataSource` deprecates getDataSources()
 
 - [Support for Lucene 5 indexes](https://github.com/yahoo/fili/pull/265)
     * Added lucene-backward-codecs.jar as a dependency to restore support for indexes built on earlier instances.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@ Current
 - [Constrained Table Support for Table Serialization](https://github.com/yahoo/fili/pull/262/files)
     * Add ConstrainedTable which closes over a physical table and an availability, caching all availability merges.
     * Add PartialDataHandler method to use `ConstrainedTable`
-    * Add SimplifiedIntervalLit.empty() to produce empty SILs
 
 - [Testing: ClassScannerSpec now supports 'discoverable' depenencies ](https://github.com/yahoo/fili/pull/262/files)
     * Creating 'supplyDependencies' method on a class's spec allows definitions of dependencies for dynamic equality testing
@@ -307,6 +306,9 @@ Current
 
 - [Moved UnionDataSource to support only single tables](https://github.com/yahoo/fili/pull/262/files)
     * `DataSource` deprecates getDataSources()
+
+- [SimplifiedIntervalList::NO_INTERVALS](https://github.com/yahoo/fili/pull/262/files)
+    * This is unsafe since it is modifiable
 
 - [Support for Lucene 5 indexes](https://github.com/yahoo/fili/pull/265)
     * Added lucene-backward-codecs.jar as a dependency to restore support for indexes built on earlier instances.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -112,12 +112,15 @@ Current
 
 - [Constrained Table Support for Table Serialization](https://github.com/yahoo/fili/pull/262/files)
     * Switched `PartialDataRequestHandler` to use the table from the query rather than the `PhysicalTableDictionary`
-    * `DruidQueryBuilder` uses constrained tables to dynamically pick between Union and Table DataSource implementations.
-    * `PartialDataHandler` has multiple different entrypoints now depending on pre or post constraint conditions.
+    * `DruidQueryBuilder` uses constrained tables to dynamically pick between Union and Table DataSource implementations
+    * `PartialDataHandler` has multiple different entrypoints now depending on pre or post constraint conditions
     * `getAvailability` moved to a `ConfigTable` interface and all configured Tables to that interface
     * DataSource implementations bind to `ConstrainedTable` and only ConstrainedTable is used after table selection
     * `PhysicalTable.getAllAvailableIntervals` explicitly rather than implicitly uses `SimplifiedIntervalList`
-    * Bound and default versions of getAvailableIntervals and getAllAvailableIntervals added to PhysicalTable interface    
+    * Bound and default versions of getAvailableIntervals and getAllAvailableIntervals added to PhysicalTable interface
+    * Package-private optimize tests in `DruidQueryBuilder` moved to protected
+    * Immutable `NoVolatileIntervalsFunction` class made final
+    
     
 - [Druid filters are now lazy.](https://github.com/yahoo/fili/pull/269)
     - The Druid filter is built when requested, NOT at DatApiRequest construction. This will
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidQueryBuilder.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidQueryBuilder.java
@@ -311,9 +311,9 @@ public class DruidQueryBuilder {
         );
 
         // Override the grain with what's set in the template if it has one set
-        if (template.getTimeGrain() != null) {
-            granularity = template.getTimeGrain().buildZonedTimeGrain(timeZone);
-        }
+        Granularity mergeGrain = (template.getTimeGrain() != null) ?
+                template.getTimeGrain().buildZonedTimeGrain(timeZone) :
+                granularity;
 
         LOG.trace("Building a single pass druid topN query");
 
@@ -321,7 +321,7 @@ public class DruidQueryBuilder {
         return new TopNQuery(
                 buildTableDataSource(table),
                 // The check that the set of dimensions has exactly one element is currently done above
-                granularity,
+                mergeGrain,
                 groupByDimension.iterator().next(),
                 filter,
                 template.getAggregations(),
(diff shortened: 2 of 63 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/PartialDataHandler.java
```diff
@@ -27,6 +28,54 @@
...
+     */
+    public SimplifiedIntervalList findMissingTimeGrainIntervals(
+            DataSourceConstraint constraint,
+            Set<PhysicalTable> physicalTables,
+            @NotNull SimplifiedIntervalList requestedIntervals,
+            Granularity granularity
+    ) {
+        SimplifiedIntervalList availableIntervals = physicalTables.stream()
+                .map(table -> table.withConstraint(constraint))
+                .map(table -> table.getAvailableIntervals())
```
QubitPi: Could this be changed to `PhysicalTable::getAvailableIntervals` ?

fili-core/src/main/java/com/yahoo/bard/webservice/data/PartialDataHandler.java
```diff
@@ -27,6 +28,54 @@
...
+     * @return subintervals of the requested intervals with incomplete data
+     */
+    public SimplifiedIntervalList findMissingTimeGrainIntervals(
+            Set<ConstrainedTable> physicalTables,
+            @NotNull SimplifiedIntervalList requestedIntervals,
+            Granularity granularity
+    ) {
+
+        SimplifiedIntervalList availableIntervals = physicalTables.stream()
+                .map(table -> table.getAvailableIntervals())
```
QubitPi: Could this be changed to `PhysicalTable::getAvailableIntervals` ?

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/datasource/TableDataSource.java
```diff
@@ -21,11 +22,28 @@
...
+     * @param physicalTable  The physical table of the data source
+     */
+    public TableDataSource(ConstrainedTable physicalTable) {
+        this(
+                physicalTable,
+                physicalTable.getDataSourceNames()
+                        .stream()
+                        .findFirst()
+                        .orElseThrow(() -> new IllegalArgumentException(
+                                "Non singleton DataSource table passed to TableDataSource constructor."))
```
QubitPi: Closing parenthesis needs to go to the next line

fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
```diff
@@ -139,4 +116,49 @@ public String getPhysicalColumnName(String logicalName) {
...
+     *
+     * @throws IllegalArgumentException If there are columns referenced by the constraint unavailable in the table
+     */
+    /**
+     * Ensure that the schema of the constraint is consistent with what the table supports.
+     *
+     * @param constraint  The constraint being tested
+     *
+     * @throws IllegalArgumentException If there are columns referenced by the constraint unavailable in the table
+     */
```
QubitPi: Same JavaDoc appeared twice

fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
```diff
@@ -139,4 +116,49 @@ public String getPhysicalColumnName(String logicalName) {
...
+    private void validateConstraintSchema(DataSourceConstraint constraint) throws IllegalArgumentException {
+        Set<String> tableColumnNames = getSchema().getColumnNames();
+        // Validate that the requested columns are answerable by the current table
+        if (!constraint.getAllColumnNames().stream().allMatch(tableColumnNames::contains)) {
+            String message = String.format(
+                    "Received invalid request requesting for columns: %s that is not available in this table: %s",
+                    constraint.getAllColumnNames().stream()
+                            .filter(name -> !tableColumnNames.contains(name))
```
QubitPi: Could this be changed to `!tableColumnNames::contains`?

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/DataSourceConstraint.java
```diff
@@ -160,4 +161,26 @@ public DataSourceConstraint withMetricIntersection(Set<String> metricNames) {
                 apiFilters
         );
     }
+
+    /**
+     * Build a constraint which should filter away no part of a given table.
+     *
+     * @param table  The table whose dimensions and metrics are to be queried
+     *
+     * @return  A constraint which should provide no restrictions
```
QubitPi: one space, lowercase A

fili-core/src/main/java/com/yahoo/bard/webservice/util/SimplifiedIntervalList.java
```diff
@@ -103,6 +104,14 @@ public static SimplifiedIntervalList simplifyIntervals(Collection<Interval>... i
                 .collect(getCollector());
     }
 
+    /**
+     * Static method to provide an empty list.
+     *
+     * @return  A new, empty list.
```
QubitPi: one space, lowercase A

fili-core/src/main/java/com/yahoo/bard/webservice/table/BasePhysicalTable.java
cdeszaq: I don't think so. Method references cannot be used in expressions like that (as far as I know)

fili-core/src/main/java/com/yahoo/bard/webservice/table/ConstrainedTable.java
```diff
@@ -0,0 +1,119 @@
...
+        if (constraint == this.constraint) {
+            return dataSourceNames;
+        }
+        return sourceTable.getDataSourceNames(constraint);
+    }
+
+    @Deprecated
+    @Override
+    public ConstrainedTable withConstraint(final DataSourceConstraint constraint) {
+        // WARN HERE
```
cdeszaq: Why this warning? Perhaps it makes sense here? If not, then a JavaDoc is probably a good addition to explain how layering warnings would work...

fili-core/src/main/java/com/yahoo/bard/webservice/table/ConstrainedTable.java
```diff
@@ -0,0 +1,119 @@
...
+    public TableName getTableName() {
+        return sourceTable.getTableName();
+    }
+
+    @Override
+    public DateTime getTableAlignment() {
+        return sourceTable.getTableAlignment();
+    }
+
+    @Deprecated
```
cdeszaq: Also, these methods should _definitely_ call out that the table they _do_ return is constrained on the original table, not this post-constrained table.

fili-core/src/main/java/com/yahoo/bard/webservice/table/ConstrainedTable.java
```diff
@@ -0,0 +1,119 @@
...
+    @Override
+    public SimplifiedIntervalList getAvailableIntervals(DataSourceConstraint constraint) {
+        // WARN HERE
+        if (constraint == this.constraint) {
```
cdeszaq: Why not use `equals(...)`? Isn't instance-equality just a perhaps faster but lower-quality way of answering the same thing?

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -102,4 +126,35 @@
...
+
+    /**
+     * Filter a map of raw intervals to those included in a physical schema.
+     *
+     * @param rawIntervals  The map of name to {@link SimplifiedIntervalList}s as the source availability
+     * @param schema  The schema describing the columns of this table
+     *
+     * @return map of column to set of available intervals
+     */
+    static Map<Column, SimplifiedIntervalList> getAllAvailableIntervals(
```
cdeszaq: Why `static` and not a default method?

cdeszaq: Also, since this method just takes a `Map<String, SIL>` and `PhysicalTableSchema`, it doesn't seem to actually care about availability at all (contrary to what the method is named). Perhaps a more utilitarian name, like `filterColumnIntervalsBySchema` would be better?

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -4,53 +4,80 @@
...
+     * Getter for all the available intervals for the corresponding columns configured on the table.
+     *
+     * @return map of column to set of available intervals
+     */
+    Map<Column, SimplifiedIntervalList> getAllAvailableIntervals();
 
-    @Override
-    PhysicalTableSchema getSchema();
+    /**
+     * Return a view of the available intervals for this table given a data source constraint.
```
cdeszaq: There is no data source constraint? Copy-paste error?

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -4,53 +4,80 @@
...
+     *
+     * @return The intervals that the table can report on, given the constraint
```
cdeszaq: No constraint...

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -4,53 +4,80 @@
...
     /**
-     * Getter for all the available intervals for the corresponding columns configured on the table.
+     * Return the {@link TableName} of the dataSources which back this table given a constraint.
      *
-     * @return map of column to set of available intervals
+     * @param constraint  A constraint which may narrow the data sources participating.
+     *
+     * @return A set of tablenames for backing dataSources
      */
-    Map<Column, List<Interval>> getAllAvailableIntervals();
```
cdeszaq: Why are we able to get rid of this method signature?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -51,6 +51,16 @@ public ConcreteAvailability(
     }
 
     @Override
+    public SimplifiedIntervalList getAvailableIntervals() {
+        return null;
```
cdeszaq: This doesn't look right... Isn't this just asking for NPE elsewhere at a random place?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -51,6 +51,16 @@ public ConcreteAvailability(
...
+    }
+
+    @Override
+    public Set<TableName> getDataSourceNames(PhysicalDataSourceConstraint constraint) {
+        return dataSourceNames;
```
cdeszaq: I don't think this override is needed, since it looks like it does what the default would be.

fili-core/src/main/java/com/yahoo/bard/webservice/util/SimplifiedIntervalList.java
```diff
@@ -30,21 +30,22 @@
  */
 public class SimplifiedIntervalList extends LinkedList<Interval> {
 
+    @Deprecated
```
cdeszaq: Javadoc explaining why / what to use instead is needed.

fili-core/src/main/java/com/yahoo/bard/webservice/util/SimplifiedIntervalList.java
```diff
@@ -103,6 +104,14 @@ public static SimplifiedIntervalList simplifyIntervals(Collection<Interval>... i
                 .collect(getCollector());
     }
 
+    /**
+     * Static method to provide an empty list.
+     *
+     * @return a new, empty list.
+     */
+    public static SimplifiedIntervalList empty() {
```
cdeszaq: If we want this to be used, we should have the constructor be private. Otherwise, we should just remove this method and make everything use the constructor. No need for both, is there?

fili-core/src/main/java/com/yahoo/bard/webservice/web/ErrorMessageFormat.java
```diff
@@ -216,6 +216,11 @@
     RESULT_MAPPING_FAILURE(
             "Error occurred while processing response data: %s"
     ),
+    INVALID_DATASOURCE_UNION(
+            "Union data source had conflicting name mappings for logical dimension '%s' with mappings of '%s' and '%s'",
+            "Union Data Source had conflicting name mappings for logical dimension '%s' with mappings of '%s' and '%s'"
```
cdeszaq: Do we need both? If the logging and exception message are exactly the same, why repeat it?

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/PartialDataRequestHandler.java
```diff
@@ -35,23 +30,19 @@
...
      * Wrap the response processor in a partial data check.
      *
      * @param next The next request handler to invoke
-     * @param physicalTableDictionary  the repository of slice data
-     * @param partialDataHandler the PartialDataHandler to use
+     * @param partialDataHandler the service to calculate partial data from table availabilities
      */
     public PartialDataRequestHandler(
             DataRequestHandler next,
-            PhysicalTableDictionary physicalTableDictionary,
```
cdeszaq: Need to deprecate constructor, not just change it. This will definitely break people w/o letting them migrate smoothly

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/aggregation/CardinalityAggregationSpec.groovy
```diff
@@ -28,11 +29,19 @@ class CardinalityAggregationSpec extends Specification {
...
     CardinalityAggregation a1
 
+    ConstrainedTable constrainedTable
+
     def setupSpec() {
         MAPPER.readTree(expectedJson)
     }
 
     def setup() {
+        constrainedTable = TableTestUtils.buildTable("table",
```
cdeszaq: `"table"` should be bumped down a line

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/handlers/PartialDataRequestHandlerSpec.groovy
```diff
@@ -55,8 +56,9 @@ class PartialDataRequestHandlerSpec extends Specification {
         groupByQuery.getDependentFieldNames() >> Collections.emptySet()
         groupByQuery.getInnermostQuery() >> groupByQuery
         groupByQuery.getDataSource() >> dataSource
-        physicalTableDictionary.get(_) >> physicalTables[0]
-        dataSource.getNames() >> ["name"]
+        //physicalTableDictionary.get(_) >> physicalTables[0]
+        //dataSource.getNames() >> ["name"]
```
cdeszaq: Should these lines be deleted, not just commented out?

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ Current
 -------
 ### Added:
 
+- [Constrained Table Support for Table Serialization](https://github.com/yahoo/fili/pull/262/files)
+    * Add ConstrainedTable which closes over a physical table and an availability, caching all availability merges.
+    * Add PartialDataHandler method to use `ConstrainedTable`
+    * Add SimplifiedIntervalLit.empty() to produce empty SILs
```
michael-mclawhorn: It deprecates but does not yet remove No Intervals, so yes, but not completely.

fili-core/src/main/java/com/yahoo/bard/webservice/data/PartialDataHandler.java
```diff
@@ -27,6 +29,52 @@
...
+    /**
+     * Find the holes in the passed in intervals at a given granularity.
+     *
+     * @param constrainedTableStream  the stream of tables whose column availabilities are checked
+     * @param requestedIntervals  The intervals that may not be fully satisfied
+     * @param granularity  The granularity at which to find missing intervals
+     *
+     * @return subintervals of the requested intervals with incomplete data
+     */
+    public SimplifiedIntervalList findMissingTimeGrainIntervals(
```
michael-mclawhorn: It's used in PartialDataRequestHandler.
There are two entry points to this partial data code, before table selection (and constraint) and after.
This is the entry point for 'after'

fili-core/src/main/java/com/yahoo/bard/webservice/data/PartialDataHandler.java
```diff
@@ -27,6 +29,52 @@
...
+            Stream<ConstrainedTable> constrainedTableStream,
+            @NotNull SimplifiedIntervalList requestedIntervals,
+            Granularity granularity
+    ) {
+
+        SimplifiedIntervalList availableIntervals = constrainedTableStream
+                .map(PhysicalTable::getAvailableIntervals)
+                .reduce(SimplifiedIntervalList::intersect)
```
michael-mclawhorn: Hm.  This is an issue.  I have to come back to it.

fili-core/src/main/java/com/yahoo/bard/webservice/data/PartialDataHandler.java
```diff
@@ -39,24 +87,17 @@
...
      *
-     * @param constraint  Constraint containing all the column names the request depends on
-     * @param physicalTables  the tables whose column availabilities are checked
+     * @param availableIntervals  the intervals available in the tables
      * @param requestedIntervals  The intervals that may not be fully satisfied
      * @param granularity  The granularity at which to find missing intervals
      *
      * @return subintervals of the requested intervals with incomplete data
      */
     public SimplifiedIntervalList findMissingTimeGrainIntervals(
```
michael-mclawhorn: it's the most general part of the contract.  I'm inclined to leave it visible, and we certainly can't make it less visible without a deprecation period.

cdeszaq: My point was more along the lines of "this method didn't exist before, since this signature is changing, so do we need to expose it". So, from that standpoint, this method here doesn't need to be deprecated.

But if we're OK supporting this contract going forward, and we think there's a high likelihood of it being useful outside this class (by someone who would already depend on this class), then it being `public` is fine by me. Otherwise, I'd lean towards hiding it until it's needed.

fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidQueryBuilder.java
```diff
@@ -202,10 +202,12 @@ protected GroupByQuery buildGroupByQuery(
                 intervals
         );
 
+        Filter mergedFilter = filter;
+
         // Override the grain with what's set in the template if it has one set
-        if (template.getTimeGrain() != null) {
-            granularity = template.getTimeGrain().buildZonedTimeGrain(timeZone);
-        }
+        Granularity mergedGranularity = template.getTimeGrain() != null ?
```
garyluoex: I think I accidentally deleted one of your comment here @cdeszaq , ops sorry. Here is the original comment by rick.

`?` should go down on the next line when wrapping, as the `:` is down on the next line.

<sub>Applies to this same fix lower in the file as well</sub>

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
```diff
@@ -4,53 +4,80 @@
...
+    }
 
     /**
-     * Get the name of the current table.
+     * Return a view of the available intervals for this table given a constraint.
      *
-     * @return name of the table as TableName
+     * @param constraint  The constraint which limits available intervals
+     *
+     * @return The intervals that the table can report on
```
garyluoex: seems like the comment is flipped with the above no argument version of this method.

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/DataSourceConstraint.java
```diff
@@ -160,4 +161,26 @@ public DataSourceConstraint withMetricIntersection(Set<String> metricNames) {
...
     }
+
+    /**
+     * Build a constraint which should filter away no part of a given table.
+     *
+     * @param table  The table whose dimensions and metrics are to be queried
+     *
+     * @return a constraint which should provide no restrictions
+     */
+    public static DataSourceConstraint emptyConstraint(PhysicalTable table) {
```
garyluoex: A little bit confusing, since it constraints to all table instead of no constraint at all.

fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidQueryBuilder.java
```diff
@@ -252,9 +254,9 @@ protected GroupByQuery buildGroupByQuery(
...
      * @return A table datasource for a fact table or a union data source for a fact table view
      */
-    private DataSource buildTableDataSource(PhysicalTable table) {
-        if (table instanceof ConcretePhysicalTable) {
-            return new TableDataSource((ConcretePhysicalTable) table);
+    private DataSource buildTableDataSource(ConstrainedTable table) {
+        if (table.getDataSourceNames().size() == 1) {
+            return new TableDataSource(table, table.getDataSourceNames().stream().findFirst().get());
         } else {
             return new UnionDataSource(Sets.newHashSet(table));
```
cdeszaq: We probably don't want to make everything Union, on the off-chance that Druid has an optimization around the single-table version vs. the union version. That said, I can definitely see this optimization being moved out of the main query builder and into a static analysis step at some point.

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/DataSourceConstraint.java
cdeszaq: "Unconstrained" would possibly be better?

fili-core/src/main/java/com/yahoo/bard/webservice/data/PartialDataHandler.java
```diff
@@ -28,36 +27,25 @@
...
      * </pre>
      * <p>
-     * This method computes the subintervals of the specified intervals for which partial data is
-     * present for a given combination of request metrics and dimensions (pulled from the API request and generated
-     * druid query) at the specified granularity.
+     * Any requested subinterval bucket that is partially unavailable is collected and returned in a simplified form.
      *
-     * @param constraint  Constraint containing all the column names the request depends on
-     * @param physicalTables  the tables whose column availabilities are checked
+     * @param availableIntervals  the intervals available in the tables
```
QubitPi: Capitalize "the"

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
michael-mclawhorn: we didn't, it moved higher in the file

fili-core/src/main/java/com/yahoo/bard/webservice/table/PhysicalTable.java
michael-mclawhorn: The 'Map<String, SIL>' is the interface of the Availability getAllAvailableIntervals() while 'Map<Colulm,SIL>' is the interface of Table.getAllAvailableIntervals().
This method, properly translates from one to the other and is used in construction so it's problematic for it to be non static.

michael-mclawhorn: alternate name: 'mapToSchemaAvailability'

Issue #246, opened 2017-04-19
DruidMetadataLoader's No Content log isn't very helpful


cdeszaq:
When the DruidMetadataLoader gets a 204 NO CONTENT response, it logs that fact (probably as a warning) since it's not expected that a Druid data source will be known but have no metadata. Unfortunately, the log message only indicates that we got a 204 response, and does not make any effort to try to explain why this is a log-worthy problem, what the cause might be, why it matters, or how to start to investigate what might be wrong / fix it.

The log message should try to be more helpful.

PR #263, opened 2017-05-01
Fix data sourec name disparity

cdeszaq:
Fixes a confusion where `TableName` was serving multi-duty as both a config concept, as well as an additional name for a `PhysicalTable` and the name referring to a Data Source in Druid. This also should allow having a `PhysicalTable` with an `Availability` that uses a different `DataSourceName` (ie. more than 1 physical table pointing at the same data source)

This PR also includes a fair amount of cleanup, both to Java code as well as the Test code. (Below is the compile commit messages for the commits included in this PR)

As part of some of this, a new `BaseMetadataAvailability` parent class to mirror the `BaseCompositeAvailability` and simplify the inheritance.

This PR also includes a fix for #246, making the log message more useful and helpful. The actual changes are in the Java cleanup I believe.

The commits are generally broken up fairly well, so it may be useful to review them 1 commit at a time. There is only a very small amount of overlap between the commits.

Commit messages:

```
Cleanup:
- Remove superfluous type indicator
- Add / Clean up / Clarify / Correct JavaDoc
- Use `Collectors.collectingAndThen` to convert to an unmodifiable collection instead of wrapping
- Unpack a lambda to get a stream into a map and then a flatMap
- Rearrange member variables to keep private static finals on top
- Make `BasePhysicalTable` take a more extension-friendly set of `PhysicalTable`s
- Clean up whitespace and variable specification in lambdas
- Remove unneeded blank lines
- Correct / Clean up line wrapping
- Small import cleanup
- Small inlining where needed
- Tidy up method ordering


Make Groovy code groovier:
- Use property access instead of getters in lots of places
- Use more list and map literals
- Use spread collections in many places instead of collects
- Compact code using ternary instead of simple conditionals and statements where readability is improved
- Removal of superfluous semicolons

Clean up tests a bit:
- Added `@Unroll` where missing
- Switch to `expect` instead of `when/then` where appropriate

Clean up `BaseDataSourceComponentSpec`
- Drop a log from `error` to `trace`
- Make JSON validation helpers return `boolean`
- Comment methods, etc.


Refactor `SegmentIntervalsHashIdGenerator`
- Make mapper map name clearer
- Tighten main processing stream and helper. Should be a little faster


-Update name of `TableName` comparator


-Force `ConcretePhysicalTable` only take a `ConcreteAvailability`


Split `DataSourceName` concept from `TableName`
- Add `DataSourceName` interface
- Make relevant places use `DataSourceName`
	- `DataSource` & children
	- `DataSourceMetadataService` & `DataSourceMetadataLoader`
	- `SegmentIntervalsHashIdGenerator`
	- `PhysicalTable` & children
	- `Availability` & children
	- `ErrorMessageFormat`
	- `SlicesApiRequest`

Add a `BaseMetadataAvailability` as a parallel to `BaseCompositeAvailability`
- `ConcreteAvailability` extends this
- `PermissiveAvailability` extends this


Remove `PhysicalTableDictionary` dependency from `SegmentIntervalsHashIdGenerator`


Make `MetricUnionAvailability` take a set of `Availability` instead of `PhysicalTable`


Remove `PhysicalTable::getTableName` to use `getName` instead
```

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/names/DataSourceName.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/names/DataSourceName.java
@@ -1,26 +1,18 @@
-// Copyright 2016 Yahoo Inc.
+// Copyright 2017 Yahoo Inc.
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.data.config.names;
 
 import java.util.Comparator;
 
 /**
- * Marker interface for objects that can be treated as a table name.
+ * Marker interface for objects that can be treated as a data source name in druid.
  */
-public interface TableName {
+public interface DataSourceName {
 
     /**
-     * Comparator to order TableNames by their asName methods, using the native String comparator.
+     * Comparator to order DataSourceNames by their asName methods, using the native String comparator.
      */
-    Comparator<TableName> AS_NAME_COMPARATOR = Comparator.comparing(TableName::asName);
-
-    /**
-     * Comparator to order TableNames by their asName methods, using the native String comparator.
-     *
-     * @deprecated due to name change. Use AS_NAME_COMPARATOR instead.
-     */
-    @Deprecated
-    Comparator<TableName> COMPARATOR = Comparator.comparing(TableName::asName);
+    Comparator<DataSourceName> AS_NAME_COMPARATOR = Comparator.comparing(DataSourceName::asName);
 
     /**
      * Return a string representation of a table name.
@@ -33,12 +25,12 @@ public interface TableName {
      * Wrap a string in an anonymous instance of TableName.
      * Rather than make heavy use of this, instead make a class.
      *
-     * @param name the name being wrapped
+     * @param name   The name being wrapped
      *
      * @return an anonymous subclass instance of TableName
      */
-    static TableName of(String name) {
-        return new TableName() {
+    static DataSourceName of(String name) {
+        return new DataSourceName() {
             @Override
             public String asName() {
                 return name;
@@ -51,8 +43,8 @@ public interface TableName {
 
             @Override
             public boolean equals(Object o) {
-                if (o != null && o instanceof TableName) {
-                    return name.equals(((TableName) o).asName());
+                if (o != null && o instanceof DataSourceName) {
+                    return name.equals(((DataSourceName) o).asName());
                 }
                 return false;
             }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/metadata/DataSourceMetadataLoader.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/metadata/DataSourceMetadataLoader.java
@@ -32,13 +32,14 @@ import javax.inject.Singleton;
 /**
  * Datasource metadata loader sends requests to the druid datasource metadata endpoint ('datasources') and returns the
  * lists of available data segments for each datasource. It then builds Datasource Metadata objects which pivot this
- * data into columns of intervals and then updates the Physical Table instances in the physical table dictionary.
+ * data into columns of intervals and then updates the {@link DataSourceMetadataService}.
  * <p>
  * Note that this uses the segmentMetadata query that touches the coordinator.
  */
 @Singleton
 public class DataSourceMetadataLoader extends Loader<Boolean> {
 
+    private static final Logger LOG = LoggerFactory.getLogger(DataSourceMetadataLoader.class);
     private static final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance();
 
     public static final String DATASOURCE_METADATA_QUERY_FORMAT = "/datasources/%s?full";
@@ -55,8 +56,6 @@ public class DataSourceMetadataLoader extends Loader<Boolean> {
     public static final String DRUID_SEG_LOADER_TIMER_DELAY_KEY =
             SYSTEM_CONFIG.getPackageVariableName("druid_seg_loader_timer_delay");
 
-    private static final Logger LOG = LoggerFactory.getLogger(DataSourceMetadataLoader.class);
-
     private final DruidWebService druidWebService;
     private final PhysicalTableDictionary physicalTableDictionary;
     private final DataSourceMetadataService metadataService;
@@ -65,9 +64,9 @@ public class DataSourceMetadataLoader extends Loader<Boolean> {
     private final FailureCallback failureCallback;
 
     /**
-     * Datasource metadata loader fetches data from the druid endpoint and updates the dimensions on that table.
+     * Datasource metadata loader fetches data from the druid coordinator and updates the datasource metadata service.
      *
-     * @param physicalTableDictionary  The physical tables to update
+     * @param physicalTableDictionary  The physical tables with data sources to update
      * @param metadataService  The service that will store the metadata loaded by this loader
      * @param druidWebService  The druid webservice to query
      * @param mapper  Object mapper to parse druid metadata
@@ -92,8 +91,7 @@ public class DataSourceMetadataLoader extends Loader<Boolean> {
         this.druidWebService = druidWebService;
         this.mapper = mapper;
         this.failureCallback = getFailureCallback();
-
-        lastRunTimestamp = new AtomicReference<>();
+        this.lastRunTimestamp = new AtomicReference<>();
     }
 
     @Override
@@ -101,13 +99,13 @@ public class DataSourceMetadataLoader extends Loader<Boolean> {
         physicalTableDictionary.values().stream()
                 .filter(table -> table instanceof ConcretePhysicalTable)
                 .map(table -> (ConcretePhysicalTable) table)
-                .peek(table -> LOG.trace("Querying metadata for datasource: {}", table))
+                .peek(dataSourceName -> LOG.trace("Querying metadata for datasource: {}", dataSourceName))
                 .forEach(this::queryDataSourceMetadata);
         lastRunTimestamp.set(DateTime.now());
     }
 
     /**
-     * Queries the data mart for updated datasource metadata and then updates the physical table.
+     * Queries Druid for updated datasource metadata and then updates the datasource metadata service.
      *
      * @param table  The physical table to be updated.
      */
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/metadata/DataSourceMetadataLoader.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/metadata/DataSourceMetadataLoader.java
@@ -13,9 +13,9 @@ import com.yahoo.bard.webservice.druid.client.DruidWebService;
 import com.yahoo.bard.webservice.druid.client.FailureCallback;
 import com.yahoo.bard.webservice.druid.client.HttpErrorCallback;
 import com.yahoo.bard.webservice.druid.client.SuccessCallback;
-import com.yahoo.bard.webservice.table.ConcretePhysicalTable;
 import com.yahoo.bard.webservice.table.PhysicalTable;
 import com.yahoo.bard.webservice.table.PhysicalTableDictionary;
+import com.yahoo.bard.webservice.table.SingleDataSourcePhysicalTable;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 
@@ -115,7 +115,7 @@ public class DataSourceMetadataLoader extends Loader<Boolean> {
      * @deprecated  Pass the DataSourceName directly, rather than via the PhysicalTable
      */
     @Deprecated
-    protected void queryDataSourceMetadata(ConcretePhysicalTable table) {
+    protected void queryDataSourceMetadata(SingleDataSourcePhysicalTable table) {
         queryDataSourceMetadata(table.getDataSourceName());
     }
 
@@ -192,7 +192,7 @@ public class DataSourceMetadataLoader extends Loader<Boolean> {
      * @deprecated  Pass the DataSourceName directly, rather than via the PhysicalTable
      */
     @Deprecated
-    protected final SuccessCallback buildDataSourceMetadataSuccessCallback(ConcretePhysicalTable table) {
+    protected final SuccessCallback buildDataSourceMetadataSuccessCallback(SingleDataSourcePhysicalTable table) {
         return buildDataSourceMetadataSuccessCallback(table.getDataSourceName());
     }
 
@@ -282,7 +282,7 @@ public class DataSourceMetadataLoader extends Loader<Boolean> {
      * @deprecated  Pass the DataSourceName directly, rather than via the PhysicalTable
      */
     @Deprecated
-    protected HttpErrorCallback getErrorCallback(ConcretePhysicalTable table) {
+    protected HttpErrorCallback getErrorCallback(SingleDataSourcePhysicalTable table) {
         return getErrorCallback(table.getDataSourceName());
     }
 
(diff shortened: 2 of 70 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/config/names/DataSourceName.java
```diff
@@ -0,0 +1,59 @@
...
+     *
+     * @return the name
+     */
+    String asName();
+
+    /**
+     * Wrap a string in an anonymous instance of TableName.
+     * Rather than make heavy use of this, instead make a class.
+     *
+     * @param name the name being wrapped
```
QubitPi: Two space between "param" and "name"; capitalized "name"

fili-core/src/main/java/com/yahoo/bard/webservice/metadata/DataSourceMetadataLoader.java
```diff
@@ -175,20 +188,78 @@ protected void queryDataSourceMetadata(ConcretePhysicalTable table) {
...
+     *
+     * @param dataSourceName  The datasource name to inject into this callback.
+     *
+     * @return The callback itself.
+     */
+    protected SuccessCallback buildDataSourceMetadataSuccessCallback(DataSourceName dataSourceName) {
+        return rootNode -> {
+            try {
+                DataSourceMetadata dataSourceMetadata = mapper.treeToValue(rootNode, DataSourceMetadata.class);
+                metadataService.update(dataSourceName, dataSourceMetadata);
```
QubitPi: Would it be better to combine this line and it's previous line into 1 line? i.e. inline dataSourceMetadata calculation

fili-core/src/main/java/com/yahoo/bard/webservice/metadata/DataSourceMetadataService.java
```diff
@@ -58,61 +58,71 @@ public DataSourceMetadataService() {
...
+     * @deprecated  Pass the DataSourceName directly, rather than via the PhysicalTable
      */
+    @Deprecated
     public void update(PhysicalTable table, DataSourceMetadata metadata) {
+        table.getDataSourceNames().forEach(dataSourceName -> update(dataSourceName, metadata));
+    }
+
+    /**
+     * Update the information with respect to the segment metadata of a particular data source.
+     * This operation should be atomic per dataSourceName.
```
cdeszaq: Other than steps 5/6, all of the rest are happening _only_ using data from the input parameters, so those operations _are_ atomic (because there's no state touched outside this method). That said, yes, I can remove the "atomic" part and replace it with "both maps are updated" (or something to that effect)

fili-core/src/main/java/com/yahoo/bard/webservice/metadata/SegmentIntervalsHashIdGenerator.java
```diff
@@ -124,9 +120,11 @@ public SegmentIntervalsHashIdGenerator(
...
      * @return A hash of the given segments
      */
-    public long getSegmentHash(Set<SortedMap<DateTime, Map<String, SegmentInfo>>> requestedSegments) {
-        return requestedSegments.stream()
-                .mapToLong(Object::hashCode)
-                .sum();
+    public Optional<Long> getSegmentHash(Stream<SortedMap<DateTime, Map<String, SegmentInfo>>> requestedSegments) {
+        return requestedSegments
+                .distinct()
+                .map(Object::hashCode)
```
cdeszaq: Unfortunately, `reduce` on a `LongStream` produces an `OptionalLong`, not an `Optional<Long>`, and those things are not compatible nor play well together. And the parent interface of the ID generator is a `QuerySigningService<Long>`, which requires the signature method to return `Optional<Long>`, so making the method return `OptionalLong` is a non-starter if we wish to keep the signing service able to return a different type than long (ie. string).

On the `make things more idiomatic` front, I like the idea. It ends up looking like this:

```java
if (
        querySigningService.getSegmentSetId(druidQuery)
                .map(id -> Objects.equals(cacheEntry.getMeta(), id))
                .orElse(false)
) {
    try {
```

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/BaseMetadataAvailability.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * An availability based on a DataSourceMetadataService, which is backed by a single data source in that service.
+ */
+public abstract class BaseMetadataAvailability implements Availability {
```
cdeszaq: It's more than that it has only 1 data source. In particular, it _must_ have a `metadataService` (as indicated in the javadoc), where the composite ones do not have one. The 1 datasource aspect is really ancillary, and not the main purpose, though it does tend to go hand-in-hand.

@michael-mclawhorn do you have thoughts here? or anyone else? I'm happy to change it, but it's more than just "not composite".

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/BaseMetadataAvailability.java
```diff
@@ -0,0 +1,56 @@
...
+        this.dataSourceNames = Collections.singleton(dataSourceName);
+    }
+
+    public DataSourceName getDataSourceName() {
+        return dataSourceName;
+    }
+
+    @Override
+    public Set<DataSourceName> getDataSourceNames() {
+        return dataSourceNames;
```
cdeszaq: I'm with you, but making the extra garbage is not something I'm a big fan of, since garbage often piles up in unexpected ways and gets collected often at the worst times. So, not making unnecesary garbage is, I think, the best route.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/table/BaseTableLoaderSpec.groovy
```diff
@@ -60,27 +58,20 @@ class BaseTableLoaderSpec extends Specification {
...
-            }
-            return new ConcretePhysicalTable(
-                    TableName.of(getName().asName()),
-                    DefaultTimeGrain.DAY.buildZonedTimeGrain(DateTimeZone.UTC),
-                    [] as Set,
-                    [:],
-                    metadataService
-            )
+        PhysicalTable build(ResourceDictionaries dictionaries, DataSourceMetadataService metadataService) {
+            !Objects.isNull(physicalTable) ?
```
cdeszaq: You are my hero

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/table/BaseTableLoaderSpec.groovy
```diff
@@ -111,19 +102,19 @@ class BaseTableLoaderSpec extends Specification {
         metricNames = TestDruidMetricName.getByLogicalTable(SHAPES)
         physDefs = TestPhysicalTableDefinitionUtils.buildShapeTableDefinitions(new TestDimensions(), metricNames)
         dimNames = TestApiDimensionName.getByLogicalTable(SHAPES)
-        tableNames = physDefs.collect {it.name} as Set
+        tableNames = physDefs.name as Set
```
cdeszaq: Groovy allows shortcutting the spread-dot operator on collections for property access of elements in the collection. So, in this case, these are equal:

```groovy
physDefs*.name == physDefs.name
```

The result is a `SpreadCollection`, that is being coerced into a `Set`

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/BaseDataServletComponentSpec.groovy
```diff
@@ -206,23 +203,22 @@ abstract class BaseDataServletComponentSpec extends Specification {
      */
     Response makeAbstractRequest(Closure queryParams=this.&getQueryParams) {
         // Set target of call
-        def httpCall = jtb.getHarness().target(getTarget())
+        def httpCall = jtb.harness.target(target)
```
cdeszaq: It would use the getter. For direct field access, you have to use the [field access operator](http://groovy-lang.org/operators.html#_direct_field_access_operator): 

```groovy
class User {
    public final String name                 
    User(String name) { this.name = name}
    String getName() { "Name: $name" }
}
def user = new User('Bob')
assert user.name == 'Name: Bob'   // Goes through the getter
assert user.@name == 'Bob' // Gets the field directly
```

fili-core/src/main/java/com/yahoo/bard/webservice/table/PermissiveConcretePhysicalTable.java
```diff
@@ -13,38 +14,73 @@
...
                 logicalToPhysicalColumnNames,
-                new PermissiveAvailability(name, dataSourceMetadataService)
+                availability
         );
+        this.dataSourceName = availability.getDataSourceName();
+    }
+
+    public DataSourceName getDataSourceName() {
+        return dataSourceName;
+    }
```
michael-mclawhorn: This redundancy strikes me as wrong.  Go ahead and pull the 'SingleDataSourcePhysicalTable' logic up into a parent class and simplify Permissive and Strict subclasses.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/BaseMetadataAvailability.java
michael-mclawhorn: I'm pretty sure the 'final' ensures that this set is a pure cache.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
```diff
@@ -64,43 +42,30 @@ public SimplifiedIntervalList getAvailableIntervals(PhysicalDataSourceConstraint
...
     public boolean equals(Object obj) {
         if (obj instanceof ConcreteAvailability) {
             ConcreteAvailability that = (ConcreteAvailability) obj;
-            return Objects.equals(name.asName(), that.name.asName())
-                    && Objects.equals(dataSourceNames, that.dataSourceNames)
+            return Objects.equals(getDataSourceName().asName(), that.getDataSourceName().asName())
                     // Since metadata service is mutable, use instance equality to ensure table equality is stable
-                    && metadataService == that.metadataService;
+                    && getDataSourceMetadataService() == that.getDataSourceMetadataService();
         }
```
michael-mclawhorn: Should permissive and concrete have distinct equals and hash code methods or can these be safely pushed up to the parent?

cdeszaq: The type matters (ie. a concrete shouldn't be considered the same as a permissive) because they have different behavior. _Some_ of this can be moved up to the parent, but not fully I don't think... Unless I'm wrong?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/BaseMetadataAvailability.java
cdeszaq: @michael-mclawhorn I'm not sure I get what you mean? I don't see a `final` on this line. Perhaps GitHub put this comment on the wrong line?

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
michael-mclawhorn: You could make class equality part of the truth in the parent.

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/BaseMetadataAvailability.java
michael-mclawhorn: I guess I don't really care either way. I think caching the prewrapped collection doesn't introduce any real two-sources-of-truth issue because the thing being wrapped is final.

So I'm gently preferring the performance optimization of skipping the extra call to Collections.singlton (not on space or GC, but on a fast-loop performance basis)

fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/ConcreteAvailability.java
cdeszaq: Makes sense. I'll make the changes to push these up to the parent (as much as I can anyways).

cdeszaq:
I think I've addressed all comments on this PR. There is some squashing work to be done once this PR is ready to merge, but I'll take that on once it gets another +1. (probably from either @michael-mclawhorn or @archolewa, since they've looked at it once already)

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/BaseDataServletComponentSpec.groovy
archolewa: Your example doesn't quite capture my concern. Here's a better one:

```groovy
class User {
    public final String name                 
    User(String name) { this.name = name}
    String getName() { "Name: $name" }
    void printName() {
         println(name)
     }
}
class OtherUser {
    String getName() { "Name: name" }
    void printName() { println(name) }
}

def user = new User('Bob')
assert user.name == 'Name: Bob'   // Goes through the getter
assert user.@name == 'Bob' // Gets the field directly
user.printName()
user.printName()
new OtherUser().printName()
```
Output:
```
Bob
Name: name
```
In other  words, inside of a class, the field takes precedence over the getter. Which means adding a private field may have very unexpected changes to behavior. 

So I would personally encourage that we should only use the shortcut outside of a class. When inside a class, we should always explicitly call the method.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/table/BaseTableLoaderSpec.groovy
archolewa: There are too many shortcuts in this language.

/EndGripe

archolewa:
This looks good. My only remaining concern is that I don't think we should use the `target <translated-by-Groovy-magic> getTarget` inside of a class, because the meaning of `target` will change if someone decides to add a field named `target`  See my [example](https://github.com/yahoo/fili/pull/263#discussion_r117328929).

I figure that is unlikely enough that I'm not going to block on it, but I do think it's a major gotcha that should be avoided whenever possible.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/BaseDataServletComponentSpec.groovy
cdeszaq: You are indeed correct, and in fact it's called out in the language spec: http://docs.groovy-lang.org/latest/html/documentation/#properties

And, if you think about it, it's required, otherwise there'd be a real possibility of an infinite loop! (or requiring something special if someone build a dumb getter).

I'll make a note of not doing this in the future, and I think Code Narc (Groovy's checkstyle equivalent) has rules to prevent this. Thx for the eagle-eye!

QubitPi:
Shall we merge this now?

cdeszaq:
Yes, this will get merged shortly. It needs the commits cleaned up / squashed where appropriate (ie. not fully squashed). Also, since this PR is rather large, I want to land it _after_ the Constraint PR, since that one (I think) matters more (and my head is in that one at the moment)

QubitPi:
@cdeszaq OK

michael-mclawhorn:
Ok, I tried to follow the rebase, and I think the SingleDataSourcePhysicalTable path was probably a mistake.  I've saved tags on my personal rebase path for this branch, but I'd like to suggest that approach be dropped in favor of a simple DataSourceName cardinality check in TableDataSource.

michael-mclawhorn:
My suggestion would be to dynamically rebase those changes off into a separate branch, keeping everything else.

michael-mclawhorn:
https://github.com/yahoo/fili/tree/fixDataSourecNameDisparityPartialRebase

cdeszaq:
@michael-mclawhorn We can deal with any additional refactoring after I think. I'd rather get this in and landed, since it's a fairly large block of changes, and conflict-prone. Merging.

PR #267, opened 2017-05-04
Merge Header Info into JsonNode

QubitPi:
2nd PR implementing https://github.com/yahoo/fili/issues/215

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,8 @@ Current
     * Add new response error messages as needed by Partial Data V2
 
 - [Merge Druid Response Header into Druid Response Body Json Node in AsyncDruidWebServiceImplV2](https://github.com/yahoo/fili/pull/267)
-    * Make the use of BinderFactory as an injected behavior. By using JSON node builder strategy, we can customize the
-    content of the JSON response
+    * Add configuration to `AsyncDruidWebServiceImpl` so that we can opt-in configuration of JSON response content.
+    * `AsyncDruidWebServiceImpl` takes a strategy for building the JSON from the entire response.
 
 - [Add MetricUnionCompositeTableDefinition](https://github.com/yahoo/fili/pull/258)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ Current
     * Add a configurable property named "druid_uncovered_interval_limit"
     * Add new response error messages as needed by Partial Data V2
 
+- [Merge Druid Response Header into Druid Response Body Json Node in AsyncDruidWebServiceImplV2](https://github.com/yahoo/fili/pull/267)
+    * Make the use of BinderFactory as an injected behavior. By using JSON node builder strategy, we can customize the
+    content of the JSON response
+
 - [Add MetricUnionCompositeTableDefinition](https://github.com/yahoo/fili/pull/258)
 
 - [Add partition availability and table](https://github.com/yahoo/fili/pull/244)
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
@@ -21,7 +21,6 @@ import com.yahoo.bard.webservice.web.handlers.RequestContext;
 
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
-import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.MappingJsonFactory;
@@ -39,11 +38,11 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.io.InputStream;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Function;
 import java.util.function.Supplier;
 
 import javax.ws.rs.core.Response.Status;
@@ -59,6 +58,7 @@ public class AsyncDruidWebServiceImpl implements DruidWebService {
     private final Meter httpErrorMeter;
     private final Meter exceptionMeter;
     private static final MetricRegistry REGISTRY = MetricRegistryFactory.getRegistry();
+
     public static final String DRUID_TIMER = "DruidProcessing";
     public static final String DRUID_QUERY_TIMER = DRUID_TIMER + "_Q_";
     public static final String DRUID_QUERY_ALL_TIMER = DRUID_QUERY_TIMER + "All";
@@ -66,8 +66,23 @@ public class AsyncDruidWebServiceImpl implements DruidWebService {
     public static final String DRUID_WEIGHTED_QUERY_TIMER = DRUID_TIMER + "_W_";
     public static final String DRUID_SEGMENT_METADATA_TIMER = DRUID_TIMER + "_S_0";
 
+    public static final Function<Response, JsonNode> DEFAULT_JSON_NODE_BUILDER_STRATEGY =
+            new Function<Response, JsonNode>() {
+
+        @Override
+        public JsonNode apply(final Response response) {
+            try {
+                return new MappingJsonFactory().createParser(response.getResponseBodyAsStream()).readValueAsTree();
+            } catch (IOException ioe) {
+                throw new IllegalStateException(ioe);
+            }
+        }
+    };
+
     private final Supplier<Map<String, String>> headersToAppend;
     private final DruidServiceConfig serviceConfig;
+
+    private final Function<Response, JsonNode> jsonNodeBuilderStrategy;
 
     /**
      * Friendly non-DI constructor useful for manual tests.
@@ -75,14 +90,38 @@ public class AsyncDruidWebServiceImpl implements DruidWebService {
      * @param serviceConfig  Configuration for the Druid Service
      * @param mapper  A shared jackson object mapper resource
      * @deprecated We now require a header supplier parameter.
-     *      Use {@link #AsyncDruidWebServiceImpl(DruidServiceConfig, ObjectMapper, Supplier)}
+     *      Use {@link #AsyncDruidWebServiceImpl(DruidServiceConfig, AsyncHttpClient, ObjectMapper, Supplier, Function)}
      */
     @Deprecated
     public AsyncDruidWebServiceImpl(
             DruidServiceConfig serviceConfig,
             ObjectMapper mapper
     ) {
-        this(serviceConfig, initializeWebClient(serviceConfig.getTimeout()), mapper, HashMap::new);
+        this(
+                serviceConfig,
+                initializeWebClient(serviceConfig.getTimeout()),
+                mapper,
+                HashMap::new,
+                DEFAULT_JSON_NODE_BUILDER_STRATEGY
+        );
+    }
+
+    /**
+     * IOC constructor.
+     *
+     * @param config  the configuration for this druid service
+     * @param asyncHttpClient  the HTTP client
+     * @param mapper  A shared jackson object mapper resource
+     * @deprecated  We now require a header supplier parameter.
+     *      Use {@link #AsyncDruidWebServiceImpl(DruidServiceConfig, AsyncHttpClient, ObjectMapper, Supplier, Function)}
+     */
+    @Deprecated
+    public AsyncDruidWebServiceImpl(
+            DruidServiceConfig config,
+            AsyncHttpClient asyncHttpClient,
+            ObjectMapper mapper
+    ) {
+        this(config, asyncHttpClient, mapper, HashMap::new, DEFAULT_JSON_NODE_BUILDER_STRATEGY);
     }
 
     /**
@@ -97,7 +136,86 @@ public class AsyncDruidWebServiceImpl implements DruidWebService {
             ObjectMapper mapper,
             Supplier<Map<String, String>> headersToAppend
     ) {
-        this(serviceConfig, initializeWebClient(serviceConfig.getTimeout()), mapper, headersToAppend);
+        this(
+                serviceConfig,
+                initializeWebClient(serviceConfig.getTimeout()),
+                mapper,
+                headersToAppend,
+                DEFAULT_JSON_NODE_BUILDER_STRATEGY
+        );
+    }
+
+    /**
+     * Friendly non-DI constructor useful for manual tests.
+     *
+     * @param serviceConfig  Configuration for the Druid Service
+     * @param mapper  A shared jackson object mapper resource
+     * @param headersToAppend Supplier for map of headers for Druid requests
+     * @param jsonNodeBuilderStrategy A function to build JSON nodes from the response
+     */
+    public AsyncDruidWebServiceImpl(
+            DruidServiceConfig serviceConfig,
+            ObjectMapper mapper,
+            Supplier<Map<String, String>> headersToAppend,
+            Function<Response, JsonNode> jsonNodeBuilderStrategy
+    ) {
+        this(
+                serviceConfig,
+                initializeWebClient(serviceConfig.getTimeout()),
+                mapper,
+                headersToAppend,
+                jsonNodeBuilderStrategy
+        );
+    }
+    /**
+     * IOC constructor.
+     *
+     * @param config  the configuration for this druid service
+     * @param asyncHttpClient  the HTTP client
+     * @param mapper  A shared jackson object mapper resource
+     * @param headersToAppend Supplier for map of headers for Druid requests
+     */
+    public AsyncDruidWebServiceImpl(
+            DruidServiceConfig config,
+            AsyncHttpClient asyncHttpClient,
+            ObjectMapper mapper,
+            Supplier<Map<String, String>> headersToAppend
+    ) {
+        this(config, asyncHttpClient, mapper, headersToAppend, DEFAULT_JSON_NODE_BUILDER_STRATEGY);
+    }
+
+    /**
+     * IOC constructor.
+     *
+     * @param config  the configuration for this druid service
+     * @param asyncHttpClient  the HTTP client
+     * @param mapper  A shared jackson object mapper resource
+     * @param headersToAppend Supplier for map of headers for Druid requests
+     * @param jsonNodeBuilderStrategy A function to build JSON nodes from the response
+     */
+    public AsyncDruidWebServiceImpl(
+            DruidServiceConfig config,
+            AsyncHttpClient asyncHttpClient,
+            ObjectMapper mapper,
+            Supplier<Map<String, String>> headersToAppend,
+            Function<Response, JsonNode> jsonNodeBuilderStrategy
+    ) {
+        this.serviceConfig = config;
+
+        if (serviceConfig.getUrl() == null) {
+            String msg = DRUID_URL_INVALID.format(config.getNameAndUrl());
+            LOG.error(msg);
+            throw new IllegalStateException(msg);
+        }
+
+        LOG.info("Configured with druid server config: {}", config.toString());
+        this.headersToAppend = headersToAppend;
+        this.webClient = asyncHttpClient;
+        this.writer = mapper.writer();
+        this.httpErrorMeter = REGISTRY.meter("druid.errors.http");
+        this.exceptionMeter = REGISTRY.meter("druid.errors.exceptions");
+
+        this.jsonNodeBuilderStrategy = jsonNodeBuilderStrategy;
     }
 
     /**
@@ -124,53 +242,6 @@ public class AsyncDruidWebServiceImpl implements DruidWebService {
         return new DefaultAsyncHttpClient(config);
     }
 
-    /**
-     * IOC constructor.
-     *
-     * @param config  the configuration for this druid service
-     * @param asyncHttpClient  the HTTP client
-     * @param mapper  A shared jackson object mapper resource
-     * @deprecated  We now require a header supplier parameter.
-     *      Use {@link #AsyncDruidWebServiceImpl(DruidServiceConfig, AsyncHttpClient, ObjectMapper, Supplier)}
-     */
-    @Deprecated
-    public AsyncDruidWebServiceImpl(
-            DruidServiceConfig config,
-            AsyncHttpClient asyncHttpClient,
-            ObjectMapper mapper
-    ) {
-        this(config, asyncHttpClient, mapper, HashMap::new);
-    }
-
-    /**
-     * IOC constructor.
-     *
-     * @param config  the configuration for this druid service
-     * @param asyncHttpClient  the HTTP client
-     * @param mapper  A shared jackson object mapper resource
-     * @param headersToAppend Supplier for map of headers for Druid requests
-     */
-    public AsyncDruidWebServiceImpl(
-            DruidServiceConfig config,
-            AsyncHttpClient asyncHttpClient,
-            ObjectMapper mapper,
-            Supplier<Map<String, String>> headersToAppend
-    ) {
-        this.serviceConfig = config;
-
-        if (serviceConfig.getUrl() == null) {
-            String msg = DRUID_URL_INVALID.format(config.getNameAndUrl());
-            LOG.error(msg);
-            throw new IllegalStateException(msg);
-        }
-
-        LOG.info("Configured with druid server config: {}", config.toString());
-        this.headersToAppend = headersToAppend;
-        this.webClient = asyncHttpClient;
-        this.writer = mapper.writer();
-        this.httpErrorMeter = REGISTRY.meter("druid.errors.http");
-        this.exceptionMeter = REGISTRY.meter("druid.errors.exceptions");
-    }
 
     /**
      * Serializes the provided query and invokes a request on the druid broker.
@@ -199,50 +270,19 @@ public class AsyncDruidWebServiceImpl implements DruidWebService {
                 new AsyncCompletionHandler<Response>() {
                     @Override
                     public Response onCompleted(Response response) {
-                        RequestLog.restore(logCtx);
-                        RequestLog.stopTiming(timerName);
-                        if (outstanding.decrementAndGet() == 0) {
-                            RequestLog.startTiming(RESPONSE_WORKFLOW_TIMER);
-                        }
                         String druidQueryId = response.getHeader("X-Druid-Query-Id");
-                        RequestLog.record(new DruidResponse(druidQueryId));
                         Status status = Status.fromStatusCode(response.getStatusCode());
-                        LOG.debug(
-                                "druid {} response code: {} {} and druid query id: {}",
-                                serviceConfig.getNameAndUrl(),
-                                status.getStatusCode(),
-                                status,
-                                druidQueryId
-                        );
+                        logRequest(logCtx, timerName, outstanding, druidQueryId, status);
 
-                        if (status != Status.OK) {
-                            httpErrorMeter.mark();
-                            LOG.debug(
-                                    "druid {} error: {} {} {} and druid query id: {}",
-                                    serviceConfig.getNameAndUrl(),
-                                    status.getStatusCode(),
-                                    status.getReasonPhrase(),
-                                    response.getResponseBody(),
-                                    druidQueryId
-                            );
-
-                            error.invoke(
-                                    status.getStatusCode(),
-                                    status.getReasonPhrase(),
-                                    response.getResponseBody()
-                            );
+                        if (hasError(status)) {
+                            markError(status, response, druidQueryId, error);
                         } else {
-                            MappingJsonFactory jsonFactory = new MappingJsonFactory();
                             try {
-                                JsonNode rootNode;
-                                try (InputStream responseStream = response.getResponseBodyAsStream();
-                                    JsonParser jp = jsonFactory.createParser(responseStream)) {
-                                    rootNode = jp.readValueAsTree();
-                                }
-                                success.invoke(rootNode);
-                            } catch (RuntimeException | IOException e) {
+                                success.invoke(jsonNodeBuilderStrategy.apply(response));
+                            } catch (RuntimeException e) {
                                 failure.invoke(e);
                             }
+
                         }
 
                         // we consumed this response, so pass null to any chains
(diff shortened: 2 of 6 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -56,8 +56,8 @@
 
     private final AsyncHttpClient webClient;
     private final ObjectWriter writer;
-    private final Meter httpErrorMeter;
-    private final Meter exceptionMeter;
+    protected final Meter httpErrorMeter;
+    protected final Meter exceptionMeter;
```
cdeszaq: Rather than exposing fields directly, we should use methods instead. In this case, methods to count errors, etc. are probably the right call, and accessors will likely work for the other variables made protected in this pr

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2.java
```diff
@@ -0,0 +1,170 @@
...
+     */
+    public AsyncDruidWebServiceImplV2(
+            DruidServiceConfig config,
+            ObjectMapper mapper,
+            Supplier<Map<String, String>> headersToAppend
+    ) {
+        super(config, mapper, headersToAppend);
+    }
+
+    @Override
```
cdeszaq: Annotations should go between the method declaration and it's Javadoc

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -356,4 +357,12 @@ public Integer getTimeout() {
     public DruidServiceConfig getServiceConfig() {
         return serviceConfig;
     }
+
+    public Meter getHttpErrorMeter() {
```
garyluoex: should be protected.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -356,4 +357,12 @@ public Integer getTimeout() {
...
+        return httpErrorMeter;
+    }
+
+    public Meter getExceptionMeter() {
```
garyluoex: should be protected.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -66,8 +67,8 @@
     public static final String DRUID_WEIGHTED_QUERY_TIMER = DRUID_TIMER + "_W_";
     public static final String DRUID_SEGMENT_METADATA_TIMER = DRUID_TIMER + "_S_0";
 
-    private final Supplier<Map<String, String>> headersToAppend;
-    private final DruidServiceConfig serviceConfig;
+    protected final Supplier<Map<String, String>> headersToAppend;
```
garyluoex: why this needs to be protected?

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -66,8 +67,8 @@
...
+    protected final DruidServiceConfig serviceConfig;
```
garyluoex: Why this needs to be protected?

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
QubitPi: changed back to private

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
QubitPi: changed back to private

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2.java
```diff
@@ -0,0 +1,173 @@
...
+                                try {
+                                    JsonNode rootNode;
+                                    ObjectNode objectNode;
+                                    try (
+                                            JsonParser jsonParser = jsonFactory.createParser(
+                                                    response.getResponseBodyAsStream()
+                                            )
+                                    ) {
+                                        rootNode = jsonParser.readValueAsTree();
+                                        objectNode = (ObjectNode) rootNode;
```
garyluoex: I would try out and consider moving the response parsing into jsonNode process into its own method in AsyncDruidWebserviceImpl.java and simply override it here instead of overriding the whole sendRequest method. If we do that, we also need to move the response status check on the top of this sendRequest method into its own method and override it here also. If you think this makes sense too, try it.

garyluoex: Also fix the unused import things

QubitPi: @garyluoex It's a long method. So yeah lets break it down

fili-core/src/main/java/com/yahoo/bard/webservice/web/ErrorMessageFormat.java
```diff
@@ -215,17 +215,6 @@
...
     RESULT_MAPPING_FAILURE(
             "Error occurred while processing response data: %s"
-    ),
-
-    DATA_AVAILABILITY_MISMATCH(
-            "Data availability expectation does not match with actual query result obtained from druid for the " +
-                    "following intervals %s where druid does not have data"
-    ),
-
-    TOO_MUCH_INTERVAL_MISSING(
```
garyluoex: Why did we remove these?

fili-core/src/main/java/com/yahoo/bard/webservice/web/ErrorMessageFormat.java
```diff
@@ -215,17 +215,6 @@
 
     RESULT_MAPPING_FAILURE(
             "Error occurred while processing response data: %s"
-    ),
-
-    DATA_AVAILABILITY_MISMATCH(
```
garyluoex: Why did we remove these?

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    protected boolean hasError(Status status) {
+        return status != Status.OK && status != Status.NOT_MODIFIED;
+    }
+
+    @Override
+    protected JsonNode constructJsonResponse(Response response) throws IOException {
+        ObjectNode objectNode = (ObjectNode) OBJECT_MAPPER.readTree(response.getResponseBody());
```
garyluoex: The reason I suggested to create a new ObjectNode is because I suspect that the JsonNode created by 
```java
(ObjectNode) OBJECT_MAPPER.readTree(response.getResponseBody());
```
is actually an `ArrayNode`, force casting it to `ObjectNode` will cause runtime error. Which I want the test to test, but it seems like the test assumes that the `response.getResponseBody()` returns a object, but I think it is actually an array.

To create a new node, 
```java
        ObjectNode node = JsonNodeFactory.instance.objectNode();
```

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -356,4 +326,105 @@ public Integer getTimeout() {
...
+    /**
+     * Log and invoke error response on non-OK druid response.
+     *
+     * @param status  The response status
+     * @param response  The druid response
+     * @param druidQueryId  The Druid query ID
+     * @param error  callback for handling http errors.
+     */
+    private void markError(Status status, Response response, String druidQueryId, HttpErrorCallback error) {
+        httpErrorMeter.mark();
```
cdeszaq: Should use the getter, not the direct field access

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -356,4 +326,105 @@ public Integer getTimeout() {
...
+    protected Meter getExceptionMeter() {
+        return exceptionMeter;
+    }
+
+    protected DruidServiceConfig getDruidServiceConfig() {
+        return serviceConfig;
+    }
+
+    /**
+     * Log request using RequestLog.
```
cdeszaq: This is doing much more than just logging the request, right? We should probably call out all of those different responsibilities of this method?

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -356,4 +326,105 @@ public Integer getTimeout() {
...
+        RequestLog.restore(logCtx);
+        RequestLog.stopTiming(timerName);
+        if (outstanding.decrementAndGet() == 0) {
+            RequestLog.startTiming(RESPONSE_WORKFLOW_TIMER);
+        }
+        RequestLog.record(new DruidResponse(druidQueryId));
+
+        LOG.debug(
+                "druid {} response code: {} {} and druid query id: {}",
+                serviceConfig.getNameAndUrl(),
```
cdeszaq: Should use the getter, not direct field access

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
cdeszaq: It's also worth noting that this isn't logging the request, but it's logging the completed response. If it was logging the request, it would not need to be triggering with pieces of the response. And if it was for any response, it should handle the failure cases as well.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -199,50 +199,20 @@ public AsyncDruidWebServiceImpl(
...
                         } else {
-                            MappingJsonFactory jsonFactory = new MappingJsonFactory();
                             try {
-                                JsonNode rootNode;
-                                try (InputStream responseStream = response.getResponseBodyAsStream();
-                                    JsonParser jp = jsonFactory.createParser(responseStream)) {
-                                    rootNode = jp.readValueAsTree();
-                                }
-                                success.invoke(rootNode);
+                                JsonNode jsonNode = constructJsonResponse(response);
```
cdeszaq: Inline

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2Spec.groovy
```diff
@@ -0,0 +1,60 @@
...
+import java.util.function.Supplier;
+
+class AsyncDruidWebServiceImplV2Spec extends Specification {
+    def "Make sure X-Druid-Response-Context and status-code are mreged into existing JsonNode"() {
+        given:
+        Response response = Mock(Response)
+        response.getResponseBody() >>
+                """
+                    {"k1":"v1"}
+                """
```
garyluoex: Also the first level should be array instead of obeject.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2Spec.groovy
```diff
@@ -0,0 +1,60 @@
...
+                    {
+                        "k1":"v1",
+                        "X-Druid-Response-Context": "{
+                            \\"uncoveredIntervals\\": [
+                                \\"2016-11-22T00:00:00.000Z/2016-12-18T00:00:00.000Z\\",
+                                \\"2016-12-25T00:00:00.000Z/2017-01-03T00:00:00.000Z\\"
+                            ],
+                            \\"uncoveredIntervalsOverflowed\\": true
+                        }",
+                        "status-code": 200
```
garyluoex: We need to check if it is 304, if it is Partial Data V2 will delegate processing responsibility directly to the next response processor which is the CacheV3ResponseProcessor

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2Spec.groovy
```diff
@@ -0,0 +1,60 @@
...
+        druidServiceConfig.getTimeout() >> 100
+        druidServiceConfig.getUrl() >> "url"
+        AsyncDruidWebServiceImplV2 asyncDruidWebServiceImplV2 = new AsyncDruidWebServiceImplV2(
+                druidServiceConfig,
+                new ObjectMapper(),
+                Mock(Supplier)
+        )
+
+        expect:
+        print asyncDruidWebServiceImplV2.constructJsonResponse(response)
```
QubitPi: I forgot to delete it

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2Spec.groovy
QubitPi: @cdeszaq We need this in the 3rd implementation of Partial Data V2 - "Check response status code, if 304, invoke next response processor directly following the rules of the last bullet point in this section, if 200, do the following"

fili-core/src/main/java/com/yahoo/bard/webservice/web/ErrorMessageFormat.java
QubitPi: They are back

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
QubitPi: This looks like a getter

fili-core/src/main/java/com/yahoo/bard/webservice/web/ErrorMessageFormat.java
```diff
@@ -215,17 +215,6 @@
...
-            "Data availability expectation does not match with actual query result obtained from druid for the " +
-                    "following intervals %s where druid does not have data"
-    ),
-
-    TOO_MUCH_INTERVAL_MISSING(
-            "More than %s interval missing information received from druid, inspect if query " +
-                    "expects more than %s missing intervals or increase " +
-                    "uncoveredIntervalsLimit configuration value"
```
QubitPi: They are back

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2Spec.groovy
QubitPi: @garyluoex Could you give an example?

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2Spec.groovy
```diff
@@ -0,0 +1,60 @@
...
+        asyncDruidWebServiceImplV2.constructJsonResponse(response).toString() ==
+                """
+                    {
+                        "k1":"v1",
+                        "X-Druid-Response-Context": "{
+                            \\"uncoveredIntervals\\": [
+                                \\"2016-11-22T00:00:00.000Z/2016-12-18T00:00:00.000Z\\",
+                                \\"2016-12-25T00:00:00.000Z/2017-01-03T00:00:00.000Z\\"
+                            ],
+                            \\"uncoveredIntervalsOverflowed\\": true
```
QubitPi: Looks like we need to escape "\", which is part of response in String representation

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2Spec.groovy
cdeszaq: Yes, why should it be an array and not an object? Are you saying that Druid's response is a top-level array?

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2Spec.groovy
cdeszaq: In a multi-line GString (ie. bounded by `"""`), we shouldn't need to escape single double-quotes. (ie. `"`). I guess I'm not understanding why we need the escapes still?

If we do need the escapes, perhaps we can use a multi-line String (ie. `'''` bounded) instead?

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2Spec.groovy
cdeszaq: Ok, so this is needed for a different feature (which _should_ be able to be fully independent, so either can be used w/o the other), right? I'm OK with that, but in general, I'd not put work into a PR that is for a future feature if it's not needed for the current feature.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -356,4 +323,97 @@ public Integer getTimeout() {
...
+    protected Meter getExceptionMeter() {
+        return exceptionMeter;
+    }
+
+    protected DruidServiceConfig getDruidServiceConfig() {
+        return serviceConfig;
+    }
+
+    /**
+     * Logs request using RequestLog and logs Druid response.
```
cdeszaq: This also restores the serialized request log info back into the RequestLog object, stops / starts some timers, counts outstanding queries, etc. We should definitely call those things out, since those are _very_ important side-effects of this method.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2.java
```diff
@@ -0,0 +1,71 @@
...
+import java.io.IOException;
+import java.util.Map;
+import java.util.function.Supplier;
+
+import javax.ws.rs.core.Response.Status;
+
+/**
+ * Represents the druid web service endpoint for partial data V2.
+ * <p>
+ * When Druid returns data with missing intervals with a non-OK status code, instead of
```
cdeszaq: Will Druid have a non-OK response code in that case? (ie. where there are missing intervals) Or does it still return as a 200 OK?

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2.java
```diff
@@ -37,14 +42,27 @@ public AsyncDruidWebServiceImplV2(
...
     }
 
     @Override
     protected JsonNode constructJsonResponse(Response response) throws IOException {
-        ObjectNode objectNode = (ObjectNode) OBJECT_MAPPER.readTree(response.getResponseBody());
+        ObjectNode objectNode = JsonNodeFactory.instance.objectNode();
+
+        JsonNode responseNode = new MappingJsonFactory().createParser(response.getResponseBodyAsStream())
+                .readValueAsTree();
+        objectNode.put("response", responseNode);
```
cdeszaq: This version of `put` is deprecated.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2Spec.groovy
```diff
@@ -9,16 +9,14 @@ import spock.lang.Specification
...
 class AsyncDruidWebServiceImplV2Spec extends Specification {
-    def "Make sure X-Druid-Response-Context and status-code are mreged into existing JsonNode"() {
+    def "Make sure X-Druid-Response-Context and status-code are merged into existing JsonNode"() {
         given:
         Response response = Mock(Response)
-        response.getResponseBody() >>
-                """
-                    {"k1":"v1"}
-                """
+        response.getResponseBodyAsStream() >> new ByteArrayInputStream("{\"k1\":\"v1\"}".getBytes(StandardCharsets.UTF_8))
```
cdeszaq: Rather than escaping the quotes, we should use Strings (instead of GStrings) if we don't need the templating behavior.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2Spec.groovy
QubitPi: OK, moving to next PR

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2Spec.groovy
QubitPi: @cdeszaq JSON parser seems to escape double-quote by adding "\", the extra escape is for this "\". I feel like this is Jackson internal

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2Spec.groovy
QubitPi: Beautiful!

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplV2Spec.groovy
```diff
@@ -0,0 +1,55 @@
...
+import org.asynchttpclient.Response
+
+import java.nio.charset.StandardCharsets
+import java.util.function.Supplier
+
+class AsyncDruidWebServiceImplV2Spec extends Specification {
+    def "Make sure X-Druid-Response-Context and status-code are merged into existing JsonNode"() {
+        given:
+        Response response = Mock(Response)
+        response.getResponseBodyAsStream() >> new ByteArrayInputStream('{"k1":"v1"}'.getBytes(StandardCharsets.UTF_8))
```
garyluoex: The body should be an array at the top level instead of object.

QubitPi: Changed

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/HeaderNestingJsonBuilderStrategy.java
```diff
@@ -0,0 +1,36 @@
...
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import org.asynchttpclient.Response;
+
+import java.util.function.Function;
+
+/**
+ * A strategy that takes a JSON Node built from a Response, nests it and makes it a peer with the Response headers.
+ */
+public class HeaderNestingJsonBuilderStrategy implements Function<Response, JsonNode> {
```
garyluoex: Can we remove the word Strategy? It makes the name longer without adding much information I think.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplWithHeaderNestingSpec.groovy
```diff
@@ -12,11 +12,12 @@ import org.asynchttpclient.Response
...
 import java.util.function.Supplier
 
-class AsyncDruidWebServiceImplV2Spec extends Specification {
+class AsyncDruidWebServiceImplWithHeaderNestingSpec extends Specification {
+
     def "Make sure X-Druid-Response-Context and status-code are merged into existing JsonNode"() {
         given:
         Response response = Mock(Response)
-        response.getResponseBodyAsStream() >> new ByteArrayInputStream('[{"k1":"v1"}]'.getBytes(StandardCharsets.UTF_8))
+        response.getResponseBodyAsStream() >> new ByteArrayInputStream('{"k1":"v1"}'.getBytes(StandardCharsets.UTF_8))
```
garyluoex: Rebase error, need to put back the list bracket.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplWithHeaderNestingSpec.groovy
```diff
@@ -31,18 +32,18 @@ class AsyncDruidWebServiceImplV2Spec extends Specification {
...
         )
 
         expect:
-        asyncDruidWebServiceImplV2.constructJsonResponse(response).toString()
-                .replaceAll("\\\\", "") ==
+        asyncDruidWebServiceImplV2.jsonNodeBuilderStrategy.apply(response).toString().replaceAll("\\\\", "") ==
                 '''
                     {
-                        "response": "[{"k1":"v1"}]",
+                        "response": {"k1":"v1"},
```
garyluoex: Rebase error, need to put back the list bracket.

CHANGELOG.md
```diff
@@ -14,6 +14,12 @@ Current
     * Add a configurable property named "druid_uncovered_interval_limit"
     * Add new response error messages as needed by Partial Data V2
 
+- [Merge Druid Response Header into Druid Response Body Json Node in AsyncDruidWebServiceImplV2](https://github.com/yahoo/fili/pull/267)
+    * Added `AsyncDruidWebServiceImplV2` which overrides `sendRequest` method in `AsyncDruidWebServiceImpl` by adding
```
garyluoex: Need to update.

CHANGELOG.md
```diff
@@ -14,6 +14,12 @@ Current
...
+    "X-Druid-Response-Context" to the response JSON
+    * AbstractBinderFactory::buildDruidWebService returns `AsyncDruidWebServiceImplV2` or `AsyncDruidWebServiceImplV2`
```
garyluoex: Need to update.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/HeaderNestingJsonBuilderStrategy.java
QubitPi: @garyluoex, No problem. @michael-mclawhorn Do I have the permission to remove it?

cdeszaq: It's naming the `Function` that this extends from, indicating how the function is being used (ie. the Strategy pattern, to determine some behavior in a broader set of processing). This name is really 2 parts: `HeaderNesting` and `JsonBuilderStrategy`. The 1st part is describing this implementation, and the 2nd part is really what purpose this `Function` is serving.

So, tl;dr: I'd keep it. But that's just my opinion.

garyluoex: Not critical, just leave it then

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ Current
     * Add a configurable property named "druid_uncovered_interval_limit"
     * Add new response error messages as needed by Partial Data V2
 
+- [Merge Druid Response Header into Druid Response Body Json Node in AsyncDruidWebServiceImplV2](https://github.com/yahoo/fili/pull/267)
```
cdeszaq: It may be worth calling out here that this is not a change to the default behavior, but is instead opt-in via configuration (and including how to configure that opt-in)

cdeszaq: Should also call out the change to the `AsyncDruidWebServiceClientImpl`, in that it now takes a strategy for building the JSON from the entire response. (since that's what the ability we're adding hinges on)

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -59,30 +58,70 @@
...
     private static final MetricRegistry REGISTRY = MetricRegistryFactory.getRegistry();
+
     public static final String DRUID_TIMER = "DruidProcessing";
     public static final String DRUID_QUERY_TIMER = DRUID_TIMER + "_Q_";
     public static final String DRUID_QUERY_ALL_TIMER = DRUID_QUERY_TIMER + "All";
     public static final String DRUID_QUERY_MAX_TIMER = DRUID_QUERY_TIMER + "Max";
     public static final String DRUID_WEIGHTED_QUERY_TIMER = DRUID_TIMER + "_W_";
     public static final String DRUID_SEGMENT_METADATA_TIMER = DRUID_TIMER + "_S_0";
 
+    public static final Function<Response, JsonNode> DEFAULT_JSON_NODE_BUILDER_STRATEGY =
```
cdeszaq: Needs JavaDoc (since this essentially a class) explaining what it does. "Default" isn't super descriptive.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -59,30 +58,70 @@
...
+            new Function<Response, JsonNode>() {
+
+        @Override
+        public JsonNode apply(final Response response) {
```
cdeszaq: `final` not needed

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -97,49 +136,52 @@ public AsyncDruidWebServiceImpl(
             ObjectMapper mapper,
             Supplier<Map<String, String>> headersToAppend
     ) {
-        this(serviceConfig, initializeWebClient(serviceConfig.getTimeout()), mapper, headersToAppend);
+        this(
+                serviceConfig,
+                initializeWebClient(serviceConfig.getTimeout()),
+                mapper,
+                headersToAppend,
+                DEFAULT_JSON_NODE_BUILDER_STRATEGY
```
cdeszaq: JavaDoc on this should call out that this constructor uses a particular strategy as the default.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -97,49 +136,52 @@ public AsyncDruidWebServiceImpl(
...
+    ) {
+        this(
+                serviceConfig,
+                initializeWebClient(serviceConfig.getTimeout()),
+                mapper,
+                headersToAppend,
+                jsonNodeBuilderStrategy
+        );
     }
-
```
cdeszaq: Keep 1 blank line between methods

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -97,49 +136,52 @@ public AsyncDruidWebServiceImpl(
...
     /**
      * IOC constructor.
```
cdeszaq: JavaDoc should call out what strategy is used as the default, since this constructor doesn't take a strategy.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -356,4 +396,104 @@ public Integer getTimeout() {
...
+
+    /**
+     * Extract relevant information from response and wrap them into a JSON node.
+     *
+     * @param response  The response from which the information is to be retrieved.
+     *
+     * @return the relevant information in a JSON node
+     * @throws IOException when there is a JSON parsing error
+     */
+    protected JsonNode constructJsonResponse(Response response) throws IOException {
```
cdeszaq: This method is never used, and seems to have been replaced by the strategy pattern.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/HeaderNestingJsonBuilderStrategy.java
```diff
@@ -0,0 +1,36 @@
...
+     * Constructor.
+     *
+     * @param baseStrategy  strategy to do initial JSON Node construction
+     */
+    public HeaderNestingJsonBuilderStrategy(Function<Response, JsonNode> baseStrategy) {
+        this.baseStrategy = baseStrategy;
+    }
+
+    @Override
+    public JsonNode apply(final Response response) {
```
cdeszaq: `final` not needed

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/HeaderNestingJsonBuilderStrategy.java
QubitPi: OK

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/HeaderNestingJsonBuilderStrategy.java
```diff
@@ -0,0 +1,36 @@
...
+        ObjectNode objectNode = JsonNodeFactory.instance.objectNode();
+        objectNode.set("response", baseStrategy.apply(response));
+        objectNode.put("X-Druid-Response-Context", response.getHeader("X-Druid-Response-Context"));
```
garyluoex: Should use `set` instead of `put`.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
QubitPi: @cdeszaq I removed `final`, but why is it not needed? I'm just curious.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/HeaderNestingJsonBuilderStrategy.java
QubitPi: @garyluoex Looks like `set` needs to have its second argument as JSON node, instead of String. I think that's why we use `put`, which saves conversion?

garyluoex: put is deprecated, we should parse it as a jsonNode

garyluoex: Modify test accordingly also.

PR #269, opened 2017-05-09
Delay filter construction

archolewa:
--Before, we were building Druid filters at DataApiRequest construction
time. However, DataApiRequestMappers are applied after DataApiRequest
construction, and one of the most common uses of mappers is to add
additional filters.

--As a result, we went to the dimension search providers to build the
Druid filter (an expensive operation), and then have to build the
filters again at least once if the customer wants to add any implicit
mappers.

--Therefore, we make druid filter construction lazy: every time
`DataApiRequest::getFilter` is invoked, we build the Druid filters. Now,
mappers are free to add implicit filters without having to worry about
keeping the Druid filter in-sync.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
@@ -107,7 +107,6 @@ public class DataApiRequest extends ApiRequest {
     private final Set<LogicalMetric> logicalMetrics;
     private final Set<Interval> intervals;
     private final Map<Dimension, Set<ApiFilter>> filters;
-    private final Filter filter;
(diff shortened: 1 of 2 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -1393,8 +1378,20 @@ public Granularity getGranularity() {
         return this.filters;
     }
 
+    /**
+    * Builds and returns the Druid filters from this request's {@link ApiFilter}s.
+    * <p>
+    * The Druid filters are built (an expensive operation) every time this method is called. Use it judiciously.
+    *
+    * @return The Druid filter
```
QubitPi: Not a big deal but "The" shall be lowercase

PR #270, opened 2017-05-12
[trivial] Make MemTupleDataCache use standard UTF8 charset

cdeszaq:
This PR is a trivial cleanup, making things more type-safe instead of stringy


QubitPi:
👍 LGTM

PR #271, opened 2017-05-12
Fix incorrect custom header provider conditional

cdeszaq:
Behavior before would do the wrong thing if, somehow, the param was set to empty string because it would try to build a custom provider using empty string as the class name. Now it will not try to build a custom provider if the param is empty string.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
@@ -948,7 +948,7 @@ public abstract class AbstractBinderFactory implements BinderFactory {
     protected Supplier<Map<String, String>> buildDruidWebServiceHeaderSupplier() {
         Supplier<Map<String, String>> supplier = HashMap::new;
         String customSupplierClassString = SYSTEM_CONFIG.getStringProperty(DRUID_HEADER_SUPPLIER_CLASS, null);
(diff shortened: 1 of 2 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -948,7 +948,7 @@ protected DruidWebService buildDruidWebService(DruidServiceConfig druidServiceCo
     protected Supplier<Map<String, String>> buildDruidWebServiceHeaderSupplier() {
         Supplier<Map<String, String>> supplier = HashMap::new;
         String customSupplierClassString = SYSTEM_CONFIG.getStringProperty(DRUID_HEADER_SUPPLIER_CLASS, null);
-        if (customSupplierClassString != null && customSupplierClassString.equals("")) {
+        if (customSupplierClassString != null && !customSupplierClassString.equals("")) {
```
QubitPi: Would it be better to use `!customSupplierClassString.isEmpty()`?

cdeszaq: Probably, yes. But it doesn't simplify the expression, since you'd still need the `null` check. All I was concerned with for this fix was correcting the logic, rather than making the check more sound (I think `isEmpty` looks for al whitespace, of any kind, to consider it as empty)

PR #272, opened 2017-05-12
Suggested refactor using a strategy rather than a subclass to overrid…

michael-mclawhorn:
…e nesting behavior.

Rather than using subclassing of the BinderFactory to override the behavior specifically around JsonNode creation, this approach makes that an injected behavior.

The default is done as an inner class because it really is very intimate with this AsyncImplementation.

The alternate implementation is implementing in a chaining way so that it can supplement any other strategy..


fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/HeaderNestingJsonBuilderStrategy.java
```diff
@@ -0,0 +1,36 @@
...
+     * Constructor.
+     *
+     * @param baseStrategy  strategy to do initial JSON Node construction
+     */
+    public HeaderNestingJsonBuilderStrategy(Function<Response, JsonNode> baseStrategy) {
+        this.baseStrategy = baseStrategy;
+    }
+
+    @Override
+    public JsonNode apply(final Response response) {
```
cdeszaq: `final` not needed

/NotABlocker

cdeszaq (approved):
Nothing worth blocking this, all little formatting trivialities.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
```diff
@@ -57,30 +58,70 @@
...
 
+    public static final Function<Response, JsonNode> DEFAULT_JSON_NODE_BUILDER_STRATEGY =
+            new Function<Response, JsonNode>() {
+
+        @Override
+        public JsonNode apply(final Response response) {
+            try {
+                return new MappingJsonFactory().createParser(response.getResponseBodyAsStream()).readValueAsTree();
+            } catch (IOException ioe) {
+                throw new IllegalStateException(ioe);
```
QubitPi: @cdeszaq Should we also log this?

ghost (approved):

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImpl.java
cdeszaq: Probably doesn't hurt, but since it's about to be thrown, isn't likely to buy us much when the parent catches and logs it.

cdeszaq:
Let's resolve the conflicts here and merge this into the primary feature branch. I think then that branch will be pretty much ready to land.

michael-mclawhorn:
Should I do that or should I let Jiaqi do that?

cdeszaq:
Doesn't really matter who does it. Once it's done, I'll review the target PR that this merges into, and then _it_ can land, so it _might_ be @QubitPi who is more motivated, since the target is his PR ;)

PR #275, opened 2017-05-17
Implement DruidPartialDataResponseProcessor

QubitPi:
3rd PR implementing https://github.com/yahoo/fili/issues/215

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/HeaderNestingJsonBuilderStrategy.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/HeaderNestingJsonBuilderStrategy.java
@@ -2,6 +2,8 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.druid.client.impl;
 
+import com.yahoo.bard.webservice.web.responseprocessors.DruidJsonResponseContentKeys;
+
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.MappingJsonFactory;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
@@ -30,14 +32,21 @@ public class HeaderNestingJsonBuilderStrategy implements Function<Response, Json
 
     @Override
     public JsonNode apply(Response response) {
+        MappingJsonFactory mappingJsonFactory = new MappingJsonFactory();
         ObjectNode objectNode = JsonNodeFactory.instance.objectNode();
-        objectNode.set("response", baseStrategy.apply(response));
+        objectNode.set(DruidJsonResponseContentKeys.RESPONSE.getName(), baseStrategy.apply(response));
         try {
             objectNode.set(
-                    "X-Druid-Response-Context",
-                    new MappingJsonFactory()
-                            .createParser(response.getHeader("X-Druid-Response-Context"))
+                    DruidJsonResponseContentKeys.DRUID_RESPONSE_CONTEXT.getName(),
+                    mappingJsonFactory
+                            .createParser(
+                                    response.getHeader(DruidJsonResponseContentKeys.DRUID_RESPONSE_CONTEXT.getName())
+                            )
                             .readValueAsTree()
+            );
+            objectNode.set(
+                    DruidJsonResponseContentKeys.STATUS_CODE.getName(),
+                    mappingJsonFactory.createParser(String.valueOf(response.getStatusCode())).readValueAsTree()
             );
         } catch (IOException ioe) {
             throw new IllegalStateException(ioe);
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidJsonResponseContentKeys.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidJsonResponseContentKeys.java
@@ -1,0 +1,30 @@
+// Copyright 2017 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.web.responseprocessors;
+
+/**
+ * Enumerates the list of keys expected to be found in the FullResponseProcessor.
+ */
+public enum DruidJsonResponseContentKeys {
+    DRUID_RESPONSE_CONTEXT("X-Druid-Response-Context"),
+    UNCOVERED_INTERVALS("uncoveredIntervals"),
+    UNCOVERED_INTERVALS_OVERFLOWED("uncoveredIntervalsOverflowed"),
+    STATUS_CODE("status-code"),
+    RESPONSE("response")
+    ;
+
+    private final String name;
+
+    /**
+     * Constructor.
+     *
+     * @param name  Name of the context key
+     */
+    DruidJsonResponseContentKeys(String name) {
+        this.name = name;
+    }
+
+    public String getName() {
+        return name;
+    }
+}
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessor.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessor.java
@@ -159,30 +159,30 @@ public class DruidPartialDataResponseProcessor implements FullResponseProcessor 
      */
     private void validateJsonResponse(JsonNode json, DruidAggregationQuery<?> query) {
         if (json.getNodeType() == JsonNodeType.ARRAY) {
-            logAndGetErrorCallback("JSON response is missing X-Druid-Response-Context and status code", query);
+            logAndGetErrorCallback(ErrorMessageFormat.CONTEXT_AND_STATUS_MISSING_FROM_RESPONSE.format(), query);
         }
 
         if (!json.has(DruidJsonResponseContentKeys.DRUID_RESPONSE_CONTEXT.getName())) {
-            logAndGetErrorCallback("JSON response is missing X-Druid-Response-Context", query);
+            logAndGetErrorCallback(ErrorMessageFormat.DRUID_RESPONSE_CONTEXT_MISSING_FROM_RESPONSE.format(), query);
             return;
         }
         JsonNode druidResponseContext = json.get(DruidJsonResponseContentKeys.DRUID_RESPONSE_CONTEXT.getName());
         if (!druidResponseContext.has(DruidJsonResponseContentKeys.UNCOVERED_INTERVALS.getName())) {
             logAndGetErrorCallback(
-                    "JSON response is missing 'uncoveredIntervals' from X-Druid-Response-Context header",
+                    ErrorMessageFormat.UNCOVERED_INTERVALS_MISSING_FROM_RESPONSE.format(),
                     query
             );
             return;
         }
         if (!druidResponseContext.has(DruidJsonResponseContentKeys.UNCOVERED_INTERVALS_OVERFLOWED.getName())) {
             logAndGetErrorCallback(
-                    "JSON response is missing 'uncoveredIntervalsOverflowed' from X-Druid-Response-Context header",
+                    ErrorMessageFormat.UNCOVERED_INTERVALS_OVERFLOWED_MISSING_FROM_RESPONSE.format(),
                     query
             );
             return;
         }
         if (!json.has(DruidJsonResponseContentKeys.STATUS_CODE.getName())) {
-            logAndGetErrorCallback("JSON response is missing response status code", query);
+            logAndGetErrorCallback(ErrorMessageFormat.STATUS_CODE_MISSING_FROM_RESPONSE.format(), query);
         }
     }
 
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessor.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessor.java
@@ -1,0 +1,261 @@
+// Copyright 2017 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.web.responseprocessors;
+
+import com.yahoo.bard.webservice.druid.client.FailureCallback;
+import com.yahoo.bard.webservice.druid.client.HttpErrorCallback;
+import com.yahoo.bard.webservice.druid.model.query.DruidAggregationQuery;
+import com.yahoo.bard.webservice.util.SimplifiedIntervalList;
+import com.yahoo.bard.webservice.web.ErrorMessageFormat;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.JsonNodeType;
+
+import org.joda.time.Interval;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import javax.ws.rs.core.Response.Status;
+
+/**
+ * Response processor for finding missing partial data in Druid
+ * <p>
+ * In druid version 0.9.0 or later, druid implemented a feature that returns missing intervals for a given query.
+ * For example
+ *
+ * <pre>
+ * {@code
+ * Content-Type: application/json
+ * 200 OK
+ * Date:  Mon, 10 Apr 2017 16:24:24 GMT
+ * Content-Type:  application/json
+ * X-Druid-Query-Id:  92c81bed-d9e6-4242-836b-0fcd1efdee9e
+ * X-Druid-Response-Context: {
+ *     "uncoveredIntervals": [
+ *         "2016-11-22T00:00:00.000Z/2016-12-18T00:00:00.000Z","2016-12-25T00:00:00.000Z/2017-
+ *         01-03T00:00:00.000Z","2017-01-31T00:00:00.000Z/2017-02-01T00:00:00.000Z","2017-02-
+ *         08T00:00:00.000Z/2017-02-09T00:00:00.000Z","2017-02-10T00:00:00.000Z/2017-02-
+ *         13T00:00:00.000Z","2017-02-16T00:00:00.000Z/2017-02-20T00:00:00.000Z","2017-02-
+ *         22T00:00:00.000Z/2017-02-25T00:00:00.000Z","2017-02-26T00:00:00.000Z/2017-03-
+ *         01T00:00:00.000Z","2017-03-04T00:00:00.000Z/2017-03-05T00:00:00.000Z","2017-03-
+ *         08T00:00:00.000Z/2017-03-09T00:00:00.000Z"
+ *     ],
+ *     "uncoveredIntervalsOverflowed": true
+ * }
+ * Content-Encoding:  gzip
+ * Vary:  Accept-Encoding, User-Agent
+ * Transfer-Encoding:  chunked
+ * Server:  Jetty(9.2.5.v20141112)
+ * }
+ * </pre>
+ *
+ * The missing intervals are indicated in "uncoveredIntervals". We compare it to the missing intervals that we expects
+ * from Partial Data V1. If "uncoveredIntervals" contains any interval that is not present in our expected
+ * missing interval list, we can send back an error response indicating the mismatch in data availability before the
+ * response is cached.
+ */
+public class DruidPartialDataResponseProcessor implements FullResponseProcessor {
+
+    private static final Logger LOG = LoggerFactory.getLogger(DruidPartialDataResponseProcessor.class);
+
+    private final ResponseProcessor next;
+
+    /**
+     * Constructor.
+     *
+     * @param next  Next ResponseProcessor in the chain
+     */
+    public DruidPartialDataResponseProcessor(ResponseProcessor next) {
+        this.next = next;
+    }
+
+    @Override
+    public ResponseContext getResponseContext() {
+        return next.getResponseContext();
+    }
+
+    @Override
+    public FailureCallback getFailureCallback(DruidAggregationQuery<?> druidQuery) {
+        return next.getFailureCallback(druidQuery);
+    }
+
+    @Override
+    public HttpErrorCallback getErrorCallback(DruidAggregationQuery<?> druidQuery) {
+        return next.getErrorCallback(druidQuery);
+    }
+
+    /**
+     * If status code is 200, do the following
+     *
+     * <ol>
+     *     <li>
+     *         Extract uncoveredIntervalsOverflowed from X-Druid-Response-Context inside the JsonNode passed into
+     *         DruidPartialDataResponseProcessor::processResponse, if it is true, invoke error response saying limit
+     *         overflowed,
+     *     </li>
+     *     <li>
+     *         Extract uncoveredIntervals from X-Druid-Response-Contex inside the JsonNode passed into
+     *         DruidPartialDataResponseProcessor::processResponse,
+     *     </li>
+     *     <li>
+     *         Parse both the uncoveredIntervals extracted above and allAvailableIntervals extracted from the union of
+     *         all the query's datasource's availabilities from DataSourceMetadataService into SimplifiedIntervalLists,
+     *     </li>
+     *     <li>
+     *         Compare both SimplifiedIntervalLists above, if allAvailableIntervals has any overlap with
+     *         uncoveredIntervals, invoke error response indicating druid is missing some data that are we are expecting
+     *         to exists.
+     *     </li>
+     * </ol>
+     *
+     * @param json  The json representing a druid data response
+     * @param query  The query with the schema for processing this response
+     * @param metadata  The LoggingContext to use
+     */
+    @Override
+    public void processResponse(JsonNode json, DruidAggregationQuery<?> query, LoggingContext metadata) {
+        validateJsonResponse(json, query);
+
+        int statusCode = json.get(DruidJsonResponseContentKeys.STATUS_CODE.getName()).asInt();
+        if (statusCode == Status.OK.getStatusCode()) {
+            checkOverflow(json, query);
+
+            SimplifiedIntervalList overlap = getOverlap(json, query);
+            if (!overlap.isEmpty()) {
+                logAndGetErrorCallback(ErrorMessageFormat.DATA_AVAILABILITY_MISMATCH.format(overlap), query);
+            }
+            if (next instanceof FullResponseProcessor) {
+                next.processResponse(json, query, metadata);
+            } else {
+                next.processResponse(json.get(DruidJsonResponseContentKeys.RESPONSE.getName()), query, metadata);
+            }
+        } else if (statusCode == Status.NOT_MODIFIED.getStatusCode() && !(next instanceof FullResponseProcessor)) {
+            logAndGetErrorCallback(
+                    "Content Not Modified(304), but no etag cache response processor is available to process " +
+                            "the 304 response",
+                    query);
+        } else {
+            next.processResponse(json, query, metadata);
+        }
+    }
+
+    /**
+     * Validates JSON response object to make sure it contains all of the following information.
+     * <ul>
+     *     <li>X-Druid-Response-Context
+     *         <ol>
+     *             <li>uncoveredIntervals</li>
+     *             <li>uncoveredIntervalsOverflowed</li>
+     *         </ol>
+     *     </li>
+     *     <li>status-code</li>
+     * </ul>
+     *
+     * @param json  The JSON response that is to be validated
+     * @param query  The query with the schema for processing this response
+     */
+    private void validateJsonResponse(JsonNode json, DruidAggregationQuery<?> query) {
+        if (json.getNodeType() == JsonNodeType.ARRAY) {
+            logAndGetErrorCallback("JSON response is missing X-Druid-Response-Context and status code", query);
+        }
+
+        if (!json.has(DruidJsonResponseContentKeys.DRUID_RESPONSE_CONTEXT.getName())) {
+            logAndGetErrorCallback("JSON response is missing X-Druid-Response-Context", query);
+            return;
+        }
+        JsonNode druidResponseContext = json.get(DruidJsonResponseContentKeys.DRUID_RESPONSE_CONTEXT.getName());
+        if (!druidResponseContext.has(DruidJsonResponseContentKeys.UNCOVERED_INTERVALS.getName())) {
+            logAndGetErrorCallback(
+                    "JSON response is missing 'uncoveredIntervals' from X-Druid-Response-Context header",
+                    query
+            );
+            return;
+        }
+        if (!druidResponseContext.has(DruidJsonResponseContentKeys.UNCOVERED_INTERVALS_OVERFLOWED.getName())) {
+            logAndGetErrorCallback(
+                    "JSON response is missing 'uncoveredIntervalsOverflowed' from X-Druid-Response-Context header",
+                    query
+            );
+            return;
+        }
+        if (!json.has(DruidJsonResponseContentKeys.STATUS_CODE.getName())) {
+            logAndGetErrorCallback("JSON response is missing response status code", query);
+        }
+    }
+
+    /**
+     * Checks and invokes error if the number of missing intervals are overflowed, i.e. more than the configured limit.
+     *
+     * @param json  The json object containing the overflow flag
+     * @param query  The query with the schema for processing this response
+     */
+    private void checkOverflow(JsonNode json, DruidAggregationQuery<?> query) {
+        if (json.get(DruidJsonResponseContentKeys.DRUID_RESPONSE_CONTEXT.getName())
+                .get(DruidJsonResponseContentKeys.UNCOVERED_INTERVALS_OVERFLOWED.getName())
+                .asBoolean()
+        ) {
+            logAndGetErrorCallback(
+                    ErrorMessageFormat.TOO_MANY_INTERVALS_MISSING.format(
+                            query.getContext().getUncoveredIntervalsLimit()
+                    ),
+                    query
+            );
+        }
+    }
+
+    /**
+     * Logs and gets error call back on the response with the provided error message.
+     *
+     * @param message  The error message passed to the logger and the exception
+     * @param query  The query with the schema for processing this response
+     */
+    private void logAndGetErrorCallback(String message, DruidAggregationQuery<?> query) {
+        LOG.error(message);
+        getErrorCallback(query).dispatch(
+                Status.INTERNAL_SERVER_ERROR.getStatusCode(),
+                "The server encountered an unexpected condition which prevented it from fulfilling the request.",
+                message);
+    }
+
+    /**
+     * Returns the overlap between uncoveredIntervals from Druid and missing intervals that Fili expects.
+     *
+     * @param json  The JSON node that contains the uncoveredIntervals from Druid, for example
+     * <pre>
+     * {@code
+     * X-Druid-Response-Context: {
+     *     "uncoveredIntervals": [
+     *         "2016-11-22T00:00:00.000Z/2016-12-18T00:00:00.000Z","2016-12-25T00:00:00.000Z/2017-
+     *         01-03T00:00:00.000Z","2017-01-31T00:00:00.000Z/2017-02-01T00:00:00.000Z","2017-02-
+     *         08T00:00:00.000Z/2017-02-09T00:00:00.000Z","2017-02-10T00:00:00.000Z/2017-02-
+     *         13T00:00:00.000Z","2017-02-16T00:00:00.000Z/2017-02-20T00:00:00.000Z","2017-02-
+     *         22T00:00:00.000Z/2017-02-25T00:00:00.000Z","2017-02-26T00:00:00.000Z/2017-03-
+     *         01T00:00:00.000Z","2017-03-04T00:00:00.000Z/2017-03-05T00:00:00.000Z","2017-03-
+     *         08T00:00:00.000Z/2017-03-09T00:00:00.000Z"
+     *     ],
+     *     "uncoveredIntervalsOverflowed": true
+     * }
+     * }
+     * </pre>
+     * @param query  The Druid query that contains the missing intervals that Fili expects
+     *
+     * @return the overlap between uncoveredIntervals from Druid and missing intervals that Fili expects.
+     */
+    private SimplifiedIntervalList getOverlap(JsonNode json, DruidAggregationQuery<?> query) {
+        List<Interval> intervals = new ArrayList<>();
+        for (JsonNode jsonNode :
+                json.get(DruidJsonResponseContentKeys.DRUID_RESPONSE_CONTEXT.getName())
+                        .get(DruidJsonResponseContentKeys.UNCOVERED_INTERVALS.getName())
+                ) {
+            intervals.add(new Interval(jsonNode.asText()));
+        }
+        SimplifiedIntervalList druidIntervals = new SimplifiedIntervalList(intervals);
+
+        return druidIntervals.intersect(
+                query.getDataSource().getPhysicalTable().getAvailableIntervals()
+        );
+    }
+}
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/FullResponseProcessor.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/FullResponseProcessor.java
@@ -1,0 +1,42 @@
+// Copyright 2017 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.web.responseprocessors;
+
+/**
+ * Response processor for that extracts header information from Druid response and put the information in our own
+ * response.
+ * <p>
+ * This is a "Full" response process in a sense that it extracts and incorporates header information. For example,
+ *
+ * <pre>
+ * {@code
+ * Content-Type: application/json
+ * 200 OK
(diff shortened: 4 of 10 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
```diff
@@ -0,0 +1,100 @@
...
+    public HttpErrorCallback getErrorCallback(DruidAggregationQuery<?> druidQuery) {
+        return next.getErrorCallback(druidQuery);
+    }
+
+    @Override
+    public void processResponse(JsonNode json, DruidAggregationQuery<?> query, LoggingContext metadata) {
+        validateJsonResponse(json);
+
+        int statusCode = json.get("status-code").asInt();
+        if (statusCode == 200) {
```
cdeszaq: We should replace this status code number with one of the StatusCode enums. We do this elsewhere in the codebase I think

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
```diff
@@ -0,0 +1,100 @@
...
+            // implementation is blocked by https://github.com/yahoo/fili/pull/262
```
cdeszaq: If this is still WIP because it's not complete (which this comment seems to indicate), we should add the WIP label. Is this PR still in-progress?

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        next.processResponse(json, query, metadata);
+    }
+
+    /**
+     * Validates JSON response object to make sure it contains all of the following information.
+     * <ul>
+     *     <li>X-Druid-Response-Context</li>
+     *     <ol>
```
cdeszaq: I don't think these list items are nesting the way you intend them to, so I don't think it will render correctly. I _think_ the closing `</li>` with the header name on it should go _after_ the closing `</ol>` for this list.

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
```diff
@@ -0,0 +1,100 @@
...
+     *         <li>uncoveredIntervals</li>
+     *         <li>uncoveredIntervalsOverflowed</li>
+     *     </ol>
+     *     <li>status-code</li>
```
cdeszaq: Why does the status code matter?

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
```diff
@@ -0,0 +1,100 @@
...
+     * </ul>
+     *
+     * @param json  The JSON response that is to be validated
+     */
+    private static void validateJsonResponse(JsonNode json) {
+        if (!json.has("X-Druid-Response-Context")) {
+            logAndThrowRunTimeException("Response is missing X-Druid-Response-Context");
+        }
+        if (!json.get("X-Druid-Response-Context").has("uncoveredIntervals")) {
+            logAndThrowRunTimeException("Response is missing \"uncoveredIntervals\" X-Druid-Response-Context");
```
cdeszaq: Use single quotes rather than escaping the double quotes.

cdeszaq: Also, all of these error messages really should go in the `ErrorMessages` enum (or whatever it's called)

cdeszaq: And lastly, _does_ this exist in the query if there are no uncovered intervals?

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+
+    /**
+     * Logs and throws RuntimeException with the provided error message.
+     *
+     * @param message  The error message passed to the logger and the exception.
+     */
+    private static void logAndThrowRunTimeException(String message) {
+        LOG.error(message);
```
cdeszaq: I _think_ this should just be a warning, and shouldn't thrown an exception. _If_, for some reason, Druid doesn't give us this info, I think this processor should warn, but not assume the result is bad.

@michael-mclawhorn or @garyluoex, do you have any thoughts on this?

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Override
+    public HttpErrorCallback getErrorCallback(DruidAggregationQuery<?> druidQuery) {
+        return next.getErrorCallback(druidQuery);
+    }
+
+    @Override
+    public void processResponse(JsonNode json, DruidAggregationQuery<?> query, LoggingContext metadata) {
+        validateJsonResponse(json);
```
cdeszaq: This doesn't handle the validator throwing an exception. If this response processor decides to fail the request, _it_ needs to handle doing that failure, rather than letting someone else handle it. Likely it should call either the error or failure callback (depending on which makes more sense).

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
QubitPi: @cdeszaq Since this PR is blocked by another PR, my thought is that we could review what we have now to save the reviewing effort latter. As a result, we should be able to land this PR quicker.

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
garyluoex: I think my original intent is to return a 500 error response or similar stating that we are not confident that the data druid gave us match the api user's expectation, after talking with tom, I think we agree that fail the query instead of giving bad result is the way to go.

garyluoex: That said, I don't  know if there is a difference between throwing an exception directly here in comparison to preparing an actual error response by calling`getErrorCallback`? (Since throwing exception means the query failed anyways?) @cdeszaq

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/FullResponseProcessor.java
```diff
@@ -0,0 +1,11 @@
+// Copyright 2017 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.web.responseprocessors;
+
+/**
+ * Response processor for partial data V2 and cache data V3.
+ * <p>
+ * The full response processor also process header information about query response.
+ */
+public interface FullResponseProcessor extends ResponseProcessor {
```
QubitPi: Explanation added

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
QubitPi: Beautiful!

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
garyluoex: It will always exist with 98% confidence, I can double check once I get my Http Requester authentication working again.

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
QubitPi: If status code is 304, do nothing. It status code is 200, we do all logics of partial data V2

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
cdeszaq: Since this PR does not really do anything, there's no point in landing it w/o what it's really supposed to do.

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
cdeszaq: Precisely. The javadoc should say that

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
QubitPi: Yeah. This PR won't merge until everything is there even after review

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
cdeszaq: In general, if a chunk of code can handle an exceptional case, it _should_ handle it, rather than making someone else handle it. The reason for this is that, if you can handle the exception, you likely have more information about what's wrong than someone else does and can provide a better error response. (ie. "something bad happened" is far less useful than "druid indicated more data was missing than we expected" and including the additional missing intervals)

QubitPi: OK. I guess we agree to keep as it is now?

cdeszaq: No, this class needs to handle it's response. It's not useful for it to just `throw` if it has a problem that _should_ fail the request.

cdeszaq: As for the log message, it's OK to leave it `error`

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
cdeszaq: :) I'm all for getting reviews on as much as possible as soon as possible, as long as what's getting reviewed is likely to be stable against the rest of the work.

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
garyluoex: @QubitPi   Fail the response instead of throw an exception, like this [here](https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/WeightCheckRequestHandler.java#L139)

QubitPi: @garyluoex @cdeszaq Thanks!

michael-mclawhorn:
What is the thinking behind having an empty interface and leaving 'v2' in the concrete class name?  Why make an empty interface?  Why not just rename the response processor?

QubitPi:
@cdeszaq @garyluoex @michael-mclawhorn Shall we avoid hardcoded string like https://github.com/yahoo/fili/blob/6fb4b5942a1c9965d81a56032c213624b95389f3/fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java#L180 ? I accessed them multiple times during implementation

garyluoex:
@QubitPi  Lets just make them a static string on the top of the class for now, I do not see a good place to put then in at this point.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplWithHeaderNestingSpec.groovy
```diff
@@ -50,7 +51,8 @@ class AsyncDruidWebServiceImplWithHeaderNestingSpec extends Specification {
                                 "2016-12-25T00:00:00.000Z/2017-01-03T00:00:00.000Z"
                             ],
                             "uncoveredIntervalsOverflowed": true
-                        }
+                        },
+                        "status-code": 200
```
michael-mclawhorn: it looks to me like this is a test of the HeaderNestingJsonBuilderStrategy, not of the webservice itself.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessorSpec.groovy
```diff
@@ -0,0 +1,87 @@
...
+        when:
+        partialDataV2ResponseProcessor.getResponseContext()
+        partialDataV2ResponseProcessor.getErrorCallback(druidAggregationQuery)
+        partialDataV2ResponseProcessor.getFailureCallback(druidAggregationQuery)
+
+        then:
+        1 * next.getResponseContext() >> responseContext
+        1 * next.getFailureCallback(druidAggregationQuery) >> failureCallback
+        1 * next.getErrorCallback(druidAggregationQuery) >> httpErrorCallback
+    }
```
michael-mclawhorn: What is this even testing?  I don't see the point of this.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessorSpec.groovy
```diff
@@ -0,0 +1,87 @@
...
+                "uncoveredIntervals": [
+                    "2016-11-22T00:00:00.000Z/2016-12-18T00:00:00.000Z",
+                    "2016-12-25T00:00:00.000Z/2017-01-03T00:00:00.000Z"
+                ],
+                "uncoveredIntervalsOverflowed": true
+            },
+            "status-code": 200
+        }
+        ''' | ["2016-11-22T00:00:00.000Z/2016-12-18T00:00:00.000Z", "2016-12-25T00:00:00.000Z/2017-01-03T00:00:00.000Z"]
+    }
```
michael-mclawhorn: This is a good test.  However if we're only doing one value, there's not much point in doing a data table.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessorSpec.groovy
```diff
@@ -0,0 +1,87 @@
...
+}
```
michael-mclawhorn: I assume there are more tests coming here.
Specifically: 
processResponse -> overflow message
processResponse -> success message 
processResponse -> data availability mismatch
validateJsonResponse error paths
checkOverflow success and failure
getUncoveredIntervalFromResponse (optional, but might make for good low level testing)

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/HeaderNestingJsonBuilderStrategy.java
```diff
@@ -30,15 +30,20 @@ public HeaderNestingJsonBuilderStrategy(Function<Response, JsonNode> baseStrateg
 
     @Override
     public JsonNode apply(Response response) {
+        MappingJsonFactory mappingJsonFactory = new MappingJsonFactory();
         ObjectNode objectNode = JsonNodeFactory.instance.objectNode();
         objectNode.set("response", baseStrategy.apply(response));
```
QubitPi: Beautiful! I assume the static string you mentioned above is not needed anymore then

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/FullResponseProcessor.java
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.web.responseprocessors;
+
+/**
+ * Response processor for that extracts header information from Druid response and put the information in our own
+ * response.
+ * <p>
+ * This is a "Full" response process in a sense that it extracts and incorporates header information.
+ */
+public interface FullResponseProcessor extends ResponseProcessor {
```
QubitPi: `processResponse` is not found to use `ObjectNode`. Do we still need this change?

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
```diff
@@ -0,0 +1,253 @@
...
+ * Server:  Jetty(9.2.5.v20141112)
+ * }
+ * </pre>
+ *
+ * The flag "uncoveredIntervalsOverflowed" being true indicates that there are more uncovered intervals in addition to
+ * the first 10 included. Using the "uncoveredIntervals", we can compare it to the missing intervals that we expects
+ * from Partial Data V1. If "uncoveredIntervals" contains any interval that is not present in our expected
+ * missing interval list, we can send back an error response indicating the mismatch in data availability before the
+ * response is cached.
+ */
```
QubitPi: Simplified.

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/PartialDataV2ResponseProcessor.java
```diff
@@ -0,0 +1,253 @@
...
+     *     </li>
+     *     <li>status-code</li>
+     * </ul>
+     *
+     * @param json  The JSON response that is to be validated
+     * @param query  The query with the schema for processing this response
+     */
+    private void validateJsonResponse(JsonNode json, DruidAggregationQuery<?> query) {
+        if (!json.has("X-Druid-Response-Context")) {
+            logAndGetErrorCallback("Response is missing X-Druid-Response-Context", query);
```
QubitPi: @garyluoex I add a check, you can see it in the next commit

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/AsyncDruidWebServiceImplWithHeaderNestingSpec.groovy
QubitPi: Yeah. I agree. I think I should rename it to make it correspond to `HeaderNestingJsonBuilderStrategy.java`

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessor.java
```diff
@@ -176,19 +152,29 @@ public void processResponse(JsonNode json, DruidAggregationQuery<?> query, Loggi
...
-        if (!json.has("X-Druid-Response-Context")) {
+        if (json instanceof ArrayNode) {
+            logAndGetErrorCallback("Response is missing X-Druid-Response-Context and status code", query);
+        }
+        if (!json.has(FullResponseContentKeys.DRUID_RESPONSE_CONTEXT.getName())) {
             logAndGetErrorCallback("Response is missing X-Druid-Response-Context", query);
         }
-        if (!json.get("X-Druid-Response-Context").has("uncoveredIntervals")) {
-            logAndGetErrorCallback("Response is missing 'uncoveredIntervals' X-Druid-Response-Context", query);
+        if (!json.get(FullResponseContentKeys.DRUID_RESPONSE_CONTEXT.getName())
```
garyluoex: It is repeated four or more times, we should take it out as its own variable

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/FullResponseContentKeys.java
```diff
@@ -0,0 +1,29 @@
+// Copyright 2017 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.web.responseprocessors;
+
+/**
+ * Enumerates the list of keys expected to be found in the FullResponseProcessor.
+ */
+public enum FullResponseContentKeys {
```
garyluoex: Lets change it to `DruidJsonResponseContentKeys`?

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/FullResponseProcessor.java
```diff
@@ -39,4 +47,21 @@
  *
  */
 public interface FullResponseProcessor extends ResponseProcessor {
+    @Override
+    default void processResponse(JsonNode json, DruidAggregationQuery<?> query, LoggingContext metadata) {
```
garyluoex: Why do we have a default?

QubitPi: The only way to override is to implement `processResponse` and `default` is needed to implement this method

QubitPi: An alternative is to tranform JSON node to Object node in `DruidPartialDataResponseProcessor`

garyluoex: Are you sure? I think just overriding the method as an interface without default should be fine. Let me know the compilation error you get if possible.

garyluoex: never mind, revert it

QubitPi: ok

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/HeaderNestingJsonBuilderStrategy.java
garyluoex: Use DruidJsonResponseContentKeys

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/HeaderNestingJsonBuilderStrategy.java
```diff
@@ -30,15 +30,20 @@ public HeaderNestingJsonBuilderStrategy(Function<Response, JsonNode> baseStrateg
...
         try {
             objectNode.set(
                     "X-Druid-Response-Context",
```
garyluoex: DruidJsonResponseContentKeys

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/HeaderNestingJsonBuilderStrategy.java
```diff
@@ -30,15 +30,20 @@ public HeaderNestingJsonBuilderStrategy(Function<Response, JsonNode> baseStrateg
...
-                    new MappingJsonFactory()
+                    mappingJsonFactory
                             .createParser(response.getHeader("X-Druid-Response-Context"))
                             .readValueAsTree()
             );
+            objectNode.set(
+                    "status-code",
```
garyluoex: DruidJsonResponseContentKeys

QubitPi: Beautiful!

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidJsonResponseContentKeys.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Enumerates the list of keys expected to be found in the FullResponseProcessor.
+ */
+public enum DruidJsonResponseContentKeys {
+    DRUID_RESPONSE_CONTEXT("X-Druid-Response-Context"),
+    UNCOVERED_INTERVALS("uncoveredIntervals"),
+    UNCOVERED_INTERVALS_OVERFLOWED("uncoveredIntervalsOverflowed"),
+    STATUS_CODE("status-code"),
+    RESPONSE("response");
```
garyluoex: semi colon to next line

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessorSpec.groovy
```diff
@@ -0,0 +1,280 @@
...
+
+        DataSource dataSource = Mock(DataSource)
+        ConstrainedTable constrainedTable = Mock(ConstrainedTable)
+
+        constrainedTable.getAvailableIntervals() >> new SimplifiedIntervalList(
+                availableIntervals.collect{it -> new Interval(it)}
+        )
+        dataSource.getPhysicalTable() >> constrainedTable
+        druidAggregationQuery.getDataSource() >> dataSource
+
```
garyluoex: extra new line

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessorSpec.groovy
```diff
@@ -0,0 +1,280 @@
...
+        expect:
+        druidPartialDataResponseProcessor.next == next;
+    }
+
+    def "getOverlap returns intersection between Druid intervals and Fili intervals in case of #caseDescription"() {
+        given:
+        JsonNode json = MAPPER.readTree(
+                jsonInString
+                        .replace(" ", "")
+                        .replace("\n", "")
```
garyluoex: why are these two replace needed?

QubitPi: Good catch! They are not needed after a refactor but I forgot to delete them

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessor.java
```diff
@@ -0,0 +1,257 @@
...
+                logAndGetErrorCallback(ErrorMessageFormat.DATA_AVAILABILITY_MISMATCH.format(overlap), query);
+            }
+            if (next instanceof FullResponseProcessor) {
+                next.processResponse(json, query, metadata);
+            } else {
+                next.processResponse(json.get(DruidJsonResponseContentKeys.RESPONSE.getName()), query, metadata);
+            }
+        }
+
+        next.processResponse(json, query, metadata);
```
garyluoex: If it is 304 and the next response processor is not fullResponseProcessor, we should throw error saying, received 304, but no etag cache response processor is available to process 304 response

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessorSpec.groovy
```diff
@@ -0,0 +1,280 @@
...
+        jsonInString | availableIntervals | expected | caseDescription
+        '''
+        {
+            "response": [{"k1":"v1"}],
+            "X-Druid-Response-Context": {
+                "uncoveredIntervals": [
+                    "2016-11-22T00:00:00.000Z/2016-12-18T00:00:00.000Z",
+                    "2016-12-25T00:00:00.000Z/2017-01-03T00:00:00.000Z"
+                ],
+                "uncoveredIntervalsOverflowed": true
```
garyluoex: Why are all over flow set to true????

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessor.java
```diff
@@ -0,0 +1,257 @@
...
+     *
+     * @return the overlap between uncoveredIntervals from Druid and missing intervals that Fili expects.
+     */
+    private SimplifiedIntervalList getOverlap(JsonNode json, DruidAggregationQuery<?> query) {
+        List<Interval> intervals = new ArrayList<>();
+        for (JsonNode jsonNode :
+                json.get(DruidJsonResponseContentKeys.DRUID_RESPONSE_CONTEXT.getName())
+                        .get(DruidJsonResponseContentKeys.UNCOVERED_INTERVALS.getName())
+                ) {
+            intervals.add(new Interval(jsonNode.asText()));
```
garyluoex: Can we add a test showing what this `jsonNode` is? Or let me know if you know the answer, is this jsonNode a arraynode or objectnode and what is the content?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessorSpec.groovy
garyluoex: And.... this test only have a give: ????

QubitPi: OK. Setting it to false makes more sense

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessor.java
QubitPi: According to Javadoc of this method

```
X-Druid-Response-Context: {
    "uncoveredIntervals": [
        "2016-11-22T00:00:00.000Z/2016-12-18T00:00:00.000Z","2016-12-25T00:00:00.000Z/2017-
        01-03T00:00:00.000Z","2017-01-31T00:00:00.000Z/2017-02-01T00:00:00.000Z","2017-02-
        08T00:00:00.000Z/2017-02-09T00:00:00.000Z","2017-02-10T00:00:00.000Z/2017-02-
        13T00:00:00.000Z","2017-02-16T00:00:00.000Z/2017-02-20T00:00:00.000Z","2017-02-
        22T00:00:00.000Z/2017-02-25T00:00:00.000Z","2017-02-26T00:00:00.000Z/2017-03-
        01T00:00:00.000Z","2017-03-04T00:00:00.000Z/2017-03-05T00:00:00.000Z","2017-03-
        08T00:00:00.000Z/2017-03-09T00:00:00.000Z"
    ],
    "uncoveredIntervalsOverflowed": true
}
```

Each is a ObjectNode containing a String like `"2016-11-22T00:00:00.000Z/2016-12-18T00:00:00.000Z"`

garyluoex: If it is a ObjectNode, it is some kind of key value mapping right? So what is the key for each individual string like `"2016-11-22T00:00:00.000Z/2016-12-18T00:00:00.000Z"`?

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessor.java
```diff
@@ -131,9 +132,14 @@ public void processResponse(JsonNode json, DruidAggregationQuery<?> query, Loggi
             } else {
                 next.processResponse(json.get(DruidJsonResponseContentKeys.RESPONSE.getName()), query, metadata);
             }
+        } else if (statusCode == Status.NOT_MODIFIED.getStatusCode() && next instanceof FullResponseProcessor) {
```
garyluoex: Are you sure this if statement is correct?

QubitPi: Maybe adding a pair of parenthesis around latter?

garyluoex: if 304, and the next response Processor is FullResponseProcessor (which can be etagCacheResponseProcessor), we throw an error?

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessor.java
QubitPi: What I mean by ObjectNode is that it is a type of JsonNode that can be serialized to text in this way

garyluoex: Can you show me some documentation or a test that confirms this behavior?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessorSpec.groovy
```diff
@@ -0,0 +1,275 @@
...
+
+        druidPartialDataResponseProcessor = new DruidPartialDataResponseProcessor(next)
+    }
+
+    def "Test constructor"() {
+        expect:
+        druidPartialDataResponseProcessor.next == next;
+    }
+
+    def "getOverlap returns intersection between Druid intervals and Fili intervals in case of #caseDescription"() {
```
garyluoex: This test only have a given: and where: which does not test anything right?

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/HeaderNestingJsonBuilderStrategySpec.groovy
```diff
@@ -50,7 +51,8 @@ class AsyncDruidWebServiceImplWithHeaderNestingSpec extends Specification {
                                 "2016-12-25T00:00:00.000Z/2017-01-03T00:00:00.000Z"
                             ],
                             "uncoveredIntervalsOverflowed": true
-                        }
+                        },
+                        "status-code": 200
                     }
                 '''.replace(" ", "").replace("\n", "")
```
garyluoex: Not sure if this is needed.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/client/impl/HeaderNestingJsonBuilderStrategySpec.groovy
```diff
@@ -28,6 +28,7 @@ class AsyncDruidWebServiceImplWithHeaderNestingSpec extends Specification {
                         "uncoveredIntervalsOverflowed": true
                     }
                 '''.replace(" ", "").replace("\n", "")
```
garyluoex: Not sure if this is needed.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessorSpec.groovy
QubitPi: Hmm...somehow a refactor removes it. Anyway it's back

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessor.java
QubitPi: I modified tests to make sure this method works as expected

michael-mclawhorn: Isn't this a text node inside a list?

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessor.java
```diff
@@ -0,0 +1,261 @@
...
+     *     <li>status-code</li>
+     * </ul>
+     *
+     * @param json  The JSON response that is to be validated
+     * @param query  The query with the schema for processing this response
+     */
+    private void validateJsonResponse(JsonNode json, DruidAggregationQuery<?> query) {
+        if (json.getNodeType() == JsonNodeType.ARRAY) {
+            logAndGetErrorCallback("JSON response is missing X-Druid-Response-Context and status code", query);
+        }
```
michael-mclawhorn: All of these error strings should be moved into constants and referenced by the constants in the tests so that they'll stay in sync.

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessor.java
```diff
@@ -0,0 +1,261 @@
...
+        if (!druidResponseContext.has(DruidJsonResponseContentKeys.UNCOVERED_INTERVALS_OVERFLOWED.getName())) {
+            logAndGetErrorCallback(
+                    "JSON response is missing 'uncoveredIntervalsOverflowed' from X-Druid-Response-Context header",
+                    query
+            );
+            return;
+        }
+        if (!json.has(DruidJsonResponseContentKeys.STATUS_CODE.getName())) {
+            logAndGetErrorCallback("JSON response is missing response status code", query);
+        }
```
michael-mclawhorn: All of these error strings should be moved into constants and referenced by the constants in the tests so that they'll stay in sync.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessorSpec.groovy
```diff
@@ -0,0 +1,258 @@
...
+        druidPartialDataResponseProcessor.checkOverflow(json, druidAggregationQuery)
+
+        then:
+        1 * httpErrorCallback.dispatch(
+                500,
+                'The server encountered an unexpected condition which prevented it from fulfilling the request.',
+                "Query is returning more than the configured limit of '10' missing intervals. " +
+                        "There may be a problem with your data.")
+    }
+
```
michael-mclawhorn: Use the constants for these errors rather than copying them into the tests.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessorSpec.groovy
```diff
@@ -0,0 +1,258 @@
...
+        {
+            "response": [{"k1":"v1"}],
+            "X-Druid-Response-Context": {
+                "uncoveredIntervals": [
+                    "2016-11-22T00:00:00.000Z/2016-12-18T00:00:00.000Z",
+                    "2016-12-25T00:00:00.000Z/2017-01-03T00:00:00.000Z"
+                ],
+                "uncoveredIntervalsOverflowed": false
+            },
+            "status-code": 200
```
michael-mclawhorn: It looks like your Json instring is the same across all the test cases. If so, lift it up out of the where clause and put it into the test.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessorSpec.groovy
QubitPi: Beautiful!

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessor.java
QubitPi: Beautiful!

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessor.java
QubitPi: Beautiful!

QubitPi:
@cdeszaq Since merging is blocked, could you take a quick look at it and approve it? Thanks!

garyluoex:
@QubitPi @cdeszaq   I dismissed the comment so that merging is not blocked since it's got 2 reviews already.

QubitPi:
@garyluoex Thanks Gary!

Issue #255, opened 2017-04-24
RFC: Cache V3


garyluoex:

cdeszaq:
Some off-line design notes:

![img_0382](https://cloud.githubusercontent.com/assets/612084/25394716/6a15467e-29a4-11e7-8a3a-8075ff505cb5.JPG)

Note: Hooks up with [Partial Data v2](215) as well.

QubitPi:
Good Design! 👍

michael-mclawhorn:
Clarifying history a little:
Cache v1 was backed by broker metadata.
Cache v2 uses the coordinator as the metadata source for query signing.
Cache v3 as a feature should probably be named 'eTagCache' or 'eTagSignedCache'.  As Rick pointed out on the PartialData v2 PRs, v2 isn't going to mean anything outside of our team and so we should probably go with meaningful names.

continued...

michael-mclawhorn:
eTagCaching requires a change in the DruidWebService to generate what I guess we're calling a 'FullResponse'.  However, writing the FullResponse from cache has to happen before any response delegation can happen in the ResponseProcessing chain.

So, rather than a response processor, we may need to either have a custom DruidWebService or a DruidWebService that can take a strategy for generating it's `FullResponse` (either attaching a cache of the body to the Full response in place of the body that comes back, or failing if no such body exists in cache.)

michael-mclawhorn:
Either way, there may be no value to having a 'ResponseProcessor' formally involved here because cache resolution has to happen FIRST on a 304 and it has to happen before any other ResponseProcessors can be called, and there's not much point in storing to cache later than when you'd pull from cache.

PR #289, opened 2017-05-19
Prepare for etag Cache

QubitPi:
The 1st PR implementing https://github.com/yahoo/fili/issues/255


michael-mclawhorn:
Per discussion on 5/22 change v3 references to be 'etag'

QubitPi:
@michael-mclawhorn Thanks for the reminder!

CHANGELOG.md
```diff
@@ -9,6 +9,13 @@ Current
 -------
 ### Added:
 
+- [Prepare for Etage Cache](https://github.com/yahoo/fili/pull/289)
+    * Add Etag Cache feature flag
```
michael-mclawhorn: spelling error on etag

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/workflow/DruidWorkflow.java
```diff
@@ -100,14 +100,12 @@ public DataRequestHandler buildWorkflow() {
...
-            } else {
-                uiHandler = new CacheRequestHandler(uiHandler, dataCache, mapper);
-                nonUiHandler = new CacheRequestHandler(nonUiHandler, dataCache, mapper);
-            }
+        if (BardFeatureFlag.DRUID_CACHE_V2.isOn()) {
+            uiHandler = new CacheV2RequestHandler(uiHandler, dataCache, querySigningService, mapper);
+            nonUiHandler = new CacheV2RequestHandler(nonUiHandler, dataCache, querySigningService, mapper);
+        } else {
+            uiHandler = new CacheRequestHandler(uiHandler, dataCache, mapper);
+            nonUiHandler = new CacheRequestHandler(nonUiHandler, dataCache, mapper);
```
michael-mclawhorn: Shouldn't the else here be NO cache request handler rather than the v1 cache request handler?

fili-core/src/test/resources/testApplicationConfig.properties
```diff
@@ -51,6 +51,9 @@ bard__druid_dim_loader_timer_delay = 60000
 # Data Cache V2 (needs the above flag set as well)
 bard__druid_cache_v2_enabled = false
 
+# Etag Cache (needs the "bard__druid_dim_loader_timer_delay" flag set as well)
+bard__druid_etag_cache_enabled = false
```
michael-mclawhorn: This comment, why is that?

QubitPi: @michael-mclawhorn Could you be more specific about your question? :)

QubitPi:
@michael-mclawhorn Regarding your question about both caches, I think they work together with no error

fili-core/src/test/resources/testApplicationConfig.properties
michael-mclawhorn: Why does etag cache need bard__druid_dim_loader_timer_delay to be set?

QubitPi: Because I noticed a comment on `bard__druid_cache_v2_enabled = false` :)

garyluoex: only either etag cache or cache v2 is allowed, not both

garyluoex:
only either etag cache or cache v2 is allowed, not both @michael-mclawhorn

fili-core/src/test/resources/testApplicationConfig.properties
QubitPi: Thanks. I've corrected it

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/HeaderNestingJsonBuilderStrategy.java
```diff
@@ -30,15 +38,27 @@ public HeaderNestingJsonBuilderStrategy(Function<Response, JsonNode> baseStrateg
...
                     "X-Druid-Response-Context",
-                    new MappingJsonFactory()
+                    mappingJsonFactory
                             .createParser(response.getHeader("X-Druid-Response-Context"))
                             .readValueAsTree()
             );
+            objectNode.set(
+                    "status-code",
+                    mappingJsonFactory.createParser(String.valueOf(response.getStatusCode())).readValueAsTree()
+            );
```
michael-mclawhorn: constants for magic strings like 'status-code'

QubitPi: @michael-mclawhorn Sure. This will happen once https://github.com/yahoo/fili/pull/275 is merged.

michael-mclawhorn: so this will pop out on rebase?

QubitPi: Yes

fili-wikipedia-example/src/test/resources/userConfig.properties.sample
```diff
@@ -6,6 +6,7 @@ bard__dimension_backend = memory
 
 bard__druid_cache_enabled = true
```
garyluoex: Do a check in AbstractBinderFactory for this flag, if set log a warning.

fili-generic-example/src/test/resources/userConfig.properties.sample
```diff
@@ -6,6 +6,7 @@ bard__dimension_backend = memory
 
 bard__druid_cache_enabled = true
 bard__druid_cache_v2_enabled = true
+bard__druid_etag_cache_enabled = false
```
garyluoex: Reason why this is false while the other is true in test?

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/cache/MemcachedCacheSpec.groovy
```diff
@@ -160,7 +160,7 @@ class MemcachedCacheSpec extends Specification {
             .request().get(String.class)
 
         then: "should read old result from cache"
-        GroovyTestUtils.compareJson(result, expected)
+        GroovyTestUtils.compareJson(result, '{"rows":[]}')
```
garyluoex: What caused the change in behavior?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/handlers/workflow/DruidWorkflowSpec.groovy
```diff
@@ -98,8 +97,7 @@ class DruidWorkflowSpec extends Specification {
 
         when:
         handlers = getHandlerChain(defaultHandler.nonUiWebServiceHandler.next)
-        isCaching = handlers.find(byClass(CacheRequestHandler)) != null |
-                handlers.find(byClass(CacheV2RequestHandler)) != null
+        isCaching = handlers.find(byClass(CacheV2RequestHandler)) != null
```
garyluoex: I think we should do find by class CacheV1RequestHandler == null

fili-wikipedia-example/src/test/resources/userConfig.properties.sample
QubitPi: Did you mean this? https://github.com/yahoo/fili/blob/ef959dab04967fd417c39fc49937654bbe47c8b7/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java#L313

fili-generic-example/src/test/resources/userConfig.properties.sample
QubitPi: Set to true

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/cache/MemcachedCacheSpec.groovy
QubitPi: "Deprecate CacheV1 components and remove from DruidWorkflow" from https://github.com/yahoo/fili/issues/255

screen shot 2017-05-31 at 11 57 20 am

fili-wikipedia-example/src/test/resources/userConfig.properties.sample
garyluoex: We need one for cache also.

fili-core/src/main/java/com/yahoo/bard/webservice/config/BardFeatureFlag.java
```diff
@@ -7,8 +7,9 @@
  */
 public enum BardFeatureFlag implements FeatureFlag {
     PARTIAL_DATA("partial_data_enabled"),
-    DRUID_CACHE("druid_cache_enabled"),
+    @Deprecated DRUID_CACHE("druid_cache_enabled"),
     DRUID_CACHE_V2("druid_cache_v2_enabled"),
```
michael-mclawhorn: I think Druid cache v2 will still be supported for now because we cannot count on fili users being on a particular version of druid and we haven't yet got a back-end-version strategy for config other than continuing to support necessary features.

In the case of V2, all the interesting druid versions support it.  In the case of eTag support, I'm not sure which versions support it, but not all of the ones since our launch.

We can deprecate it, but not until we have a clear policy around which versions of druid are supported.

michael-mclawhorn:
I promise a proper review shortly.

fili-core/src/main/java/com/yahoo/bard/webservice/config/BardFeatureFlag.java
garyluoex: @QubitPi  collapse the cache config parameter into one called `druid_query_response_caching_strategy`, this parameter takes a string of one of the following: `none`, `v2`. `etag`. Put comment on the config parameter.

cdeszaq: Per conversation yesterday, the values for the new caching strategy config param:
- ETag
- LocalSignature
- TtlOnly
- None

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/workflow/DruidWorkflow.java
```diff
@@ -115,14 +114,9 @@ public DataRequestHandler buildWorkflow() {
         }
 
         // If query caching is enabled, the cache is checked before sending the request
-        if (BardFeatureFlag.DRUID_CACHE.isOn()) {
-            if (BardFeatureFlag.DRUID_CACHE_V2.isOn()) {
-                uiHandler = new CacheV2RequestHandler(uiHandler, dataCache, querySigningService, mapper);
-                nonUiHandler = new CacheV2RequestHandler(nonUiHandler, dataCache, querySigningService, mapper);
-            } else {
-                uiHandler = new CacheRequestHandler(uiHandler, dataCache, mapper);
```
cdeszaq: If we're only deprecating cache v1, rather than removing it, we should leave this conditional here and log a warning if someone uses it, rather than removing it.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/cache/MemcachedCacheSpec.groovy
cdeszaq: Per my comment above about removal vs. deprecation, this behavior shouldn't be changing yet.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/workflow/CachingMode.java
```diff
@@ -0,0 +1,95 @@
...
+    private static final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance();
+    private static final int DRUID_UNCOVERED_INTERVAL_LIMIT = SYSTEM_CONFIG.getIntProperty(
+            SYSTEM_CONFIG.getPackageVariableName("druid_uncovered_interval_limit"),
+            0
+    );
+    private static final String NO_CACHING = "No caching";
+    private static final String TTL_CACHING = "TTL caching only";
+    private static final String LOCAL_SIGNATURE_CACHING = "Local signature caching only";
+    private static final String ETAG_CACHING = "Etag caching only";
+
```
michael-mclawhorn: I would like to suggest short, no spaces, and camelized names for these property values.  We may want to put more than one into a property at some point and this certainly is verbose for hand editting in a properties file.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/workflow/CachingMode.java
```diff
@@ -0,0 +1,95 @@
...
+        } else if (CACHE_V1 && !CACHE_V2 && DRUID_UNCOVERED_INTERVAL_LIMIT <= 0) {
+            return TTL_CACHING;
+        } else if (!CACHE_V1 && CACHE_V2 && DRUID_UNCOVERED_INTERVAL_LIMIT <= 0) {
+            return LOCAL_SIGNATURE_CACHING;
+        } else if (!CACHE_V1 && !CACHE_V2 && DRUID_UNCOVERED_INTERVAL_LIMIT > 0) {
+            return ETAG_CACHING;
+        } else if ((CACHE_V1 || CACHE_V2) && DRUID_UNCOVERED_INTERVAL_LIMIT > 0) {
+            String message = "Etag caching cannot coexist with TTL or local signature caching";
+            LOG.error(message);
+            throw new RuntimeException(message);
```
michael-mclawhorn: Why is the druid uncovered limit relevant here?

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/workflow/CachingMode.java
```diff
@@ -0,0 +1,95 @@
...
+
+        /**
+         * Returns the caching mode in String representation.
+         *
+         * @return the caching mode in String representation
+         */
+        public String getMode() {
+            return mode;
+        }
+    }
```
michael-mclawhorn: Ok, here's what I'd like to suggest as an alternative.  
Make a DefaultCachingMode (or DefaultCacheMode) (or DefaultDruidCacheMode) enum which implements a CachingMode (or CacheMode) interface.

* Use the camelized names of the Enum values as your names for the CacheModes.

* I don't think  DRUID_UNCOVERED_INTERVAL_LIMIT is the name for the cacheMode property in config.  There should be an explicit property which has one of the cache mode values.

* Make a protected 'getCacheMode' method in the AbstractBinderFactory so people can extend their cache mode selection.  
* Make the method you have at CachingMode.getCachingMode() take a system configuration so you don't have to pull all the config values out statically.  
*  Move the validation to the top of the method, so that the error case doesn't get disguised by an earlier match.
* Make the method return an Optional<CacheMode>, and don't have it return NoCache, but rather an empty optional instead and have the method in AbstractBinder Factory unpack the Optional defaulting to NoCache.  This way if people want to write their own CacheModes, they can cascade from the defaultCacheMode selection (checking for empty optional) and continue on into their own cache mode resolution.

fili-core/src/main/resources/moduleConfig.properties
```diff
@@ -60,6 +60,13 @@ bard__druid_max_response_length_to_cache = 9223372036854775807
 # Data Cache V2 (needs the above flag set as well)
 bard__druid_cache_v2_enabled = true
 
+# Caching strategy, which is one of the following
+# 1. ETag
+# 2. LocalSignature (cache V2)
+# 3. TtlOnly (cache V1)
+# 4. None (no caching)
+druid_query_response_caching_strategy = None
```
michael-mclawhorn: Do these values get used somewhere?

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/workflow/CachingMode.java
michael-mclawhorn: It looks like the property you meant to use was: druid_query_response_caching_strategy

fili-wikipedia-example/src/main/resources/applicationConfig.properties
```diff
@@ -19,6 +19,7 @@ bard__dimension_backend=memory
 # Data Cache
 bard__druid_cache_enabled = false
 bard__druid_cache_v2_enabled = false
+bard__druid_etag_cache_enabled = false
```
michael-mclawhorn: We shouldn't be adding this is we're only allowing new or old config.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -754,6 +755,7 @@ protected final void setupDruidDimensionsLoader(
                 throw new IllegalStateException(e);
             }
         } else if (BardFeatureFlag.DRUID_CACHE.isOn()) {
+            LOG.warn("Cache V1 has been deprecated and is no longer supported. Cache V2 is the default cache now.");
```
garyluoex: Instead we should say something like, this feature flag is deprecated, use the new configuration parameter to set desired caching strategy.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/workflow/CacheMode.java
```diff
@@ -0,0 +1,94 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Optional;
+
+/**
+ * Caching config analyzer.
+ */
+public abstract class CacheMode {
```
garyluoex: I think abstract class does not a whole lot of sense here. There are several options, 1. make it final class 2. make it a class and remove static modifier on getCacheMode to allow for override.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/workflow/CacheMode.java
```diff
@@ -0,0 +1,94 @@
...
+    private static final String LOCAL_SIGNATURE_CACHE = "LocalSignature";
+    private static final String ETAG_CACHE = "ETag";
+
+    /**
+     * Returns the caching mode in String.
+     *
+     * @return the caching mode in String
+     */
+    public static Optional<String> getCacheMode() {
+        if (CACHE_V1 && !CACHE_V2 && QUERY_RESPONSE_CACHING_STRATEGY.equals(NO_CACHE)) {
```
garyluoex: I'm pretty sure we can re-organize these if logics to something more organized, and use nested if statement to reduce number of if statements? for example take out QUERY_RESPONSE_CACHING_STRATEGY.equals(NO_CACHE) to a top level if statement

garyluoex: also, we don't need else statements if there are return statements already.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/workflow/CacheMode.java
```diff
@@ -0,0 +1,94 @@
...
+    private static final boolean CACHE_V2 = BardFeatureFlag.DRUID_CACHE_V2.isOn();
+    private static final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance();
+    private static final String QUERY_RESPONSE_CACHING_STRATEGY = SYSTEM_CONFIG.getStringProperty(
+            SYSTEM_CONFIG.getPackageVariableName("query_response_caching_strategy"),
+            "None"
+    );
+    private static final String NO_CACHE = "None";
+    private static final String TTL_CACHE = "TtlOnly";
+    private static final String LOCAL_SIGNATURE_CACHE = "LocalSignature";
+    private static final String ETAG_CACHE = "ETag";
```
garyluoex: I would simply put these as hardcoded string inside enum, these variable serve very little purpose.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/workflow/CacheMode.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private static final boolean CACHE_V1 = BardFeatureFlag.DRUID_CACHE.isOn();
+    private static final boolean CACHE_V2 = BardFeatureFlag.DRUID_CACHE_V2.isOn();
+    private static final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance();
+    private static final String QUERY_RESPONSE_CACHING_STRATEGY = SYSTEM_CONFIG.getStringProperty(
+            SYSTEM_CONFIG.getPackageVariableName("query_response_caching_strategy"),
+            "None"
+    );
+    private static final String NO_CACHE = "None";
+    private static final String TTL_CACHE = "TtlOnly";
```
garyluoex: why is it called TTL?

QubitPi: @michael-mclawhorn @cdeszaq ?

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -754,6 +755,9 @@ protected final void setupDruidDimensionsLoader(
                 throw new IllegalStateException(e);
             }
         } else if (BardFeatureFlag.DRUID_CACHE.isOn()) {
+            LOG.warn("Cache V1 feature flag is deprecated," +
+                    "use the new configuration parameter to set desired caching strategy"
+            );
```
michael-mclawhorn: You should check whether or not the the underlying system property is set, not whether the feature flag is on.  And you should check both parameters.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/workflow/CacheMode.java
```diff
@@ -0,0 +1,108 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Optional;
+
+/**
+ * Caching config analyzer.
+ */
+public final class CacheMode {
```
michael-mclawhorn: I really wanted you to make CacheMode an interface and what you call Mode inside to be 'DefaultCacheMode'.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/workflow/CacheMode.java
```diff
@@ -0,0 +1,108 @@
...
+            if (CACHE_V1.isEmpty()) {
+                return Optional.of(Mode.LOCAL_SIGNATURE);
+            } else if (CACHE_V2.isEmpty()) {
+                return Optional.of(Mode.TTL_ONLY);
+            } else if ("true".equalsIgnoreCase(CACHE_V1) && "true".equalsIgnoreCase(CACHE_V2)) {
+                return Optional.of(Mode.LOCAL_SIGNATURE);
+            } else if ("true".equalsIgnoreCase(CACHE_V1) &&
+                    "false".equalsIgnoreCase(CACHE_V2)) {
+                return Optional.of(Mode.TTL_ONLY);
+            } else {
```
michael-mclawhorn: You can use the feature flag 'isOn' methods here.

fili-core/src/test/resources/testApplicationConfig.properties
```diff
@@ -51,6 +51,13 @@ bard__druid_dim_loader_timer_delay = 60000
 # Data Cache V2 (needs the above flag set as well)
 bard__druid_cache_v2_enabled = false
 
+# Caching strategy, which is one of the following
+# 1. ETag
+# 2. LocalSignature (cache V2)
+# 3. TtlOnly (cache V1)
+# 4. None (no caching)
+druid__query_response_caching_strategy = None
+
```
michael-mclawhorn: We discussed if caching strategy is set, then cache_enabled and cache_v2_enabled would be removed.

fili-generic-example/src/main/resources/applicationConfig.properties
```diff
@@ -21,6 +21,13 @@ bard__dimension_backend=memory
 bard__druid_cache_enabled = false
 bard__druid_cache_v2_enabled = false
 
+# Caching strategy, which is one of the following
+# 1. ETag
+# 2. LocalSignature (cache V2)
+# 3. TtlOnly (cache V1)
+# 4. None (no caching)
+bard__query_response_caching_strategy = None
+
```
michael-mclawhorn: Ditto

fili-generic-example/src/test/resources/userConfig.properties.sample
```diff
@@ -7,6 +7,13 @@ bard__dimension_backend = memory
 bard__druid_cache_enabled = true
 bard__druid_cache_v2_enabled = true
 
+# Caching strategy, which is one of the following
+# 1. ETag
+# 2. LocalSignature (cache V2)
+# 3. TtlOnly (cache V1)
+# 4. None (no caching)
+bard__query_response_caching_strategy = LocalSignature
+
```
michael-mclawhorn: Ditto

fili-wikipedia-example/src/main/resources/applicationConfig.properties
```diff
@@ -20,6 +20,13 @@ bard__dimension_backend=memory
 bard__druid_cache_enabled = false
 bard__druid_cache_v2_enabled = false
 
+# Caching strategy, which is one of the following
+# 1. ETag
+# 2. LocalSignature (cache V2)
+# 3. TtlOnly (cache V1)
+# 4. None (no caching)
+bard__query_response_caching_strategy = None
+
```
michael-mclawhorn: Ditto

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/workflow/CacheMode.java
```diff
@@ -0,0 +1,17 @@
...
+ * Cache Mode represents a strategy for Druid caching.
+ */
+public interface CacheMode {
+    /**
+     * Returns true if a particular cache mode is set.
+     *
+     * @return true if a particular cache mode is set
+     */
+    Optional<Boolean> isSet();
+}
```
michael-mclawhorn: This optional isn't exactly what @cdeszaq and I were talking about adding 'isSet' we were actually thinking of adding it to the FeatureFlag interface so you could determine whether or not it was true.  Given that this 'isSet' is always defaulted to false, it could simply return boolean rather than optional boolean.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -754,6 +757,9 @@ protected final void setupDruidDimensionsLoader(
                 throw new IllegalStateException(e);
             }
         } else if (BardFeatureFlag.DRUID_CACHE.isOn()) {
+            LOG.warn("Cache V1 feature flag is deprecated, " +
+                    "use the new configuration parameter to set desired caching strategy"
+            );
```
michael-mclawhorn: what about binding if cache v3 is set via the new configuration setting?  Under this code, the StubCache would be built, which I think is wrong.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -744,31 +745,76 @@ protected final void setupDruidDimensionsLoader(
...
-        } else if (BardFeatureFlag.DRUID_CACHE.isOn()) {
-            try {
-                DataCache<String> cache = new HashDataCache<>(new MemDataCache<HashDataCache.Pair<String, String>>());
-                LOG.info("MemcachedClient started {}", cache);
-                return cache;
-            } catch (IOException e) {
-                LOG.error("MemcachedClient failed to start {}", e);
-                throw new IllegalStateException(e);
+        } else { // new caching strategy
+            if (CacheFeatureFlag.TTL.isOn()) {
```
michael-mclawhorn: This is unreachable if the above if isn't changed.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -744,31 +745,76 @@ protected final void setupDruidDimensionsLoader(
...
 
     /**
+     * Builds and returns an instance of local signature cache.
+     *
+     * @return the instance of local signature cache
+     */
+    private DataCache<?> buildLocalSignatureCache() {
+        LOG.warn("Cache V2 feature flag is deprecated, " +
+                "use the new configuration parameter to set desired caching strategy"
+        );
```
michael-mclawhorn: We're no longer necessarily coming here via a set Cache V2 feature flag.  We may be coming here via the new, non deprecated, config.  LocalSignature and TTL could the live for a while yet.   You should only give this warning if the old configs are set.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -744,31 +745,76 @@ protected final void setupDruidDimensionsLoader(
...
+    /**
+     * Builds and returns an instance of TTL cache.
+     *
+     * @return the instance of TTL cache
+     */
+    private DataCache<?> buildTtlCache() {
+        LOG.warn("Cache V1 feature flag is deprecated, " +
+                "use the new configuration parameter to set desired caching strategy"
+        );
+        try {
```
michael-mclawhorn: Same comment as v2 above.

fili-core/src/main/java/com/yahoo/bard/webservice/config/CacheFeatureFlag.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    public String getName() {
+        return propertyName;
+    }
+
+    @Override
+    public boolean isOn() {
+        // TODO: Remove this if conditional after cache V1 & V2 are removed
+        if (BardFeatureFlag.DRUID_CACHE.isSet() || BardFeatureFlag.DRUID_CACHE_V2.isSet()) {
```
michael-mclawhorn: Update comment 'after cache V1 & V2 configuration flags are removed'

fili-core/src/main/java/com/yahoo/bard/webservice/config/CacheFeatureFlag.java
```diff
@@ -0,0 +1,65 @@
...
+            return (this.value.equals("Ttl") && !BardFeatureFlag.DRUID_CACHE_V2.isOn())
+                    || (this.value.equals("LocalSignature") && BardFeatureFlag.DRUID_CACHE_V2.isOn());
+        }
+        return QUERY_RESPONSE_CACHING_STRAGEGY.equalsIgnoreCase(value);
+    }
+
+    @Override
+    public void setOn(Boolean newValue) {
+        return;
```
michael-mclawhorn: You could actually implement 'setOn' by manipulating SystemConfig.  It might be a bad idea, though.  

However, since anyone attempting to do this is NOT getting what they want done, you should absolutely log an error here before doing nothing.

fili-core/src/main/java/com/yahoo/bard/webservice/config/CacheFeatureFlag.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Override
+    public boolean isOn() {
+        // TODO: Remove this if conditional after cache V1 & V2 are removed
+        if (BardFeatureFlag.DRUID_CACHE.isSet() || BardFeatureFlag.DRUID_CACHE_V2.isSet()) {
+            return (this.value.equals("Ttl") && !BardFeatureFlag.DRUID_CACHE_V2.isOn())
+                    || (this.value.equals("LocalSignature") && BardFeatureFlag.DRUID_CACHE_V2.isOn());
+        }
+        return QUERY_RESPONSE_CACHING_STRAGEGY.equalsIgnoreCase(value);
```
michael-mclawhorn: There's a third path through the above boolean.  Someone has CacheV1 flag explicitly set to false (so it triggers is set), in which case None is a legal value here.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -744,31 +745,76 @@ protected final void setupDruidDimensionsLoader(
...
-        if (BardFeatureFlag.DRUID_CACHE_V2.isOn()) {
-            try {
-                MemTupleDataCache<String> cache = new MemTupleDataCache<>();
-                LOG.info("MemcachedClient Version 2 started {}", cache);
-                return cache;
-            } catch (IOException e) {
-                LOG.error("MemcachedClient Version 2 failed to start {}", e);
-                throw new IllegalStateException(e);
+        if (CacheFeatureFlag.TTL.isOn() || CacheFeatureFlag.LOCAL_SIGNATURE.isOn()) { // old caching strategy
+            if (CacheFeatureFlag.LOCAL_SIGNATURE.isOn()) {
```
michael-mclawhorn: This defaulting behavior is handled in the CacheFlags below.  You can simply do a set of flag if/else's here.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
QubitPi: TTL and localSignature could be set in new config or old config. I think there is no way to tell which config is used. In addition, LocalSignature and TTL are all marked as deprecated, which means they should always be warned, I think?

fili-core/src/main/java/com/yahoo/bard/webservice/config/CacheFeatureFlag.java
QubitPi: @michael-mclawhorn Right, in this case V1Flat.isSet() returns false, which seems like a right return value? If we want to check for NoCache, we can call None.isSet() again.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
michael-mclawhorn: You can check via 'isSet' on the FeatureFlag

fili-core/src/main/java/com/yahoo/bard/webservice/config/CacheFeatureFlag.java
```diff
@@ -0,0 +1,73 @@
...
+                    && !BardFeatureFlag.DRUID_CACHE.isOn()
+            )
+                    || (this.value.equals("LocalSignature") && BardFeatureFlag.DRUID_CACHE_V2.isOn());
+        }
+        return QUERY_RESPONSE_CACHING_STRAGEGY.equalsIgnoreCase(value);
+    }
+
+    @Override
+    public void setOn(Boolean newValue) {
+        LOG.warn("com.yahoo.bard.webservice.config.CacheFeatureFlag#setOn(Boolean) does not apply in " +
```
garyluoex: Do you thin the long package name is necessary?

fili-core/src/main/java/com/yahoo/bard/webservice/config/FeatureFlag.java
```diff
@@ -6,6 +6,9 @@
  * Feature flags bind an object to a system configuration name.
  */
 public interface FeatureFlag {
+
+    SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance();
```
garyluoex: No need to define a variable here maybe? I would personally prefer put it in the default method. And if we do happen to choose a not so common practice of putting a variable here, it should be accessible by all that implement this interface and therefore no need to define their own `SYSTEM_CONFIG` it seems.

fili-core/src/main/java/com/yahoo/bard/webservice/config/CacheFeatureFlag.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public boolean isOn() {
+        // TODO: Remove this if conditional after cache V1 & V2 configuration flags are removed
+        if (BardFeatureFlag.DRUID_CACHE.isSet() || BardFeatureFlag.DRUID_CACHE_V2.isSet()) {
+            return (this.value.equals("Ttl")
+                    && !BardFeatureFlag.DRUID_CACHE_V2.isOn()
+                    && !BardFeatureFlag.DRUID_CACHE.isOn()
+            )
+                    || (this.value.equals("LocalSignature") && BardFeatureFlag.DRUID_CACHE_V2.isOn());
```
garyluoex: We need to consider the case where `DRUID_CACHE` is not set but `DRUID_CACHE_V2` is set to true case for `LocalSignature` cache.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -744,31 +745,70 @@ protected final void setupDruidDimensionsLoader(
...
+        }
+    }
+
+    /**
+     * Builds and returns an instance of eTag cache.
+     *
+     * @return the instance of eTag cache
+     */
+    private DataCache<?> buildETagCahe() {
+        return buildTtlCache();
```
garyluoex: We need to have a implementation of the data cache that we can use for the `EtagCacheResponseProcessor` i.e. with the signature of <String, String, String>

QubitPi: Sounds reasonable. I'm adding a new data cache in this PR

```java
public class eTagDataCache<V extends Serializable> implements TupleDataCache<String, String, String>
```

fili-core/src/main/java/com/yahoo/bard/webservice/config/CacheFeatureFlag.java
```diff
@@ -0,0 +1,73 @@
...
+    ETAG("query_response_caching_strategy", "ETag");
+
+    private static final Logger LOG = LoggerFactory.getLogger(CacheFeatureFlag.class);
+    private static final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance();
+    private static final String QUERY_RESPONSE_CACHING_STRAGEGY = SYSTEM_CONFIG.getStringProperty(
+            SYSTEM_CONFIG.getPackageVariableName("query_response_caching_strategy"),
+            "None"
+    );
+
+    private final String propertyName;
```
QubitPi: As I discussed with @michael-mclawhorn , it might become useful in future

fili-core/src/main/java/com/yahoo/bard/webservice/config/FeatureFlag.java
QubitPi: can do that

fili-core/src/main/java/com/yahoo/bard/webservice/config/CacheFeatureFlag.java
michael-mclawhorn: I think perhaps there's a missed case here.  Perhaps a ternary would express it better:

CacheV1   CacheV2  Expected
True           True         LocalSignature
True            False       TTL
False           True        LocalSignature
False           False       None

michael-mclawhorn: Probably that exact table should be in the where of a test.

fili-core/src/main/java/com/yahoo/bard/webservice/config/CacheFeatureFlag.java
```diff
@@ -0,0 +1,79 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Binds a caching strategy to a system configuration name.
+ */
+public enum CacheFeatureFlag implements FeatureFlag {
+    /**
+     * No cache.
+     */
+    NONE("query_response_caching_strategy", "NoCache"),
```
michael-mclawhorn: yeah, let's refactor the reused magic strings into constants.

fili-core/src/main/java/com/yahoo/bard/webservice/config/CacheFeatureFlag.java
michael-mclawhorn: I'm going to change my vote on this, let's pull it out.  Clearly it's not needed now.

fili-core/src/test/groovy/com/yahoo/bard/webservice/config/CacheFeatureFlagSpec.groovy
```diff
@@ -0,0 +1,41 @@
...
+        CacheFeatureFlag.NONE.isOn() == NoCacheOn
+        CacheFeatureFlag.TTL.isOn() == ttlOn
+        CacheFeatureFlag.LOCAL_SIGNATURE.isOn() == localSignatureOn
+        CacheFeatureFlag.ETAG.isOn() == eTagOn
+
+        where:
+        ttlIsOn | localSignatureIsOn | NoCacheOn | ttlOn  | localSignatureOn  | eTagOn
+        true    | false              | false     | true   | false             | false
+        false   | true              | false     | false  | true              | false
+        true    | true               | false     | false  | true              | false
```
michael-mclawhorn: Per outside dicsussion, I got the legacy requirements wrong.

false/true should be no cache, not local signature.

fili-generic-example/src/main/resources/applicationConfig.properties
```diff
@@ -17,9 +17,12 @@ bard__fili_port=9998
...
 
-# Data Cache
-bard__druid_cache_enabled = false
-bard__druid_cache_v2_enabled = false
+# Data Cache strategy, which is one of the following
+# 1. ETagOnly
+# 2. LocalSignatureOnly (cache V2)
+# 3. TtlOnly (cache V1)
+# 4. None (no caching)
+bard__query_response_caching_strategy = None
```
garyluoex: `None` or `NoCache`? Need to fix all the comments on all the .properties files also.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/impl/HeaderNestingJsonBuilderStrategy.java
```diff
@@ -48,6 +49,16 @@ public JsonNode apply(Response response) {
                     DruidJsonResponseContentKeys.STATUS_CODE.getName(),
                     mappingJsonFactory.createParser(String.valueOf(response.getStatusCode())).readValueAsTree()
             );
+            if (CacheFeatureFlag.ETAG.isOn()) {
+                objectNode.set(
+                        DruidJsonResponseContentKeys.ETAG.getName(),
+                        mappingJsonFactory
+                                .createParser(
+                                        response.getHeader(DruidJsonResponseContentKeys.ETAG.getName())
```
garyluoex: There is a bug here.

QubitPi: What do you think is wrong?

garyluoex: `If-None-Match` is never present on the response, in addition to it not being present. It the etag related header only present when we receive a 200, there will be no etag related header in response in 300.

michael-mclawhorn: https://github.com/yahoo/fili/issues/255
Look at the headers for Cache hit and Cache miss responses.
We don't always get the Etag header back, and If-None-Match is not the response header for etags.

PR #309, opened 2017-05-26
Fix DruidNavigator loading multiple tables

kevinhinterlong:
It appears the locking mechanism was forced to time out while loading
multiple tables because the lock for getting all the tables was still active.

Now only the lock for getting all the tables (`loadAllDatasources()`) is used.

### DIFF
--- a/fili-generic-example/README.md
+++ b/fili-generic-example/README.md
@@ -25,7 +25,10 @@ mvn -pl fili-generic-example exec:java
 - Note that if your setup is different you can adjust it by changing the default parameters below
 
     ```bash
-    mvn -pl fili-generic-example exec:java -Dbard__fili_port=9998 -Dbard__druid_coord=http://localhost:8081/druid/coordinator/v1
+    mvn -pl fili-generic-example exec:java -Dbard__fili_port=9998 \
+    -Dbard__druid_coord=http://localhost:8081/druid/coordinator/v1 \
+    -Dbard__non_ui_druid_broker=http://localhost:8082/druid/v2 \
+    -Dbard__ui_druid_broker=http://localhost:8082/druid/v2
     ```
 
 From another window, run a test query against the default druid data.
--- a/fili-generic-example/README.md
+++ b/fili-generic-example/README.md
@@ -1,9 +1,15 @@
 Fili Generic Loader Application
 ==================================
 
-This application will automatically configure fili to work with **any** instance of Druid and show the basic metrics and dimensions. This lets you test what it's like using Fili without putting any effort into setting it up.
+This application will automatically configure fili to work with **any** instance
+ of Druid and show the basic metrics and dimensions. This lets you test what it's
+  like using Fili without putting any effort into setting it up.
 
-In order to set up, this will connect to druid at  [http://localhost:8081/druid/coordinator/v1](http://localhost:8081/druid/coordinator/v1). If your set up is different, you'll have to change the `bard__druid_coord` url in `applicationConfig.properties`.
+In order to set up, this will connect to druid at  [http://localhost:8081/druid/coordinator/v1](http://localhost:8081/druid/coordinator/v1).
+ If your set up is different, you'll have to change the `bard__druid_coord`,
+  `bard__non_ui_druid_broker`, `bard__ui_druid_broker` url in `applicationConfig.properties`.
+  
+Note that this was last tested using [version 0.9.1](https://github.com/yahoo/fili/tree/0.9.1)
 
 ## Setup and Launching
 
--- a/fili-generic-example/README.md
+++ b/fili-generic-example/README.md
@@ -15,12 +15,11 @@ In order to set up, this will connect to druid at  [http://localhost:8081/druid/
     ```
 3. Use Maven to install and launch the Fili Generic example:
 
-
-```bash
-cd fili
-mvn install
-mvn -pl fili-generic-example exec:java
-```
+    ```bash
+    cd fili
+    mvn install
+    mvn -pl fili-generic-example exec:java
+    ```
 
 - Note that if your setup is different you can adjust it by changing the default parameters below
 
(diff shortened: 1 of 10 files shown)


cdeszaq (dismissed):
A few things here:

1. Changelog entry needed (probably in the `Fixes` section)
2. It would be nice to have a test in place that shows the issue and that this change fixes it.
3. I'm not sure I understand how this is fixing the problem. The main change I see is to move the `get` to the caller. I don't really understand how that solves the problem which, as I understand it, is that the follow-on calls are trying to run while the 1st call isn't finished. Given how the AsyncHttpClient works, multiple calls in flight at once shouldn't be a problem. And there's no lock involved (other than the implicit one in `Future::get`), so I'm a little confused there as well.
4. Don't we still want to have the later calls timeout or at least finish executing before the rest of the application starts up? I think this change would allow the app to start w/o having any of the tables configured...

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/dimension/GenericDimensionConfigs.java
```diff
@@ -17,39 +17,46 @@
...
+import java.util.HashMap;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 /**
  * Hold all the dimension configurations for a generic druid configuration.
  */
 public class GenericDimensionConfigs {
-    private final Set<DimensionConfig> dimensionConfigs;
+    private final Map<String, Set<DimensionConfig>> dimensionConfigs;
```
michael-mclawhorn: It's not clear what the String in this map is.

fili-generic-example/src/test/groovy/com/yahoo/wiki/webservice/web/endpoints/MultipleDatasourceDruidConfigLoaderSpec.groovy
```diff
@@ -0,0 +1,80 @@
...
+        druidWebService = new TestDruidWebService("testInstance")
+        druidWebService.jsonResponse = {
+            if (druidWebService.lastUrl.equals("/datasources/")) {
+                return new ObjectMapper().writeValueAsString(datasources)
+            } else if (druidWebService.lastUrl.equals("/datasources/table1/?full")) {
+                return getFullTable(datasources[0], table1_metrics, table1_dimensions)
+            } else if (druidWebService.lastUrl.equals("/datasources/table2/?full")) {
+                return getFullTable(datasources[1], table2_metrics, table2_dimensions)
+            }
+            println "No response for " + druidWebService.lastUrl
```
QubitPi: Maybe you forgot to delete it?

fili-generic-example/src/test/groovy/com/yahoo/wiki/webservice/web/endpoints/AutomaticDruidConfigLoaderSpec.groovy
```diff
@@ -53,7 +52,7 @@ public class AutomaticDruidConfigLoaderSpec extends Specification {
             } else if (druidWebService.lastUrl == '/datasources/' + datasource + "/?full") {
                 return expectedMetricsAndDimensions
             }
-            return "Unexpected URL"
+            println "No response for " + druidWebService.lastUrl
```
QubitPi: Maybe you forgot to delete this?

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/dimension/GenericDimensionConfigs.java
```diff
@@ -17,39 +17,46 @@
...
+                                    )
+                            ).collect(
+                                    Collectors.collectingAndThen(
+                                            Collectors.toSet(),
+                                            Collections::unmodifiableSet
+                                    ));
+
+                    dimensionConfigs.put(dataSourceConfiguration.getName(), tableDimensionConfigs);
+                });
+        ;
```
QubitPi: What's this for?

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/dimension/GenericDimensionConfigs.java
```diff
@@ -58,7 +65,21 @@ public GenericDimensionConfigs(Supplier<List<? extends DataSourceConfiguration>>
...
+        return dimensionConfigs.values().stream()
+                .flatMap(Set::stream)
+                .distinct()
+                .collect(Collectors.toSet());
+    }
+
+    /**
+     * Get all the dimension configurations associated with this datasource.
+     *
+     * @param dataSourceConfiguration the datasource configuration's dimensions to load
```
QubitPi: Two spaces before description. Capitalize `the`

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/table/GenericTableLoader.java
```diff
@@ -21,29 +21,23 @@
...
-
 import org.joda.time.DateTimeZone;
 
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Set;
+import javax.validation.constraints.NotNull;
+import java.util.*;
```
QubitPi: Avoid start import

fili-generic-example/src/test/groovy/com/yahoo/wiki/webservice/web/endpoints/MultipleDatasourceDruidConfigLoaderSpec.groovy
```diff
@@ -0,0 +1,80 @@
...
+                    "type": "none"
+                },
+            "binaryVersion": 9,
+            "size": 5537610,
+            "identifier": "${name}_2015-09-12T00:00:00.000Z_2015-09-13T00:00:00.000Z_2017-02-27T03:06:09.422Z"
+        }
+    ]
+}
+
+"""
```
QubitPi: Align indentation

ghost (changes requested):

michael-mclawhorn (approved):
Looks pretty good

ghost (approved):

cdeszaq (commented):
Partial review. Didn't get as far as `GenericDimensionConfigs`.

fili-generic-example/README.md
```diff
@@ -90,4 +95,6 @@ Here are some sample queries that you can run to verify your server:
 
 1. In IntelliJ, go to `File -> Open`
 2. Select the `pom.xml` file at the root of the project
-3. Run `GenericMain` which can be found in `fili-generic-example` (e.g. right click and choose run)
+3. Under `src/main/resources/applicationConfig.properties`, change `bard__non_ui_druid_broker`, 
+`bard__ui_druid_broker`, `bard__druid_coord`, and other properties as needed.
```
kevinhinterlong: Yep these are set correctly for running it locally and they work with the wikipedia walkthrough. 

I added the following note in `README.md` to clarify:

> **NOTE:** if you're running this locally and haven't changed any settings (like the Wikipedia example) you can **skip step 3**.

kevinhinterlong:
I added a note in the readme of the version this was last tested with. 
I think this should be good to merge now.

garyluoex (approved):
Rebase, and Im ok with it

Issue #306, opened 2017-05-25
Add availability to Table resource


cdeszaq:
Currently the tables/table/timeGrain response lacks an indication of the time range(s) for which the table has data (ie. can answer questions). Having this data would be super useful when it comes to building valid or useful queries.

The baseline ask here is to simply expose the intervals for which the logical table could answer _at least 1_ question. 

However, there is a richer capability that would be nice to have, which is to expand the capabilities of the Table resource so that it can take pretty much the same inputs that the Data resource takes, and would use those inputs to constrain / restrict the available time ranges (and even available schema, etc.) of the logical table.

The relevant components from the Data resource that we'd want to re-use are:

- Grouping dimensions
- Metrics
- Filters
- Date / Time (possibly)

The way I see this working is that Table requests could now look like this:

```
/tables/myTable/week/dim1/dim2?metrics=myMetric&filters=dim3|id-in[foo,bar]
```

Which would result in a table response with the metrics, dimensions, and available intervals restricted down to the set of items that are still "reachable" given the constraints in the query (`dim1`, `dim2`, `dim3`, and `myMetric`, in this case). So, if the table normally indicates that `dim7` is one of it's dimensions, but there isn't a backing physical table for `myTable` that has `dim1`, `dim2`, `dim3`, and `myMetric` along with `dim7`, then `dim7` would not be in the dimension list returned in the response. <sub>(If this prose description is too hard to follow, let me know and I can put together sample unconstrained and constrained responses along with some sample config info)</sub>

I think the bulk of this is already pretty doable, since it's really just doing part of the same work that the Data servlet is doing, except it's stopping once the satisfying tables have been identified. The metrics, dimensions, and availabilities from these tables are then unioned, and filtered down against the configuration of the Logical table.

PR #310, opened 2017-05-26
RFC for Issue #306

michael-mclawhorn:
https://github.com/yahoo/fili/issues/306

### DIFF
--- a/rfc/tableAvailabilityMarkdown.md
+++ b/rfc/tableAvailabilityMarkdown.md
@@ -106,3 +106,15 @@ Add `availability` as a list of serialized intervals to the default table format
       ]
     }
 ```
+
+## Implementation
+
+### 1st Milestone - Table-wide Availability
+In `getLogicalTableFullView`, take the TablesApiRequest to calculate the availability for the logical table (calculate
+directly on logical table(union the availability for the logical table) without taking the TablesApiRequest into account)
+
+### 2nd Milestone - Have Tables Endpoint Support (but not use) Additional Query Parameters
+Make the availability consider the TablesApiRequest by passing it into the `getLogicalTableFullView` method
+
+### 3rd Milestone - Filter Using Dimensions, Metrics, and Filters
+Have TablesApiRequest accept metrics and dimensions from the API request
--- a/rfc/tableAvailabilityMarkdown.md
+++ b/rfc/tableAvailabilityMarkdown.md
@@ -1,0 +1,108 @@
+# Table resource supporting filtered availability
+
+This RFC implements [Github Issue 306](https://github.com/yahoo/fili/issues/306)
+
+## Goals
+
+1. Show the queryable range of time as part of the Logical Table endpoint view.
+2. Allow the queryable range to reflect constraints as would be found in a data request.
+  * Required metrics and dimensions
+  * Applied filters
+
+
+## Calculating availability
+
+Availability for a logical table shall be defined as the maximal set of intervals that the table may be able to respond
+ to.  Or put another way, the table can certainly not respond to any intervals beyond the range marked available.  As
+ such the calculation will involve unioning the available intervals for all physical tables backing the logical table.
+ 
+ Constrained availability will be defined as the availability for a table given a set of query constraints.  This will
+ be the result of first filtering out physical tables on the logical table which cannot answer the query.  This may end
+ up excluding ALL physical tables, meaning an empty availability.
+ 
+ Existing code for constrained physical tables should be able to be repurposed:
+
+1. Build a `QueryPlanningConstraint` from the TableApiRequest
+    From: `DruidQueryBuilder`
+    ```java 
+        TableGroup group = logicalTable.getTableGroup();
+    
+        // Resolve the table from the the group, the combined dimensions in request, and template time grain
+        QueryPlanningConstraint constraint = new QueryPlanningConstraint(request, template);
+    
+    ```
+    * In the case of unconstrained queries, ensure that an 'unconstrained' constraint is built.
+    * `QueryPlanningConstraint` will need more constructor richness to handle making metric constraints optional
+    
+2.  Get the tables from the `LogicalTable`'s `tableGroup`.
+3.  Use matchers to filter the tables from the table group given the `QueryPlanningConstraint`
+From: `DefaultPhysicalTableResolver`
+    ```java
+        public List<PhysicalTableMatcher> getMatchers(QueryPlanningConstraint requestConstraint) {
+            return Arrays.asList(
+                    new SchemaPhysicalTableMatcher(requestConstraint),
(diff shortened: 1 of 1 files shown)


archolewa:
If we're going to do this, then we may want to add an additional `ApiRequest`, `FilterableApiRequest` that is an `ApiRequest` with filters, that the `Dimensions`, `Data` and `Table` requests can all extend from. As soon as you add filters to a request, there opens up the doors to being able to add implicit filters. Having to either duplicate the implicit filter for each concrete request, or making heavy use of `instanceof` and casting are mildly annoying with two filterable requests. Adding a third makes it even worse.

michael-mclawhorn:
@archolewa That makes sense.  

One motivating desire to leverage the query planning capabilities of a PartitionedTable by applying ApiFilters to the query should go under the same query rewrite restrictions.

We could roll this out with no api support for filters to simplify security concerns.  But internally we definitely want to be able to attach filters.  It's possible, in conjunction with your suggestion, that the right thing to do would be to have the ApiRequest do a lookup for any pertinent security filters from a registry, and rewrite itself with any filters that match it.

michael-mclawhorn:
e.g. If it's an APIRequest that accepts FilterValidation, it should get all security rules that target FilterValidation and map through each of them.

We really need to figure out the default security model sooner than later.

michael-mclawhorn:
@archolewa I've got a change about to be in a PR where I create an additional ChainingApiRequestMapper to support richer security support.
This isn't directly pertinent to your recommendation about commonality on 'has filter' being in the type structure of the ApiRequest, but I thought you might want to know, as you have the largest current use of transforming ApiRequests.

rfc/tableAvailabilityMarkdown.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+The table metadata endpoint should now accept an optional list of path separated grouping dimensions, an optional list
+of metrics, and an optional filter clause.
+
+Code currently in `DataApiRequest` will be shared with `TablesApiRequest` to support parsing these elements.
+
+## Response Format Changes:
+
+Add `availability` as a list of serialized intervals to the default table format.
```
cdeszaq: Is this the label for the new field? It doesn't match the example...

cdeszaq:
This looks pretty good to me. Here are the general thoughts I have:

1. We should leverage as much of the existing code in DataApiRequest as we can. And rather than inheritance, we should look more at composing I think. We're already pretty inherit-centric in many places and it's caused more pain than I think we anticipated as we've tried to change things. Perhaps moving the common logic into helper utilities would get us enough separation and re-use?

2. Along with code reuse, I think leaning on interfaces for "this thing can have filters added to it via a  `with*` method is probably a better route than inheriting from a common parent.

3. We should definitely figure out a general security model / framework, starting with the needs. I'll try to start a discussion in the dev group and likely an issue as well. But let's not hold this effort here up on not having a stronger sense of security needs.

cdeszaq (approved):

QubitPi:
I can implement this RFC.

rfc/tableAvailabilityMarkdown.md
```diff
@@ -0,0 +1,108 @@
+# Table resource supporting filtered availability
+
+This RFC implements [Github Issue 306](https://github.com/yahoo/fili/issues/306)
+
+## Goals
+
+1. Show the queryable range of time as part of the Logical Table endpoint view.
+2. Allow the queryable range to reflect constraints as would be found in a data request.
+  * Required metrics and dimensions
+  * Applied filters
```
aklish: Do we really need filters?  It seems like the only information they convey are a set of dimensions that are included in the filter request.  The API to return availability seems odd having to specify data filters.  Does the operator matter?  What about the filter values?  Would I get a different result back if I use startsWith vs contains?

michael-mclawhorn: The filters can impact availability if dimension /values/ impact the query planning.  (right now that's only a concern on physical tables with dimension value partitions)

rfc/tableAvailabilityMarkdown.md
```diff
@@ -0,0 +1,108 @@
...
+1. Build a `QueryPlanningConstraint` from the TableApiRequest
+    From: `DruidQueryBuilder`
+    ```java 
+        TableGroup group = logicalTable.getTableGroup();
+    
+        // Resolve the table from the the group, the combined dimensions in request, and template time grain
+        QueryPlanningConstraint constraint = new QueryPlanningConstraint(request, template);
+    
+    ```
+    * In the case of unconstrained queries, ensure that an 'unconstrained' constraint is built.
```
QubitPi: How do you know if a query is unconstrained?

ghost (commented):
Looks like 1. 2. 3. have already been implemented. Are we implementing a set of new classes that utilizes those?

rfc/tableAvailabilityMarkdown.md
michael-mclawhorn: By unconstrainted I mean that it's constrained only by fields specified in the ApiRequest.  So if the ApiRequest doesn't list dimensions, no dimensional constraints are applied, and so forth.

michael-mclawhorn:
I'm not sure where you mean 1 is implemented.  Can you clarify where the code in question is?

QubitPi:
By 1, 2, 3, I mean 

**1**. Build a `QueryPlanningConstraint` from the TableApiRequest
    From: `DruidQueryBuilder`
    ```java 
        TableGroup group = logicalTable.getTableGroup();
    
        // Resolve the table from the the group, the combined dimensions in request, and template time grain
        QueryPlanningConstraint constraint = new QueryPlanningConstraint(request, template);
    
    ```
    * In the case of unconstrained queries, ensure that an 'unconstrained' constraint is built.
    * `QueryPlanningConstraint` will need more constructor richness to handle making metric constraints optional
    
**2**.  Get the tables from the `LogicalTable`'s `tableGroup`.
**3**.  Use matchers to filter the tables from the table group given the `QueryPlanningConstraint`
From: `DefaultPhysicalTableResolver`
    ```java
        public List<PhysicalTableMatcher> getMatchers(QueryPlanningConstraint requestConstraint) {
            return Arrays.asList(
                    new SchemaPhysicalTableMatcher(requestConstraint),
                    new AggregatableDimensionsMatcher(requestConstraint),
                    new TimeAlignmentPhysicalTableMatcher(requestConstraint)
            );
        }
    ```

    From: `BasePhysicalTableResolver`
    ```java
        public Set<PhysicalTable> filter(
                Collection<PhysicalTable> candidateTables,
                QueryPlanningConstraint requestConstraint
        ) throws NoMatchFoundException {
            return filter(candidateTables, getMatchers(requestConstraint));
        }
    ```

michael-mclawhorn:
This was implemented

PR #312, opened 2017-06-01
Implement etag cache request handler

QubitPi:
The 3rd PR implementing https://github.com/yahoo/fili/issues/255


michael-mclawhorn:
Isn't the context stripped from the request before caching?

michael-mclawhorn:
Let me clarify, is the etag in the context of the cached document?  I thought it was going to be the signature on the cache value.  I thought for normalization purposes the context got stripped for caching.

michael-mclawhorn:
It looks like I am wrong about that.  nevermind.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/EtagCacheRequestHandler.java
```diff
@@ -0,0 +1,174 @@
...
+
+                            return true;
+
+                        } catch (IOException e) {
+                            LOG.warn("Error processing cached value: ", e);
+                        }
+                    } else {
+                        context
+                                .getHeaders()
+                                .putSingle(DruidJsonResponseContentKeys.ETAG.getName(), "non-existing-etag");
```
cdeszaq: Shouldn't this be happening if there is a cache miss, not if there's a signature miss-match? I think the logic is wrong here around what happens for a cache hit vs. a cache miss.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/EtagCacheRequestHandler.java
```diff
@@ -0,0 +1,174 @@
...
+                            if (context.getNumberOfOutgoing().decrementAndGet() == 0) {
+                                RequestLog.record(new BardQueryInfo(druidQuery.getQueryType().toJson(), true));
+                                RequestLog.stopTiming(REQUEST_WORKFLOW_TIMER);
+                            }
+
+                            if (context.getNumberOfIncoming().decrementAndGet() == 0) {
+                                RequestLog.startTiming(RESPONSE_WORKFLOW_TIMER);
+                            }
+
+                            CACHE_HITS.mark(1);
```
cdeszaq: Shouldn't we be counting hits based on what Druid comes back with? (ie. it's not a cache hit if we have a stale entry)

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/EtagCacheRequestHandler.java
```diff
@@ -0,0 +1,174 @@
...
+                                            .get(DruidJsonResponseContentKeys.ETAG.getName())
+                                            .asText()
+                            );
+                    // Make sure that if the optional return value is empty, the statement always evaluates to false
+                    // Metadata type needs to be int.
+                    if (
+                            querySigningService.getSegmentSetId(druidQuery)
+                                    .map(id -> Objects.equals(cacheEntry.getMeta(), id))
+                                    .orElse(false)
+                            ) {
```
cdeszaq: Why are we looking at the locally-computed signature for this? Why not use _only_ the etags?

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/EtagCacheRequestHandler.java
```diff
@@ -0,0 +1,174 @@
...
+                                RequestLog.startTiming(RESPONSE_WORKFLOW_TIMER);
+                            }
+
+                            CACHE_HITS.mark(1);
+                            RequestLog logCtx = RequestLog.dump();
+                            nextResponse.processResponse(
+                                    mapper.readTree(cacheEntry.getValue()),
+                                    druidQuery,
+                                    new LoggingContext(logCtx)
+                            );
```
cdeszaq: This doesn't look right. Since the eTag cache mechanism can only determine a cache hit if it asks Druid, it can't trigger the response processing unless it goes to Druid, so this handler should _never_ do this, right?

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/EtagCacheRequestHandler.java
```diff
@@ -0,0 +1,174 @@
...
+    ) {
+        ResponseProcessor nextResponse = response;
+
+        String cacheKey = null;
+        try {
+            cacheKey = getKey(druidQuery);
+
+            if (context.isReadCache()) {
+                final TupleDataCache.DataEntry<String, Long, String> cacheEntry = dataCache.get(cacheKey);
+                CACHE_REQUESTS.mark(1);
```
michael-mclawhorn: Long is now out of sync with the other PR

michael-mclawhorn:
@QubitPi I will review this after @cdeszaq 's  comments have been substantively responded too.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/EtagCacheRequestHandler.java
```diff
@@ -0,0 +1,174 @@
...
+        this.dataCache = (TupleDataCache<String, Long, String>) dataCache;
+        this.querySigningService = (QuerySigningService<Long>) querySigningService;
+    }
+
+    @Override
+    public boolean handleRequest(
+            final RequestContext context,
+            final DataApiRequest request,
+            final DruidAggregationQuery<?> druidQuery,
+            final ResponseProcessor response
```
QubitPi: I modified and it's needed outside in constructing `EtagCacheResponseProcessor `

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/EtagCacheRequestHandler.java
```diff
@@ -0,0 +1,141 @@
...
+            DruidAggregationQuery<?> druidQuery,
+            ResponseProcessor response
+    ) {
+        ResponseProcessor nextResponse = response;
+
+        String cacheKey = null;
+        try {
+            if (context.isReadCache()) {
+                cacheKey = getKey(druidQuery);
+                final TupleDataCache.DataEntry<String, String , String> cacheEntry = dataCache.get(cacheKey);
```
kevinhinterlong: No need for `final`

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/EtagCacheRequestHandler.java
```diff
@@ -0,0 +1,141 @@
...
+                    );
+
+                    CACHE_MISSES.mark(1);
+                }
+            }
+        } catch (Exception e) {
+            LOG.warn("Cache key cannot be built: ", e);
+        }
+
+        nextResponse = new EtagCacheResponseProcessor(
```
garyluoex: So... it is possible that we don't put the etag header and still have the `EtagResponseProcessor` in the response chain when `isReadCache` is false?

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/DruidJsonResponseContentKeys.java
```diff
@@ -11,7 +11,9 @@
     UNCOVERED_INTERVALS_OVERFLOWED("uncoveredIntervalsOverflowed"),
     STATUS_CODE("status-code"),
     RESPONSE("response"),
-    ETAG("Etag")
+    ETAG("Etag"),
+    REQUEST_ETAG("If-None-Match"),
+    NON_EXISTING_ETAG_VALUE("non-existing-etag")
```
garyluoex: We should start a new `DruidRequestRequestContentKeys` and put the request stuff there.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/workflow/DruidWorkflow.java
```diff
@@ -115,14 +117,15 @@ public DataRequestHandler buildWorkflow() {
...
-            }
+        if (CacheFeatureFlag.TTL.isOn()) {
+            uiHandler = new CacheRequestHandler(uiHandler, dataCache, mapper);
+            nonUiHandler = new CacheRequestHandler(nonUiHandler, dataCache, mapper);
+        } else if (CacheFeatureFlag.LOCAL_SIGNATURE.isOn()) {
+            uiHandler = new CacheV2RequestHandler(uiHandler, dataCache, querySigningService, mapper);
+            nonUiHandler = new CacheV2RequestHandler(nonUiHandler, dataCache, querySigningService, mapper);
+        } else if (CacheFeatureFlag.ETAG.isOn()) {
+            uiHandler = new EtagCacheRequestHandler(uiHandler, dataCache, querySigningService, mapper);
+            nonUiHandler = new EtagCacheRequestHandler(uiHandler, dataCache, querySigningService, mapper);
```
garyluoex: so.. where is our etag data cache? <String, String, String>.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/handlers/workflow/DruidWorkflowSpec.groovy
```diff
@@ -56,24 +62,23 @@ class DruidWorkflowSpec extends Specification {
...
         splittingStatus = QUERY_SPLIT.isOn()
 
         systemConfig = SystemConfigProvider.getInstance()
         uncoveredKey = SYSTEM_CONFIG.getPackageVariableName("druid_uncovered_interval_limit")
     }
 
     def cleanup() {
-        DRUID_CACHE.setOn(cacheStatus)
-        DRUID_CACHE_V2.setOn(cacheV2Status)
+        SYSTEM_CONFIG.clearProperty(TTL_CACHE_CONFIG_KEY)
```
garyluoex: We need to store the property value in setup, and restore the value in cleanup time so that this test don't cause any side affect to other tests.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/EtagCacheRequestHandler.java
```diff
@@ -0,0 +1,141 @@
...
+     *
+     * @param next  The next handler in the chain
+     * @param dataCache  The cache instance
+     * @param querySigningService  The service to generate query signatures
+     * @param mapper  The mapper for all JSON processing
+     */
+    public EtagCacheRequestHandler(
+            @NotNull DataRequestHandler next,
+            @NotNull DataCache<?> dataCache,
+            @NotNull QuerySigningService<?> querySigningService,
```
QubitPi: Hmm...I think I forgot to delete it after a refactor

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/EtagCacheRequestHandler.java
QubitPi: Good catch! Thanks!

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/workflow/DruidWorkflow.java
QubitPi: Let's case it and add a etag dataCache in a separate PR

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/handlers/workflow/DruidWorkflowSpec.groovy
QubitPi: Looks like `clearProperty `does remove **runtime** configs

garyluoex: @QubitPi  We need to fix both test then.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/workflow/DruidWorkflow.java
garyluoex: We should implement in this PR since the code will break, and this PR should begin allowing people to use the cache. @michael-mclawhorn  what do you think?

QubitPi:
@garyluoex Thanks for the approval!
Sure, let me try to make that change in this PR

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/EtagCacheRequestHandler.java
```diff
@@ -0,0 +1,136 @@
...
+    public boolean handleRequest(
+            RequestContext context,
+            DataApiRequest request,
+            DruidAggregationQuery<?> druidQuery,
+            ResponseProcessor response
+    ) {
+        ResponseProcessor nextResponse = response;
+
+        String cacheKey = null;
+        try {
```
michael-mclawhorn: Per @cdeszaq before, move inside the try block.

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/EtagCacheRequestHandler.java
```diff
@@ -0,0 +1,136 @@
...
+                        RequestLog.startTiming(RESPONSE_WORKFLOW_TIMER);
+                    }
+                } else { // Current query is not in data cache
+                    // Insert "If-None-Match" header into RequestContext; the value a random pre-defined string
+                    context.getHeaders().putSingle(
+                            eTagInRequest,
+                            DruidJsonRequestContentKeys.NON_EXISTING_ETAG_VALUE.getName()
+                    );
+
+                    CACHE_MISSES.mark(1);
```
michael-mclawhorn: Shouldn't there be a matching CACHE_HITS mark somewhere in here?

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/EtagCacheRequestHandler.java
```diff
@@ -0,0 +1,136 @@
...
+                CACHE_REQUESTS.mark(1);
+
+                String eTagInRequest = DruidJsonRequestContentKeys.ETAG.getName();
+                if (cacheEntry != null) { // Current query is in data cache
+                    // Insert "If-None-Match" header into RequestContext; the value is etag of the corresponding cache
+                    context.getHeaders().putSingle(
+                            eTagInRequest,
+                            mapper.readTree(cacheEntry.getValue()).get(DruidJsonResponseContentKeys.ETAG.getName())
+                                    .asText()
+                            );
```
michael-mclawhorn: Suggestion: I don't like the wrapping here.  how about putting the .get down to align with the .asText

QubitPi: Sure. That looks better indeed

michael-mclawhorn:
:shipit:

PR #313, opened 2017-06-02
Translate top README to Chinese

QubitPi:


kevinhinterlong:
Do we want to have a `translations` folder or just keep it in the root?

michael-mclawhorn:
It might make sense to put alternate languages in subdirectories by their i18n locale.
http://www.oracle.com/technetwork/java/javase/java8locales-2095355.html
That way any additional translations can be checked for and routed programmatically as contributions arise.

cdeszaq:
To facilitate deeper discussion with a broader audience, I've created [a topic in the Developer's group](https://groups.google.com/forum/#!topic/fili-developers/7wFhLBb8mwg).

saiyan86:
Very good translation. Keep up the good work!

michael-mclawhorn:
What's the relationship between this PR and https://github.com/yahoo/fili/pull/398 ?

QubitPi:
@michael-mclawhorn I think, instead of having a markdown README, @saiyan86 wants a [org-mode](https://en.wikipedia.org/wiki/Org-mode) README, which can be automatically translated to HTML. I know @saiyan86 is a big fan of Emacs org-mode.

So the relationship is that this PR is a standard README translation, whereas https://github.com/yahoo/fili/pull/398 tries to transform README from markdown file to Emacs org-mode file.

But I'm not sure this is the way that core team wants.

saiyan86:
Actually it's a org-mode source file which can be exported to markdown. It is totally OK for me if you guys don't want that. I have reviewed this PR. The translation is very good.

QubitPi:
Duplicate of https://github.com/yahoo/fili/pull/444

PR #315, opened 2017-06-05
Finalize v0.8.x dev line

cdeszaq:
Lots of small cleanups within the changelog for v0.8.x, but otherwise just the "normal" release steps, including:
- Summary for the release
- New section for vNext
- Update the readme pointer for the stable version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,13 +7,82 @@ pull request if there was one.
 
 Current
 -------
+
 ### Added:
+
+
+
+### Changed:
+
+
+
+### Deprecated:
+
+
+
+### Fixed:
+
+
+
+### Known Issues:
+
+
+
+### Removed:
+
+
+
+v0.8.69 - 2017/06/06
+--------------------
+
+The main changes in this version are changes to the Table and Schema structure, including a major refactoring of
+Physical Table. The concept of Availability was split off from Physical Table, allowing Fili to better reason about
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,73 @@ pull request if there was one.
 
 Current
```
cdeszaq: `Current` is explicitly unversioned because we couldn't specify a full version triplet (ie. `x.y.z`, since the `.z` portion will advance with every push to `master`. 

The README _does_ have a dynamic badge that points at the most recent version in Bintray, so that (I think) should satisfy some of the usefulness you're looking for here, but out of curiosity, what use do you think including a version number here would provide? It's quite likely that there are uses I've not thought of 😄!

CHANGELOG.md
```diff
@@ -7,6 +7,73 @@ pull request if there was one.
...
+a number of other small concerns. Here are some of the _highlights_ beyond the main changes around Physical Tables:
+
+Fixes:
+- Unicode characters are now properly sent back to Druid
+- Druid client now follows redirects
+
+New Capabilities & Enhancements:
+- Can sort on `dateTime`
+- Class Scanner Spec can discover dependencies, making it's dynamic equality testing easier to use
+- There's an example application that shows how to slurp configuration from an existing Druid instance
```
michael-mclawhorn: 'its'

PR #316, opened 2017-06-05
Initialize 0 9 x line

cdeszaq:
This builds on the previous one, it's really just the pom file changes.


cdeszaq:
If it's a bug, as in, something that was working is now broken, we can fix it, but if it's an enhancement or something that allows for new behavior, I'd much rather see it in the v0.9.x line, rather than having the v0.8.x line continue to get new things.

PR #317, opened 2017-06-06
Add dimension dictionary to metric loader without breaking

garyluoex:


michael-mclawhorn:
Definitely needs a changelog.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/MetricLoader.java
```diff
@@ -16,4 +17,13 @@
      */
     void loadMetricDictionary(MetricDictionary metricDictionary);
```
cdeszaq: Can we deprecate the "old" method, since we want to allow dimensions always anyways?

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/MetricLoader.java
```diff
@@ -14,7 +14,10 @@
      * Load metrics and populate the metric dictionary.
      *
      * @param metricDictionary  The dictionary that will be loaded with metrics
+     *
+     * @deprecated in favor of loadMetricDictionary(MetricDictionary, DimensionDictionary)
```
cdeszaq: Needs changelog entry about the deprecation 😉

garyluoex: I think it is in the changelog already

cdeszaq: Indeed it is. Somehow I missed that, or perhaps GitHub was overly aggressive with it's cache ;)

cdeszaq:
Please hold off on merging until #315 and #316 land.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/MetricLoader.java
michael-mclawhorn: I don't really agree with this deprecation but I don't feel strongly.

cdeszaq: My thoughts on this are that we've come to the realization that some metric definitions require dimensions. The best way (in my opinion) to make this dependency clear is to pass the "up-stream" components into the downstream ones so that it's obvious that the dimensions need to be set up before the metrics. Deprecating (and eventually removing) the unclear method call takes us that direction.

PR #335, opened 2017-06-08
Fix MetricLoader Interface and DimensionalFilter Constructor

garyluoex:


fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/MetricLoader.java
```diff
@@ -18,7 +21,12 @@
      * @deprecated in favor of loadMetricDictionary(MetricDictionary, DimensionDictionary)
      */
     @Deprecated
-    void loadMetricDictionary(MetricDictionary metricDictionary);
+    default void loadMetricDictionary(MetricDictionary metricDictionary) {
+        String message = "The method loadMetricDictionary(MetricDictionary, DimensionDictionary) must be implemented " +
+                "for MetricLoader";
+        LoggerFactory.getLogger(DimensionToNameSerializer.class).error(message);
```
cdeszaq: Should be `warn`, not `error`.

cdeszaq (dismissed):
Changelog needed

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/MetricLoader.java
QubitPi: `MetricLoader.class`

ghost (changes requested):

michael-mclawhorn:
I concur with the requested changes and have no additional issues.  I can provide a 👍 in place of Rick's once the changes are made.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/MetricLoader.java
```diff
@@ -18,7 +21,12 @@
...
+        throw new RuntimeException(message);
```
michael-mclawhorn: 👍

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/MetricLoader.java
michael-mclawhorn: 'must' is too strong here as well.  There are many perfectly healthy MetricLoaders that don't implement the new method, and I don't really think we should be strong arming them to do so.  Also, this should not throw a generic runtime exception.
https://docs.oracle.com/javase/7/docs/api/java/lang/UnsupportedOperationException.html

garyluoex: @michael-mclawhorn so...? you are saying user don't have to implement `loadMetricDictionary` method at all in the `MetricLoader` and the loader is still perfectly healthy?

fili-core/src/test/java/com/yahoo/bard/webservice/druid/model/filter/TestDimensionalFilter.java
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.druid.model.filter;
+
+import com.yahoo.bard.webservice.data.dimension.Dimension;
+
+/**
+ * A dummy test class that simply constructs a dimensional filter.
+ */
+public class TestDimensionalFilter extends DimensionalFilter {
+
```
michael-mclawhorn: Why do we need this class?

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/MetricLoader.java
```diff
@@ -18,7 +20,11 @@
      * @deprecated in favor of loadMetricDictionary(MetricDictionary, DimensionDictionary)
      */
     @Deprecated
-    void loadMetricDictionary(MetricDictionary metricDictionary);
+    default void loadMetricDictionary(MetricDictionary metricDictionary) {
+        String message = "The method loadMetricDictionary(MetricDictionary, DimensionDictionary) should be used " +
+                "instead";
+        LoggerFactory.getLogger(MetricLoader.class).warn(message);
+    }
```
michael-mclawhorn: I would lift this logger up to a static final on the class just to keep with the pattern used everywhere else

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/MetricLoader.java
michael-mclawhorn: I was actually suggesting that while some metric loaders depend on filtered metrics and thus dimension dictionaries, not everyone would. 
But I don't care that much, we can ask them to do so via deprecation.

fili-core/src/test/java/com/yahoo/bard/webservice/druid/model/filter/TestDimensionalFilter.java
garyluoex: I don't want to add dependency on sub-class to a test on the base class. Inline class was not working, due to some weird groovy error that I think might be a bug.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/MetricLoader.java
garyluoex: Actually I don't think I want to do that, because I don't want people implementing this interface to use this Logger and this Logger can not be private.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
```diff
@@ -209,13 +209,11 @@ protected ConfigPhysicalTable buildPhysicalTableWithDependency(
...
                                         tableName.asName(),
-                                        buildPhysicalTableWithDependency(
-                                                tableName.asName(),
-                                                availableTableDefinitions,
-                                                dictionaries
-                                        )
+                                        availableTableDefinitions,
+                                        dictionaries
+
                                 );
```
michael-mclawhorn: There shouldn't be a blank line here.

fili-core/src/test/java/com/yahoo/bard/webservice/druid/model/filter/TestDimensionalFilter.java
```diff
@@ -0,0 +1,29 @@
...
+    private FilterType type;
```
michael-mclawhorn: Local variable not needed?

michael-mclawhorn (approved):
Other than two trivial changes, I'm ok with this.

PR #396, opened 2017-06-13
Enhance SearchFilter.QueryType and FailedFuture

kevinhinterlong:
This adds some minor enhancements to FailedFuture and allows for a QueryType to be found from its serialized name.


fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/SearchFilter.java
```diff
@@ -42,6 +42,20 @@
         QueryType(String type) {
             this.type = type;
         }
+
+        /**
+         * Get the QueryType enum fromType it's search type.
+         * @param type  Type of the query type (for serialization)
+         * @return the enum QueryType
```
QubitPi: Empty lines before `@return`

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/SearchFilter.java
```diff
@@ -42,6 +42,20 @@
         QueryType(String type) {
             this.type = type;
         }
+
+        /**
+         * Get the QueryType enum fromType it's search type.
+         * @param type  Type of the query type (for serialization)
```
QubitPi: Empty lines before `@param`

fili-core/src/main/java/com/yahoo/bard/webservice/util/CompletedFuture.java
```diff
@@ -0,0 +1,60 @@
+// Copyright 2016 Yahoo Inc.
```
QubitPi: 2017

ghost (changes requested):

asifmansoora (approved):
This PR looks good to me 👍

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/SearchFilter.java
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016 Yahoo Inc.
+// Copyright 2017 Yahoo Inc.
```
QubitPi: Looks like this file is there since 2016 (we put based on when file is initialized)

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/SearchFilter.java
```diff
@@ -42,6 +42,22 @@
...
         }
+
+        /**
+         * Get the QueryType enum from it's search type.
+         *
+         * @param type  The type of query (i.e. "insensitive_contains")
+         *
+         * @return the enum QueryType
+         */
+        public static QueryType fromType(String type) {
```
QubitPi: There is a standard way(simpler and more readable) of achieving your purpose. Example will be https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/config/CacheFeatureFlag.java or http://grepcode.com/file_/repo1.maven.org/maven2/javax.ws.rs/jsr311-api/1.1/javax/ws/rs/core/Response.java/?v=source

Instead of `fromType`, you have 
```java
        public String getType() {
            return type;
        }
```
And you check type by having "contains".equals(QueryType.Contains.getType()). That's how we check type of enum (as the two link example does, as well)

ghost (changes requested):
CHANGELOG needed; make sure to explain why you add `CompletedFuture` and delete the other

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/SearchFilter.java
kevinhinterlong: I'm not sure I understand what you mean. Do you want me to enumerate each possibility to find which type the enum is instead of using the loop?

From the second example you gave they did this which is nearly identical to the way I checked the type.

```java
public static Status fromStatusCode(final int statusCode) {
    for (Status s : Status.values()) {
        if (s.code == statusCode) {
            return s;
        }
    }
    return null;
}
```

fili-core/src/main/java/com/yahoo/bard/webservice/util/CompletedFuture.java
```diff
@@ -0,0 +1,60 @@
...
+        return true;
+    }
+
+    @Override
+    public E get() throws ExecutionException {
+        if (this.throwable != null) {
+            throw new ExecutionException(throwable);
+        } else {
+            return item;
+        }
```
QubitPi: `else` is not necessary

fili-core/src/main/java/com/yahoo/bard/webservice/util/CompletedFuture.java
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * @param <E>  The type to be returned by this future
+ */
+public class CompletedFuture<E> implements Future<E> {
+    private final E item;
+    private final Throwable throwable;
+
+    /**
+     * Creates a completed future with an item to return or
+     * if a nonnull throwable is passed it it will throw an
```
QubitPi: passed it -> provided ?

fili-core/src/main/java/com/yahoo/bard/webservice/util/CompletedFuture.java
```diff
@@ -0,0 +1,60 @@
...
+     * execution exception.
```
QubitPi: `{@link java.util.concurrent.ExecutionException}` makes it more readable

fili-core/src/main/java/com/yahoo/bard/webservice/util/CompletedFuture.java
```diff
@@ -0,0 +1,60 @@
...
+ * as soon as it is called.
+ *
+ * @param <E>  The type to be returned by this future
+ */
+public class CompletedFuture<E> implements Future<E> {
+    private final E item;
+    private final Throwable throwable;
+
+    /**
+     * Creates a completed future with an item to return or
```
QubitPi: `an item to return on on {@link #get()}`

fili-core/src/test/java/com/yahoo/bard/webservice/models/druid/client/impl/TestDruidWebService.java
```diff
@@ -90,7 +90,7 @@ public TestDruidWebService() {
     /**
      * If TestDruidWebService#throwable is set, invokes the failure callback.  Otherwise invokes success or failure
      * dependent on whether TestDruidWebService#statusCode equals 200.
-     * Note that since this doesn't send requests to druid all the responses will be null or {@link FailedFuture}.
+     * Note that since this doesn't send requests to druid all the responses will be null or {@link CompletedFuture}.
```
QubitPi: Since this class is in a different package

`{@link com.yahoo.bard.webservice.util.CompletedFuture}`

fili-core/src/main/java/com/yahoo/bard/webservice/util/CompletedFuture.java
```diff
@@ -0,0 +1,60 @@
...
+     * if a nonnull throwable is passed it it will throw an
+     * execution exception.
+     *
+     * @param item  The item to return when called.
+     * @param throwable  A non null throwable if an exception should be thrown.
+     */
+    public CompletedFuture(E item, Throwable throwable) {
```
QubitPi: I see all constructor is called everywhere with `item = null`, like `new CompletedFuture<>(null, throwable)`. This makes code less readable, because people don't know that `null` means. A better approach as I learned from Rick's suggestion is

```java
public CompletedFuture(E item, Throwable throwable) {
    this.item = item;
    this.throwable = throwable;
}

public CompletedFuture(Throwable throwable) {
    this(null, throwable);
}
```

That way, you do `new CompletedFuture<>(throwable)`.

ghost (changes requested):

fili-core/src/main/java/com/yahoo/bard/webservice/util/CompletedFuture.java
kevinhinterlong: I think this is clear from the context and adding this could make it unclear what happens when `{@link #get(long, TimeUnit)}` is called.

fili-core/src/test/java/com/yahoo/bard/webservice/models/druid/client/impl/TestDruidWebService.java
kevinhinterlong: It's available from context because of 
```java
import com.yahoo.bard.webservice.util.CompletedFuture;
```


do you want me to add the fully qualified name?

fili-core/src/main/java/com/yahoo/bard/webservice/util/CompletedFuture.java
kevinhinterlong: I've decided to replace this with [Either](https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/util/Either.java) and give one constructor for the item and one for the throwable so it's clear you can only do one or the other.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/SearchFilter.java
kevinhinterlong: What's the reason for adding `final`? 
I've been told in previous PRs to remove `final` from parameters

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/SearchFilter.java
```diff
@@ -42,6 +42,22 @@
         QueryType(String type) {
             this.type = type;
         }
+
+        /**
+         * Get the QueryType enum from it's search type.
```
kevinhinterlong: I think this is clear from context as you can only pass in a `String` to the method.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/SearchFilter.java
```diff
@@ -42,6 +42,22 @@
...
+         *
+         * @param type  The type of query (i.e. "insensitive_contains")
```
kevinhinterlong: Same as above

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/SearchFilter.java
```diff
@@ -42,6 +42,22 @@
...
         }
+
+        /**
+         * Get the QueryType enum from it's search type.
+         *
+         * @param type  The type of query (i.e. "insensitive_contains").
+         *
+         * @return the enum QueryType.
+         */
+        public static QueryType fromType(String type) {
```
archolewa: I would argue that this should be returning an `Optional` instead of a `null`, or throwing an `IllegalArgumentException` if the `type` is not a recognized `QueryType`. 

Returning an Optional makes it clear to people using the code that they may not get a QueryType back, and need to deal with that case. This makes sense if it's totally valid to have a string type that cannot be translated to a QueryType. 

The second case assumes that it's an error if we get an unrecognized type.

I'm assuming that it's an error to receive an unrecognized type, but I don't know yet what it's meant to be used for, so I can't say for certain.

archolewa (changes requested):

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/SearchFilter.java
kevinhinterlong: @QubitPi Are you fine with me reverting to the [IllegalArgumentException](https://github.com/yahoo/fili/pull/396/commits/0ced60b2512fc2eb621603e92d84bac4ca110395#r123132356)/returning optional?

For context, this was made because `SearchFilter` doesn't have a reference to its `QueryType` and I have to process a `SearchFilter` differently for each case (and if not found, throw an exception).

Additionally, if `SearchFilter` must be one of those `QueryType` then it should probably be verified in the [constructor](https://github.com/yahoo/fili/blob/369787a86097e19c672072c8fe4e5b76e4fc012a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/SearchFilter.java#L67) that accepts a `String` queryType.

fili-core/src/main/java/com/yahoo/bard/webservice/util/CompletedFuture.java
```diff
@@ -0,0 +1,47 @@
...
+     * Private constructor - all methods static.
+     */
+    private CompletedFuture() {
+
+    }
+
+    /**
+     * Creates a completed {@link java.util.concurrent.Future} which will return this item.
+     *
+     * @param item  The item to return when called.
```
QubitPi: no period

fili-core/src/main/java/com/yahoo/bard/webservice/util/CompletedFuture.java
```diff
@@ -0,0 +1,47 @@
...
+     * @param <E>  The type to be returned by this future.
```
QubitPi: no period

fili-core/src/main/java/com/yahoo/bard/webservice/util/CompletedFuture.java
```diff
@@ -0,0 +1,47 @@
...
+     *
+     * @return a completed {@link CompletableFuture} which will successfully return an item.
```
QubitPi: no period and all others as well

ghost (changes requested):
I promise a deep review on CompletedFuture later today

fili-core/src/main/java/com/yahoo/bard/webservice/util/CompletedFuture.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Creates a completed {@link java.util.concurrent.Future} which will return this item.
+     *
+     * @param item  The item to return when called
+     * @param <E>  The type to be returned by this future
+     *
+     * @return a completed {@link CompletableFuture} which will successfully return an item.
+     */
+    public static <E> CompletableFuture<E> returning(E item) {
```
QubitPi: Where is this method used?

ghost (changes requested):

fili-core/src/main/java/com/yahoo/bard/webservice/util/CompletedFuture.java
kevinhinterlong: I was using this in #430 but it's no longer needed and can be deleted if that makes sense.

fili-core/src/test/java/com/yahoo/bard/webservice/models/druid/client/impl/TestDruidWebService.java
QubitPi: This is not for readability. The reason to put complete path is for Javadoc plugin to be able to recognize and inject HTML link so that points to the correct class.

fili-core/src/main/java/com/yahoo/bard/webservice/util/CompletedFuture.java
```diff
@@ -0,0 +1,47 @@
...
+        //Intentionally left blank.
+    }
+
+    /**
+     * Creates a completed {@link java.util.concurrent.Future} which will return this item.
+     *
+     * @param item  The item to return when called
+     * @param <E>  The type to be returned by this future
+     *
+     * @return a completed {@link CompletableFuture} which will successfully return an item.
```
QubitPi: no period at end

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/SearchFilter.java
```diff
@@ -42,6 +43,22 @@
         QueryType(String type) {
             this.type = type;
         }
+
+        /**
+         * Get the QueryType enum from it's search type.
+         *
+         * @param type  The type of query (i.e. "insensitive_contains")
```
QubitPi: `(i.e. "insensitive_contains")` indicates this is the only QueryType? I would remove it.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/SearchFilter.java
```diff
@@ -78,18 +95,26 @@ private SearchFilter(Dimension dimension, String type, String value) {
         return query;
     }
 
+    private String getQueryType() {
+        return query.get("type");
```
QubitPi: `type` is used multiple times. It's more maintainable to have it as a static final String.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/SearchFilter.java
```diff
@@ -78,18 +95,26 @@ private SearchFilter(Dimension dimension, String type, String value) {
...
+    }
+
+    private String getQueryValue() {
+        return query.get("value");
```
QubitPi: same

fili-core/src/main/java/com/yahoo/bard/webservice/util/CompletedFuture.java
```diff
@@ -0,0 +1,47 @@
+// Copyright 2017 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.util;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * A utility class to create futures which have already been completed
+ * and will return as soon as it is called.
+ *
```
QubitPi: extra line

fili-core/src/main/java/com/yahoo/bard/webservice/util/CompletedFuture.java
```diff
@@ -0,0 +1,47 @@
+// Copyright 2017 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.util;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * A utility class to create futures which have already been completed
+ * and will return as soon as it is called.
```
QubitPi: can be move to the previous line

fili-core/src/main/java/com/yahoo/bard/webservice/util/CompletedFuture.java
```diff
@@ -0,0 +1,47 @@
...
+     * {@link java.util.concurrent.ExecutionException}.
+     *
+     * @param throwable  A throwable to be included in the {@link java.util.concurrent.ExecutionException}
+     * @param <E>  The type to be returned by this future
+     *
+     * @return a completed {@link CompletableFuture} which will fail and throw an exception
+     */
+    public static <E> CompletableFuture<E> throwing(Throwable throwable) {
+        CompletableFuture<E> completedFuture = CompletableFuture.supplyAsync(() -> null);
+        completedFuture.completeExceptionally(throwable);
```
QubitPi: @michael-mclawhorn @garyluoex @archolewa I'm little concerned about this statement. According to its [doc](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#completeExceptionally-java.lang.Throwable-)

"If **not already completed**, causes invocations of get() and related methods to throw the given exception."

Looks like this method wants to throw exception **on complete** instead (Or Javadoc of this method is simply wrong?).

ghost (changes requested):
Getting there. Code looks very good!

fili-core/src/test/java/com/yahoo/bard/webservice/models/druid/client/impl/TestDruidWebService.java
kevinhinterlong: The javadoc plugin is able to correctly recognize this and inject the html link in other classes I've tried (like in CompletedFuture). Did you generate the javadoc and it didn't work correctly?

ghost (approved):

archolewa (approved):

PR #397, opened 2017-06-13
Add ability to create druid responses

kevinhinterlong:
Needed to create druid responses from sql results in #364

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponse.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponse.java
@@ -1,4 +1,4 @@
-// Copyright 2016 Yahoo Inc.
+// Copyright 2017 Yahoo Inc.
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.druid.model.response;
 
@@ -12,7 +12,11 @@ import java.util.List;
  */
 @JsonSerialize(using = DruidResponseSerializer.class)
 public class DruidResponse {
-    private final List<DruidResultRow> results = new ArrayList<>();
+    private final List<DruidResultRow> results;
+
+    public DruidResponse() {
+        results = new ArrayList<>();
+    }
 
     /**
      * Adds a row to the list of results.
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponse.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponse.java
@@ -14,6 +14,9 @@ import java.util.List;
 public class DruidResponse {
     private final List<DruidResultRow> results;
 
+    /**
+     * Construct a Druid Response which will store rows of results.
+     */
     public DruidResponse() {
         results = new ArrayList<>();
     }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponse.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponse.java
@@ -1,0 +1,34 @@
+// Copyright 2016 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.druid.model.response;
+
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A response which serializes to an equivalent DruidResponse.
+ */
+@JsonSerialize(using = DruidResponseSerializer.class)
+public class DruidResponse {
+    private final List<DruidResultRow> results = new ArrayList<>();
+
+    /**
+     * Adds a row to the list of results.
+     *
+     * @param row  The row to be added.
+     */
+    public void add(DruidResultRow row) {
+        results.add(row);
+    }
+
+    /**
+     * Gets the list of results for this response.
+     *
+     * @return the results.
+     */
+    public List<DruidResultRow> getResults() {
+        return results;
+    }
+}
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponse.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponse.java
@@ -17,7 +17,7 @@ public class DruidResponse {
     /**
      * Construct a Druid Response which will store rows of results.
      */
-    public DruidResponse() {
+    DruidResponse() {
         results = new ArrayList<>();
     }
 
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponseSerializer.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponseSerializer.java
@@ -3,7 +3,6 @@
 package com.yahoo.bard.webservice.druid.model.response;
 
 import com.fasterxml.jackson.core.JsonGenerator;
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonSerializer;
 import com.fasterxml.jackson.databind.SerializerProvider;
 
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponseSerializer.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponseSerializer.java
@@ -1,22 +1,25 @@
-// Copyright 2016 Yahoo Inc.
+// Copyright 2017 Yahoo Inc.
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.druid.model.response;
 
 import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonSerializer;
 import com.fasterxml.jackson.databind.SerializerProvider;
 
 import java.io.IOException;
 
 /**
- * Serializes {@link DruidResponse} into a json equivalent to Druid's actual responses.
+ * Serializes a {@link DruidResponse} into the same format as Druid's responses.
  */
-public class DruidResponseSerializer extends JsonSerializer {
+public class DruidResponseSerializer extends JsonSerializer<DruidResponse> {
 
     @Override
-    public void serialize(Object o, JsonGenerator jsonGenerator, SerializerProvider serializerProvider)
-            throws IOException {
-        DruidResponse druidResponse = (DruidResponse) o;
+    public void serialize(
+            DruidResponse druidResponse,
+            JsonGenerator jsonGenerator,
+            SerializerProvider serializerProvider
+    ) throws IOException {
         jsonGenerator.writeStartArray();
         for (Object result : druidResponse.getResults()) {
             jsonGenerator.writeObject(result);
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponseSerializer.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponseSerializer.java
@@ -9,7 +9,8 @@ import com.fasterxml.jackson.databind.SerializerProvider;
 import java.io.IOException;
 
 /**
- * Serializes a {@link DruidResponse} into the same format as Druid's responses.
+ * Serializes a {@link com.yahoo.bard.webservice.druid.model.response.DruidResponse} into the
+ * same format as Druid's responses.
  */
 public class DruidResponseSerializer extends JsonSerializer<DruidResponse> {
 
@@ -20,8 +21,8 @@ public class DruidResponseSerializer extends JsonSerializer<DruidResponse> {
             SerializerProvider serializerProvider
     ) throws IOException {
         jsonGenerator.writeStartArray();
-        for (Object result : druidResponse.getResults()) {
-            jsonGenerator.writeObject(result);
+        for (DruidResultRow resultRow : druidResponse.getResults()) {
+            jsonGenerator.writeObject(resultRow);
         }
         jsonGenerator.writeEndArray();
     }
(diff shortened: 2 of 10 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponse.java
```diff
@@ -0,0 +1,34 @@
+// Copyright 2016 Yahoo Inc.
```
QubitPi: 2017

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponse.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A response which serializes to an equivalent DruidResponse.
+ */
+@JsonSerialize(using = DruidResponseSerializer.class)
+public class DruidResponse {
+    private final List<DruidResultRow> results = new ArrayList<>();
```
QubitPi: It makes more sense to have a constructor that does the assignment.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponseSerializer.java
```diff
@@ -0,0 +1,26 @@
+// Copyright 2016 Yahoo Inc.
```
QubitPi: 2017

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponseSerializer.java
```diff
@@ -0,0 +1,26 @@
...
+package com.yahoo.bard.webservice.druid.model.response;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+
+import java.io.IOException;
+
+/**
+ * Serializes {@link DruidResponse} into a json equivalent to Druid's actual responses.
```
QubitPi: No sure what you mean by "Serializes X into Y to Z"

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponseSerializer.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+public class DruidResponseSerializer extends JsonSerializer {
+
+    @Override
+    public void serialize(Object o, JsonGenerator jsonGenerator, SerializerProvider serializerProvider)
```
QubitPi: `o` doesn't tell what this variable does

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponseSerializer.java
```diff
@@ -0,0 +1,26 @@
...
+            throws IOException {
```
QubitPi: I think JSON parsing exception should be handled inside the method, which means we should throw in a `try-catch`; you don't want to pass this method-logic to API callers.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResultRow.java
```diff
@@ -0,0 +1,39 @@
+// Copyright 2016 Yahoo Inc.
```
QubitPi: 2017

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResultRow.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+
+/**
+ * A row of results in a Druid Response.
+ */
+public abstract class DruidResultRow {
+    @JsonIgnore
+    public final DateTime timestamp;
```
QubitPi: Why do we make it public?

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResultRow.java
```diff
@@ -0,0 +1,39 @@
...
+     *
+     * @param timestamp  The timestamp to be included in serialization of a response.
+     */
+    public DruidResultRow(DateTime timestamp) {
+        this.timestamp = timestamp;
+    }
+
+    @JsonProperty
+    public String getTimestamp() {
+        return timestamp.toDateTime(DateTimeZone.UTC).toString();
```
QubitPi: It's more efficient to move `timestamp.toDateTime(DateTimeZone.UTC).toString()` to a instance variable if this method will be called multiple times.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/GroupByResultRow.java
```diff
@@ -0,0 +1,75 @@
+// Copyright 2016 Yahoo Inc.
```
QubitPi: 2017

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/GroupByResultRow.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.Map;
+
+/**
+ * A row of results in a GroupByQuery.
+ * todo: use and test
+ */
+@JsonPropertyOrder({"version", "timestamp", "event"})
+public class GroupByResultRow extends DruidResultRow {
+    @JsonIgnore
+    public final Version version;
```
QubitPi: Why do we make it public?

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/GroupByResultRow.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @JsonProperty
+    private final Map<String, Object> event = new HashMap<>();
```
QubitPi: It makes more sense to have a constructor that does the assignment.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/TimeseriesResultRow.java
```diff
@@ -0,0 +1,34 @@
+// Copyright 2016 Yahoo Inc.
```
QubitPi: 2017

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/TimeseriesResultRow.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Row of results in a {@link com.yahoo.bard.webservice.druid.model.query.TimeSeriesQuery}.
+ */
+@JsonPropertyOrder({"timestamp", "result"})
+public class TimeseriesResultRow extends DruidResultRow {
+    @JsonProperty
+    private final Map<String, Object> result = new HashMap<>();
```
QubitPi: Constructor

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/response/DruidResponseSerializationSpec.groovy
```diff
@@ -0,0 +1,61 @@
+package com.yahoo.bard.webservice.druid.model.response
```
QubitPi:
```
// Copyright 2017 Yahoo Inc.
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/response/DruidResponseSerializationSpec.groovy
```diff
@@ -0,0 +1,61 @@
...
+import spock.lang.Unroll
+
+class DruidResponseSerializationSpec extends Specification {
+    static DateTime DATE_TIME = DateTime.parse("2017-06-13T00:00:00.000Z");
+    static ObjectMapper objectMapper = new ObjectMapper();
+    static key1 = "key1"
+    static key2 = "key2"
+
+    @Unroll
+    def "Check Timeseries druid response serialization"() {
```
QubitPi: If Unrolls it, it's common to add a description for each different test case. For example https://github.com/yahoo/fili/blob/master/fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/MetricUnionAvailabilitySpec.groovy#L90

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/response/DruidResponseSerializationSpec.groovy
```diff
@@ -0,0 +1,61 @@
...
+        where:
+        keys               | values          | serialized
+        asList(key1)       | asList("value") | """[{"timestamp":"2017-06-13T00:00:00.000Z","result":{"key1":"value"}}]"""
+        asList(key1)       | asList(1)       | """[{"timestamp":"2017-06-13T00:00:00.000Z","result":{"key1":1}}]"""
+        asList(key1)       | asList(1.0D)    | """[{"timestamp":"2017-06-13T00:00:00.000Z","result":{"key1":1.0}}]"""
+        asList(key1, key2) | asList(1, 2)    | """[{"timestamp":"2017-06-13T00:00:00.000Z","result":{"key1":1,"key2":2}}]"""
+    }
+
+    @Unroll
+    def "Check GroupBy druid response serialization"() {
```
QubitPi: Description

ghost (changes requested):

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResponseSerializer.java
kevinhinterlong: Yea I was just following the same way as `DimensionToNameSerializer` did it.

As far as I can tell this logic never gets passed to callers directly (unless they instantiate a `DruidResponseSerializer`) because they go through the `ObjectMapper` which I'm guessing is handling the exception. 

Let me know if you still want me to catch it because I'm not sure which way is expected with jackson.

kevinhinterlong:
My comment got collapsed after making changes so I'll put it down here @QubitPi 

> Yea I was just following the same way as DimensionToNameSerializer did it.
> 
> As far as I can tell this logic never gets passed to callers directly (unless they instantiate a DruidResponseSerializer) because they go through the ObjectMapper which I'm guessing is handling the exception.
> 
> Let me know if you still want me to catch it because I'm not sure which way is expected with jackson.

Same thing with #399, I'm not sure what the expected behavior is for a `JsonSerializer`

garyluoex:
@QubitPi  Does it make sense to catch it? If we catch it, what should we do in the catch block?

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResultRow.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+
+/**
+ * A row of results in a Druid Response.
+ */
+public abstract class DruidResultRow {
+    @JsonIgnore
+    private final String timestamp;
```
QubitPi: Having searched Fili codebase, I see timestamp is modeled using `DateTime`, instead of String, everywhere such as `Result`, `TopNResultSetMapper`, `DataSourceMetadataLoader`.

Having it as `DateTime` is also better so that child class can select different timezone instead of fixing to `DateTimeZone.UTC`

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResultRow.java
```diff
@@ -0,0 +1,39 @@
...
+        return timestamp;
+    }
+
+    /**
+     * Adds a json key/value pair to the row.
+     *
+     * @param key  The key to be added.
+     * @param value  The value of the key.
+     */
+    public abstract void add(String key, Object value);
```
QubitPi: `public abstract void add(String key, V value);`

`public abstract class DruidResultRow<V> {`

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/GroupByResultRow.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * Enum to clarify the version being used.
+     */
+    public enum Version {
+        V1("v1"),
+        V2("v2");
+
+        private String name;
```
QubitPi: `final`

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/GroupByResultRow.java
```diff
@@ -0,0 +1,75 @@
...
+
+        /**
+         * Creates the version with a name.
+         *
+         * @param name  The name to be shown in results.
+         */
+        Version(String name) {
```
QubitPi: `final`

ghost (changes requested):
CHANGELOG needed

@michael-mclawhorn @garyluoex Does it make sense to have a Response.Druid package for all new files? Since Druid won't be the only data store in the future

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/DruidResultRow.java
kevinhinterlong: This has to be able to store `String` and `Number` types. If you would like I could expand it to
```java
public abstract void add(String key, String value);
public abstract void add(String key, Number value);
```
but there is not a specific Type which can be associated with a row.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/response/GroupByResultRow.java
kevinhinterlong: Did you mean this to be applied to and not the parameter?
```java
private String name;
```

kevinhinterlong:
Druid responses are parsed into a result set and by wrapping all queries in a `AbstractDruidAggregationQuery` which overrides the query type to be `GROUP_BY` we can use a single format for druid responses (at least for the DruidResponseParser) so this is no longer needed.

Issue #390, opened 2017-06-09
Verify SearchProvider behavior for empty-string as a filter value


cdeszaq:
We don't have any tests for it, but filtering for empty string (ie. `dim|id-contains[""]`) is valid, and should match everything I think. We should verify this behavior by adding tests to the SearchProvider TCK.

PR #400, opened 2017-06-15
Add a SearchProvider test for matching an empty string

kevinhinterlong:
This was brought up in #390 and verifies that searching for an empty string will match all results

### DIFF
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/SearchProviderSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/SearchProviderSpec.groovy
@@ -25,9 +25,6 @@ import org.joda.time.DateTime
 
 import spock.lang.Specification
 
-
-import java.util.stream.Collectors;
-
 /**
  * Specification of behavior that all SearchProviders should share.
  */
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/SearchProviderSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/SearchProviderSpec.groovy
@@ -553,8 +553,7 @@ abstract class SearchProviderSpec<T extends SearchProvider> extends Specificatio
 
         and: "We get only {'owl', 'hawk', 'eagle', 'kumquat'}"
         resultsPage.numResults == 4
-        Set<String> actualOutcomes = resultsPage.pageOfData.stream().map { it.getKeyValue() }.collect(Collectors.toSet())
-        actualOutcomes.equals(expectedOutcomes)
+        expectedOutcomes == resultsPage.pageOfData.collect{ it.getKeyValue() } as Set
     }
 
     /**
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/SearchProviderSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/SearchProviderSpec.groovy
@@ -526,6 +526,31 @@ abstract class SearchProviderSpec<T extends SearchProvider> extends Specificatio
         resultsPage.numResults == 3
     }
 
+    def "An empty string filter value should return all results"() {
+        setup: "Given a filter that will match every row"
+        /* Expected rows, not necessarily in this order:
+                name: "owl", description: "this is an owl"
+                name: "hawk", description: "this is a raptor"
+                name: "eagle", description: "this is a raptor"
+                name: "kumquat", description: "this is not an animal"
+        */
+        Set<ApiFilter> filters = [
+                buildFilter("animal|desc-startswith[this]"),
+                buildFilter("animal|desc-contains[\"\"]")
+        ]
+        and: "We get the second page, where each page has two rows (so both pages should have two results)"
+        PaginationParameters parameters = new PaginationParameters(2, 2)
+
+        when: "We query the search provider"
+        Pagination<DimensionRow> resultsPage = searchProvider.findFilteredDimensionRowsPaged(filters, parameters)
+
+        then: "We get only the last page of results (which has two)"
+        resultsPage.getPageOfData().size() == 2
+
+        and: "The pagination metadata includes the correct number of total results"
+        resultsPage.numResults == 4
+    }
+
     /**
      * Checks that this search provider's indices have been cleared.
      *
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/SearchProviderSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/SearchProviderSpec.groovy
@@ -2,10 +2,10 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.data.dimension.impl
 
-import static com.yahoo.bard.webservice.data.dimension.BardDimensionField.ID
 import static com.yahoo.bard.webservice.data.dimension.BardDimensionField.DESC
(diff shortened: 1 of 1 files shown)


michael-mclawhorn (approved):
Looks good.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/SearchProviderSpec.groovy
```diff
@@ -526,6 +526,31 @@ abstract class SearchProviderSpec<T extends SearchProvider> extends Specificatio
...
+        setup: "Given a filter that will match every row"
+        /* Expected rows, not necessarily in this order:
+                name: "owl", description: "this is an owl"
+                name: "hawk", description: "this is a raptor"
+                name: "eagle", description: "this is a raptor"
+                name: "kumquat", description: "this is not an animal"
+        */
+        Set<ApiFilter> filters = [
+                buildFilter("animal|desc-startswith[this]"),
+                buildFilter("animal|desc-contains[\"\"]")
```
QubitPi: Avoid escaping quotes - `'animal|desc-contains[""]'`

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/SearchProviderSpec.groovy
```diff
@@ -526,6 +526,31 @@ abstract class SearchProviderSpec<T extends SearchProvider> extends Specificatio
...
+        ]
+        and: "We get the second page, where each page has two rows (so both pages should have two results)"
```
QubitPi: `We get the second page, where each page has two rows` sounds good enough already

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/SearchProviderSpec.groovy
```diff
@@ -526,6 +526,31 @@ abstract class SearchProviderSpec<T extends SearchProvider> extends Specificatio
...
+        PaginationParameters parameters = new PaginationParameters(2, 2)
+
+        when: "We query the search provider"
+        Pagination<DimensionRow> resultsPage = searchProvider.findFilteredDimensionRowsPaged(filters, parameters)
+
+        then: "We get only the last page of results (which has two)"
```
QubitPi: `We get only the last page of results` sounds good enough already

ghost (changes requested):

QubitPi:
Looks pretty good. Minor comments

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/SearchProviderSpec.groovy
```diff
@@ -526,6 +526,31 @@ abstract class SearchProviderSpec<T extends SearchProvider> extends Specificatio
...
     }
 
+    def "An empty string filter value should return all results"() {
+        setup: "Given a filter that will match every row"
+        /* Expected rows, not necessarily in this order:
+                name: "owl", description: "this is an owl"
+                name: "hawk", description: "this is a raptor"
+                name: "eagle", description: "this is a raptor"
+                name: "kumquat", description: "this is not an animal"
+        */
```
QubitPi: It's good to list expected rows. It's gonna be even better if you could code them into test, i.e. making sure the 4 rows are equal to those

ghost (changes requested):

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/SearchProviderSpec.groovy
kevinhinterlong: I don't think there's any guarantee  about order they're returned in. Do you want me to do something like
```
Set<String> expectedOutcomes = ["owl", "hawk", "eagle", "kumquat"]
...

Set<String> actualOutcomes = resultsPage.pageOfData.stream().map { it.getKeyValue() }.collect(Collectors.toSet())
actualOutcomes.equals(expectedOutcomes)
```

QubitPi: You could try
```
Collections.sort(actualOutcomes);
Collections.sort(expectedOutcomes);
```

kevinhinterlong: So you want me to make them both lists, sort them like
```java
List<String> actualOutcomes = resultsPage.pageOfData.stream().map { it.getKeyValue() }.sorted().collect(Collectors.toList())
Collections.sort(expectedOutcomes);
```
then check for equality?
I prefer the Set way because there's no need to rely on sorting the results

QubitPi: You can do that

michael-mclawhorn: Ideally we'd like the code going through the system to preserve or coerce ordering.  Without that we'll be inefficient in terms of creating cache keys.  
I'd suggest a jira ticket to create a Comparator on ApiFilter, probably sorting on DimensionName, DimensionFieldName, OperationName, and then lexically sorted values.  Then the ApiFilters could be stored in a SortedSet, which should make test cases far easier to write.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/SearchProviderSpec.groovy
```diff
@@ -526,6 +529,34 @@ abstract class SearchProviderSpec<T extends SearchProvider> extends Specificatio
...
+        Pagination<DimensionRow> resultsPage = searchProvider.findFilteredDimensionRowsPaged(filters, parameters)
+
+        then: "We get only one page of results"
+        resultsPage.getPageOfData().size() == 4
+
+        and: "We get only {'owl', 'hawk', 'eagle', 'kumquat'}"
+        resultsPage.numResults == 4
+        Set<String> actualOutcomes = resultsPage.pageOfData.stream().map { it.getKeyValue() }.collect(Collectors.toSet())
+        actualOutcomes.equals(expectedOutcomes)
+    }
```
michael-mclawhorn: To make this test more idiomatic for groovy and spock, lines 564-5 could be rewritten as follows:
expectedOutcomes == resultsPage.pageOfData.collect { it.getKeyValue() } as Set

This uses the simpler groovy collect without going through java streams of collectors library and the coercion operator to control the collection type.
The use of '==' is (sort of) equivalent to .equals but leverages the power assert capability for more useful debugging info.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/SearchProviderSpec.groovy
michael-mclawhorn: I'm fine with set comparison here.  More on how to streamline the test below.

michael-mclawhorn (changes requested):
I'd love to see the groovy-ification of the tests.

michael-mclawhorn (approved):

archolewa (approved):

michael-mclawhorn:
@QubitPi  please update your feedback

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/SearchProviderSpec.groovy
```diff
@@ -25,6 +25,9 @@ import org.joda.time.DateTime
 
 import spock.lang.Specification
 
+
```
QubitPi: delete new line

ghost (approved):

PR #401, opened 2017-06-15
Fix doc error

QubitPi:
Redundant content - two same pieces appear one after the other


michael-mclawhorn:
👍

michael-mclawhorn:
While I do approve this, do we have any business process that allows for one-review on these?

PR #403, opened 2017-06-16
Translate end-user-api.md and docs/README.md to Chinese

QubitPi:


QubitPi:
Merged into https://github.com/yahoo/fili/pull/585

PR #405, opened 2017-06-19
Simple security example

michael-mclawhorn:


fili-security/src/main/java/com/yahoo/bard/webservice/web/security/RoleBasedRoutingRequestMapper.java
```diff
@@ -0,0 +1,69 @@
...
+        SecurityContext securityContext = context.getSecurityContext();
+        RequestMapper<T> mapper = prioritizedRoleBasedMappers.keySet().stream()
+                .filter(securityContext::isUserInRole)
+                .peek(it -> System.out.println(it))
+                .map(prioritizedRoleBasedMappers::get)
+                .peek(it -> System.out.println(it))
+                .findFirst().orElse(defaultMapper);
+        return (mapper == null) ?
+                request :
+                mapper.apply(request, context);
```
archolewa: I don't think we want to treat a `null` mapper as the identity. This effectively means that if a user doesn't have any of the specified roles, they get FULL ACCESS TO EVERYTHING FOREVER.  People concerned about security usually get twitchy when the default is FULL ACCESS TO EVERYTHING FOREVER.

Instead, we should probably throw an exception here (an `IllegalArgumentException` or something).

fili-security/src/main/java/com/yahoo/bard/webservice/web/security/RoleDimensionApiFilterRequestMapper.java
```diff
@@ -0,0 +1,161 @@
...
+        SecurityContext securityContext = context.getSecurityContext();
+
+        Set<ApiFilter> mergedSecurityFilters = unionMergeFilterValues(
+                roleApiFilters.keySet().stream()
+                        .filter(securityContext::isUserInRole)
+                        .map(roleApiFilters::get)
+                        .flatMap(Set::stream)
+        );
+
+        if (mergedSecurityFilters.size() == 0) {
```
archolewa: I don't think we should rely on the `mergedSecurityFilters` being zero to signal that the user does not have a valid role. What if a customer wants to have some sort of god role that can see everything? Then they'd probably much rather not attach any filters at all, then be forced to attach a filter that contains _everything_.

While it's less elegant, we're probably better off first finding out if the user has at least one of the specified roles and then merging the api filters only if they have one of those roles.

fili-security/src/main/java/com/yahoo/bard/webservice/web/security/RoleDimensionApiFilterRequestMapper.java
```diff
@@ -0,0 +1,161 @@
...
+            );
+        }
+        Function<Map.Entry<Dimension, Set<ApiFilter>>, Set<ApiFilter>> substituteFilters = entry ->
+                entry.getKey().equals(dimension) ?
+                        StreamUtils.setMerge(entry.getValue(), mergedSecurityFilters)
+                        : entry.getValue();
+
+        Map<Dimension, Set<ApiFilter>> newMap =  request.getFilters().entrySet().stream()
+                .map(entry -> new AbstractMap.SimpleEntry<>(entry.getKey(), substituteFilters.apply(entry)))
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
```
archolewa: I think you're better off using `Collectors.toMap`:

```java
        entrySet().stream().collect(Collectors.toMap(Map.Entry::getKey,   substituteFilters::apply);
```

fili-security/src/main/java/com/yahoo/bard/webservice/web/security/RoleDimensionApiFilterRequestMapper.java
```diff
@@ -0,0 +1,161 @@
...
+        if (mergedSecurityFilters.size() == 0) {
+            String name = securityContext.getUserPrincipal().getName();
+            LOG.warn(DIMENSION_MISSING_MANDATORY_ROLE.logFormat(name, dimension.getApiName()));
+            throw new RequestValidationException(
+                    Response.Status.FORBIDDEN,
+                    unauthorizedHttpMessage,
+                    UNAUTHORIZED_USER_MESSAGE
+            );
+        }
+        Function<Map.Entry<Dimension, Set<ApiFilter>>, Set<ApiFilter>> substituteFilters = entry ->
```
archolewa: Much as I love me some in-line functions, we might be better off making this an instance method that takes the entry and `mergedSecurityFilters` as arguments. It'll save us having to build this function over and over again when building security filters. Normally I wouldn't care, but if customers are using it, this code will probably get called on _every_ request, and since I think both an instance method and an in-line function are equally simple and readable in this case...

fili-security/src/main/java/com/yahoo/bard/webservice/web/security/RoleDimensionApiFilterRequestMapper.java
```diff
@@ -0,0 +1,161 @@
...
+            new ImmutableTriple<>(filter.getDimension(), filter.getDimensionField(), filter.getOperation());
+
+        Map<Triple<Dimension, DimensionField, FilterOperation>, Set<String>> filterMap =
+                filterStream.collect(Collectors.groupingBy(
+                        filterGroupingIdentity::apply,
+                        Collectors.mapping(
+                                ApiFilter::getValues,
+                                Collectors.reducing(Collections.emptySet(), StreamUtils::setMerge)
+                        )
+                ));
```
archolewa: It's so pretty...

archolewa:
Also, I'm going to send this PR to @aklish in case he some time to take a look.

fili-security/src/main/java/com/yahoo/bard/webservice/web/security/RoleDimensionApiFilterRequestMapper.java
```diff
@@ -0,0 +1,161 @@
...
+     * Constructor.
+     *
+     * @param resourceDictionaries  The dictionaries to use for request mapping.
+     * @param dimension  The dimension whose roles are being matched
+     * @param roleApiFilters  ApiFilters by role for a given dimension
+     */
+    public RoleDimensionApiFilterRequestMapper(
+            ResourceDictionaries resourceDictionaries,
+            Dimension dimension,
+            Map<String, Set<ApiFilter>> roleApiFilters
```
archolewa: Here's a question: Rather than mapping from user role to a set of filters, do we instead want to map from user role to a function? I'm thinking something that takes the DataApiRequest, and `ContainerRequestContext`. Customers may want to do something more complex than a static collection of filters for each role.

fili-core/src/main/java/com/yahoo/bard/webservice/web/ApiFilter.java
```diff
@@ -172,6 +173,26 @@ public FilterOperation getOperation() {
...
+     *
+     * @param one  The first ApiFilter
+     * @param two  The second ApiFilter
+     *
+     * @return an ApiFilter with the union of values
+     */
+    public static ApiFilter union(ApiFilter one, ApiFilter two) {
+        if (!Objects.equals(one.getDimension(), two.getDimension())
+                && Objects.equals(one.getDimensionField(), two.getDimensionField())
+                && Objects.equals(one.getOperation(), two.getOperation())
```
michael-mclawhorn: The first '!' is incorrect, it should have been a not distributed over the whole boolean.

fili-security/src/main/java/com/yahoo/bard/webservice/web/security/RoleBasedRoutingRequestMapper.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public T apply(T request, ContainerRequestContext context)
+            throws RequestValidationException {
+        SecurityContext securityContext = context.getSecurityContext();
+        RequestMapper<T> mapper = prioritizedRoleBasedMappers.keySet().stream()
+                .filter(securityContext::isUserInRole)
+                .peek(it -> System.out.println(it))
+                .map(prioritizedRoleBasedMappers::get)
+                .peek(it -> System.out.println(it))
```
michael-mclawhorn: Missed those on cleanup

fili-security/src/main/java/com/yahoo/bard/webservice/web/security/RoleDimensionApiFilterRequestMapper.java
michael-mclawhorn: Per conversation I'm going to punt on making a change around this.
The proposed change would involve allowing a Map<Role, Provider<Set<ApiFilter>> option in place of Map<Role, Set<ApiFilter>> (or a more n-ity general function than provider), to allow more complicated security arrangements.  Other code would not need to change.

fili-security/src/test/groovy/com/yahoo/bard/webservice/web/security/RoleBasedRoutingRequestMapperSpec.groovy
```diff
@@ -0,0 +1,67 @@
...
+    RequestMapper<ApiRequest> mapperB = Mock(RequestMapper)
+    RequestMapper<ApiRequest> mapperC = Mock(RequestMapper)
+
+    ContainerRequestContext containerRequestContext = Mock(ContainerRequestContext)
+    SecurityContext securityContext = Mock(SecurityContext)
+    RoleBasedRoutingRequestMapper mapper
+    Map<String, RequestMapper<ApiRequest>> mappers = [a: mapperA, b: mapperB, c:mapperC]
+
+    def setup() {
+        mapper = new RoleBasedRoutingRequestMapper(Mock(ResourceDictionaries), prioritizedRoleBasedMappers, mapperC)
```
garyluoex: Why there is two of this line? one here and one below?

fili-security/src/test/groovy/com/yahoo/bard/webservice/web/security/RoleBasedRoutingRequestMapperSpec.groovy
```diff
@@ -0,0 +1,67 @@
...
+
+import javax.ws.rs.container.ContainerRequestContext
+import javax.ws.rs.core.SecurityContext
+
+class RoleBasedRoutingRequestMapperSpec extends Specification {
+
+    LinkedHashMap<String, RequestMapper<ApiRequest>> prioritizedRoleBasedMappers;
+    RequestMapper<ApiRequest> mapperA = Mock(RequestMapper)
+    RequestMapper<ApiRequest> mapperB = Mock(RequestMapper)
+    RequestMapper<ApiRequest> mapperC = Mock(RequestMapper)
```
garyluoex: Are we using mapperC?

fili-security/src/test/groovy/com/yahoo/bard/webservice/web/security/RoleBasedRoutingRequestMapperSpec.groovy
```diff
@@ -0,0 +1,67 @@
...
+        ApiRequest apiRequest = Mock(ApiRequest)
+        setupRoles(roles)
+        RequestMapper delegate = prioritizedRoleBasedMappers[nextMapper]
+
+        [mapperA, mapperB, mapperC].each {
+            int count = mappers[nextMapper] == it ? 1 : 0
+            count * it.apply(apiRequest, containerRequestContext) >> apiRequest
+        }
+
+        expect:
```
garyluoex: This test is tricky.... we should add description here what we are actually expecting.

fili-security/src/test/groovy/com/yahoo/bard/webservice/web/security/RoleDimensionApiFilterRequestMapperSpec.groovy
```diff
@@ -0,0 +1,143 @@
...
+
+    def "validateSecurityFilters throws exception on empty security filters"() {
+        setup:
+        filterDimension.getApiName() >> "TestApiName"
+
+        when:
+        mapper.validateSecurityFilters(userPrincipal, [] as Set)
+
+        then:
+        thrown(RequestValidationException)
```
garyluoex: I would try to match the message

fili-navi/src/test/resources/testApplicationConfig.properties
```diff
@@ -1,7 +1,7 @@
-# Copyright 2017 Yahoo Inc.
+# Copyright 2016 Yahoo Inc.
 # Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 
-#The name of this module (required)
-moduleName = fili-navi
```
michael-mclawhorn: Module name belongs in moduleConfig.properties not testApplicationConfig.properties.  This was a bug.

fili-security/src/test/resources/testApplicationConfig.properties
```diff
@@ -0,0 +1,7 @@
+# Copyright 2017 Yahoo Inc.
+# Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+
+# Application configuration must specify a package name.  This prefix applies to all config properties.
+package_name = bard
+
+moduleDependencies = fili-core
```
michael-mclawhorn: Yep, this was a mistake.

fili-security/src/test/groovy/com/yahoo/bard/webservice/web/security/RoleBasedRoutingRequestMapperSpec.groovy
michael-mclawhorn: It's used down in line 29 building the mapper, and in the next mapper test at the bottom.

fili-security/src/test/groovy/com/yahoo/bard/webservice/web/security/RoleBasedRoutingRequestMapperSpec.groovy
michael-mclawhorn: This was a mistake.

fili-navi/src/test/resources/testApplicationConfig.properties
garyluoex: right

PR #407, opened 2017-06-19
Fix a bug where table loader uses nested compute if absent

garyluoex:


ghost (approved):
@michael-mclawhorn Not sure if we should make it more readable by iteratively building physical table instead of recursively.

michael-mclawhorn:
This makes a lot of sense to me.

michael-mclawhorn:
There should be tests around BaseTableLoader#buildPhysicalTableWithDependency.
I don't insist on them being added on this PR since it isn't changing, but I think we should make an issue to come back and add tests that might have caught this bug.

michael-mclawhorn (approved):
Approved on the condition of a tech debt issue to test this code.

garyluoex:
@michael-mclawhorn   technically the buildDimensionSpanningTableGroup tests the build dependency capability already, but we can discuss if anything is missing.

garyluoex:
@QubitPi   dependency cause building tables iteratively inefficient since it requires several pass to build everything, let me know if you know a way to get around that.

PR #408, opened 2017-06-20
Correct cache config comment

QubitPi:
* Make comment list out all caching strategy values that align with `CacheFeatureFlag` enum values
* Explain what each cache strategy means

### DIFF
--- a/fili-core/src/main/resources/moduleConfig.properties
+++ b/fili-core/src/main/resources/moduleConfig.properties
@@ -54,10 +54,10 @@ bard__memcached_expiration_seconds = 3600
 # If using memcached, make sure this aligns with the setting memcached is using, overridden with it's -I parameter
 bard__druid_max_response_length_to_cache = 9223372036854775807
 
-# Data Cache strategy, which is one of the following
-# 1. ETagOnly
-# 2. LocalSignatureOnly (cache V2, needs the above flag set as well)
-# 3. TtlOnly (cache V1)
+# Data Cache strategy, whose value is one of the following
+# 1. ETag
+# 2. LocalSignature (cache V2)
+# 3. Ttl (cache V1)
 # 4. NoCache
 druid__query_response_caching_strategy = LocalSignature
 
--- a/fili-core/src/main/resources/moduleConfig.properties
+++ b/fili-core/src/main/resources/moduleConfig.properties
@@ -57,7 +57,7 @@ bard__druid_max_response_length_to_cache = 9223372036854775807
 # Data Cache strategy, whose value is one of the following
 # 1. ETag
 #    This is the same thing as LocalSignature except that Etag optimizes to avoid the possibility of cache out-of-sync
-#    problem
+#    problem. ETag relies on Druid to generate the signature(hash) instead of Fili
 # 2. LocalSignature (cache V2, deprecated)
 #    Maintains a mapping of (query, druidSegmentHash) -> DruidResponse. When a query comes in, Fili finds and computes
 #    hash of all segments involved in that query. The query itself and the segment hash are combined together as the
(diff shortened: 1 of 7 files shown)


fili-core/src/main/resources/moduleConfig.properties
```diff
@@ -54,10 +54,10 @@ bard__memcached_expiration_seconds = 3600
...
 bard__druid_max_response_length_to_cache = 9223372036854775807
 
-# Data Cache strategy, which is one of the following
-# 1. ETagOnly
-# 2. LocalSignatureOnly (cache V2, needs the above flag set as well)
-# 3. TtlOnly (cache V1)
+# Data Cache strategy, whose value is one of the following
+# 1. ETag
+# 2. LocalSignature (cache V2, needs the above flag set as well)
+# 3. Ttl (cache V1)
```
michael-mclawhorn: I'm pretty sure that the 'needs the above flag set as well' comment is wrong.

fili-core/src/main/resources/moduleConfig.properties
```diff
@@ -54,10 +54,10 @@ bard__memcached_expiration_seconds = 3600
...
 bard__druid_max_response_length_to_cache = 9223372036854775807
 
-# Data Cache strategy, which is one of the following
-# 1. ETagOnly
-# 2. LocalSignatureOnly (cache V2, needs the above flag set as well)
-# 3. TtlOnly (cache V1)
+# Data Cache strategy, whose value is one of the following
+# 1. ETag
+# 2. LocalSignature (cache V2)
+# 3. Ttl (cache V1)
```
archolewa: What does `Ttl` stand for?

archolewa: Some more detailed explanations about all three of these options would be useful. In particular we should specify for each one:

1. What is used as the cache key for each strategy.
2. What is involved in computing each key.
3. The limitations that each strategy possesses.

So that customers have some idea of what they're picking.

fili-core/src/test/resources/testApplicationConfig.properties
```diff
@@ -48,10 +48,21 @@ bard__druid_seg_loader_timer_delay = 60000
...
 
-# Data Cache strategy, which is one of the following
-# 1. ETagOnly
-# 2. LocalSignatureOnly (cache V2, needs the above flag set as well)
-# 3. TtlOnly (cache V1)
+# Data Cache strategy, whose value is one of the following
+# 1. ETag
+#    This is the same thing as LocalSignature except that Etag optimizes to avoid the possibility of cache out-of-sync
+#    problem
+# 2. LocalSignature (cache V2, deprecated)
```
michael-mclawhorn: I would make the point out that ETag relies on Druid to generate the signature instead of Fili.

fili-core/src/main/resources/moduleConfig.properties
```diff
@@ -54,10 +54,21 @@ bard__memcached_expiration_seconds = 3600
...
+#    Maintains a mapping of (query, druidSegmentHash) -> DruidResponse. When a query comes in, Fili finds and computes
+#    hash of all segments involved in that query. The query itself and the segment hash are combined together as the
+#    cache key. If it's a new query or Druid reprocessed so that the segments change, there won't be a cache key in the
+#    mapping and Fili talks to Druid to get new result. There is a possibility of cache out-of-sync problem when Druid
+#    reprocesses and the segment scanner updates
+# 3. Ttl (cache V1, deprecated)
+#    "Ttl" stands for "Time to live". This strategy does NOT account for reprocessing. If you run a query, a cache is
+#    generated with a TTL period. Until the cache entry expires(TTL > 0), same result will be sent back on the same
+#    query; this is the case even when Druid has updated its data(result). Client using this cache strategy can
+#    potentially receive staled result until the cache entry for that query expires.
```
wcekan: I would say `stale results`

Issue #392, opened 2017-06-09
Better programmatic generation of metadata json in tests


cdeszaq:
We should rework metadata tests to be more generated from strings and more pluggable to support heavier and more expressive testing. This would allow for more consistency, as well as make it easier to test more cases if we do it right.

QubitPi:
I'm working on it

QubitPi:
Addressed by https://github.com/yahoo/fili/pull/412

PR #412, opened 2017-06-23
Better programmatic generation of metadata json in tests

QubitPi:
Address issue https://github.com/yahoo/fili/issues/392

The strategy is to have all metadata specs under `com.yahoo.bard.webservice.metadata` to extend `BaseDataSourceMetadataSpec`, which is a base specification of all metadata tests. All metadata Specs override `childSetupSpec()`, and call/override` generate*()` methods to initiate metadata related resources.

Take "intervals" as an example:

1. you don't need any of the resources(including "intervals") defined in base Spec, you do nothing
2. you need "intervals" in your test, you override `childSetupSpec()` and call its generation method in your Spec as follows

```java
@Override
def childSetupSpec() {
    intervals = generateIntervals()
}
```
3. you need "intervals" in your test, but you need a different set of intervals:

```java
@Override
def childSetupSpec() {
    intervals = generateIntervals()
}

@Override
Map<String, Interval> generateIntervals() {
    [
        "interval1": Interval.parse("2015-01-01T00:00:00.000Z/2015-01-02T00:00:00.000Z"),
        "interval2": Interval.parse("2015-01-02T00:00:00.000Z/2015-01-03T00:00:00.000Z"),
        "interval3": Interval.parse("2015-01-03T00:00:00.000Z/2015-01-04T00:00:00.000Z"),
        "interval123": Interval.parse("2015-01-01T00:00:00.000Z/2015-01-04T00:00:00.000Z")
    ]
}
```

### DIFF
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/BaseDataSourceMetadataSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/BaseDataSourceMetadataSpec.groovy
@@ -256,8 +256,8 @@ abstract class BaseDataSourceMetadataSpec extends Specification {
                         intervals.interval1,
                         versions.version1,
                         null,
-                        dimensions.keySet().toList(),
-                        metrics.keySet().toList(),
+                        dimensions.keySet() as List,
+                        metrics.keySet() as List,
                         partition1,
                         binaryVersions.binaryVersion1,
                         sizes.size1
@@ -267,8 +267,8 @@ abstract class BaseDataSourceMetadataSpec extends Specification {
                         intervals.interval1,
                         versions.version2,
                         null,
-                        dimensions.keySet().toList(),
-                        metrics.keySet().toList(),
+                        dimensions.keySet() as List,
+                        metrics.keySet() as List,
                         partition2,
                         binaryVersions.binaryVersion1,
                         sizes.size2
@@ -278,8 +278,8 @@ abstract class BaseDataSourceMetadataSpec extends Specification {
                         intervals.interval2,
                         versions.version1,
                         null,
-                        dimensions.keySet().toList(),
-                        metrics.keySet().toList(),
+                        dimensions.keySet() as List,
+                        metrics.keySet() as List,
                         partition1,
                         binaryVersions.binaryVersion1,
                         sizes.size1
@@ -289,8 +289,8 @@ abstract class BaseDataSourceMetadataSpec extends Specification {
                         intervals.interval2,
                         versions.version2,
                         null,
-                        dimensions.keySet().toList(),
-                        metrics.keySet().toList(),
+                        dimensions.keySet() as List,
+                        metrics.keySet() as List,
                         partition2,
                         binaryVersions.binaryVersion1,
                         sizes.size2
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/BaseDataSourceMetadataSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/BaseDataSourceMetadataSpec.groovy
@@ -29,7 +29,7 @@ import spock.lang.Specification
  * Take "intervals" as an example
  * <ul>
  *     <li>
- *         you don't need any of the resources(including "intervals") defined in metadata tests, you do nothing and
+ *         you don't need any of the resources(including "intervals") defined in this Spec, you do nothing and
  *         leave childSetupSpec() empty as it is in this Spec
  *     </li>
  *     <li>
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/BaseDataSourceMetadataSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/BaseDataSourceMetadataSpec.groovy
@@ -20,82 +20,129 @@ import io.druid.timeline.DataSegment
 import spock.lang.Shared
 import spock.lang.Specification
 
-class BaseDataSourceMetadataSpec extends Specification {
-    @Shared
-    String tableName = TestDruidTableName.ALL_PETS.asName()
-
-    @Shared
-    String intervalString1
-    @Shared
-    String intervalString2
-    @Shared
-    String intervalString12
-
-    @Shared
-    Interval interval1
-    @Shared
-    Interval interval2
-    @Shared
-    Interval interval12
-
-    @Shared
-    RangeSet<DateTime> rangeSet12
-
-    @Shared
-    String version1
-    @Shared
-    String version2
-
-    @Shared
-    DataSegment segment1
-
-    @Shared
-    DataSegment segment2
-
-    @Shared
-    DataSegment segment3
-
-    @Shared
-    DataSegment segment4
-
-    @Shared
-    List<DataSegment> segments
-
+/**
+ * Base specification of all metadata tests.
+ * <p>
+ * Any metadata Specs should extend this Spec, override childSetupSpec(), and call/override generate*() methods to
+ * initiate metadata related resources.
+ * <p>
+ * Take "intervals" as an example
+ * <ul>
+ *     <li>
+ *         you don't need any of the resources(including "intervals") defined in metadata tests, you do nothing and
+ *         leave childSetupSpec() empty as it is in this Spec
+ *     </li>
+ *     <li>
+ *         you need "intervals" in your test, you override childSetupSpec() and call its generation method in your Spec
+ *         as follows
+ *         <pre>
+ *             {@code
+ *             @Override
+ *             def childSetupSpec() {
+ *                 intervals = generateIntervals()
+ *             }
+ *             }
+ *         </pre>
+ *     </li>
+ *     <li>
+ *         you need "intervals" in your test, but you need a different set of intervals:
+ *         <pre>
+ *             {@code
+ *             @Override
+ *             def childSetupSpec() {
+ *                 intervals = generateIntervals()
+ *             }
+ *
+ *             @Override
+ *             Map<String, Interval> generateIntervals() {
+ *                 [
+ *                     "interval1": Interval.parse("2015-01-01T00:00:00.000Z/2015-01-02T00:00:00.000Z"),
+ *                     "interval2": Interval.parse("2015-01-02T00:00:00.000Z/2015-01-03T00:00:00.000Z"),
+ *                     "interval3": Interval.parse("2015-01-03T00:00:00.000Z/2015-01-04T00:00:00.000Z"),
+ *                     "interval123": Interval.parse("2015-01-01T00:00:00.000Z/2015-01-04T00:00:00.000Z")
+ *                 ]
+ *             }
+ *             }
+ *         </pre>
+ *     </li>
+ * </ul>
+ */
+abstract class BaseDataSourceMetadataSpec extends Specification {
     @Shared
     DateTimeZone currentTZ
 
     @Shared
-    List<String> dimensions123
-
+    String tableName
     @Shared
-    List<String> metrics123
+    Map<String, Interval> intervals
+    @Shared
+    RangeSet<DateTime> rangeSet
+    @Shared
+    List<TestApiDimensionName> dimensions
+    @Shared
+    List<TestApiMetricName> metrics
+    @Shared
+    Integer binversion1
+    @Shared
+    long size1
+    @Shared
+    long size2
+    @Shared
+    List<String> versions
+    @Shared
+    List<DataSegment> segments
 
     def setupSpec() {
         currentTZ = DateTimeZone.getDefault()
         DateTimeZone.setDefault(DateTimeZone.UTC)
+        binversion1 = 9
+        size1 = 1024
+        size2 = 512
 
-        intervalString1 = "2015-01-01T00:00:00.000Z/2015-01-02T00:00:00.000Z"
-        intervalString2 = "2015-01-02T00:00:00.000Z/2015-01-03T00:00:00.000Z"
-        intervalString12 = "2015-01-01T00:00:00.000Z/2015-01-03T00:00:00.000Z"
-        interval1 = Interval.parse(intervalString1)
-        interval2 = Interval.parse(intervalString2)
-        interval12 = Interval.parse(intervalString12)
-        rangeSet12 = TreeRangeSet.create()
-        rangeSet12.add(Range.closedOpen(interval12.getStart(), interval12.getEnd()))
-        version1 = DateTimeFormat.fullDateTime().print(DateTime.now().minusDays(1))
-        version2 = DateTimeFormat.fullDateTime().print(DateTime.now())
+        childSetupSpec()
+    }
 
-        TestApiDimensionName dim1 = TestApiDimensionName.BREED
-        TestApiDimensionName dim2 = TestApiDimensionName.SPECIES
-        TestApiDimensionName dim3 = TestApiDimensionName.SEX
+    def childSetupSpec() {}
 
-        dimensions123 = [dim1, dim2, dim3]*.asName()
+    def shutdownSpec() {
+        DateTimeZone.setDefault(currentTZ)
+    }
 
-        TestApiMetricName met1 = TestApiMetricName.A_ROW_NUM
-        TestApiMetricName met2 = TestApiMetricName.A_LIMBS
-        TestApiMetricName met3 = TestApiMetricName.A_DAY_AVG_LIMBS
+    String generateTableName() {
+        TestDruidTableName.ALL_PETS.asName()
+    }
 
-        metrics123 = [met1, met2, met3]*.asName()
+    Map<String, Interval> generateIntervals() {
+        [
+                "interval1": Interval.parse("2015-01-01T00:00:00.000Z/2015-01-02T00:00:00.000Z"),
+                "interval2": Interval.parse("2015-01-02T00:00:00.000Z/2015-01-03T00:00:00.000Z"),
+                "interval12": Interval.parse("2015-01-01T00:00:00.000Z/2015-01-03T00:00:00.000Z")
+        ]
+    }
+
+    RangeSet<DateTime> generateRangeSet() {
+        rangeSet = TreeRangeSet.create()
+        rangeSet.add(Range.closedOpen(intervals["interval12"].getStart(), intervals["interval12"].getEnd()))
+        rangeSet
+    }
+
+    List<TestApiDimensionName> generateDimensions() {
+        [TestApiDimensionName.BREED, TestApiDimensionName.SPECIES, TestApiDimensionName.SEX]
+    }
+
+    List<TestApiMetricName> generateMetrics() {
+        [TestApiMetricName.A_ROW_NUM, TestApiMetricName.A_LIMBS, TestApiMetricName.A_DAY_AVG_LIMBS]
+    }
+
+    List<String> generateVersions() {
+        [
+                DateTimeFormat.fullDateTime().print(DateTime.now().minusDays(1)),
+                DateTimeFormat.fullDateTime().print(DateTime.now())
+        ]
+    }
+
+    List<DataSegment> generateSegments() {
+        versions = generateVersions()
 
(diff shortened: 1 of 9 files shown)


fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/BaseDataSourceMetadataSpec.groovy
```diff
@@ -20,82 +20,128 @@ import io.druid.timeline.DataSegment
...
-    Interval interval12
-
+    List<TestApiDimensionName> dimensions
     @Shared
-    RangeSet<DateTime> rangeSet12
-
+    List<TestApiMetricName> metrics
     @Shared
-    String version1
+    Integer binversion1
```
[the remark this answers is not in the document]
archolewa: And why is it an `Integer` rather than an `int`?

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/BaseDataSourceMetadataSpec.groovy
```diff
@@ -20,82 +20,128 @@ import io.druid.timeline.DataSegment
...
     @Shared
-    String version2
-
+    long size1
     @Shared
-    DataSegment segment1
-
+    long size2
```
archolewa: What are these the size of?

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/BaseDataSourceMetadataSpec.groovy
```diff
@@ -20,82 +20,128 @@ import io.druid.timeline.DataSegment
...
+        childSetupSpec()
+    }
 
-    @Shared
-    DateTimeZone currentTZ
+    def childSetupSpec() {}
 
-    @Shared
-    List<String> dimensions123
+    def shutdownSpec() {
```
archolewa: This should be `cleanupSpec`. Also if we're going to have a `childSetupSpec` then we should have a `childCleanupSpec` as well, so that people can cleanup whatever they setup.

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/BaseDataSourceMetadataSpec.groovy
```diff
@@ -20,82 +20,128 @@ import io.druid.timeline.DataSegment
...
+    }
 
-    def setupSpec() {
-        currentTZ = DateTimeZone.getDefault()
-        DateTimeZone.setDefault(DateTimeZone.UTC)
+    Map<String, Interval> generateIntervals() {
+        [
+                "interval1": Interval.parse("2015-01-01T00:00:00.000Z/2015-01-02T00:00:00.000Z"),
+                "interval2": Interval.parse("2015-01-02T00:00:00.000Z/2015-01-03T00:00:00.000Z"),
+                "interval12": Interval.parse("2015-01-01T00:00:00.000Z/2015-01-03T00:00:00.000Z")
```
archolewa: You don't need to wrap quotes around String keys in Groovy:

```groovy
     interval1: Interval.parse(...),
     interval2: Interval.parse(...),
     ...
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/BaseDataSourceMetadataSpec.groovy
```diff
@@ -20,82 +20,128 @@ import io.druid.timeline.DataSegment
...
-        interval1 = Interval.parse(intervalString1)
-        interval2 = Interval.parse(intervalString2)
-        interval12 = Interval.parse(intervalString12)
-        rangeSet12 = TreeRangeSet.create()
-        rangeSet12.add(Range.closedOpen(interval12.getStart(), interval12.getEnd()))
-        version1 = DateTimeFormat.fullDateTime().print(DateTime.now().minusDays(1))
-        version2 = DateTimeFormat.fullDateTime().print(DateTime.now())
+    RangeSet<DateTime> generateRangeSet() {
+        rangeSet = TreeRangeSet.create()
+        rangeSet.add(Range.closedOpen(intervals["interval12"].getStart(), intervals["interval12"].getEnd()))
```
archolewa: Similarly, when accessing string keys in a map, you can just write `map.key`:

```groovy
        rangeSet.add(Range.closedOpen(intervals.interval12.getStart(), intervals.intervals12.getEnd
```
Though since you're pulling out interval12 twice, I'd probably pull it out into a variable:

```groovy
       Interval interval12 = intervals.interval12
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/BaseDataSourceMetadataSpec.groovy
```diff
@@ -105,63 +151,51 @@ class BaseDataSourceMetadataSpec extends Specification {
...
-        segments = [segment1, segment2, segment3, segment4]
-
-    }
-
-    def shutdownSpec() {
-        DateTimeZone.setDefault(currentTZ)
+        [
+                new DataSegment(
+                        tableName,
+                        intervals["interval1"],
```
archolewa: Groovy has a syntactic shortcut for accessing string keys: `intervals.interval1`

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/BaseDataSourceMetadataSpec.groovy
```diff
@@ -105,63 +151,51 @@ class BaseDataSourceMetadataSpec extends Specification {
...
-                size2
-        )
-
-        segments = [segment1, segment2, segment3, segment4]
-
-    }
-
-    def shutdownSpec() {
-        DateTimeZone.setDefault(currentTZ)
+        [
```
archolewa: Personally, I prefer to use an explicit `return` statement in functions/methods that are longer than one or two lines. Makes it clearer what you're doing with that last line.

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/DataSourceMetadataLoadTaskSpec.groovy
```diff
@@ -170,10 +145,10 @@ class DataSourceMetadataLoadTaskSpec extends Specification {
         druidWS.jsonResponse = {fullDataSourceMetadataJson}
 
         expectedIntervalsMap = [:]
-        dimensions123.each {
-            expectedIntervalsMap.put(new DimensionColumn(dimensionDict.findByApiName(it)), [interval123] as Set)
+        dimensions.each {
```
archolewa: Rather than building an empty map, and then using an `each` to populate it, I would initialize it from the beginning:

```groovy
expectedIntervalsMap = dimensions123
              .collect { dimensionDict.findByApiName(it) }
              .collectEntries{[(it): [interval123] as Set]} 
```

And then use assignment syntax for the metrics:

```groovy
    metrics123
            .collect { new MetricColumn(it)}
            .each { expectedIntervalsMap[it] = [intervals.interval123] as Set }
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/BaseDataSourceMetadataSpec.groovy
```diff
@@ -20,82 +20,128 @@ import io.druid.timeline.DataSegment
...
+    RangeSet<DateTime> generateRangeSet() {
+        rangeSet = TreeRangeSet.create()
+        rangeSet.add(Range.closedOpen(intervals["interval12"].getStart(), intervals["interval12"].getEnd()))
+        rangeSet
+    }
 
-        TestApiDimensionName dim1 = TestApiDimensionName.BREED
-        TestApiDimensionName dim2 = TestApiDimensionName.SPECIES
-        TestApiDimensionName dim3 = TestApiDimensionName.SEX
+    List<TestApiDimensionName> generateDimensions() {
```
archolewa: Rather than making this a `list`, I would make this a `Map<String, TestApiDimensionName>`. Then, in extending classes, you can use `dimensions.dim3.asName()` rather than `dimensions[2].asName()`. This allows people writing specs to give their dimensions meaningful names and then use them (though admittedly the names we typically use aren't all that more meaningful than `dimensions[2]`).

The same applies to metrics.

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/DataSourceMetadataLoadTaskSpec.groovy
```diff
@@ -24,46 +21,36 @@ import com.yahoo.bard.webservice.table.ConstrainedTable
...
-    TestApiDimensionName dim1 = TestApiDimensionName.BREED
-    TestApiDimensionName dim2 = TestApiDimensionName.SPECIES
-    TestApiDimensionName dim3 = TestApiDimensionName.SEX
+    @Override
+    def childSetupSpec() {
+        tableName = generateTableName()
+        intervals = generateIntervals()
+        dimensions = generateDimensions()
+        metrics = generateMetrics()
+        versions = generateVersions()
```
archolewa: We keep doing this in children classes. Since this simply assigns fields that are in the parent class, why not move this into the parent class?

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/BaseDataSourceMetadataSpec.groovy
QubitPi: @archolewa Beautiful! :+1: This is exactly what I was considering before looking at this comment

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/BaseDataSourceMetadataSpec.groovy
QubitPi: @archolewa Thank you very much for the tip!

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/BaseDataSourceMetadataSpec.groovy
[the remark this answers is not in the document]
QubitPi: @archolewa A binary version is used as a required argument to construct a `DataSegment`. That argument is an `Integer`

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/BaseDataSourceMetadataSpec.groovy
QubitPi: @archolewa Same as `binVersion` above

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/DataSourceMetadataLoadTaskSpec.groovy
QubitPi: @archolewa Would you like to move `versions = generateVersions()` or all other `generate*()`'s ?

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/DataSourceMetadataLoadTaskSpec.groovy
QubitPi: @archolewa Thank you very much for teaching me this good practice! I'll take a note on this :wink:

QubitPi:
@archolewa I applied the dictionary method to all shared variables, because I think your suggestion is such a good idea that it should be applied more broadly. 

In addition, in the last commit, I cleaned up all meta data tests a little bit(organizing init stuff to `setup()` method and applied more Groovy smart syntax, etc)

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/DataSourceMetadataLoadTaskSpec.groovy
archolewa: I meant the entire method, sorry I wasn't clear.

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/DataSourceMetadataLoadTaskSpec.groovy
```diff
@@ -46,109 +70,73 @@ class DataSourceMetadataLoadTaskSpec extends BaseDataSourceMetadataSpec {
...
-                            "partitions": $partitions
-                        },
-                        "binaryVersion": $binVersion,
-                        "size": $size,
-                        "identifier": ""
-        }"""
-    }
+    def setup() {
+        dimensions13 = [
+                dimensions.(TestApiDimensionName.BREED.asName()).asName(),
```
archolewa: I didn't know you could do that (wrap `TestApiDimensionName.BREED.asName()` in parenthesis to use the dot syntax, rather than the `[]` syntax).

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/DataSourceMetadataLoadTaskSpec.groovy
```diff
@@ -46,109 +70,73 @@ class DataSourceMetadataLoadTaskSpec extends BaseDataSourceMetadataSpec {
...
+                dimensions.(TestApiDimensionName.SEX.asName()).asName()
+        ]
+        quotedDimensions = dimensions13.collect() { '"' + it + '"'}
```
archolewa: Any particular reason you can't do `'"$it"'`?

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/DataSourceMetadataLoadTaskSpec.groovy
QubitPi: @archolewa I did that just incase if people change `BREED` API name in the future, they don't break this test 😌  Does that look good to you?

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/DataSourceMetadataLoadTaskSpec.groovy
QubitPi: I should do that, yes. Thanks for the doc link 😉

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/DataSourceMetadataLoadTaskSpec.groovy
QubitPi: Well, If we move all of these and (in some cases) we override `generateIntervals()` in child Spec, for example, then other `generate` method might throw an error. For instance

```java
    @Override
    Map<String, Interval> generateIntervals() {
        [
                "interval1": Interval.parse(intervalString1),
                "interval3": Interval.parse(intervalString3),
                "interval13": Interval.parse(intervalString13),
                "interval23": Interval.parse(intervalString23)
        ]
    }
```

Then this default in base Spec won't work, because `intervals.interval2` does not exist

```java
    Map<String, DataSegment> generateSegments() {
        NumberedShardSpec partition1 = Stub(NumberedShardSpec) {
            getPartitionNum() >> 0
        }

        NumberedShardSpec partition2 = Stub(NumberedShardSpec) {
            getPartitionNum() >> 1
        }

        return [
                segment1: new DataSegment(
                        tableName,
                        intervals.interval1,
                        versions.version1,
                        null,
                        dimensions.keySet().toList(),
                        metrics.keySet().toList(),
                        partition1,
                        binaryVersions.binaryVersion1,
                        sizes.size1
                ),
                segment2: new DataSegment(
                        tableName,
                        intervals.interval1,
                        versions.version2,
                        null,
                        dimensions.keySet().toList(),
                        metrics.keySet().toList(),
                        partition2,
                        binaryVersions.binaryVersion1,
                        sizes.size2
                ),
                segment3: new DataSegment(
                        tableName,
                        intervals.interval2,
                        versions.version1,
                        null,
                        dimensions.keySet().toList(),
                        metrics.keySet().toList(),
                        partition1,
                        binaryVersions.binaryVersion1,
                        sizes.size1
                ),
                segment4: new DataSegment(
                        tableName,
                        intervals.interval2,
                        versions.version2,
                        null,
                        dimensions.keySet().toList(),
                        metrics.keySet().toList(),
                        partition2,
                        binaryVersions.binaryVersion1,
                        sizes.size2
                )
        ]
    }
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/metadata/DataSourceMetadataLoadTaskSpec.groovy
```diff
@@ -24,121 +22,23 @@ import com.yahoo.bard.webservice.table.ConstrainedTable
...
-    String tableName = TestDruidTableName.ALL_PETS.asName()
+    Interval interval1
+    Interval interval2
+    Interval interval3
 
-    Interval interval1 = Interval.parse("2015-01-01T00:00:00.000Z/2015-01-02T00:00:00.000Z")
-    Interval interval2 = Interval.parse("2015-01-02T00:00:00.000Z/2015-01-03T00:00:00.000Z")
-    Interval interval3 = Interval.parse("2015-01-03T00:00:00.000Z/2015-01-04T00:00:00.000Z")
-    Interval interval123 = Interval.parse("2015-01-01T00:00:00.000Z/2015-01-04T00:00:00.000Z")
+    List<String> dimensions13
```
tarrantzhang: In general, one thing your strategy confuse me is that in the original Spec, we have something like:
```
List<String> dimensions123 = [dim1, dim2, dim3]*.asName()		
List<String> dimensions13 = [dim1, dim3]*.asName()
```
Your strategy choose to use `generateDimensions()` to populate your new `dimensions123` as the shared variable  `dimensions`, but for `dimensions13` it's populate in the `setup()`. It's a bit inconsistent for me that two similar variables in the old Spec get to be set in different ways. 

I might be wrong, but here is my suggestion:
1. Define all `generateBlah()` in `BaseDataSourceMetadataSpec` and for each `generateBlah()`, populate the map with all possible data you will be using in all child Spec, you can rename the keys to distinguish between which child Spec will be using it:
```
    Map<String, Interval> generateIntervals() {
        return [
                loadTaskSpecInterval1: Interval.parse("2015-01-01T00:00:00.000Z/2015-01-02T00:00:00.000Z"),
                loadTaskSpecInterval2: Interval.parse("2015-01-02T00:00:00.000Z/2015-01-03T00:00:00.000Z"),
                loadTaskSpecInterval3: Interval.parse("2015-01-01T00:00:00.000Z/2015-01-03T00:00:00.000Z"),
                SegmentSpecInterval1: Interval.parse("2015-01-01T00:00:00.000Z/2015-01-04T00:00:00.000Z"),
                ...
        ]
    }
```

2. Don't override any `generateBlah()` in your child Spec, then in your `setup()`:
```
def setup() {
        Interval loadTaskSpecInterval1 = intervals.loadTaskSpecInterval1
        Interval loadTaskSpecInterval2 = intervals.loadTaskSpecInterval2
        ....
```

With these change you can avoid calling those `generateBlah()` functions over and over again in your `childSetupSepc()` as mentioned by @archolewa . You don't even need a `childSetupSepc()` because the parent Spec does all the population for you.  😼

QubitPi: `dimensions123` is gone. `dimensions` serves as a resource container for all dimensions that are to be used in a Spec

The reason for not generating these resources but having child extend and override is that, for example, if you override `generateInterval`, `generateSegment` throws error in other Specs, since segments depends on generated intervals

If we initialize all the resources  in the way you recommended, we didn't solve the old problem because we we just changing variable into strings.

tarrantzhang: Sounds reasonable

QubitPi:
Thanks @archolewa !
Thanks @tarrantzhang for helping to review at late night 👍 !

PR #413, opened 2017-06-26
Partial implementation of a sql backend.

kevinhinterlong:
- Generates/Evaluates Having, Filters, and PostAggregations (with tests)
- Generates aggregations for sql
- Minor helper code for making connections and building sql queries
- Helper code for making tests quickly

**NOTE:** This is actually around 2500 lines and a lot of it is fairly simple helper code or generated code.


fili-core/src/main/java/com/yahoo/bard/webservice/sql/evaluator/FilterEvaluator.java
```diff
@@ -0,0 +1,281 @@
...
+     * @param filter  The filter to be evaluated.
+     * @param dimensions  The list of dimensions already found.
+     *
+     * @return a RexNode containing an equivalent filter to the one given.
+     */
+    private static RexNode evaluate(RelBuilder builder, Filter filter, List<String> dimensions) {
+        if (filter == null) {
+            return null;
+        }
+        return DispatchUtils.dispatch(
```
archolewa: Why are we doing this? Why not just build (take as an argument) an instance of `FilterEvaluator` and invoke the method directly?

fili-core/src/main/java/com/yahoo/bard/webservice/sql/helper/DatabaseHelper.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Finds the name of the Timestamp column in a sql database.
+     *
+     * @param connection  The connection to the database.
+     * @param tableWithSchema  The name of the table with schema (i.e. "public"."table").
+     *
+     * @return the name of the timestamp column.
+     *
+     * @throws SQLException if failed while reading database.
```
kevinhinterlong: Yea, I have yet to test different date types in SQL so I've limited it to only being a timestamp, but was planning on trying out `CAST` later.

I've logged the column name/documented the exception for now, but I agree it might be best to just remove this, keep the requirement for a timestamp column, and require that the column is explicitly specified.

fili-core/src/main/java/com/yahoo/bard/webservice/sql/helper/CalciteHelper.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /**
+     * Gets the name of a table with it's schema (i.e. table to "schema"."table").
+     *
+     * @param sqlTableName  The name of the table to escape.
+     *
+     * @return the escaped schema/table
+     */
+    public String escapeTableName(String sqlTableName) {
```
kevinhinterlong: `getTimestampColumn(...)` now takes the schema, and table name separately and unescaped for `DataBaseMetadata::getColumns` so this now just returns `schemaName` 👍

fili-core/src/main/java/com/yahoo/bard/webservice/sql/helper/TimeConverter.java
```diff
@@ -0,0 +1,197 @@
...
+    public static DateTime parseDateTime(int offset, String[] row, Granularity granularity) {
+        DefaultTimeGrain timeGrain = getDefaultTimeGrain(granularity);
+        DateTime resultTimeStamp = new DateTime(0, DateTimeZone.UTC);
+
+        List<SqlDatePartFunction> times = TIMEGRAIN_TO_GROUPBY.get(timeGrain);
+        for (int i = 0; i < times.size(); i++) {
+            int value = Integer.parseInt(row[offset + i]);
+            SqlDatePartFunction fn = times.get(i);
+            resultTimeStamp = setDateTime(value, fn, resultTimeStamp);
+        }
```
kevinhinterlong: I'm not sure I understand the

> That would save us having to use the `SqlDatePartFunction`

Wouldn't we have to use it for the delimiters?

Also is there a delimiter which corresponds to `WEEK`?

fili-core/src/main/java/com/yahoo/bard/webservice/sql/helper/SqlAggregationType.java
```diff
@@ -0,0 +1,89 @@
...
+import org.apache.calcite.sql.SqlAggFunction;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.tools.RelBuilder;
+
+import java.util.Locale;
+
+/**
+ * All the aggregation types supported for use with a sql backend.
+ */
+public enum SqlAggregationType {
```
kevinhinterlong: Ok thanks for the suggestion! I've implemented this strategy, but haven't yet hooked it up in the BinderFactory (I'll do that in the next pull request)

fili-core/src/main/java/com/yahoo/bard/webservice/sql/helper/SqlTimeConverter.java
```diff
@@ -0,0 +1,130 @@
...
+     * @return the datetime for the start of the interval.
+     */
+    default DateTime getIntervalStart(int offset, String[] row, TimeGrain timeGrain) {
+        DateTime resultTimeStamp = new DateTime(0, DateTimeZone.UTC);
+
+        List<SqlDatePartFunction> times = timeGrainToDatePartFunctions(timeGrain);
+        for (int i = 0; i < times.size(); i++) {
+            int value = Integer.parseInt(row[offset + i]);
+            SqlDatePartFunction fn = times.get(i);
+            resultTimeStamp = setDateTime(value, fn, resultTimeStamp);
```
kevinhinterlong: Not sure if there is a great way to do this with writing out the date as a string and then parsing it (as andrew suggested), but  if there is it would probably be in [ISODateTimeFormat](http://joda-time.sourceforge.net/apidocs/org/joda/time/format/ISODateTimeFormat.html).

fili-core/src/main/java/com/yahoo/bard/webservice/sql/helper/DatabaseHelper.java
```diff
@@ -0,0 +1,47 @@
...
+     * @param connection  The connection to the database.
+     * @param schema  The name of the schema (i.e. "PUBLIC").
+     * @param table  The name of the table (i.e. "table").
+     *
+     * @return the name of the timestamp column.
+     *
+     * @throws SQLException if failed while reading database.
+     * @throws IllegalStateException if no {@link JDBCType#TIMESTAMP} column could be found.
+     */
+    public static String getTimestampColumn(Connection connection, String schema, String table)
```
kevinhinterlong: While it would be nice to require that the Timestamp column is specified in the application config, it would get pretty difficult to set this up to work with multiple tables (having different names for their timestamp columns)

This might be better suited using the AbstractBinderFactory? @archolewa 
This could also take the role of specifying the schema for a table

fili-core/src/main/java/com/yahoo/bard/webservice/sql/helper/CalciteHelper.java
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+     * Initialize the helper with a datasource and it's schema.
+     *
+     * @param dataSource  The datasource to make connections with.
+     * @param schemaName  The name of the schema where the tables are stored.
+     *
+     * @throws SQLException if failed while making a connection to the database.
+     */
+    public CalciteHelper(DataSource dataSource, String schemaName)
```
kevinhinterlong: This subtly forces all tables to be within the same schema which might not be the case and could cause issues. It might be best to follow the AbstractBinder method and provide a interface mapping there for configuring the schema

fili-core/src/main/java/com/yahoo/bard/webservice/sql/helper/DatabaseHelper.java
archolewa: That's because specifying the name of the timestamp column doesn't belong in `application.config`. That file is for system wide properties and as you point out, the name of the timestamp column is specific to each table.

However, we already have a mechanism for configuring things that are specific to tables. It's called the table configuration (i.e. the `TableLoader`)! 

I propose we extend `Schema` to include the concept of a timestamp column. The current `Schema` would become the `DruidSchema` or something similar, since Druid tables don't need the timestamp column specified.

We could also include a default value, so that customers don't have to manually specify the timestamp column for every SQL table if they follow some convention (i.e. the table's name is timestamp). @michael-mclawhorn @cdeszaq thoughts?

PR #415, opened 2017-06-28
Fix wikipedia-example

### LINKED ISSUE #346: Replace Apache with AsyncHttpClient in Wikipedia and other examples
Currently, `WikiMain` uses Apache's http library. We should have that use Ning instead, because Ning is already used in the Fili ecosystem, while Apache's http is not.

kevinhinterlong:
Rename wikipedia data dimensions to match druid names
- The api names for accessing the druid dimensions are now all named correctly
to match the physical dimensions in Druid.
- Manually tested to make sure names are correct (can't make a test without integration testing)

Also fixed #346

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ Current
 
 
 ### Changed:
+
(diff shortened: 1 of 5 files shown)


michael-mclawhorn:
Please add a changelog explaining the change.

Issue #340, opened 2017-06-08
Remove dependency on json library


cdeszaq:
We have 2 different JSON processing libraries we depend on, `json.org` and `Jackson`. `json.org` causes problems for a number of reasons (it's not packaged in a way that's Java 8-friendly, and it has a non-standard versioning scheme), and Jackson is generally better (ie. more performant), so we should remove the dependency on `json.org`.

PR #416, opened 2017-06-28
Remove org.json dependency

kevinhinterlong:
Addresses issue #340 

- All org.json uses have been replaced with jackson equivalent

### DIFF
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/web/MetricParserSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/web/MetricParserSpec.groovy
@@ -2,26 +2,27 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.web
 
-import org.json.JSONArray
+import com.fasterxml.jackson.databind.ObjectMapper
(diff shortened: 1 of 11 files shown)


kevinhinterlong:
It looks like a lot of the code around `MetricParser` is using json where it doesn't need to be. If you think I should change it let me know

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/MetricParserSpec.groovy
```diff
@@ -2,26 +2,27 @@
...
 
 import spock.lang.Specification
 import spock.lang.Unroll
 
 class MetricParserSpec extends Specification {
-
-    JSONArray expectedJsonObj1
-    JSONArray expectedJsonObj2
-    JSONArray expectedJsonObj3
+    ObjectMapper mapper = new ObjectMapper()
```
archolewa: Rather than building an `ObjectMapper` directly, I would suggest going through the [ObjectMappersSuite](https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/application/ObjectMappersSuite.java). The ObjectMappersSuite That way you're using ObjectMappers with the same configuration as we do throughout the rest of the codebase.

michael-mclawhorn:
👍

Issue #375, opened 2017-06-09
Include ETags on responses


cdeszaq:
Letting clients know that a response is cached, etc. opens up a lot of ability for optimizing clients who poll for updated results.

QubitPi:
Addressed by https://github.com/yahoo/fili/pull/417

PR #417, opened 2017-06-28
Include ETags on responses

QubitPi:
https://github.com/yahoo/fili/issues/375

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,9 @@ Current
 
 ### Changed:
 
+- [Include ETags on responses](https://github.com/yahoo/fili/pull/417)
+    * Modify FullResponse JSON Objects to contain a flag showing whether a response is new or fetched from cache.
+
 - [Fix wrong default druid url and broken getInnerMostQuery](https://github.com/yahoo/fili/pull/528)
     * Comment out the wrong default druid broker url in module config that break old url config compatibility, add check for validate url in `DruidClientConfigHelper `
     * Fix broken `getInnermostQuery` method in `DruidQuery`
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,10 @@ Current
 
 ### Changed:
 
+- [Include ETags on responses](https://github.com/yahoo/fili/pull/417)
+    * Let clients know that weather a response is cached(new) or fetched from cache offers a lot of ability for
```
kevinhinterlong: know whether*

michael-mclawhorn: How about a description that includes the context that this only applies to FullResponses.
"'Modify FullResponse JSON Objects to contain a flag showing whether a response is new or fetched from cache."

michael-mclawhorn:
@QubitPi IS this PR finishable?

QubitPi:
@michael-mclawhorn It's definitely finishable. I'm asking @archolewa to take a look.

fili-core/src/main/java/com/yahoo/bard/webservice/web/responseprocessors/EtagCacheResponseProcessor.java
```diff
@@ -113,6 +113,11 @@ public void processResponse(JsonNode json, DruidAggregationQuery<?> druidQuery,
             }
         }
 
+        ((ObjectNode) json).put(
```
archolewa: Wouldn't this type of metadata belong in the header, rather than the body of the response?

QubitPi: Yes, this type of metadata is actually in the header at the moment. If I were going to put it in body of the response, I would have done something like 

```
body = json.get(DruidJsonResponseContentKeys.RESPONSE.getName())
body.put("cachedResponse", true/false)
```

archolewa: sorry I got a bit confused. It looks like we are augmenting the Druid response, not the Fili response. This comment no longer applies.

Issue #377, opened 2017-06-09
Add case-insensitive "contains" filter


cdeszaq:
It would be really nice to have a case-insensitive `contains` filter operator.

michael-mclawhorn:
We can't make this change without also adding a case sensitive contains filter.
Also, this behavior should probably have an update in the api-docs.

PR #418, opened 2017-06-28
Add case-insensitive contains filter

QubitPi:
https://github.com/yahoo/fili/issues/377

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Current
 
 ### Added:
 
-- [Add case-insensitive "contains" filter](https://github.com/yahoo/fili/issues/376)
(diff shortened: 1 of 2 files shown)


kevinhinterlong:
I read this issue as a new `containsi` operation in `FilterOperation` which would match the value of the filter case-insensitively as opposed to matching the `operationName` case-insensitively for "contains"

michael-mclawhorn:
No, the intent was to figure out how to do a case-insensitive contains operation against the Lucene search provider and possibly provide another named operator to indicate that case-insensitive mapping should apply.

We may actually want to make contains, by default be insensitive and move the case sensitive behavior to a new operator.

michael-mclawhorn:
This doesn't really address the issue.  I'm going to close this and it can be reopened again later.

Issue #359, opened 2017-06-09
Add Code Narc to validate Groovy style


cdeszaq:
Checkstyle is great, but it doesn't process Groovy. Code Narc is Checkstyle for Groovy, so we should totally use it.

QubitPi:
I'm working on it.

QubitPi:
Addressed by https://github.com/yahoo/fili/pull/420

PR #420, opened 2017-07-06
Add Code Narc to validate Groovy style

QubitPi:
https://github.com/yahoo/fili/issues/359


michael-mclawhorn:
👍 
I'm worried about whether this is going to blow us up with errors, but I'm willing to give it a try.

QubitPi:
I think it won't :) because it runs well on this master-based branch

PR #430, opened 2017-07-14
Sql backend implementation

kevinhinterlong:
@archolewa @michael-mclawhorn Here's the WIP Part 2 of https://github.com/yahoo/fili/pull/413

Here's what I can think of that needs to be done:
- [ ] [CLEANUP] Create a `SqlDimensionLoader` instead of relying on a backing druid instance to sql data
- [ ] [CLEANUP] Remove schema/timestampColumn from code and #435 
- [ ] [CLEANUP] Fix the `SqlRequestHandler` once we decide on a schema
- [ ] [DOCS] Mention that tables can have a custom timezone, but for your sanity you should only use UTC
- [ ] [DOCS] Fix instructions for configuring schema/timestamp column 
- [ ] [TESTS] Check that multiple intervals works fine more extensively (probably working)
- [ ] [CONFIG] Add support for custom `DruidQueryToSqlConverter` (`AbstractBinderFactory`? Custom `Filter`/`Having`/`PostAggregation` evaluator?)

Not yet implemented:
- [ ] [DruidAggregationQuery] Nested `GroupByQuery`
- [ ] [DruidAggregationQuery] `LookbackQuery`
- [ ] [DruidQuery] SearchQuery
- [ ] [DruidQuery] SelectQuery 
- [ ] [DruidQuery] DataSourceMetadataQuery 
- [ ] [DruidQuery] TimeBoundaryQuery


fili-core/src/main/java/com/yahoo/bard/webservice/sql/DruidQueryToSqlConverter.java
```diff
@@ -0,0 +1,444 @@
...
+                            )
+                            .sortLimit(
+                                    NONE,
+                                    getThreshold(druidQuery),
+                                    getSort(localBuilder, druidQuery, apiToFieldMapper)
+                            );
+                    String sql = writeSql(sqlWriter, relToSql, localBuilder);
+                    return "(" + sql + ")";
+                })
+                .collect(Collectors.joining("\nUNION ALL\n"));
```
kevinhinterlong: Ok so after looking into it further, it appears this is supported by MySQL/PostgreSQL (not sure what else), but is not supported by Standard SQL. So for example SQLite will fail with this method.

[Relevant Stackoverflow](https://stackoverflow.com/questions/4653124/what-does-the-sql-standard-say-about-parentheses-in-sql-union-except-intersect-s)

So it looks like the best way to accomplish the `ORDER BY` and `LIMIT` functionality as subqueries (these are used by TopN over each bucket and the results are unioned together) would be to make multiple queries and aggregate the results manually like @michael-mclawhorn suggested.

I've filed a [bug report with Calcite](https://issues.apache.org/jira/browse/CALCITE-1892), but since I'm not sure how much can be done for this issue.

docs/sql-support.md
```diff
@@ -0,0 +1,41 @@
+# Using a SQL Backend
+Fili has initial support for using a sql database as the backend instead of druid. Please
+note that there are some restrictions since Fili is optimized for Druid.
+
+## Setup
```
kevinhinterlong: Maybe mention that `DruidDimensionLoader` should be turned off?

fili-core/src/main/java/com/yahoo/bard/webservice/sql/SqlAggregationQuery.java
```diff
@@ -0,0 +1,156 @@
...
+     * @param dimensions  The dimensions
+     * @param filter  The filter
+     * @param aggregations  The aggregations
+     * @param postAggregations  The post-aggregations
+     * @param intervals  The intervals
+     * @param context  The context
+     * @param doFork  true to fork a new context and bump up the query id, or false to create an exact copy of the
+     * context.
+     */
+    private SqlAggregationQuery(
```
kevinhinterlong: I think this constructor could be removed

fili-core/src/main/java/com/yahoo/bard/webservice/sql/aggregation/DruidSqlAggregationConverter.java
```diff
@@ -0,0 +1,24 @@
...
+package com.yahoo.bard.webservice.sql.aggregation;
+
+import com.yahoo.bard.webservice.druid.model.aggregation.Aggregation;
+
+import java.util.Optional;
+
+/**
+ * Provides a mapping from Druid's {@link Aggregation} to a {@link SqlAggregationBuilder}.
+ */
+public interface DruidSqlAggregationConverter {
```
kevinhinterlong: @archolewa I've done something similar to what you mentioned in #413, but tied it specifically to aggregations, Let me know if you have any concerns with this approach

fili-core/src/main/java/com/yahoo/bard/webservice/sql/evaluator/FilterEvaluator.java
```diff
@@ -0,0 +1,254 @@
...
+import java.util.Locale;
+import java.util.stream.Collectors;
+
+/**
+ * Evaluates filters to find all the dimensions used in them and
+ * to build a {@link RexNode} with an equivalent sql filter.
+ */
+public class FilterEvaluator implements ReflectiveVisitor {
+    private RelBuilder builder;
+    private final List<String> dimensions;
```
kevinhinterlong: This should probably be a `Set`

fili-core/src/main/java/com/yahoo/bard/webservice/sql/evaluator/PostAggregationEvaluator.java
```diff
@@ -0,0 +1,133 @@
...
+                return prod;
+            case MINUS:
+                Double sub = dispatcher.invoke(arithmeticPostAggregation.getFields().get(0));
+                for (int i = 1; i < arithmeticPostAggregation.getFields().size(); i++) {
+                    PostAggregation postAgg = arithmeticPostAggregation.getFields().get(i);
+                    sub -= dispatcher.invoke(postAgg);
+                }
+                return sub;
+            case DIVIDE:
+                if (arithmeticPostAggregation.getFields().size() != 2) {
```
kevinhinterlong: This PR needs to be rebased on #413, this was changed there

fili-core/src/main/java/com/yahoo/bard/webservice/sql/helper/CalciteHelper.java
```diff
@@ -0,0 +1,161 @@
...
+import java.sql.SQLException;
+import java.util.List;
+
+import javax.sql.DataSource;
+
+/**
+ * Small utility class to help with connection to databases, building, and writing sql.
+ */
+public class CalciteHelper {
+    public static final String DEFAULT_SCHEMA = "PUBLIC";
```
kevinhinterlong: I don't think this is true for most but not all databases. Another reason this should be specified in the Table logic and not here

fili-core/src/main/java/com/yahoo/bard/webservice/sql/helper/DatabaseHelper.java
```diff
@@ -0,0 +1,47 @@
...
+     * @param connection  The connection to the database.
+     * @param schema  The name of the schema (i.e. "PUBLIC").
+     * @param table  The name of the table (i.e. "table").
+     *
+     * @return the name of the timestamp column.
+     *
+     * @throws SQLException if failed while reading database.
+     * @throws IllegalStateException if no {@link JDBCType#TIMESTAMP} column could be found.
+     */
+    public static String getTimestampColumn(Connection connection, String schema, String table)
```
kevinhinterlong: Should also be specified in the Table, repurpose this to check that the `time` column is actually a `TIMESTAMP`

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/SqlRequestHandler.java
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * Initializes the connection to the sql backend and prepares
+     * the {@link SqlBackedClient} for converting queries.
+     *
+     * @param mapper  The mapper for all JSON processing.
+     */
+    private void initializeSqlBackend(ObjectMapper mapper) {
+        String dbUrl = SYSTEM_CONFIG.getStringProperty(DATABASE_URL);
+        String driver = SYSTEM_CONFIG.getStringProperty(DATABASE_DRIVER);
+        String schema = SYSTEM_CONFIG.getStringProperty(DATABASE_SCHEMA);
```
kevinhinterlong: remove schema from here as well

fili-core/src/test/groovy/com/yahoo/bard/webservice/sql/DefaultSqlBackedClientSpec.groovy
```diff
@@ -0,0 +1,262 @@
...
+
+import spock.lang.Specification
+import spock.lang.Unroll
+
+class DefaultSqlBackedClientSpec extends Specification {
+    private static SqlBackedClient sqlBackedClient = new DefaultSqlBackedClient(Database.getDataSource(), new ObjectMapper())
+    private static final String TRUE = "TRUE"
+    private static final String FALSE = "FALSE"
+    private static final String FIRST_COMMENT = "added project"
+    //this is the first result in the database
```
kevinhinterlong: clarify "this"

fili-core/src/main/java/com/yahoo/bard/webservice/sql/aggregation/DefaultDruidSqlAggregationConverter.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public Optional<SqlAggregationBuilder> fromDruidType(Aggregation aggregation) {
+        String aggregationType = aggregation.getType().toLowerCase(Locale.ENGLISH);
+        Optional<SqlAggregationBuilder> sqlAggregationBuilder = Arrays.stream(DefaultSqlAggregationType.values())
+                .filter(sqlAggregationType -> aggregationType.contains(sqlAggregationType.type))
+                .map(sqlAggregationType -> sqlAggregationType.with(aggregation))
+                .findFirst();
+
+        if (!sqlAggregationBuilder.isPresent()) {
```
kevinhinterlong: It would make more sense to log where this is used

fili-core/src/main/java/com/yahoo/bard/webservice/web/ResponseFormatType.java
```diff
@@ -8,6 +8,7 @@
  * Response format type.
  */
 public enum ResponseFormatType {
+    SQL,
```
kevinhinterlong: Needs #435

docs/sql-support.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Limitations
+- Character sets. Basically don't try to query unicode characters unless your database supports them.
+- MYSQL databases are **much** slower than druid at doing TopN queries.
+ *Do **NOT** make TopN queries over large amounts of time with small buckets.* (For example, don't ask for the TopN from every hour over an entire year.) 
+
+### Supported queries
+* timeseries
+* groupBy
+* topN
```
kevinhinterlong: Mention that `bard__top_n_enabled` should be off

PR #432, opened 2017-07-17
added between and not between operators.

deepakb91:
I would like someone to review and provide feedback. I have set the default type to LESS_THAN for between and not between operators for now as there is no specific type on the Druid end.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidHavingBuilder.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidHavingBuilder.java
@@ -84,9 +84,10 @@ public class DruidHavingBuilder {
         List<Having> havings = new ArrayList<>();
         AndHaving andHaving = null;
         if (operation.equals(HavingOperation.between) || operation.equals(HavingOperation.notBetween)) {
-            if (values.size() != 2) {
+            final int SIZE = 2;
+            if (values.size() != SIZE) {
                 throw new UnsupportedOperationException(HAVING_CUSTOM_OPERATOR_TOO_FEW_PARAMETERS.format
-                        (operation.name(), operation.name(), 2));
+                        (operation.name(), operation.name(), SIZE));
             }
             else {
                 double firstValue = values.get(0);
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidHavingBuilder.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidHavingBuilder.java
@@ -30,7 +30,7 @@ import static com.yahoo.bard.webservice.web.ErrorMessageFormat.HAVING_OPERATOR_W
  */
 public class DruidHavingBuilder {
     private static final Logger LOG = LoggerFactory.getLogger(DruidHavingBuilder.class);
-    private static final int HAVING_PARAM_LENGTH = 2;
+    private static final int HAVING_RANGE_PARAM_LENGTH = 2;
     /**
      * Build a having model that ANDs together having queries for each of the metrics.
      *
@@ -87,9 +87,9 @@ public class DruidHavingBuilder {
 
         Having newHaving;
         if (operation.equals(HavingOperation.between) || operation.equals(HavingOperation.notBetween)) {
-            if (values.size() != HAVING_PARAM_LENGTH) {
+            if (values.size() != HAVING_RANGE_PARAM_LENGTH) {
                 throw new UnsupportedOperationException(HAVING_OPERATOR_WRONG_NUMBER_OF_PARAMETERS.format
-                        (operation.name(), operation.name(), HAVING_PARAM_LENGTH, values.size()));
+                        (operation.name(), operation.name(), HAVING_RANGE_PARAM_LENGTH, values.size()));
             }
             double lowerValue = values.get(0);
             double upperValue = values.get(1);
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidHavingBuilder.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidHavingBuilder.java
@@ -30,11 +30,13 @@ import static com.yahoo.bard.webservice.web.ErrorMessageFormat.HAVING_CUSTOM_OPE
  */
 public class DruidHavingBuilder {
     private static final Logger LOG = LoggerFactory.getLogger(DruidHavingBuilder.class);
-
+    private static final int SIZE = 2;
+    
     /**
      * Build a having model that ANDs together having queries for each of the metrics.
      *
      * @param metricMap  A map of logical metric to the set of havings for that metric
+     *
      * @return The having clause to appear in the Druid query. Returns null if the metricMap is empty or null.
      */
     public static Having buildHavings(Map<LogicalMetric, Set<ApiHaving>> metricMap) {
@@ -57,6 +59,7 @@ public class DruidHavingBuilder {
      *
      * @param metric  Metric for the having query
      * @param havings  All having queries belonging to that metric
+     *
      * @return A druid query having object representing the having clause on a given metric
      */
     public static Having buildMetricHaving(LogicalMetric metric, Set<ApiHaving> havings) {
@@ -74,6 +77,7 @@ public class DruidHavingBuilder {
      *
      * @param metric  The metric that the operation applied to.
      * @param having  The ApiHaving object
+     *
      * @return A single having representing the API Filter
      */
     public static Having buildHaving(LogicalMetric metric, ApiHaving having) {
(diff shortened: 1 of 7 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidHavingBuilder.java
```diff
@@ -28,8 +29,7 @@
     /**
      * Build a having model that ANDs together having queries for each of the metrics.
      *
-     * @param metricMap  A map of logical metric to the set of havings for that metric
-     *
+     * @param metricMap A map of logical metric to the set of havings for that metric
```
michael-mclawhorn: two spaces between @param names and comments is correct to our style guide.  Your formatter reduced to a single space incorrectly.

michael-mclawhorn: This applied throughout these changes.

fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidHavingBuilder.java
```diff
@@ -68,15 +67,34 @@ public static Having buildMetricHaving(LogicalMetric metric, Set<ApiHaving> havi
...
+     * @param metric The metric that the operation applied to.
+     * @param having The ApiHaving object
      * @return A single having representing the API Filter
      */
     public static Having buildHaving(LogicalMetric metric, ApiHaving having) {
         LOG.trace("Building having using metric: {} and API Having: {}", metric, having);
 
         HavingOperation operation = having.getOperation();
+
+        Set<Double> values = having.getValues();
```
michael-mclawhorn: Since order will matter here, we're going to need to change the internals of ApiHaving to use a List, not a Set for it's values list.

fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidHavingBuilder.java
```diff
@@ -68,15 +67,34 @@ public static Having buildMetricHaving(LogicalMetric metric, Set<ApiHaving> havi
...
+        }
+        else if (operation.equals(HavingOperation.notBetween)) {
+            double firstValue = values.stream().min(Double::compareTo).get();
+            double secondValue = values.stream().max(Double::compareTo).get();
+            List<Having> havings = new ArrayList<>();
+            havings.add(new NumericHaving(Having.DefaultHavingType.GREATER_THAN, metric.getName(), firstValue));
+            havings.add(new NumericHaving(Having.DefaultHavingType.LESS_THAN, metric.getName(), secondValue));
+            AndHaving andHaving = new AndHaving(havings);
+            return new NotHaving(andHaving);
+        }
```
michael-mclawhorn: There's way more code duplication here than makes sense.

fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidHavingBuilder.java
michael-mclawhorn: Let's actually get a little more firm in parsing the values list.  Right now your code will work if between lists three values, which is nice, except that if someone asks the question with two values, they've probably made a bad mistake.  Better to validate both count and order on a having-style operation and error if they are wrong.

michael-mclawhorn: ErrorMessageFormat is where you'll want to put the new validation error message(s) on wrong count  and wrong order.

fili-generic-example/src/main/resources/applicationConfig.properties
```diff
@@ -9,10 +9,10 @@ moduleDependencies = fili
 bard__version = Local Mode
 bard__resource_binder=com.yahoo.wiki.webservice.application.GenericBinderFactory
 
-bard__non_ui_druid_broker=http://localhost:8082/druid/v2
-bard__ui_druid_broker=http://localhost:8082/druid/v2
-bard__druid_coord=http://localhost:8081/druid/coordinator/v1
-bard__fili_port=9998
+bard__non_ui_druid_broker=http://badedisobeyed.corp.gq1.yahoo.com:8082/druid/v2
```
kevinhinterlong: Should probably `git rebase -i` and remove this

fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidHavingBuilder.java
```diff
@@ -29,7 +35,6 @@
      * Build a having model that ANDs together having queries for each of the metrics.
      *
      * @param metricMap  A map of logical metric to the set of havings for that metric
-     *
```
michael-mclawhorn: Space between @param and @return is correct.

fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidHavingBuilder.java
```diff
@@ -70,16 +74,46 @@ public static Having buildMetricHaving(LogicalMetric metric, Set<ApiHaving> havi
...
 
         HavingOperation operation = having.getOperation();
-        List<Having> havings = having.getValues().stream()
-                .map(value -> new NumericHaving(operation.getType(), metric.getName(), value))
-                .collect(Collectors.toList());
+        List<Double> values = having.getValues();
+        List<Having> havings = new ArrayList<>();
+        AndHaving andHaving = null;
+        if (operation.equals(HavingOperation.between) || operation.equals(HavingOperation.notBetween)) {
+            int size = 2;
```
michael-mclawhorn: This as a local variable doesn't make much sense.  Either just use '2' or make it a constant at the class level.

fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidHavingBuilder.java
```diff
@@ -96,7 +130,17 @@ public static Having buildHaving(LogicalMetric metric, ApiHaving having) {
         //     !(metric1 > 2 || metric1 > 4 || metric1 > 8)
         // . In both cases, the second method is correct.
 
-        Having newHaving = havings.size() == 1 ? havings.get(0) : new OrHaving(havings);
-        return operation.isNegated() ? new NotHaving(newHaving) : newHaving;
+        Having newHaving;
+        if (operation.equals(HavingOperation.between)) {
```
michael-mclawhorn: You can dramatically reduce the cyclomatic complexity of this method by restructuring it a little.  Use the same variable (say 'newHaving') in both of your initial if/else.  (Moving the if count == 1 inside the else.
Then you can just return on the final line with the not-transform not even needing to be assigned to a local variable.

fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidHavingBuilder.java
```diff
@@ -70,16 +74,46 @@ public static Having buildMetricHaving(LogicalMetric metric, Set<ApiHaving> havi
...
-                .collect(Collectors.toList());
+        List<Double> values = having.getValues();
+        List<Having> havings = new ArrayList<>();
+        AndHaving andHaving = null;
+        if (operation.equals(HavingOperation.between) || operation.equals(HavingOperation.notBetween)) {
+            int size = 2;
+            if (values.size() != size) {
+                throw new UnsupportedOperationException(HAVING_CUSTOM_OPERATOR_TOO_FEW_PARAMETERS.format
+                        (operation.name(), operation.name(), size));
+            }
```
michael-mclawhorn: Technically there could be too few or too many.  You should probably make the message 'wrong number of operators for ${operation}, required '%n', found '%x'.

fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidHavingBuilder.java
```diff
@@ -15,16 +15,22 @@
...
+import static com.yahoo.bard.webservice.web.ErrorMessageFormat.HAVING_CUSTOM_OPERATOR_IMPROPER_RANGE;
+import static com.yahoo.bard.webservice.web.ErrorMessageFormat.HAVING_CUSTOM_OPERATOR_WRONG_NUMBER_OF_PARAMETERS;
+
 /**
  * Class to hold static methods to build druid query model objects from ApiHaving.
  */
 public class DruidHavingBuilder {
     private static final Logger LOG = LoggerFactory.getLogger(DruidHavingBuilder.class);
-
+    private static final int SIZE = 2;
```
michael-mclawhorn: Size is a little out of context here.  Perhaps 'HAVING_PARAM_LENGTH'?

fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidHavingBuilder.java
```diff
@@ -77,9 +83,36 @@ public static Having buildHaving(LogicalMetric metric, ApiHaving having) {
...
+
+        Having newHaving;
+        if (operation.equals(HavingOperation.between) || operation.equals(HavingOperation.notBetween)) {
+            if (values.size() != SIZE) {
+                throw new UnsupportedOperationException(HAVING_CUSTOM_OPERATOR_WRONG_NUMBER_OF_PARAMETERS.format
+                        (operation.name(), operation.name(), SIZE, values.size()));
+            }
+            double firstValue = values.get(0);
+            double secondValue = values.get(1);
+            if (secondValue < firstValue) {
```
michael-mclawhorn: Maybe contextualize the variable names here as lowerValue and upperValue

fili-core/src/main/java/com/yahoo/bard/webservice/web/ErrorMessageFormat.java
```diff
@@ -252,7 +252,10 @@
     ),
 
     TOO_MANY_BACKING_DATA_SOURCES("TableDataSource built with too many backing data sources: %s"),
-    TOO_FEW_BACKING_DATA_SOURCES("TableDataSource built with insufficient backing data sources: %s")
+    TOO_FEW_BACKING_DATA_SOURCES("TableDataSource built with insufficient backing data sources: %s"),
+    HAVING_CUSTOM_OPERATOR_WRONG_NUMBER_OF_PARAMETERS("Wrong number of parameters provided for %s operator. " +
+                                              "%s operator needs %d parameters but found %d"),
+    HAVING_CUSTOM_OPERATOR_IMPROPER_RANGE("Upper range cannot be less than the lower range for %s operator.")
     ;
```
michael-mclawhorn: I think 'CUSTOM_OPERATOR' isn't informative here.  It really isn't a custom operator at this point, it's just an operator, the same as the others.

fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidHavingBuilder.java
```diff
@@ -15,16 +15,22 @@
...
+import static com.yahoo.bard.webservice.web.ErrorMessageFormat.HAVING_OPERATOR_IMPROPER_RANGE;
+import static com.yahoo.bard.webservice.web.ErrorMessageFormat.HAVING_OPERATOR_WRONG_NUMBER_OF_PARAMETERS;
+
 /**
  * Class to hold static methods to build druid query model objects from ApiHaving.
  */
 public class DruidHavingBuilder {
     private static final Logger LOG = LoggerFactory.getLogger(DruidHavingBuilder.class);
-
+    private static final int HAVING_PARAM_LENGTH = 2;
```
michael-mclawhorn: I actually gave you bad advice on the variable name.  There are at least three having signatures that make sense, >1 (ins), exactly 2 (ranges), and 1 (comparisons).
Please change this to reflect that it's range parameters.  As in HAVING_RANGE_PARAM_LENGTH.

I'm +1ing the PR because it's a simple change. 

You just need one more review from someone.

PR #437, opened 2017-07-19
Have TablesApiRequest Support Additional Query Parameters

QubitPi:
The 2nd milestone of https://github.com/yahoo/fili/pull/310

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ Current
 
 - [Have Tables Endpoint Support (but not use) Additional Query Parameters](https://github.com/yahoo/fili/pull/437)
     * Make the availability consider the TablesApiRequest by passing it into the getLogicalTableFullView method
+    * Move auxiliary methods from `DataApiRequest` to `ApiRequest` in order to make them sharable between
+    `DataApiRequest` and `TableApiRequest`.
 
 - [Fili-security module](https://github.com/yahoo/fili/pull/405)
     * Added security module for fili data security filters
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,12 @@ Current
 
 ### Added:
 
+- [Have Tables Endpoint Support (but not use) Additional Query Parameters](https://github.com/yahoo/fili/pull/437)
+    * Make the availability consider the TablesApiRequest by passing it into the getLogicalTableFullView method
+
 - [Fili-security module](https://github.com/yahoo/fili/pull/405)
     * Added security module for fili data security filters
-    * Created `ChainingRequestMapper`, and a set of mappers for gatekeeping on security roles and whitelisting dimension filters. 
+    * Created `ChainingRequestMapper`, and a set of mappers for gatekeeping on security roles and whitelisting dimension filters.
 
 - [Add Table-wide Availability](https://github.com/yahoo/fili/pull/414)
     * Add `availableIntervals` field to tables endpoint by union the availability for the logical table without taking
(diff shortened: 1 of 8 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/TablesServlet.java
```diff
@@ -368,7 +368,12 @@ public Response getTablesFullView(
      * @param uriInfo  UriInfo of the request
      *
      * @return Full view of the logical table
+     *
+     * @deprecated Inorder to display constrained data availability in table resource, this method needs to accept a
```
michael-mclawhorn: spelling -> In order

QubitPi:
@michael-mclawhorn I moved some methods from `DataApiRequest` to `ApiRequest` so that they can be shared between `DataApiRequest` and `TableApiRequest`

CHANGELOG.md
```diff
@@ -10,9 +10,12 @@ Current
...
+- [Have Tables Endpoint Support (but not use) Additional Query Parameters](https://github.com/yahoo/fili/pull/437)
+    * Make the availability consider the TablesApiRequest by passing it into the getLogicalTableFullView method
+
 - [Fili-security module](https://github.com/yahoo/fili/pull/405)
     * Added security module for fili data security filters
-    * Created `ChainingRequestMapper`, and a set of mappers for gatekeeping on security roles and whitelisting dimension filters. 
+    * Created `ChainingRequestMapper`, and a set of mappers for gatekeeping on security roles and whitelisting dimension filters.
 
 - [Add Table-wide Availability](https://github.com/yahoo/fili/pull/414)
     * Add `availableIntervals` field to tables endpoint by union the availability for the logical table without taking
```
dayamr: It would be nice to explain the scope of this PR since it is not the complete feature implementation. Otherwise changes looks good.

michael-mclawhorn:
I know this is going to seem like a really petty test, but I'd like to test that Tables API constructs with various partial parameters.  Just construct one with all of dimensions,intervals,filters,etc and one with just filters and one with just intervals.  Just a few mix-and-match to make sure that the fields are not interdependent.

QubitPi:
@michael-mclawhorn I agree with you, but we will have to pass everything fully constructed into `TableApiRequest` constructor because the constructor calls validation methods on dimension, metrics, filters, etc. I can add **one** test to make sure dimension, metrics, etc are constructed correctly if you are OK with it.

michael-mclawhorn:
I think we should change the design a little.  I think we may be pulling too many things into API Request base class.  In fact, I think we're going to need to make big changes to API Request soon and this will make it harder.

I feel like we should have a superclass in common between DataApiRequest and TableApiRequest to hold the common code/state.  Any suggestions as to a name?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/TablesApiRequestSpec.groovy
```diff
@@ -5,12 +5,18 @@ package com.yahoo.bard.webservice.web
...
+import com.yahoo.bard.webservice.data.time.DefaultTimeGrain
+import com.yahoo.bard.webservice.data.time.GranularityParser
 import com.yahoo.bard.webservice.data.time.StandardGranularityParser
+import com.yahoo.bard.webservice.druid.model.query.Granularity
 import com.yahoo.bard.webservice.table.LogicalTable
 import com.yahoo.bard.webservice.table.LogicalTableDictionary
 import com.yahoo.bard.webservice.table.TableGroup
+import com.yahoo.bard.webservice.table.TableIdentifier
 import com.yahoo.bard.webservice.web.endpoints.TablesServlet
+import com.yahoo.bard.webservice.web.util.BardConfigResources
```
michael-mclawhorn: These are now all excess imports

PR #442, opened 2017-07-25
RFC for custom Response creation.

archolewa:
A design for support of custom response serialization.


rfc/resultSetSerializability.md
```diff
@@ -0,0 +1,73 @@
...
+    a. A `ResponseData` class that contains all of the data and helper methods
+        currently living on `Response`.
+    b. An implementation of `ResponseWriter` for each of
+        `Response::writeJsonResponse`, `Response::writeJsonApiResponse`,
+        `Response::writeCsvResponse`.
+
+3. Define an interface `ResponseWriterSelector` with a single method:
+    ```java
+        ResponseWriter select(DataApiRequest request);
+    ```
```
michael-mclawhorn: Should this be optional?

QubitPi:
Why not consider `TableApiRequest` and `DimensionApiRequest`, as well?

archolewa:
@QubitPi I've updated the interface for `ResponseWriter` to take an `ApiRequest` rather than a `DataApiRequest`, because you're right, we should allow the ability to fully customize the other servlets.

However, fully generalizing this to work for the other endpoints will require more than just having the interface take an `ApiRequest`. For one thing, we'd have to generalize `ResponseData` somehow to accomodate for the very differently structured responses that the other servlets return. So, adding support for rewriting the other endpoints is outside the scope of this RFC.

archolewa:
@kevinhinterlong Should probably go under docs.

archolewa:
@kevinhinterlong I've moved the rfc folder under `docs`.

Issue #428, opened 2017-07-13
Fili Generic Example Segment Metadata


kevinhinterlong:
Load `List<DataSegment>` into the `DatasourceMetadataService` in the [GenericTableLoader](https://github.com/yahoo/fili/blob/0e599c9ba939d510ff53cbc44a83a93457794896/fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/table/GenericTableLoader.java#L86-L86)

The results returned from [druid](localhost:8081/druid/coordinator/v1/datasources/wikiticker/?full) in `DruidNavigator` provide them so we should use them.

```json
{
   ...
   "segments":[
      {
         "dataSource":"wikiticker",
         "interval":"2015-09-12T00:00:00.000Z/2015-09-13T00:00:00.000Z",
         "version":"2017-06-12T19:09:29.629Z",
         "loadSpec":{
            "type":"local",
            "path":"/home/hinterlong/druid-0.9.0/var/druid/segments/wikiticker/wikiticker/2015-09-12T00:00:00.000Z_2015-09-13T00:00:00.000Z/2017-06-12T19:09:29.629Z/0/index.zip"
         },
         "dimensions":"channel,cityName,comment,countryIsoCode,countryName,isAnonymous,isMinor,isNew,isRobot,isUnpatrolled,metroCode,namespace,page,regionIsoCode,regionName,user",
         "metrics":"count,added,deleted,delta,user_unique",
         "shardSpec":{
            "type":"none"
         },
         "binaryVersion":9,
         "size":5537575,
         "identifier":"wikiticker_2015-09-12T00:00:00.000Z_2015-09-13T00:00:00.000Z_2017-06-12T19:09:29.629Z"
      }
   ]
}
```

PR #445, opened 2017-07-25
Use all available segment metadata for fili-generic-example

kevinhinterlong:
- The `fili-generic-example` was previously only using the first segment in the response from Druid but it may as well use all of them
- This also provides these segments to the MetadataService

Addresses #428

### DIFF
--- a/fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/DataSourceConfiguration.java
+++ b/fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/DataSourceConfiguration.java
@@ -35,14 +35,14 @@ public interface DataSourceConfiguration {
     /**
      * Gets the names of all the metrics for the current datasource.
      *
-     * @return a list of names of metrics for the current datasource.
+     * @return a set of names of metrics for the current datasource.
      */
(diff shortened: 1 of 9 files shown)


kevinhinterlong:
@michael-mclawhorn yes

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/DataSourceConfiguration.java
```diff
@@ -34,14 +37,14 @@
      *
      * @return a list of names of metrics for the current datasource.
```
garyluoex: comment out of date.

fili-generic-example/src/main/java/com/yahoo/wiki/webservice/data/config/auto/DataSourceConfiguration.java
```diff
@@ -34,14 +37,14 @@
...
      */
-    List<String> getMetrics();
+    Set<String> getMetrics();
 
     /**
      * Gets the names of all the dimensions for the current datasource.
      *
      * @return a list of names of dimensions for the current datasource.
```
garyluoex: same

PR #447, opened 2017-07-26
Custom Response creation part1

tarrantzhang:
This PR implements step 6,7 in [RFC for custom response creation](https://github.com/yahoo/fili/pull/442)

- Make `httpResponseMaker` injectable.
- Change signature of `httpResponseChannel` constructor to take `ApiRequest`.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
@@ -35,6 +35,7 @@ import com.yahoo.bard.webservice.config.SystemConfig;
 import com.yahoo.bard.webservice.config.SystemConfigProvider;
 import com.yahoo.bard.webservice.data.DruidQueryBuilder;
 import com.yahoo.bard.webservice.data.DruidResponseParser;
+import com.yahoo.bard.webservice.data.HttpResponseMaker;
 import com.yahoo.bard.webservice.data.PartialDataHandler;
 import com.yahoo.bard.webservice.data.PreResponseDeserializer;
 import com.yahoo.bard.webservice.data.cache.DataCache;
@@ -301,6 +302,8 @@ public abstract class AbstractBinderFactory implements BinderFactory {
                 bind(getAsynchronousProcessBuilder()).to(AsynchronousWorkflowsBuilder.class);
 
                 bind(getClock()).to(Clock.class);
+
+                bind(getHttpResponseMaker()).to(HttpResponseMaker.class);
 
                 if (DRUID_DIMENSIONS_LOADER.isOn()) {
                     DruidDimensionsLoader druidDimensionsLoader = buildDruidDimensionsLoader(
@@ -904,6 +907,15 @@ public abstract class AbstractBinderFactory implements BinderFactory {
      */
     protected JobPayloadBuilder buildJobPayloadBuilder() {
         return new DefaultJobPayloadBuilder();
+    }
+
+    /**
+     * Initializes the factory that builds HttpResponseMaker.
+     *
+     * @return an instance of the {@link HttpResponseMaker}
+     */
+    protected Class<? extends HttpResponseMaker> getHttpResponseMaker() {
+        return HttpResponseMaker.class;
     }
 
     /**
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/HttpResponseChannel.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/HttpResponseChannel.java
@@ -6,9 +6,8 @@ import static com.yahoo.bard.webservice.web.handlers.workflow.DruidWorkflow.RESP
 
 import com.yahoo.bard.webservice.async.ResponseException;
 import com.yahoo.bard.webservice.logging.RequestLog;
-import com.yahoo.bard.webservice.web.DataApiRequest;
+import com.yahoo.bard.webservice.web.ApiRequest;
 import com.yahoo.bard.webservice.web.PreResponse;
-import com.yahoo.bard.webservice.web.ResponseFormatType;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -17,7 +16,6 @@ import rx.Observer;
 
 import javax.ws.rs.container.AsyncResponse;
 import javax.ws.rs.core.Response;
-import javax.ws.rs.core.UriInfo;
 
 /**
  * Converts preResponse/errorResponse into HTTP Responses, and ships them immediately to the client.
@@ -28,28 +26,7 @@ public class HttpResponseChannel implements Observer<PreResponse> {
 
     private final AsyncResponse asyncResponse;
     private final HttpResponseMaker httpResponseMaker;
-    private final ResponseFormatType responseFormatType;
-    private final UriInfo uriInfo;
-
-    /**
-     * Constructor.
-     *
-     * @param asyncResponse  An async response that we can use to respond asynchronously
-     * @param httpResponseMaker  Helper class instance to prepare the response object
-     * @param responseFormatType  The format of the response returned to the user
-     * @param uriInfo  UriInfo of the request
-     */
-    public HttpResponseChannel(
-            AsyncResponse asyncResponse,
-            HttpResponseMaker httpResponseMaker,
-            ResponseFormatType responseFormatType,
-            UriInfo uriInfo
-    ) {
-        this.asyncResponse = asyncResponse;
-        this.httpResponseMaker = httpResponseMaker;
-        this.responseFormatType = responseFormatType;
-        this.uriInfo = uriInfo;
-    }
+    private final ApiRequest apiRequest;
 
     /**
      * Constructor.
@@ -58,16 +35,15 @@ public class HttpResponseChannel implements Observer<PreResponse> {
      * @param apiRequest  Api request object with all the associated info with it
      * @param httpResponseMaker  Helper class instance to prepare the response object
      *
-     * @deprecated  The ResponseFormatType and UriInfo should be passed explicitly, rather than implicitly via the
-     * DataApiRequest.
      */
-    @Deprecated
     public HttpResponseChannel(
             AsyncResponse asyncResponse,
-            DataApiRequest apiRequest,
+            ApiRequest apiRequest,
             HttpResponseMaker httpResponseMaker
     ) {
-        this(asyncResponse, httpResponseMaker, apiRequest.getFormat(), apiRequest.getUriInfo());
+        this.asyncResponse = asyncResponse;
+        this.httpResponseMaker = httpResponseMaker;
+        this.apiRequest = apiRequest;
     }
 
     @Override
@@ -104,7 +80,7 @@ public class HttpResponseChannel implements Observer<PreResponse> {
 
     @Override
     public void onNext(PreResponse preResponse) {
-        publishResponse(httpResponseMaker.buildResponse(preResponse, responseFormatType, uriInfo));
+        publishResponse(httpResponseMaker.buildResponse(preResponse, apiRequest));
     }
 
     /**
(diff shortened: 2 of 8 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/HttpResponseMaker.java
```diff
@@ -63,13 +66,15 @@ public HttpResponseMaker(ObjectMappersSuite objectMappers, DimensionDictionary d
...
+     * @param apiRequest apiRequest needed for serialization writer
      *
      * @return Completely built response with headers and result set
      */
     public javax.ws.rs.core.Response buildResponse(
             PreResponse preResponse,
             ResponseFormatType responseFormatType,
-            UriInfo uriInfo
+            UriInfo uriInfo,
+            ApiRequest apiRequest
```
tarrantzhang: Not so sure if `UriInfo` and `ResponseFormatType` still needed as we can get them from ApiRequest.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -907,6 +910,15 @@ protected JobPayloadBuilder buildJobPayloadBuilder() {
     }
 
     /**
+     * Initializes the factory that builds HttpResponseMaker.
+     *
+     * @return An instance of the {@link HttpResponseMaker}
```
QubitPi: Lowercase `An` (This fili doc standard)

fili-core/src/main/java/com/yahoo/bard/webservice/data/HttpResponseChannel.java
```diff
@@ -2,23 +2,21 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
 import rx.Observer;
 
 import javax.ws.rs.container.AsyncResponse;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriInfo;
 
+import static com.yahoo.bard.webservice.web.handlers.workflow.DruidWorkflow.RESPONSE_WORKFLOW_TIMER;
```
QubitPi: static import goes to the top

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/HttpResponseMakerSpec.groovy
```diff
@@ -19,17 +15,16 @@ import com.yahoo.bard.webservice.web.PreResponse
...
 import rx.subjects.Subject
 import spock.lang.Specification
 
 import javax.ws.rs.container.AsyncResponse
-import javax.ws.rs.core.HttpHeaders
-import javax.ws.rs.core.MultivaluedMap
-import javax.ws.rs.core.PathSegment
-import javax.ws.rs.core.Response
-import javax.ws.rs.core.UriInfo
+import javax.ws.rs.core.*
```
QubitPi: Avoid start import

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/HttpResponseMakerSpec.groovy
```diff
@@ -19,17 +15,16 @@ import com.yahoo.bard.webservice.web.PreResponse
...
+
+import static com.yahoo.bard.webservice.data.time.DefaultTimeGrain.DAY
+import static com.yahoo.bard.webservice.web.responseprocessors.ResponseContextKeys.API_METRIC_COLUMN_NAMES
+import static com.yahoo.bard.webservice.web.responseprocessors.ResponseContextKeys.REQUESTED_API_DIMENSION_FIELDS
```
QubitPi: static import goes to the top

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/DataServlet.java
```diff
@@ -404,6 +405,7 @@ public void getData(
                     apiRequest.getAsyncAfter(),
                     apiRequest.getFormat(),
                     uriInfo,
+                    apiRequest,
```
archolewa: Since we have to pass the entire apiRequest into this method anyway, we might as well drop the parameters that we pull directly from the apiRequest (i.e. asyncAfter, format, the uriInfo). It'll make this method's signature a little bit smaller, which is always nice when you can get away with it.

archolewa: I don't really have a good place to put this, because github stupidly insists on limiting what lines I can comment on. Right now, we're not actually using the `HttpResponseMaker` that is being injected. You'll note that on [line 397](https://github.com/yahoo/fili/pull/447/files#diff-33f581382b3b5c860a25432e57a66854R397) we are still manually building the `HttpResponseMaker`. So it doesn't matter what customers put in the `AbstractBinderFactory`, we'll still build and use the version defined in  Fili!

So, In addition to providing an injection point in the `AbstractBinderFactory`, you _also_ need to make the `HttpResponseMaker`  a constructor argument for `DataServlet`. 

Since `DataServlet` has the `@Inject` annotation, HK2 will build it for us, and as a part of that, it will build all of the constructor arguments and pass them in to the DataServlet.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
```diff
@@ -463,9 +463,8 @@ protected void handleResultsResponse(
                 .subscribe(
                         new HttpResponseChannel(
                                 asyncResponse,
-                                httpResponseMaker,
-                                apiRequest.getFormat(),
-                                apiRequest.getUriInfo()
+                                apiRequest,
```
archolewa: [We also need to inject the `HttpResponseMaker` into the `JobsServlet`](https://github.com/yahoo/fili/pull/447/files#diff-33f581382b3b5c860a25432e57a66854R408)

fili-core/src/main/java/com/yahoo/bard/webservice/data/HttpResponseMaker.java
```diff
@@ -61,21 +64,19 @@ public HttpResponseMaker(ObjectMappersSuite objectMappers, DimensionDictionary d
...
-            ResponseFormatType responseFormatType,
-            UriInfo uriInfo
+            ApiRequest apiRequest
     ) {
         ResponseBuilder rspBuilder = createResponseBuilder(
                 preResponse.getResultSet(),
                 preResponse.getResponseContext(),
-                responseFormatType,
-                uriInfo
+                apiRequest.getFormat(),
```
archolewa: The response builder should also take the full `ApiRequest` instead of just the `format` and `uriInfo`, or are you planning to make that change in the next PR as part of splitting `Response`? If so, feel free to disregard this comment.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -302,6 +303,8 @@ protected void configure() {
 
                 bind(getClock()).to(Clock.class);
 
+                bind(getHttpResponseMaker()).to(HttpResponseMaker.class);
```
QubitPi: This needs a test. At least testing that it's been binded like `1 * dc.bind({ it.implementation.contains(HttpResponseMaker.canonicalName) }, _)` in `"test configure bindings"()`

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/DataServlet.java
```diff
@@ -447,24 +447,20 @@ public void getData(
     /**
      * Builds the asynchronous workflows, and subscribes the appropriate channels to the appropriate workflows.
      *
-     * @param asyncAfter  How long the user is willing to wait for a synchronous request
-     * @param responseFormat  The requested format for the response
-     * @param uriInfo  The URI of the request
+     * @param apiRequest The apiRequest
```
QubitPi: Param description doesn't help too much. People already know that from param name. Descriptions talks about what it is and how it's gonna be used

fili-core/src/main/java/com/yahoo/bard/webservice/data/HttpResponseMaker.java
```diff
@@ -61,21 +64,18 @@ public HttpResponseMaker(ObjectMappersSuite objectMappers, DimensionDictionary d
      * Build complete response.
      *
      * @param preResponse  PreResponse object which contains result set, response context and headers
-     * @param responseFormatType  The format in which the response should be returned to the user
-     * @param uriInfo  UriInfo of the request
+     * @param apiRequest apiRequest needed for serialization writer
```
QubitPi: 2 spaces between param name and param desc. Capitalize first word in description (This is fili doc standard) ¯\_(ツ)_/¯

fili-core/src/main/java/com/yahoo/bard/webservice/data/HttpResponseMaker.java
```diff
@@ -95,18 +95,18 @@ public HttpResponseMaker(ObjectMappersSuite objectMappers, DimensionDictionary d
      *
      * @param resultSet  The result set being processed
      * @param responseContext  A meta data container for the state gathered by the web container
-     * @param responseFormatType  The format in which the response should be returned to the user
-     * @param uriInfo  UriInfo of the request
+     * @param apiRequest apiRequest needed for serialization writer
```
QubitPi: 2 spaces between param name and param desc. Capitalize first word in description

fili-core/src/main/java/com/yahoo/bard/webservice/data/HttpResponseChannel.java
```diff
@@ -58,16 +35,15 @@ public HttpResponseChannel(
      * @param apiRequest  Api request object with all the associated info with it
      * @param httpResponseMaker  Helper class instance to prepare the response object
      *
-     * @deprecated  The ResponseFormatType and UriInfo should be passed explicitly, rather than implicitly via the
```
QubitPi: @michael-mclawhorn @archolewa Is really necessary to pass the giant ApiRequest object? It de-separate layer abstraction and makes it more difficult to set up a test. Does passing more variables from ApiRequest solves the problem as well?

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
archolewa: I disagree. This is a dirt-simple binding. All we'd be testing is the HK2 injection framework. Far as I can tell we test hardly any of the other similarly simple bindings (i.e. we don't verify that the `ObjectMapperSuite` is being bound correctly.

fili-core/src/main/java/com/yahoo/bard/webservice/data/HttpResponseChannel.java
archolewa: The primary purpose of making the `HttpResponseChannel` injectable is so that customers can inject their own `HttpResponseChannel` so that they can customize how they send back the response. 

There are situations (like the one that motivated this change) where customers will want to customize the response format based on something in the request. However, we can't very well predict what components of the request customers will use to determine how to send the response.

archolewa:
@QubitPi The steps are mislabeled in the RFC (there are two 5's). He's implementing [the second five](src//test/resources/com/flurry/gatekeeper/tests/prod/purplebox/public-v1/PurpleboxDataEndpointPublicV1.feature)

fili-core/src/main/java/com/yahoo/bard/webservice/data/HttpResponseChannel.java
archolewa: And it actually doesn't it make that much harder to setup a test since Groovy ignores access modifiers and the `DataApiRequest` (at least) has a private default constructor. So you can build an empty `ApiRequest` and then use the request's withers to initialize only the pieces you care about for the test.

tarrantzhang:
@QubitPi Sorry didn't sync up the PR description with RFC, it should be steps 6 & 7.

PR #451, opened 2017-07-28
Fix: Etag error test case fails on depending on system timezone settings

mpardesh:


fili-core/src/test/groovy/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessorSpec.groovy
```diff
@@ -16,7 +16,7 @@ import com.fasterxml.jackson.databind.node.ArrayNode
 import com.fasterxml.jackson.datatype.jdk8.Jdk8Module
 
 import org.joda.time.Interval
-
+import spock.lang.Shared
```
archolewa: We can drop this import. It's a holdover from a previous attempted fix.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/responseprocessors/DruidPartialDataResponseProcessorSpec.groovy
```diff
@@ -125,9 +125,10 @@ class DruidPartialDataResponseProcessorSpec extends Specification {
         DataSource dataSource = Mock(DataSource)
         ConstrainedTable constrainedTable = Mock(ConstrainedTable)
 
-        constrainedTable.getAvailableIntervals() >> new SimplifiedIntervalList(
-                [new Interval("2016-11-22T00:00:00.000Z/2016-12-18T00:00:00.000Z")]
-        )
+        Interval interval = new Interval("2016-11-22T00:00:00.000Z/2016-12-18T00:00:00.000Z")
```
kevinhinterlong: Make this string a constant as well as the other interval

michael-mclawhorn:
I've changed the name of the PR to reflect the scope more narrowly of the fix.
I was initially having trouble understanding what was being fixed.

Issue #453, opened 2017-07-28
Stricter checks in travis/deploy.bash


kevinhinterlong:
In the script we have 
```bash
#!/bin/bash
PUBLISH_WHITELIST="master"

# ...

# Only publish whitelisted branches
WHITELISTED=$(echo ${PUBLISH_WHITELIST} | grep ${TRAVIS_BRANCH})
if [[ "${WHITELISTED}" == "" ]]; then
    echo "INFO Do not flag for publication, this is not a whitelisted branch: ${TRAVIS_BRANCH}"
    exit 0
fi
```

but this fails if you have a branch called `mas` or some substring of `master`. If master is likely our only whitelisted branch then we could just check if it's equal. So in the case of the push build, this will tag and publish triggering a deploy.

PR #454, opened 2017-07-28
Stabilizing Travis and Maven builds

michael-mclawhorn:
https://github.com/yahoo/fili/issues/453

We have an issue where if a branch is a /subtring/ of a whitelisted travis branch( e.g. branch 'ast' and our whitelist branch 'master'), it will be treated like master for purpose of triggering release on push.

### DIFF
--- a/travis/deploy.bash
+++ b/travis/deploy.bash
@@ -48,8 +48,8 @@ if [[ ${MAVEN_RETURN_CODE} -ne 0 ]]; then
 fi
 
 # Only publish whitelisted branches
-WHITELISTED=$(echo ${PUBLISH_WHITELIST} | grep ${TRAVIS_BRANCH})
-if [[ "${WHITELISTED}" == "" ]]; then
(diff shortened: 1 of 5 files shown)


travis/deploy.bash
```diff
@@ -48,8 +48,8 @@ if [[ ${MAVEN_RETURN_CODE} -ne 0 ]]; then
 fi
 
 # Only publish whitelisted branches
-WHITELISTED=$(echo ${PUBLISH_WHITELIST} | grep ${TRAVIS_BRANCH})
-if [[ "${WHITELISTED}" == "" ]]; then
+WHITELISTED=$(for a in ${PUBLISH_WHITELIST}; do if [ ${TRAVIS_BRANCH} = ${a} ]; then echo true; fi; done;)
```
kevinhinterlong: It'd be nice if `a` was something like `branch`, but 👍

kevinhinterlong:
It appears the builds are having problems. [The containers they run in have 4GB of memory available.](https://docs.travis-ci.com/user/reference/precise/#Virtualization-environments) 
We [may need to set](https://stackoverflow.com/questions/12114174/what-does-xxmaxpermsize-do) `-XX:MaxPermSize=` as well in order to stabilize this, I'll try it out

PR #455, opened 2017-07-28
Custom Response creation part2

tarrantzhang:
Continuing on [RFC for custom response creation](https://github.com/yahoo/fili/pull/442). 
 [Part 1](https://github.com/yahoo/fili/pull/447).

Implementing steps 1~5:

- Define interface `ResponseWriter` and its default implementation
- Refactor `Response` class, splitting into `ResponseData` and three implementations of `ResponseWriter` 
- Define interface `ResponseWriterSelector` and its default implementation
- Implement 
    ```java
        void addResponseType(ResponseType type, ResponseWriter writer);
    ```
- Hook up the new serialization logic with `HttpResponseMaker` to replace the old one


fili-core/src/main/java/com/yahoo/bard/webservice/web/ResponseData.java
```diff
@@ -88,7 +87,8 @@ public Response(
             SimplifiedIntervalList volatileIntervals,
             Map<String, URI> paginationLinks,
             Pagination pagination,
-            ObjectMappersSuite objectMappers
+            ObjectMappersSuite objectMappers,
+            ApiRequest apiRequest
```
tarrantzhang: I really hesitate to add `apiRequest` to ResponseData, but it's need by write function in [line 183](https://github.com/tarrantzhang/fili/blob/bca02a125026e5568c077c6f46fac4620ecbaf00/fili-core/src/main/java/com/yahoo/bard/webservice/web/ResponseData.java#L183)

fili-core/src/main/java/com/yahoo/bard/webservice/web/CsvResponseWriter.java
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Writes CSV response.
+     *
+     * @param request  ApiRequest object with all the associated info in it
+     * @param responseData  data object containing all the result information
+     * @param os  OutputStream
+     *
+     * @throws IOException if a problem is encountered writing to the OutputStreamC
+     */
+    public void write(
```
archolewa: Should be annotated with [`@Override`](https://stackoverflow.com/questions/94361/when-do-you-use-javas-override-annotation-and-why)

fili-core/src/main/java/com/yahoo/bard/webservice/web/FiliResponseWriter.java
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * Selects a writer from mapping and writes response.
+     *
+     * @param request  ApiRequest object with all the associated info in it
+     * @param responseData  Data object containing all the result information
+     * @param os  OutputStream
+     *
+     * @throws IOException if a problem is encountered writing to the OutputStreamC
+     */
+    public void write(
```
archolewa: Annotated with `@Override`.

fili-core/src/main/java/com/yahoo/bard/webservice/web/FiliResponseWriter.java
```diff
@@ -0,0 +1,43 @@
...
+            ApiRequest request,
+            ResponseData responseData,
+            OutputStream os
+    ) throws IOException {
+        ResponseWriter writer = responseWriterSelector.select(request);
+        writer.write(request, responseData, os);
```
archolewa: We should do a null check on `writer` here and throw a useful error message if the response type isn't in the map (i.e. isn't recognized). Otherwise, if at some point down the line, someone adds a custom response format, but forgets to register the writer, then they'll just get a vague `NullPointerException`.

fili-core/src/main/java/com/yahoo/bard/webservice/web/FiliResponseWriterSelector.java
```diff
@@ -0,0 +1,52 @@
...
+        writers.put(ResponseFormatType.JSON, new JsonResponseWriter());
+    }
+
+    /**
+     * Selects a ReponseWriter given the format type from request.
+     *
+     * @param request  ApiRequest object with all the associated info in it
+     * @return Response writer for the given format type
+     */
+    public ResponseWriter select(ApiRequest request) {
```
archolewa: Should be annotated with `@Override`.

fili-core/src/main/java/com/yahoo/bard/webservice/web/FiliResponseWriter.java
archolewa: Also, since this method isn't changing the contract of the method it's implementing, JavaDoc isn't necessary.

fili-core/src/main/java/com/yahoo/bard/webservice/web/CsvResponseWriter.java
archolewa: Also, since this method isn't changing the contract of the method it's implementing, JavaDoc isn't necessary.

fili-core/src/main/java/com/yahoo/bard/webservice/web/FiliResponseWriterSelector.java
```diff
@@ -0,0 +1,52 @@
+// Copyright 2017 Oath Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.web;
+
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Response writer selector object.
```
archolewa: This is the place where I would specify that this selector selects the `ResponseWriter` based on the format type in the `ApiRequest`, rather than on the constructor or method signature.

fili-core/src/main/java/com/yahoo/bard/webservice/web/ResponseWriterSelector.java
```diff
@@ -0,0 +1,19 @@
...
+ * Response writer selector object.
+ */
+public interface ResponseWriterSelector {
+    /**
+     * Select ResponseWriter given certain type of format from DataApiRequest.
+     *
+     * @param request  ApiRequest object with all the associated info in it
+     * @return  Writer of given format type
+     */
+    default ResponseWriter select(DataApiRequest request) {
```
archolewa: This should absolutely not be defaulted. In my opinion, default implementations should only be used for methods that are added to preserve backwards compatibility, or maybe as convenience methods for invoking the other, meatier methods on the interface.

We definitely shouldn't be using a default implementation that relies on one of the implementations of the interface.

fili-core/src/main/java/com/yahoo/bard/webservice/web/JsonApiResponseWriter.java
```diff
@@ -0,0 +1,129 @@
...
+     *   "linkName1" : "http://uri1",
+     *   "linkName2": "http://uri2",
+     *   ...
+     *   "linkNameN": "http://uriN"
+     * }
+     * }
+     * </pre>
+     *
+     * Where "linkName1" ... "linkNameN" are the N keys in paginationLinks, and "http://uri1" ... "http://uriN" are the
+     * associated URI's.
```
archolewa: I would push this description of the JSON format onto the class level description, since this is the entire purpose of the class.

fili-core/src/main/java/com/yahoo/bard/webservice/web/JsonApiResponseWriter.java
```diff
@@ -0,0 +1,129 @@
...
+     *
+     * @param request  ApiRequest object with all the associated info in it
+     * @param responseData  Data object containing all the result information
+     * @param os  OutputStream
+     *
+     * @throws IOException if a problem is encountered writing to the OutputStream
+     */
+    public void write(
```
archolewa: Should be annotated with `@Override`.

fili-core/src/main/java/com/yahoo/bard/webservice/web/JsonResponseWriter.java
```diff
@@ -0,0 +1,71 @@
...
+     * },
+     *   "linkName1" : "http://uri1",
+     *   "linkName2": "http://uri2",
+     *   ...
+     *   "linkNameN": "http://uriN"
+     * }
+     * </pre>
+     *
+     * Where "linkName1" ... "linkNameN" are the N keys in paginationLinks, and "http://uri1" ... "http://uriN" are the
+     * associated URI's.
```
archolewa: I would push this description up onto the class level.

fili-core/src/main/java/com/yahoo/bard/webservice/web/JsonResponseWriter.java
```diff
@@ -0,0 +1,71 @@
...
+     *
+     * @param request  ApiRequest object with all the associated info in it
+     * @param responseData  Data object containing all the result information
+     * @param os  OutputStream
+     *
+     * @throws IOException if a problem is encountered writing to the OutputStream
+     */
+    public void write(
```
archolewa: Should be annotated with `@Override`.

fili-core/src/main/java/com/yahoo/bard/webservice/web/ResponseData.java
archolewa: We should just remove that function. It no longer belongs on this class, and this is so deep into the guts of Fili, that I don't think we have to worry about removing it breaking anybody.

fili-core/src/main/java/com/yahoo/bard/webservice/web/ResponseData.java
```diff
@@ -543,7 +370,7 @@ private static String getDimensionColumnName(Dimension dimension, DimensionField
      *
      * @throws IOException if the generator throws an IOException.
      */
-    private void writeMetaObject(
+    public void writeMetaObject(
```
archolewa: `writeMetaObject` should be pushed onto the two `JSON` request writers, since this is a JSON-specific thing. We may want to combine the JSON and JSONAPI writers into a single hierarchy with a shared parent class, which can have this method.

fili-core/src/main/java/com/yahoo/bard/webservice/web/ResponseData.java
archolewa: Basically, anything related to creating JSON or CSV no longer belongs on this class, and should be removed. That includes the `ObjectMappers` and the `jsonFactory` and `csvMapper`.

archolewa: We can probably also drop the `ResponseFormatType` parameter, since the writers are pulling that off the `ApiRequest`.

fili-core/src/main/java/com/yahoo/bard/webservice/web/ResponseWriter.java
```diff
@@ -0,0 +1,25 @@
+// Copyright 2017 Oath Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.web;
+
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+/**
+ * Response writer interface.
```
archolewa: We should provide a better class level JavaDoc. Nothing here says anything that the class declaration doesn't already.

In general, when writing JavaDoc you should focus on the `what`, and possibly the `why` depending on context (and how much tolerance your co-developers have for large class docs ;). 

For example, I might write something like the following:
```
The interface for objects that write fully-processed ResultSets back to the user. This allows customers 
to fully customize how they choose to serialize the results from Fili based on the request.
```

fili-core/src/main/java/com/yahoo/bard/webservice/web/ResponseWriter.java
```diff
@@ -0,0 +1,25 @@
...
+    /**
+     * Interface for all responseWriter. Default to use a FiliResponseWriter to writer response according to format
+     * type from request.
+     *
+     * @param request  ApiRequest object with all the associated info in it
+     * @param responseData  Data object containing all the result information
+     * @param os  OutputStream
+     * @throws IOException if a problem is encountered writing to the OutputStream
+     */
+    default void write(ApiRequest request, ResponseData responseData, OutputStream os)throws IOException {
```
archolewa: Missing a space: `OuputStream os)throws`

fili-core/src/main/java/com/yahoo/bard/webservice/web/ResponseWriterSelector.java
```diff
@@ -0,0 +1,19 @@
+// Copyright 2017 Oath Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.web;
+
+
+/**
+ * Response writer selector object.
```
archolewa: Again, this should provide a bit more insight into what this exists, and why it should be used. Maybe something like:
```
Typically, a Fili program's response writing infrastructure will consist of a collection of unrelated 
ResponseWriters each responsible for serializing the ResultSet into a specific format, and a single 
ResponseWriter that chooses the correct one based on the current state of the DataApiRequest.

This interface allows customers provide a clean interface to the logic that makes that selection. Its 
sole purpose is to take a `DataApiRequest` and return the `ResponseWriter` that should be used 
to  write the response. 
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/ResponseSpec.groovy
```diff
@@ -107,7 +107,11 @@ class ResponseSpec extends Specification {
 
     Set<Column> columns
     Set<LogicalMetric> testLogicalMetrics
-    Response response
+    ResponseData response
```
archolewa: At a minimum, we should turn this into the `FiliResponseWriterSpec`, since that's what this is testing now.

However, we really should break this apart. Anything that tests only one format (i.e. just CSV, or just JSON) should be moved into a spec for the appropriate `ResponseWriter`. Anything that tests the getters in `ResponseData` should be moved into `ResponseDataSpec`.

In my mind, testing should be the following:
1. There should be dedicated `Spec` files for each writer, which test that they serialize the format correclty.
2. There should be a dedicated `Spec` that tests those `ResponseData` utility methods that are interesting.
3. There should be a `Spec` for `FiliResponseWriter` that ensures it selects the right writer (heh).

fili-core/src/main/java/com/yahoo/bard/webservice/data/HttpResponseMaker.java
```diff
@@ -136,7 +136,7 @@ private ResponseBuilder createResponseBuilder(
                                 LinkedHashMap::new
                         ));
 
-        Response response = new Response(
+        ResponseData response = new ResponseData(
```
archolewa: You're building a `ResponseData` object, but you're not actually using a `ResponseWriter` anywhere (except through the `ResponseData::write` method that should be removed). In order to be effective, the `ResponseWriter` needs to be injected, just like the `HttpResponseMaker`:

1. Add a binding of `ResponseWriter` to the `AbstractBinderFactory` that defaults to the `FiliResponseWriter`.
2. Pass the `ResponseWriter` into the `HttpResponseMaker`. The injection framework should then get the writer into the maker.
3. Use the writer to write the response to the outputstream.
4. Pass the output stream to the [response builder](https://github.com/yahoo/fili/pull/455/files#diff-7bdeb7a580261f6a4a65dd6bf01df1abR154)

tarrantzhang: @archolewa I have trouble making `ResponseWriter` injectable. 

The default `ResponseWriter` requires `ResponseWriterSelector` which requires all three pre-defined `ResponseWriter` to be instantiated  (JSON, JsonApi and Csv) at binding time. But `JsonResponseWriter` and `JsonApiResponseWriter` needs two variables: `responseData` and `paginationLinks` that are not available at binding time.  Is there a way to work around this.

tarrantzhang:
Most line changes are due to refactoring large test cases and `Response` class, the number itself looks scary though.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/HttpResponseMakerSpec.groovy
```diff
@@ -54,8 +56,13 @@ class HttpResponseMakerSpec extends Specification {
         paramMap = Mock(MultivaluedMap)
         DimensionDictionary dimensionDictionary = Mock(DimensionDictionary)
         Dimension dim1 = Mock(Dimension)
+        responseWriter = Mock(ResponseWriter)
 
-        httpResponseChannel = new HttpResponseChannel(Mock(AsyncResponse), apiRequest, new HttpResponseMaker(MAPPERS, dimensionDictionary))
+        httpResponseChannel = new HttpResponseChannel(Mock(AsyncResponse), apiRequest, new HttpResponseMaker(
```
archolewa: When anything needs to be wrapped, everything should be wrapped. So instead of having the first two arguments on the first line, and then wrapping the last, every argument should be on its own line:

```java
     new HttpResponseChannel(
                Mock(AsyncResponse),
                apiRequest,
                new HttpResponseMaker(...)
      )
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/HttpResponseMakerSpec.groovy
```diff
@@ -54,8 +56,13 @@ class HttpResponseMakerSpec extends Specification {
...
+                MAPPERS,
+                dimensionDictionary,
+                responseWriter,
```
archolewa: I guess having a trailing comma is legal in Groovy. Still, we should remove it.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/HttpResponseMakerSpec.groovy
```diff
@@ -82,7 +89,7 @@ class HttpResponseMakerSpec extends Specification {
         resultSet = Mock(ResultSet)
         resultSet.getSchema() >> schema
 
-        httpResponseMaker = new HttpResponseMaker(MAPPERS, dimensionDictionary)
+        httpResponseMaker = new HttpResponseMaker(MAPPERS, dimensionDictionary, responseWriter,)
```
archolewa: Drop the trailing comma after `responseWriter`

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/FiliResponseWriterSpec.groovy
```diff
@@ -0,0 +1,167 @@
...
+        paginating << [false] + [true] * LINK_NAMES_LIST.size()
+        linkNames << [''] + LINK_NAMES_LIST
+        bodyLinks << [[:] as Map] + BODY_LINKS_LIST
+        bodyLinksAsJson << [''] + BODY_LINKS_AS_JSON_LIST
+        arePaginating = paginating ? "paginating with pagination links $linkNames" : "not paginating"
+    }
+
+
+
+    def "Test dimension column name builds correctly"() {
```
archolewa: This is a test on a method in `ResponseData`, not in `FiliResponseWriter`, so it should be in the `ResponseDataSpec`, not the `FiliResponseWriterSpec`.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/FiliResponseWriterSpec.groovy
```diff
@@ -0,0 +1,167 @@
...
+
+        and: "With values to be used to build the string"
+        dimension.getApiName() >> "Foo"
+        dimensionField.getName() >> "Bar"
+
+        expect: "Verify that the correct string is built"
+        ResponseData.getDimensionColumnName(dimension, dimensionField) == "Foo|Bar"
+    }
+
+    def "Test getDimensionColumnName updates cache"() {
```
archolewa: [ditto](https://github.com/yahoo/fili/pull/455/files#diff-41cf17a9bab2c5bd24665e02857fa783R119)

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/FiliResponseWriterSpec.groovy
```diff
@@ -0,0 +1,167 @@
...
+
+        when: "Build, cache and retrieve the name"
+        String name = ResponseData.getDimensionColumnName(dimension, dimensionField)
+
+        then: "Verify that the name is correct and that the name is in the cache"
+        name == "Foo|Bar"
+        ResponseData.DIMENSION_FIELD_COLUMN_NAMES.get(dimension).get(dimensionField) == name
+    }
+
+    def "Test getDimensionColumnName retrieves from cache"() {
```
archolewa: [ditto](https://github.com/yahoo/fili/pull/455/files#diff-41cf17a9bab2c5bd24665e02857fa783R119)

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/FiliResponseWriterSpec.groovy
```diff
@@ -0,0 +1,167 @@
...
+import com.yahoo.bard.webservice.util.SimplifiedIntervalList
+
+import org.joda.time.Interval
+
+import spock.lang.Unroll
+
+class FiliResponseWriterSpec extends ResponseWriterSpec {
+
+    @Unroll
+    def "test for requested numeric metrics in the API response with the #linkNames links"() {
```
archolewa: This should probably be put on the spec for the Json and JsonApi specifications, since the whole metadata block only applies to the json/jsonapi mappers.

/NotABlocker

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/FiliResponseWriterSpec.groovy
```diff
@@ -0,0 +1,167 @@
...
+        GroovyTestUtils.compareJson(os.toString(), expectedJson)
+
+        where:
+        linkNames << LINK_NAMES_LIST
+        bodyLinks << BODY_LINKS_LIST
+        bodyLinksAsJson << BODY_LINKS_AS_JSON_LIST
+    }
+
+    @Unroll
+    def "test for existence of missing intervals in response when #arePaginating"() {
```
archolewa: [ditto](https://github.com/yahoo/fili/pull/455/files#diff-41cf17a9bab2c5bd24665e02857fa783R19)

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/ResponseWriterSelectorSpec.groovy
```diff
@@ -0,0 +1,97 @@
...
+
+    static class DebugWriter implements ResponseWriter {
+        @Override
+        void write(final ApiRequest request, final ResponseData responseData, final OutputStream os)
+                throws IOException {
+
+        }
+    }
+
+    def "test selector selecting correct writer for #TypeNames"() {
```
archolewa: Should add the `Unroll` annotation. The `Unroll` annotation expands each test case into its own separate test (substituting whatever variables need to be substituted in the process).

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/ResponseWriterSelectorSpec.groovy
```diff
@@ -0,0 +1,97 @@
...
+        filiResponseWriterSelector = new FiliResponseWriterSelector(csvResponseWriter, jsonResponseWriter, jsonApiResponseWriter)
+        ResponseWriter actualWriter = filiResponseWriterSelector.select(apiRequest)
+
+        expect:
+        writers.isInstance(actualWriter)
+
+        where:
+        TypeNames << TYPE_NAMES_LIST
+        writers << WRITER_LIST
+        Types << TYPE_LIST
```
archolewa: The variables should all be lowercase, not upper case like `TypeNames` and `Types`.

The variable names should in fact be singular, since Spock iteratively takes an element out of each list and assigns it to the variable. Also, since this is the only place using these values, it will actually be clearer to make a table (like in Cucumber):

```groovy
  typeName | writer                            | type"
 Json"     | JsonResponseWriter                | ResponseFormatType.JSON
 ...
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/ResponseWriterSpec.groovy
```diff
@@ -0,0 +1,327 @@
...
+import org.joda.time.DateTimeZone
+
+import spock.lang.Specification
+
+import javax.ws.rs.core.UriBuilder
+
+/**
+ * This class host common fields and functions shared by all ResponseWriterSpec.
+ */
+abstract class ResponseWriterSpec extends Specification {
```
archolewa: At some point, we'll probably also want to add a suite of tests to this class. Then, implementors can write a spec file that extends this one, and Spock will run these tests for that implementation without the customer having to rewrite them.

I'm not going to block on that though. This is adequate for now, and I'm not even sure what kinds of tests we could write, since the whole point is to customize the output!

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletReactiveChainforResultsEndpointSpec.groovy
```diff
@@ -46,13 +47,15 @@ class JobsServletReactiveChainforResultsEndpointSpec extends Specification {
...
     HttpResponseMaker httpResponseMaker
+    ResponseWriter responseWriter
 
     def setup() {
         objectMappersSuite = Mock(ObjectMappersSuite)
         ObjectMapper objectMapper = Mock(ObjectMapper)
         ObjectWriter objectWriter = Mock(ObjectWriter)
         objectMappersSuite.getMapper() >> objectMapper
         objectMapper.writer() >> objectWriter
+        responseWriter = Mock(ResponseWriter)
```
archolewa: I would probably just build a simple ResponseWriter here (maybe the `JsonResponseWriter`). I generally prefer to use mocks only when it's either impossible, or a truly MASSIVE pain to build an actual object, and that object isn't pertinent to the test. Since it's pretty simple to build a ResponseWriter, I don't see any reason not to use a real one.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletReactiveChainforResultsEndpointSpec.groovy
```diff
@@ -62,7 +65,7 @@ class JobsServletReactiveChainforResultsEndpointSpec extends Specification {
 
         preResponseStore = new HashPreResponseStore()
         broadcastChannel = new SimpleBroadcastChannel<>(PublishSubject.create())
-        httpResponseMaker = new HttpResponseMaker(objectMappersSuite, dimensionDictionary)
+        httpResponseMaker = new HttpResponseMaker(objectMappersSuite, dimensionDictionary, responseWriter,)
```
archolewa: Should drop the trailing comma after `responseWriter`

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/responseprocessors/MappingResponseProcessorSpec.groovy
```diff
@@ -31,14 +32,20 @@ class MappingResponseProcessorSpec extends Specification{
...
         apiRequest = Mock(DataApiRequest)
         objectMappers = new ObjectMappersSuite()
+        responseWriter = Mock(ResponseWriter)
 
         AsyncResponse asyncResponse = Mock(AsyncResponse)
-        HttpResponseChannel httpResponseChannel = new HttpResponseChannel(asyncResponse, apiRequest, new HttpResponseMaker(objectMappers, Mock(DimensionDictionary)))
+        HttpResponseChannel httpResponseChannel = new HttpResponseChannel(asyncResponse, apiRequest, new HttpResponseMaker(
+                objectMappers,
+                Mock(DimensionDictionary),
+                responseWriter,
```
archolewa: [ditto](https://github.com/yahoo/fili/pull/455/files#diff-fd9bfd667dcfa0b4c6d56f006840de39R68)

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/responseprocessors/MappingResponseProcessorSpec.groovy
```diff
@@ -31,14 +32,20 @@ class MappingResponseProcessorSpec extends Specification{
     DataApiRequest apiRequest
     ObjectMappersSuite objectMappers
     Subject<PreResponse, PreResponse> mappingResponseChannel
+    ResponseWriter responseWriter
 
     def setup() {
         groupByQuery = Mock(GroupByQuery)
         apiRequest = Mock(DataApiRequest)
         objectMappers = new ObjectMappersSuite()
+        responseWriter = Mock(ResponseWriter)
```
archolewa: [ditto](https://github.com/yahoo/fili/pull/455/files#diff-fd9bfd667dcfa0b4c6d56f006840de39R58)

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/responseprocessors/ResultSetResponseProcessorSpec.groovy
```diff
@@ -90,8 +91,9 @@ class ResultSetResponseProcessorSpec extends Specification {
         uriInfo = Mock(UriInfo)
         pathSegment = Mock(PathSegment)
         paramMap = Mock(MultivaluedMap)
+        responseWriter = Mock(ResponseWriter)
```
archolewa: [ditto](https://github.com/yahoo/fili/pull/455/files#diff-fd9bfd667dcfa0b4c6d56f006840de39R58)

fili-core/src/main/java/com/yahoo/bard/webservice/web/FiliResponseWriter.java
```diff
@@ -0,0 +1,49 @@
...
+public class FiliResponseWriter implements ResponseWriter {
+
+    private final FiliResponseWriterSelector responseWriterSelector;
+
+    /**
+     * Constructor.
+     *
+     * @param responseWriterSelector  Used to decided writer type when called write.
+     */
+    public FiliResponseWriter(FiliResponseWriterSelector responseWriterSelector) {
```
tarrantzhang: @archolewa This is a very educative explanation! Stupid me, can't believe I thought `com.java.astrology` exist for a few seconds.

fili-core/src/main/java/com/yahoo/bard/webservice/web/FiliResponseWriterSelector.java
```diff
@@ -38,16 +39,12 @@ public FiliResponseWriterSelector(
...
         ResponseFormatType format = request.getFormat();
         if (format == null) {
             format = ResponseFormatType.JSON;
         }
-        return writers.get(format);
-    }
-
-    public Map<ResponseFormatType, ResponseWriter> getWriters() {
-        return writers;
+        return Optional.of(writers.get(format));
```
archolewa: This should be `Optional.ofNullable`. `Optional.of` explodes if the value it attempts to write is `null` (which will happen if the format the user provided isn't in the map). `Optional.ofNullable` meanwhile will return an `Optional.empty` if the wrapped value is `null`.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/CsvResponseWriterSpec.groovy
```diff
@@ -0,0 +1,155 @@
...
+                "platform_version",
+                "product",
+                "product_family",
+                "pty_country",
+                "user_state"
+        ]
+
+        Map<Dimension, Set<DimensionField>> defaultDimensionFieldsToShow = [:] as LinkedHashMap
+
+        DataApiRequest apiRequest1 = Mock(DataApiRequest) {
```
kevinhinterlong: Maybe use a different name here

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/FiliResponseWriterSpec.groovy
```diff
@@ -0,0 +1,151 @@
...
+                  "timeSpent": 10
+                }
+              ]
+            }"""
+
+        and: "The format type for serialization"
+        apiRequest.getFormat() >> ResponseFormatType.JSON
+
+        when: "Serialize the data with FiliResponseWriter "
+        FiliResponseWriter filiResponseWriter = new FiliResponseWriter(new FiliResponseWriterSelector(
```
kevinhinterlong: Probably make `FiliResponseWriter` a field

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/UIJsonResponseSpec.groovy
```diff
@@ -108,20 +114,23 @@ class UIJsonResponseSpec extends Specification {
...
+                [:]
         )
 
         ByteArrayOutputStream os = new ByteArrayOutputStream()
-        jro.write(os)
+        csvResponseWriter = new CsvResponseWriter(MAPPERS)
+        jsonResponseWriter = new JsonResponseWriter(MAPPERS)
+        jsonApiResponseWriter = new JsonApiResponseWriter(MAPPERS)
+        filiResponseWriterSelector = new FiliResponseWriterSelector(csvResponseWriter, jsonResponseWriter, jsonApiResponseWriter);
+        filiResponseWriter = new FiliResponseWriter(filiResponseWriterSelector)
```
kevinhinterlong: Same thing here, probably make it a field

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/ResponseDataSpec.groovy
```diff
@@ -0,0 +1,237 @@
+// Copyright 2016 Yahoo Inc.
```
kevinhinterlong: fix copyright

fili-core/src/main/java/com/yahoo/bard/webservice/web/ResponseData.java
```diff
@@ -0,0 +1,304 @@
+// Copyright 2016 Yahoo Inc.
```
kevinhinterlong: Honestly not sure on this one, but maybe update

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/ResponseDataSpec.groovy
tarrantzhang: good catch thanks

archolewa:
@kevinhinterlong I agree that having custom response types would be nice to have, but it's outside the scope of this PR. When someone wants custom response types, they can come in and make the refactors to the response type (i.e. hiding the enum behind an interface) that will be necessary.

kevinhinterlong:
@archolewa I agree I think this PR is ready. I'll go ahead and make an issue for this

tarrantzhang:
Rebased and squashed, ready for merge.

PR #459, opened 2017-07-29
use 'maven-enforcer-plugin' instead of maven version prerequisites (in order to eliminate build warnings)

dejan2609:
…


dejan2609:
Excellent catch @kevinhinterlong !  

`<enforcer.skip>true</enforcer.skip>` is part of the project from a start (i.e. since initial commit) and it does skips [maven-enforcer-plugin execution](http://maven.apache.org/enforcer/maven-enforcer-plugin/enforce-mojo.html#skip). 
However, property presence/value comes into force only in the presence of maven-enforcer-plugin (without that plugin this property is useless).

Given a fact that my initial lower maven interval endpoint was too low I failed to recognize property skip presence. I'll change my commit jut a little bit, herewith a changes:

1. 
`<enforcer.skip>false</enforcer.skip>` 
  
2.  
```
   <requireMavenVersion>
       <version>[3.3.9,)</version>
    </requireMavenVersion>
```

dejan2609:
Note: I rebased few times, sorry for additional noise.

dejan2609:
@will-lauer and @michael-mclawhorn: please review.

dejan2609:
@will-lauer: please review.

Issue #462, opened 2017-07-31
Make functional tests not lock on the same files


kevinhinterlong:
The `LuceneSearchProvider` requires a lot of file locks and tests using it operate on the same the same `lucene_index_path` for reading/writing. Since all these tests operate in the same directory, it prevents multiple of these tests running at the same time and provides a potential place for test interference.

The end goal being would be to make tests able to run in parallel, whether or not that actually happens by default will probably depend on how well travis handles it.

QubitPi:
Addressed

Issue #356, opened 2017-06-09
Make functional tests use a dynamic port


cdeszaq:
Using a dynamic port makes the tests less brittle in the event that more than 1 test is running on the box. We can tell Jersey to use an open port. Here's how:

Jersey docs: https://jersey.java.net/apidocs/2.5.1/jersey/org/glassfish/jersey/test/TestProperties.html#CONTAINER_PORT

And a sample for how to do it:

```java
/***** Old *****/
// Set Base URL to Grizzly container
System.setProperty( Configuration.BASE_URL_KEY, "http://localhost:9998" )

/***** New *****/
// Setting jersey's port to 0 causes it to pass 0 to its call to bind, which forces bind to look
// for an unused port and use that. This allows us to run jersey on an arbitrary unused port and avoid
// collisions with any other service already running on the system. This avoids port collisions when 
// running multiple builds simultaneously.
System.setProperty("jersey.config.test.container.port", "0");
jtb = new JerseyTestBinder(resourceClasses)

// Set Base URL to Grizzly container. Make sure to get the current jersey port, since a random unused port
// was allocated when the JerseyTest object was created. 
System.setProperty( Configuration.BASE_URL_KEY, "http://localhost:" + Integer.toString(jtb.harness.getPort()))
```

PR #463, opened 2017-07-31
Use dynamic ports in JerseyTestBinder for tests

kevinhinterlong:
Addresses #356 (The next issue which needs to be accomplished before tests could _probably_ be parallel would be #462)

Convert static bindings to port 9998 into dynamically assigned ports

Ideally we would be able to specify a range, but we don't have any interface to do this and checking manually if a port is available is error prone since jersey has to actually bind to it.

### DIFF
--- a/fili-core/src/test/java/com/yahoo/bard/webservice/web/endpoints/JobsEndpointResources.java
+++ b/fili-core/src/test/java/com/yahoo/bard/webservice/web/endpoints/JobsEndpointResources.java
@@ -49,7 +49,7 @@ public class JobsEndpointResources {
         fieldValueMap1.put(DefaultJobField.JOB_TICKET, "ticket1");
         fieldValueMap1.put(DefaultJobField.DATE_CREATED, "2016-01-01");
         fieldValueMap1.put(DefaultJobField.DATE_UPDATED, "2016-01-01");
-        fieldValueMap1.put(DefaultJobField.QUERY, "https://localhost:9998/v1/data/QUERY");
+        fieldValueMap1.put(DefaultJobField.QUERY, "https://localhost:PORT/v1/data/QUERY");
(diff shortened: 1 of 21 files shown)


fili-core/src/test/java/com/yahoo/bard/webservice/web/endpoints/JobsEndpointResources.java
```diff
@@ -49,7 +49,7 @@ public static ApiJobStore getApiJobStore() {
         fieldValueMap1.put(DefaultJobField.JOB_TICKET, "ticket1");
         fieldValueMap1.put(DefaultJobField.DATE_CREATED, "2016-01-01");
         fieldValueMap1.put(DefaultJobField.DATE_UPDATED, "2016-01-01");
-        fieldValueMap1.put(DefaultJobField.QUERY, "https://localhost:9998/v1/data/QUERY");
+        fieldValueMap1.put(DefaultJobField.QUERY, "https://localhost:PORT/v1/data/QUERY");
```
kevinhinterlong: The only places this refactor gets weird is when `JobsEndpointResources` is used (i.e. `JobsServletSpec.groovy`).

This is because we need to build the binder (including `JobsEndpointResources`) to start jersey, but we need to start jersey to know the port and I couldn't think of a way to inject the port.

tarrantzhang: Correct me if I'm wrong. I think one way you can do it without using a placeholder port is to put the expected result resource(e.g. `JobsEndpointResources.java`) in the `JobsServletSpec.groovy` so you can initialize your resource after setting up Jersey.

tarrantzhang:
👍 lgtm.

Issue #301, opened 2017-05-24
Add Jackson's Afterburner module for better performance


cdeszaq:
This module boosts serialization / deserialization speed by 30-40%, so we should see if we get a boost: https://github.com/FasterXML/jackson-module-afterburner

We ingest potentially large JSON and output it as well, so it's likely that we'll see a boost.

Here's a doc about Jackson performance best practices that includes Afterburner and we might want to consider when we enable it: https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance

kevinhinterlong:
Now that #466 has been merged this issue is just to follow [Jackson's best practices guide](https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance).

kevinhinterlong:
One place that deserves cleanup is `ScanSearchProvider`. 

Right now it creates a new `ObjectMapper` every time it is constructed. Instead we could create a single `ObjectMapper` in `ScanSearchProviderManager` and provide it as a parameter

PR #466, opened 2017-07-31
Add jackson afterburner module

kevinhinterlong:
Addresses #301


michael-mclawhorn:
👍

kevinhinterlong:
After this is merged, I'll open another issue for "Jackson best practices" so that it' still recorded.

Issue #362, opened 2017-06-09
Move timing in servlets into finally blocks to ensure it's counted and use suppliers to return


cdeszaq:
This simplifies the code a bit, and makes sure timing is respected, even as the code changes. Here's an example:

```java
   ) {
        Supplier<Response> responseSender;
        try {
            RequestLog.startTiming(this);
            RequestLog.record(new TableRequest(tableName, grain));

            TablesApiRequest apiRequest = new TablesApiRequest(
                    // ...
            );

            Map<String, Object> result = getLogicalTableFullView(apiRequest.getTable(), uriInfo);
            String output = mapper.writeValueAsString(result);
            LOG.debug("Tables endpoint Response: {}", output);
            responseSender = () -> Response.status(Response.Status.OK).entity(output).build();
        } catch (IOException e) {
            String msg = String.format("Internal server error. IOException : %s", e.getMessage());
            LOG.error(msg, e);
            responseSender = () -> Response.status(INTERNAL_SERVER_ERROR).entity(msg).build();
        } catch (Error | Exception e) {
            String msg = String.format("Exception processing request: %s", e.getMessage());
            LOG.info(msg, e);
            responseSender = () -> Response.status(Response.Status.BAD_REQUEST).entity(msg).build();
        } finally {
            RequestLog.stopTiming(this);
        }
        return responseSender.get();
    }
```

PR #471, opened 2017-07-31
Cleanup RequestLog timing usage on servlets

kevinhinterlong:
Addresses #362 

- Calls to stop `RequestLog` timers should be in finally blocks

**NOTES:** 

- I think there were some problems with the handling of `RequestLog` in `JobsServlet`, I've made the most significant changes there
- I'm not super familiar with ReactiveX so there's probably something I could have done better.


kevinhinterlong:
I think it would make sense for #496 to land first. Then I could redo this adding those annotations while keeping the cleanup

QubitPi:
@kevinhinterlong This PR is missing a CHANGELOG

PR #474, opened 2017-08-02
Fixing the performance problems around feature flags by making them c…

michael-mclawhorn:
…ache.

The performance on these objects is terrible because the entire underlying config map gets copied on access.  This protects against that problem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,11 @@ Current
 
 ### Fixed:
 
+- [Fix performance bug around feature flag](https://github.com/yahoo/fili/issues/473)
+    * BardFeatureFlag, when used in a tight loop, is very expensive.  Underlying map configuration copies the config map on each access.
+    * Switching to lazy value evaluation
+    * Added reset contract so changes to feature flags can be directly reverted rather than going through the `SystemConfig` directly
+
 - [Fix deploy branch issue where substrings of whitelisted branches could be released](https://github.com/yahoo/fili/issues/453)
 
 - [Fix availability testing utils to be compatible with composite tables](https://github.com/yahoo/fili/pull/419)
@@ -68,11 +73,6 @@ Current
 
 - [Fix metric and dimension names for wikipedia-example](https://github.com/yahoo/fili/pull/415)
     * The metrics and dimensions configured in the `fili-wikipedia-example` were different from those in Druid and as a result the queries sent to Druid were incorrect
-
-- [Fix performance bug around feature flag](https://github.com/yahoo/fili/issues/473)
-    * BardFeatureFlag, when used in a tight loop, is very expensive.  Underlying map configuration copies the config map on each access.
-    * Switching to lazy value evaluation
-    * Added reset contract so changes to feature flags can be directly reverted rather than going through the `SystemConfig` directly
 
 
 ### Known Issues:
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -69,6 +69,11 @@ Current
 - [Fix metric and dimension names for wikipedia-example](https://github.com/yahoo/fili/pull/415)
     * The metrics and dimensions configured in the `fili-wikipedia-example` were different from those in Druid and as a result the queries sent to Druid were incorrect
 
+- [Fix performance bug around feature flag](https://github.com/yahoo/fili/issues/473)
```
garyluoex: We should put it on the top instead of bottom I think

garyluoex: We should put it on the top instead of bottom I think

fili-core/src/main/java/com/yahoo/bard/webservice/config/BardFeatureFlag.java
```diff
@@ -38,11 +39,22 @@ public String getName() {
...
 
     @Override
     public void setOn(Boolean newValue) {
         SYSTEM_CONFIG.setProperty(SYSTEM_CONFIG.getPackageVariableName(propertyName), newValue.toString());
+        isOn = null;
+        isOn();
+    }
+
+    @Override
+    public void reset() {
```
garyluoex: It is unclear whether this reset only reset cache or clear the config itself.

fili-core/src/main/java/com/yahoo/bard/webservice/config/CacheFeatureFlag.java
```diff
@@ -47,31 +48,49 @@ public String getName() {
...
+                return (this.value.equals("Ttl")
+                        && !BardFeatureFlag.DRUID_CACHE_V2.isOn()
+                        && BardFeatureFlag.DRUID_CACHE.isOn()
                 )
-        );
+                        || (this.value.equals("LocalSignature")
+                        && BardFeatureFlag.DRUID_CACHE.isOn()
+                        && BardFeatureFlag.DRUID_CACHE_V2.isOn());
+            }
+            on = value.equalsIgnoreCase(SYSTEM_CONFIG.getStringProperty(
```
garyluoex: Assuming caching strategy is immutable during runtime.

fili-core/src/main/java/com/yahoo/bard/webservice/config/BardFeatureFlag.java
```diff
@@ -21,6 +21,7 @@
     static final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance();
 
     private final String propertyName;
+    private Boolean on = null;
```
QubitPi: How about putting assignment in constructor?

fili-core/src/main/java/com/yahoo/bard/webservice/config/BardFeatureFlag.java
```diff
@@ -38,11 +39,22 @@ public String getName() {
...
+            on = SYSTEM_CONFIG.getBooleanProperty(SYSTEM_CONFIG.getPackageVariableName(propertyName), false);
+        }
+        return on;
     }
 
     @Override
     public void setOn(Boolean newValue) {
         SYSTEM_CONFIG.setProperty(SYSTEM_CONFIG.getPackageVariableName(propertyName), newValue.toString());
+        on = null;
+        isOn();
```
QubitPi: If we are trying to update `on`, how about `on = newValue`?

fili-core/src/main/java/com/yahoo/bard/webservice/config/CacheFeatureFlag.java
```diff
@@ -30,6 +30,7 @@
     private static final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance();
 
     private final String value;
+    Boolean on = null;
```
QubitPi: How about putting assignment in constructor?

PR #475, opened 2017-08-02
RFC for Sql Support in Fili

kevinhinterlong:


rfc/sqlDesignDocs.md
```diff
@@ -0,0 +1,84 @@
...
+Customers will be able to configure their connection to a SQL database similar to Druid through the `applicationConfig.properties` and build a table in Fili specifically backed by SQL. They can then depend on the `fili`  and `fili-sql`module as opposed for configuring their application with SQL support.
+
+## Implementation
+
+1. Create a `fili-sql` module.
+
+2. Define a `SqlBackedClient` 
+
+   ```java
+   Future<JsonNode> executeQuery(
```
archolewa: If I remember correctly, we don't use a future in the Druid backend. Everything is handled in the success callback. Any particular reason we're not doing the same thing here?

I'm not opposed to using a Future, I just think it's generally easier to follow the same convention for the same kinds of things unless there's a good reason not to.

rfc/sqlDesignDocs.md
```diff
@@ -0,0 +1,84 @@
...
+      AllGranularity.INSTANCE -> ()
+      DefaultTimeGrain.YEAR   -> (YEAR)
+      DefaultTimeGrain.MONTH  -> (YEAR, MONTH)
+      DefaultTimeGrain.WEEK   -> (YEAR, WEEK)
+      DefaultTimeGrain.DAY    -> (YEAR, DAYOFYEAR)
+      DefaultTimeGrain.HOUR   -> (YEAR, DAYOFYEAR, HOUR)
+      DefaultTimeGrain.MINUTE -> (YEAR, DAYOFYEAR, HOUR, MINUTE)
+      ```
+
+   2. Define an `ApiToFieldMapper` to handle converting from API names to physical names.
```
archolewa: Remind me again, why is this necessary? Why can't the physical table backing the SQL backend handle this, like it does for Druid?

rfc/sqlDesignDocs.md
kevinhinterlong: The [`DruidWebService`](https://github.com/yahoo/fili/blob/2bd3ce3af7942ec485bb4bc84a37f7de21ea402a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/DruidWebService.java) got changed to return a future in #254 to clean up the blocking calls in `fili-generic-example`

rfc/sqlDesignDocs.md
kevinhinterlong: This may have been unnecessarily abstracted then because my implementation of `ApiToFieldMapper` is using the physical table. 

This is just a helper class that seemed useful to me, but it might not have been useful to put it in the RFC

kevinhinterlong:
Move to `docs/`

kevinhinterlong:
This `SqlBackedClient` interface is missing the `RequestContext` like in [`DruidWebService`](https://github.com/kevinhinterlong/fili/blob/2bd3ce3af7942ec485bb4bc84a37f7de21ea402a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/client/DruidWebService.java#L29) which is needed to stop/start the `RequestLog` phases.

PR #476, opened 2017-08-02
Part 1 of RFC for SQL Support

kevinhinterlong:
Initialize fili-sql module
@michael-mclawhorn @archolewa 

RFC: #475

### DIFF
--- a/fili-sql/src/main/resources/moduleConfig.properties
+++ b/fili-sql/src/main/resources/moduleConfig.properties
@@ -1,0 +1,21 @@
+# Copyright 2017 Yahoo Inc.
+# Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+
+#The name of this module (required)
+moduleName = fili-sql
+
+moduleDependencies = fili-core
+
+# Disabling the druid dimension loader by default
+# When set to true you will also need to populate the list of dimensions using druid_dim_loader_dimensions
+bard__druid_dimensions_loader_enabled = false
+bard__druid_dim_loader_row_limit = 1000
+
+# Add support for SQL backend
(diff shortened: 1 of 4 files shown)


fili-sql/src/main/resources/moduleConfig.properties
```diff
@@ -0,0 +1,153 @@
...
+bard__request_limit_global = 70
+bard__request_limit_per_user = 2
+bard__request_limit_ui = 52
+
+# setting for druid query weight limit
+bard__query_weight_limit = 100000
+
+# Default the timeout to 10 minutes, in milliseconds
+bard__druid_request_timeout = 600000
+
```
michael-mclawhorn: Clean out config that isn't specific to your module.
Add module dependency to fili-core

kevinhinterlong: Updated

fili-sql/src/main/resources/moduleConfig.properties
```diff
@@ -0,0 +1,21 @@
...
+#The name of this module (required)
+moduleName = fili-sql
+
+moduleDependencies = fili-core
+
+# Disabling the druid dimension loader by default
+# When set to true you will also need to populate the list of dimensions using druid_dim_loader_dimensions
+bard__druid_dimensions_loader_enabled = false
+bard__druid_dim_loader_row_limit = 1000
```
kevinhinterlong: These aren't relevant until later in the rfc, should I take it out @michael-mclawhorn ?

michael-mclawhorn: Nah. It'll arrive in time.

PR #478, opened 2017-08-02
Part 3 (partial-1) of RFC for SQL Support

kevinhinterlong:
Adds 
- support for converting `Aggregation` to Sql aggregations
- adds a time converter for grouping/filtering on intervals
- An ApiToField mapper so that the correct metrics/dimensions are accessed on the sql table

RFC: #475


fili-sql/src/main/java/com/yahoo/bard/webservice/sql/ApiToFieldMapper.java
```diff
@@ -0,0 +1,39 @@
...
+        this.physicalTableSchema = physicalTableSchema;
+    }
+
+    @Override
+    public String apply(String input) {
+        return physicalTableSchema.getPhysicalColumnName(input);
+    }
+
+    /**
+     * Finds the api name given a field name and if none are found then it returns the given inpu.
```
archolewa: Should be `input` not `inpu`.

fili-sql/src/main/java/com/yahoo/bard/webservice/sql/aggregation/DruidSqlAggregationConverter.java
```diff
@@ -0,0 +1,70 @@
...
+     * @param aggregation  The druid aggregation, i.e.
+     * {@link com.yahoo.bard.webservice.druid.model.aggregation.LongSumAggregation}.
+     * @param apiToFieldMapper  the mapping from api name to field name of the aggregation.
+     *
+     *  @return the supported sql aggregation type.
+     */
+    @Override
+    public Optional<SqlAggregation> apply(Aggregation aggregation, ApiToFieldMapper apiToFieldMapper) {
+        String aggregationType = aggregation.getType();
+        return druidToSqlAggregation.containsKey(aggregationType) ?
```
archolewa: FYI, you can also it as follows:

```java
  return Optional
          .ofNullable(druidToSqlAggregation.get(aggregationType))
          .map(sqlAggregationType -> sqlAggregationType.getSqlAggregation(aggregation, apiToFieldMapper));
```
A bit cleaner in my opinion, because you only have to deal with the case where the aggregation is present. `Optional` will naturally handle the `null` case.

PR #484, opened 2017-08-03
Move avroDimensionRowParsing to a stream

michael-mclawhorn:
Performance from pulling the entire file into memory is pretty unacceptable.  This change makes the caller responsible for pulling the values off the stream and expects the stream to get closed to notify the file to close.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
@@ -107,7 +107,7 @@ public class AvroDimensionRowParser {
     }
 
     /**
-     * Returns a stream which parses avro records into dimension rows
+     * Returns a stream which parses avro records into dimension rows.
      *
      * @param dataFileReader  An open file reader for avro records
      * @param dimension  The dimension object used to configure the dimension
@@ -135,7 +135,7 @@ public class AvroDimensionRowParser {
     }
 
     /**
-     * Parses the avro file and sends dimension rows to a consumer
+     * Parses the avro file and sends dimension rows to a consumer.
      *
      * @param dimension  The dimension object used to configure the dimension
      * @param avroFilePath  The path of the AVRO data file (.avro)
@@ -160,7 +160,7 @@ public class AvroDimensionRowParser {
     }
 
     /**
-     * Parses the avro file and returns the dimension rows
+     * Parses the avro file and returns the dimension rows.
      *
      * @param dimension The dimension object used to configure the dimension
      * @param avroFilePath The path of the AVRO data file (.avro)
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
@@ -123,7 +123,10 @@ public class AvroDimensionRowParser {
             Runnable fileCloser = () -> {
                 try {
                     dataFileReader.close();
-                } catch (IOException ignore) {
+                } catch (IOException e) {
+                    String msg = String.format("Error closing avro file, at the location %s", avroFilePath);
+                    LOG.error(msg, e);
+                    throw new IllegalArgumentException(msg, e);
                 }
             };
             // Generates a set of dimension Rows after retrieving the appropriate fields
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
@@ -19,7 +19,9 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
 /**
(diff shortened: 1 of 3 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -93,34 +95,45 @@ private boolean doesSchemaContainAllDimensionFields(Dimension dimension, Schema
...
-                                 )
-                            )
-                    )
+                    .map(recordMapFunction)
                     .map(dimension::parseDimensionRow)
-                    .collect(Collectors.toSet());
+                    .onClose(fileCloser);
 
         } catch (IOException e) {
             String msg = String.format("Unable to process the file, at the location %s", avroFilePath);
```
kevinhinterlong: I think we should call `fileCloser.run()` here to be safe

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -93,34 +95,45 @@ private boolean doesSchemaContainAllDimensionFields(Dimension dimension, Schema
...
-                            .collect(
-                                 Collectors.toMap(
-                                     DimensionField::getName,
-                                     dimensionField -> resolveRecordValue(
-                                             genericRecord, dimensionFieldNameMapper.convert(dimension, dimensionField))
-                                 )
-                            )
-                    )
+                    .map(recordMapFunction)
                     .map(dimension::parseDimensionRow)
```
kevinhinterlong: Before if we threw an `IllegalArgumentException` here it would be closed before the `catch` block.

Now if it happens (if I'm reading [the javadoc](http://docs.oracle.com/javase/8/docs/api/java/util/stream/BaseStream.html#onClose-java.lang.Runnable-) correctly) we don't end up closing the file. So I think we'll need a `try`/`catch` in here

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -93,34 +95,45 @@ private boolean doesSchemaContainAllDimensionFields(Dimension dimension, Schema
...
+    public Stream<DimensionRow> parseAvroFileDimensionRows(Dimension dimension, String avroFilePath)
         throws IllegalArgumentException {
 
         // Creates an AVRO DatumReader object
         DatumReader<GenericRecord> datumReader = new GenericDatumReader<>();
 
         // Creates an AVRO DataFileReader object that reads the AVRO data file one record at a time
-        try (DataFileReader<GenericRecord> dataFileReader = new DataFileReader<>(new File(avroFilePath), datumReader)) {
+        try {
+            DataFileReader<GenericRecord> dataFileReader = new DataFileReader<>(new File(avroFilePath), datumReader);
```
will-lauer: What happens if this generates an error? Previously, it would have closed the file, and now it won't. That will effectively leak the file, to be closed later during GC. I'm not sure if we care about waiting that long or not.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -93,34 +95,45 @@ private boolean doesSchemaContainAllDimensionFields(Dimension dimension, Schema
...
+                                    DimensionField::getName,
+                                    dimensionField -> resolveRecordValue(
+                                            genericRecord, dimensionFieldNameMapper.convert(dimension, dimensionField))
+                            )
+                    );
+            Runnable fileCloser = () -> {
+                try {
+                    dataFileReader.close();
+                } catch (IOException e) {
+                    String msg = String.format("Error closing avro file, at the location %s", avroFilePath);
```
will-lauer: Generally, if the file is already closed when the closer gets called, any exception that would be generated should probably be swallowed and not propagated. Other errors other than file already closed should propagate. This prevents weird errors from occuring if the file is closed multitple times.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -93,34 +95,45 @@ private boolean doesSchemaContainAllDimensionFields(Dimension dimension, Schema
...
-                                     DimensionField::getName,
-                                     dimensionField -> resolveRecordValue(
-                                             genericRecord, dimensionFieldNameMapper.convert(dimension, dimensionField))
-                                 )
-                            )
-                    )
+                    .map(recordMapFunction)
                     .map(dimension::parseDimensionRow)
-                    .collect(Collectors.toSet());
+                    .onClose(fileCloser);
```
will-lauer: Is it reasonable to add the .onClose first before the map to ensure that things get closed properly if map fails?

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
michael-mclawhorn: onClose takes a runnable
I'm not sure I understand this feedback

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
michael-mclawhorn: That makes sense

kevinhinterlong: If the map fails `onClose` is not called

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
kevinhinterlong: Oh I should have been more clear about the context. I think we should call `fileCloser.run()` inside of
```java
catch (IOException e) {
    // here
}
```
just to ensure we close the file if something forced us to exit

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
kevinhinterlong: Ok so I played around with it a little and now I understand what `onClose` actually does. Just note that since we're actually giving the stream a resource now we have to close it.

`onClose` is only called if you call `close` so in the test cases we should call `close` on the stream. 

This might be worth adding to the javadoc as well.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -83,6 +85,96 @@ private boolean doesSchemaContainAllDimensionFields(Dimension dimension, Schema
...
+
+        // Generates a set of dimension Rows after retrieving the appropriate fields
+        return StreamSupport.stream(dataFileReader.spliterator(), false)
+                .map(recordMapFunction)
+                .map(dimension::parseDimensionRow);
+
+    }
+
+    /**
+     * Parses the avro file and populates the dimension rows after validating the schema.
```
kevinhinterlong: Update javadoc

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -83,6 +85,96 @@ private boolean doesSchemaContainAllDimensionFields(Dimension dimension, Schema
...
+                                DimensionField::getName,
+                                dimensionField -> resolveRecordValue(
+                                        genericRecord,
+                                        dimensionFieldNameMapper.convert(dimension, dimensionField)
+                                )
+                        ));
+    }
+
+    /**
+     * Parses the avro file and populates the dimension rows after validating the schema.
```
kevinhinterlong: Update javadoc

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -83,6 +85,96 @@ private boolean doesSchemaContainAllDimensionFields(Dimension dimension, Schema
...
+     *
+     * @param dimension  The dimension object used to configure the dimension
+     * @param avroFilePath  The path of the AVRO data file (.avro)
+     * @param consumer  A consumer to process records from the avro file
+     *
+     * @throws IllegalArgumentException thrown if JSON object `fields` is not present
+     */
+    public void parseAvroFileDimensionRows(Dimension dimension, String avroFilePath, Consumer<DimensionRow> consumer) {
```
kevinhinterlong: Bring the `throws IllegalArgumentException` with

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/AvroDimensionRowParser.java
```diff
@@ -83,6 +85,96 @@ private boolean doesSchemaContainAllDimensionFields(Dimension dimension, Schema
...
+     *
+     * @param dataFileReader  An open file reader for avro records
+     * @param dimension  The dimension object used to configure the dimension
+     * @param avroFilePath  The path of the AVRO data file (.avro)
+     *
+     * @return A stream over the open file which produces dimension rows
+     *
+     * @throws IllegalArgumentException thrown if JSON object `fields` is not present
+     */
+    private Stream<DimensionRow> streamDimensionRows(
```
kevinhinterlong: Bring the `throws IllegalArgumentException` with

Issue #351, opened 2017-06-09
Make getInnerQuery return Optional instead of null


cdeszaq:
`Optional` is more correct. Returning `null` is asking for trouble.

PR #485, opened 2017-08-03
DruidQuery::getInnerQuery and Datasource::getQuery return Optional

kevinhinterlong:
Addresses #351 and applied to `Datasource::getQuery` as well.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/FilteredAggregationMaker.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/FilteredAggregationMaker.java
@@ -66,7 +66,7 @@ public class FilteredAggregationMaker extends MetricMaker {
                 new TemplateDruidQuery(
                         ImmutableSet.of(filteredAggregation),
                         Collections.emptySet(),
-                        sourceMetric.getTemplateDruidQuery().getInnerQuery()
+                        sourceMetric.getTemplateDruidQuery().getInnerQueryUnchecked()
                 ),
                 sourceMetric.getCalculation(),
                 metricName
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/FilteredAggregationMaker.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/FilteredAggregationMaker.java
@@ -66,7 +66,7 @@ public class FilteredAggregationMaker extends MetricMaker {
                 new TemplateDruidQuery(
                         ImmutableSet.of(filteredAggregation),
                         Collections.emptySet(),
-                        sourceMetric.getTemplateDruidQuery().getInnerQueryUnchecked()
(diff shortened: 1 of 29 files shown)


kevinhinterlong:
This is going to need a doc update at least for `configuring-metrics.md`

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/TemplateDruidQuery.java
```diff
@@ -43,7 +44,7 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(TemplateDruidQuery.class);
 
-    private final TemplateDruidQuery nestedQuery;
+    private final Optional<TemplateDruidQuery> nestedQuery;
```
archolewa: It's generally considered best practice not to wrap a field in `Optional`. Instead, the field is the raw value, and the guts of the class deal with `null`. Then, when you want to return a reference to the value, you wrap the return value in an `Optional`: https://zeroturnaround.com/rebellabs/java-8-best-practices-cheat-sheet/

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/TemplateDruidQuery.java
```diff
@@ -196,6 +202,9 @@ public boolean isTimeGrainValid() {
     public TemplateDruidQuery merge(TemplateDruidQuery sibling) {
 
         // TODO: Handle merging with a null TDQ
+//        if (!nestedQuery.isPresent()) {
```
archolewa: If we don't need this code, we should just delete it. Let version control remember it.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/FilteredAggregationMaker.java
```diff
@@ -66,7 +66,7 @@ protected LogicalMetric makeInner(String metricName, List<String> dependentMetri
                 new TemplateDruidQuery(
                         ImmutableSet.of(filteredAggregation),
                         Collections.emptySet(),
-                        sourceMetric.getTemplateDruidQuery().getInnerQuery()
+                        sourceMetric.getTemplateDruidQuery().getInnerQuery().orElse(null)
```
tarrantzhang: Maybe other expert on `Optional` can correct me. But isn't the point of using optional is to avoid null check? `orElse()` provides a chance to supply a default value to a nullable object then wouldn't feeding it with a null counteract the merit of `Optional`?

kevinhinterlong: See [Andrew's comment](https://github.com/yahoo/fili/pull/485#discussion_r131694877)

Although I do agree it gets a little awkward here - it could probably use a comment on the constructor at least.

tarrantzhang: Then that's fine I guess. I don't have any other comment.

Issue #394, opened 2017-06-09
Slices endpoint gives api name instead of druid name


cdeszaq:
Since it's main purpose in life is to surface debug info about segment info pulled in from Druid, it should stay in Physical land primarily, and _include_ (not replace) logical info (ie. dimension API names)

QubitPi:
@cdeszaq I can work on this. Could you add little more description on how to implement this?

cdeszaq:
Since the slices endpoints are tied to the physical table, they should be showing the physical names for the dimensions, not the API names of the dimensions. Right now, the endpoints are showing the API names, which isn't as useful. So, find the place where the Slices output is using API names, and make them use physical names from the tables instead.

QubitPi:
@cdeszaq Are you talking about changing names for dimensions for names for table?

kevinhinterlong:
@QubitPi If you're not working on it I think I'd like to take this one

PR #491, opened 2017-08-07
Return druid name of dimension on slices servlet instead of api name

kevinhinterlong:
Addresses #394


kevinhinterlong:
@michael-mclawhorn @QubitPi same build issue from #535 on here

PR #492, opened 2017-08-08
Allow Webservice to Configure Metric Long Name

QubitPi:


fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/LogicalMetric.java
```diff
@@ -80,6 +80,28 @@ public LogicalMetric(TemplateDruidQuery templateDruidQuery, ResultSetMapper calc
...
     }
 
+    /**
+     * Constructor. Builds a Logical Metric whose instance variables are provided by a LogicalMetricInfo object.
+     *
+     * @param templateDruidQuery  Query the metric needs
+     * @param calculation  Mapper for the metric
+     * @param logicalMetricInfo  Logical Metric info provider
+     */
+    public LogicalMetric(
```
kevinhinterlong: It might be nice to implement this constructor and have everything else call this one.

This would require changing this constructor `LogicalMetricInfo(String name, String longName)` to `LogicalMetricInfo(String name, String description)` but that might be what we want since that's what `LogicalMetric` allows

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
```diff
@@ -132,11 +133,29 @@ public LogicalMetric make(String metricName, String dependentMetric) {
...
 
     /**
+     * Delegated to for actually making the metric.
+     *
+     * @param logicalMetricInfo  Logical metric info provider
+     * @param dependentMetrics  Metrics this metric depends on
+     *
+     * @return the new logicalMetric
+     */
+    protected LogicalMetric makeInner(LogicalMetricInfo logicalMetricInfo, List<String> dependentMetrics) {
```
kevinhinterlong: It looks like this was only implemented for `AggregationAverageMaker`, shouldn't it be done for all subclasses of `MetricMaker`?

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/LogicalMetric.java
kevinhinterlong: Also, would we want to mark the other constructors with `@Deprecated` telling people to use a `LogicalMetricInfo`?

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
```diff
@@ -85,6 +86,21 @@ protected LogicalMetric makeInner(String metricName, List<String> dependentMetri
         return new LogicalMetric(outerQuery, NO_OP_MAPPER, metricName);
     }
 
+    @Override
+    protected LogicalMetric makeInner(LogicalMetricInfo logicalMetricInfo, List<String> dependentMetrics) {
+        // Get the Metric that is being averaged over
```
michael-mclawhorn: Would it make sense to proxy the other method into this one by building a default LogicalMetricInfo that is more or less backwards compatible?

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/LogicalMetric.java
michael-mclawhorn: Let's go ahead and do this to push people toward the LogicalMetricInfo constructor

michael-mclawhorn:
Once you've addressed @kevinhinterlong  comments I'll dismiss his review unless he jumps on here before then,

michael-mclawhorn:
I don't understand why we didn't go through and add makeInner(LogicalMetricInfo) to all the subclasses of MetricMaker.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
QubitPi: @michael-mclawhorn Yes! This would be a brilliant idea! 👍

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
QubitPi: @kevinhinterlong Yes. `AggregationAverageMaker` is for experimenting. All subclasses will be modified

QubitPi:
@michael-mclawhorn You are right. I'll add makeInner(LogicalMetricInfo) to all subclasses. The reason for not adding previously is to have an example implementation in `AggregationAverageMaker.java` for you to evaluate whether the implementation is good or not

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/MetricMaker.java
```diff
@@ -132,9 +133,25 @@ public LogicalMetric make(String metricName, String dependentMetric) {
...
+    /**
+     * Delegated to for actually making the metric.
+     *
+     * @param logicalMetricInfo  Logical metric info provider
+     * @param dependentMetrics  Metrics this metric depends on
+     *
+     * @return the new logicalMetric
      */
-    protected abstract LogicalMetric makeInner(String metricName, List<String> dependentMetrics);
+    protected abstract LogicalMetric makeInner(LogicalMetricInfo logicalMetricInfo, List<String> dependentMetrics);
```
garyluoex: We should give it a default implementation, if not other peoples metric maker will break because they don't have this `makeInner` implemented.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/MetricInstance.java
```diff
@@ -23,6 +24,7 @@
  */
 public class MetricInstance {
 
+    private final LogicalMetricInfo logicalMetricInfo;
     private final String metricName;
```
garyluoex: I don't think we need this anymore.

garyluoex: Need CHANGELOG

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/MetricInstance.java
```diff
@@ -47,9 +66,29 @@ public MetricInstance(String metricName, MetricMaker maker, String... dependency
...
+     * @deprecated logical metric needs more config-richness to not just configure metric name, but also metric long
+     * name, description, etc. Use {@link #MetricInstance(LogicalMetricInfo, MetricMaker, FieldName...)} instead.
      */
+    @Deprecated
     public MetricInstance(FieldName metricName, MetricMaker maker, FieldName... dependencyFields) {
-        this.metricName = metricName.asName();
+        this.logicalMetricInfo = new LogicalMetricInfo(metricName.asName());
+        this.maker = maker;
+        this.dependencyMetricNames = new ArrayList<>();
+        for (FieldName fieldName : dependencyFields) {
```
kevinhinterlong: Maybe make this streamy using `Arrays.stream(...).map(...).collect(...)`

kevinhinterlong:
I only skimmed through it so you may want to get a third review/dismiss mine, but from what I saw 👍

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ Current
 
 ### Added:
 
+- [Allow Webservice to Configure Metric Long Name](https://github.com/yahoo/fili/pull/492)
+    * Logical metric needs more config-richness to not just configure metric name, but also metric long name,
+      description, etc. MetricInstance is now created by accepting a LogicalMetricInfo which contains all those
+      informations, instead of accepting a metric name.
```
michael-mclawhorn: which contains all these fields in addition to metric name.

Issue #111, opened 2016-11-29
Make `RequestLog` annotation-based


archolewa:
While the [RequestLog](https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/logging/RequestLog.java) object provides some valuable insight into requests (in particular the timings that it provides), it does so in a rather invasive manner.

For example, if you wish to time a segment of code, you really need to wrap the code in a try-finally block in order to ensure that the timer gets turned off properly even in the case of an exception (while this is already done in places where exceptions are expected, it is not done everywhere, even though exceptions could spring up anywhere).

To make this less invasive, we should make the `RequestLog` annotation-based (for example, a `@Timing` annotation we can apply to methods that automatically wraps all method invocations in a try-finally block that starts and stops a timer). This will likely require us to introduce [aspect-oriented programming](https://en.wikipedia.org/wiki/Aspect-oriented_programming) via the [AspjectJ](https://eclipse.org/aspectj/) framework or the [Spring AOP](http://docs.spring.io/spring/docs/current/spring-framework-reference/html/aop.html) framework.

PR #496, opened 2017-08-09
Use AOP with AspectJ to allow for Annotation based RequestLog timings

kevinhinterlong:
Addresses #111 

AspectJ is definitely new to me so I tried to document things as best I could.

- Uses compile time weaving
- **Extremely small** difference between annotation based and handcoded timings
- Can be used in dependent modules (this should make generifying the backend in the RequestWorkflows easier)

Here's the necessary plugin+configuration to use aspects from `fili-core` in other modules
```xml
<plugin>
    <groupId>org.codehaus.mojo</groupId>
    <artifactId>aspectj-maven-plugin</artifactId>
    <configuration>
        <aspectLibraries>
            <aspectLibrary>
                <groupId>com.yahoo.fili</groupId>
                <artifactId>fili-core</artifactId>
            </aspectLibrary>
        </aspectLibraries>
    </configuration>
</plugin>
```


kevinhinterlong:
Ugh so after trying this rebased onto #471 it appears to not weave in the calls to stop the timer now that they are using lambdas/method references. 

So I'd have to revert this to just wrapping the call to the method in a `RequestLog` timing block

```java
try (TimedPhase timer = RequestLog.startTiming(timerName)) {
    // call the method
}
```

We lose out a little bit on timing but it's still a lot cleaner. @michael-mclawhorn?

michael-mclawhorn:
We can reopen this if it gets picked up again.

Issue #352, opened 2017-06-09
Support first and last keywords for pagination page IDs


cdeszaq:
It would be great if one could ask for the First or Last page as keywords, rather than just as page numbers.

PR #502, opened 2017-08-13
Add first and last pagination parameters

mpardesh:
I have added the "first" and "last" pagination parameters. #352

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
@@ -563,7 +563,6 @@ public class LuceneSearchProvider implements SearchProvider {
             throws PageNotFoundException {
         int perPage = paginationParameters.getPerPage();
         validatePerPage(perPage);
-        int requestedPageNumber = paginationParameters.getPage();
 
         TreeSet<DimensionRow> filteredDimRows;
         int documentCount;
@@ -578,11 +577,11 @@ public class LuceneSearchProvider implements SearchProvider {
                         luceneIndexSearcher,
                         null,
                         query,
-                        perPage,
-                        requestedPageNumber
+                        perPage
                 );
                 hits = hitDocs.scoreDocs;
                 documentCount = hitDocs.totalHits;
+                int requestedPageNumber = paginationParameters.getPage(documentCount);
                 if (hits.length == 0) {
                     if (requestedPageNumber == 1) {
                         return new SinglePagePagination<>(Collections.emptyList(), paginationParameters, 0);
@@ -592,7 +591,7 @@ public class LuceneSearchProvider implements SearchProvider {
                 }
                 for (int currentPage = 1; currentPage < requestedPageNumber; currentPage++) {
                     ScoreDoc lastEntry = hits[hits.length - 1];
-                    hits = getPageOfData(luceneIndexSearcher, lastEntry, query, perPage, requestedPageNumber).scoreDocs;
+                    hits = getPageOfData(luceneIndexSearcher, lastEntry, query, perPage).scoreDocs;
                     if (hits.length == 0) {
                         throw new PageNotFoundException(requestedPageNumber, perPage, 0);
                     }
@@ -652,7 +651,6 @@ public class LuceneSearchProvider implements SearchProvider {
      * search after this entry (if lastEntry is null, the indexSearcher will begin its search from the beginning)
      * @param query  The Lucene query used to locate the desired dimension metadata
      * @param perPage  The number of entries per page
-     * @param currentPage  The desired page number
      *
      * @return The desired page of dimension metadata
      */
@@ -660,15 +658,14 @@ public class LuceneSearchProvider implements SearchProvider {
             IndexSearcher indexSearcher,
             ScoreDoc lastEntry,
             Query query,
-            int perPage,
-            int currentPage
+            int perPage
     ) {
         TimeLimitingCollectorManager manager = new TimeLimitingCollectorManager(searchTimeout, lastEntry, perPage);
         lock.readLock().lock();
         try {
             return indexSearcher.search(query, manager);
         } catch (IOException e) {
-            String errorMessage = "Unable to find dimension rows for page " + currentPage;
+            String errorMessage = "Unable to find dimension rows for specified page.";
             LOG.error(errorMessage);
             throw new RuntimeException(errorMessage);
         } catch (TimeLimitingCollector.TimeExceededException e) {
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/util/PaginationParameters.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/util/PaginationParameters.java
@@ -104,12 +104,12 @@ public class PaginationParameters {
     }
 
     /**
-     * Verifies that the passed parameter is greater than 0 or is -1 (last page).
+     * Verifies that the passed parameter is greater than 0.
      *
      * @param parameter  The parameter to be validated.
      * @param parameterName  The name of the parameter to appear in the error message
      *
-     * @throws BadPaginationException if 'parameter' is not greater than 0 or is not -1.
+     * @throws BadPaginationException if 'parameter' is not greater than 0.
      */
     private static void validate(int parameter, String parameterName) throws BadPaginationException {
         if (parameter < MINIMAL_VALUE) {
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/util/PaginationParameters.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/util/PaginationParameters.java
@@ -56,10 +56,7 @@ public class PaginationParameters {
      * @throws BadPaginationException If at least one of 'perPage' or 'page' is not a positive integer.
      */
     public PaginationParameters(String perPage, String page) throws BadPaginationException {
-        this.perPage = parseParameter(perPage, "perPage");
-        this.page = parseParameter(page, "page");
-        validate(this.perPage, "perPage");
-        validatePage(this.page, "page");
+        this(parseParameter(perPage, "perPage"), parseParameter(page, "page"));
     }
 
     /**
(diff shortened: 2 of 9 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/web/util/PaginationParameters.java
```diff
@@ -90,6 +93,11 @@ private int parseParameter (String parameter, String parameterName) throws BadPa
             throw new BadPaginationException(errorMessage.format(parameterName));
         }
         try {
+            if (parameter.equals(FIRST)) {
+                return 1;
+            } else if (parameter.equals(LAST)) {
+                return LAST_PAGE;
```
kevinhinterlong: I think it might be better to make pagination parameters use an `Either<Integer, PageResultsFlag>`. I know this is only really needed for the `LAST` page case, but using `-1` as a special flag could lead to some confusion later.
```java
enum PageResultsFlag {
    FIRST("first"),
    LAST("last");
    // You could also have a function to handle #getPage(resultSize) in here for both cases
}
```

fili-core/src/main/java/com/yahoo/bard/webservice/web/util/PaginationParameters.java
```diff
@@ -99,15 +107,15 @@ private int parseParameter (String parameter, String parameterName) throws BadPa
...
      *
      * @param parameter  The parameter to be validated.
      * @param parameterName  The name of the parameter to appear in the error message
      *
-     * @throws BadPaginationException if 'parameter' is not greater than 0.
+     * @throws BadPaginationException if 'parameter' is not greater than 0 or is not -1.
      */
     private void validate(int parameter, String parameterName) throws BadPaginationException {
-        if (parameter < MINIMAL_VALUE) {
+        if (parameter < MINIMAL_VALUE && parameter != LAST_PAGE) {
```
kevinhinterlong: For example one place this gets confusing is the validation. As it is now this lets us pass in `perPage=-1` as a valid query parameter

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/util/PaginationParametersSpec.groovy
```diff
@@ -19,12 +19,15 @@ class PaginationParametersSpec extends Specification {
 
         then: "The handler's perPage and page are as expected"
         handler.getPerPage() == expectedPerPage
-        handler.getPage() == expectedPage
+        handler.getPage(10) == expectedPage
```
kevinhinterlong: Make this a parameter in the `where:` section to add more tests

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/util/PaginationParametersSpec.groovy
```diff
@@ -130,7 +133,16 @@ class PaginationParametersSpec extends Specification {
         thrown(BadPaginationException)
 
         where:
-        invalidPerPage << ["0", "-1", "1.0", "2.3", "$Long.MAX_VALUE", "1a", "AndrewIsMostlyHarmless."]
+        invalidPerPage << ["0", "-2", "1.0", "2.3", "$Long.MAX_VALUE", "1a", "AndrewIsMostlyHarmless."]
+    }
+
+    def "If the result set is empty and the page is last then getPage returns the first page"() {
```
kevinhinterlong: You could combine this with `build the correct handler when 'perPage' is #perPage and 'page' is #page` if you make the `resultSize` a parameter

fili-core/src/main/java/com/yahoo/bard/webservice/web/util/PaginationParameters.java
archolewa: I don't know. I think using `-1` opens up the possibility for doing python-style paging, where -1 is the last page, -2 is the second to last page, etc.

fili-core/src/main/java/com/yahoo/bard/webservice/web/util/PaginationParameters.java
archolewa: Which makes sense if you think about negatives as counting backwards.

fili-core/src/main/java/com/yahoo/bard/webservice/web/util/PaginationParameters.java
kevinhinterlong: Ok if we introduce python-style paging I would definitely be ok with this approach. I don't want to start using negative values as special flags though

fili-core/src/main/java/com/yahoo/bard/webservice/web/util/PaginationParameters.java
kevinhinterlong: For sure but I would imagine this should still only apply to what `page` your asking for not the elements `perPage`

mpardesh: Yeah, I wasn't thinking about `perPage` when I changed this method. I will fix it 😄

fili-core/src/main/java/com/yahoo/bard/webservice/web/util/PaginationParameters.java
mpardesh: I wasn't sure what to do about this, so I just left it for now. I would be happy to make any changes.

archolewa: @kevinhinterlong I don't think we should officially support python-style pages just yet. I think that's a big enough change that we would want to discuss it with others first. Plus, as I think about it, I'm honestly not sure exactly how useful it would be. Really, I think we should hold off on that until some asks for the ability to easily get the "second-to-last" page.

I don't want to add an enum for the first and last page, because then we either need to have some sort of awkward either object that stores either a number or an enum, or we need to have two fields, one of which is guaranteed to be `null` at any one time.

In my opinion, the first one introduces way too much awkward infrastructure for something that is _very_ internal to the `PaginationParameters` object. Meanwhile, `-1` is a commonly used sentinel value to represent the "last" number whatever that may mean(for example, it's often used to represent infinity in configuration parameters). 

If we were passing the raw page number to other parts of the system, then the extra infrastructure would be worth it.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -578,11 +577,11 @@ private Query getFilterQuery(Set<ApiFilter> filters) {
                         luceneIndexSearcher,
                         null,
                         query,
-                        perPage,
-                        requestedPageNumber
+                        perPage
                 );
                 hits = hitDocs.scoreDocs;
                 documentCount = hitDocs.totalHits;
+                int requestedPageNumber = paginationParameters.getPage(documentCount);
```
archolewa: As I think about it, I'm not sure this will work.

We'll need to dig into the Lucene library some to be sure, but I'm afraid that the `hitDocs` contains only the first page of data, so `hitDocs.totalHits <= perPage`, not the total number of possible documents.

fili-core/src/main/java/com/yahoo/bard/webservice/web/util/PaginationParameters.java
```diff
@@ -114,11 +122,37 @@ private void validate(int parameter, String parameterName) throws BadPaginationE
         }
     }
 
+    /**
+     * Verifies that the page is positive or is -1 (last page).
```
archolewa: Rather than saying "or is -1" we should say "or is the last page."

As far as this method is concerned, it shouldn't really care what value we use to represent the last page.

fili-core/src/main/java/com/yahoo/bard/webservice/web/util/PaginationParameters.java
kevinhinterlong: Ok I wasn't entirely sure how much the idea of pagination was exposed in the other parts of the system which is why my preference was towards the `Either`. 

I think `-1` will be okay since it's not exposed but it should definitely be internal only and not be allowed in the api.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
mpardesh: Yay! `totalHits` is really the total number of results, so this will work.

archolewa:
@mpardesh Can you rebase this on top of master?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/DimensionsServletComponentSpec.groovy
```diff
@@ -906,7 +908,7 @@ class DimensionsServletComponentSpec extends Specification {
 
     def "test negative page number is handled correctly"() {
         setup:
-        Map<String, String> queryParams = [page: -1, perPage: 3]
+        Map<String, String> queryParams = [page: -2, perPage: 3]
```
kevinhinterlong: Couldn't this be reverted to `-1`

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/util/PaginationParametersSpec.groovy
```diff
@@ -38,11 +45,11 @@ class PaginationParametersSpec extends Specification {
 
         where:
         perPage | page  || expectedError
-        "-1"  | "-1"    || errorMessage("perPage", "-1")
```
kevinhinterlong: Same for these test cases

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/util/PaginationParametersSpec.groovy
```diff
@@ -115,7 +122,7 @@ class PaginationParametersSpec extends Specification {
         thrown(BadPaginationException)
 
         where:
-        invalidPage << ["0", "-1", "1.0", "2.3", "$Long.MAX_VALUE", "1a", "AndrewIsMostlyHarmless."]
+        invalidPage << ["0", "-2", "1.0", "2.3", "$Long.MAX_VALUE", "1a", "AndrewIsMostlyHarmless."]
```
kevinhinterlong: And here

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/util/PaginationParametersSpec.groovy
```diff
@@ -130,7 +137,7 @@ class PaginationParametersSpec extends Specification {
         thrown(BadPaginationException)
 
         where:
-        invalidPerPage << ["0", "-1", "1.0", "2.3", "$Long.MAX_VALUE", "1a", "AndrewIsMostlyHarmless."]
+        invalidPerPage << ["0", "-2", "1.0", "2.3", "$Long.MAX_VALUE", "1a", "AndrewIsMostlyHarmless."]
```
kevinhinterlong: And here

archolewa:
@mpardesh Could you please rebase this on top of master?

PR #505, opened 2017-08-14
Fix datasource name and physical table name mismatching

tarrantzhang:
This PR fixes the use of datasource name querying `physicalTableDictionary` which use physical table name as key in `VolatileDataRequestHandler`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,9 @@ Current
 
 ### Fixed:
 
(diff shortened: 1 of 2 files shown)


michael-mclawhorn:
This should get a CHANGELOG entry.  Otherwise it's fine.

archolewa:
I'm going to merge this in. @michael-mclawhorn didn't use the fancy-pants github approve button, but he still basically approved it, assuming a CHANGELOG (which has since been added) and he's on vacation right now.

Issue #497, opened 2017-08-10
Filter variables and methods in DataApiRequest are confusing


tarrantzhang:
ApiFilter is call `filters` while Druid filter is call `filter` in `DataApiRequest`. This is very confusing for people trying to understand the code. In my opinion, they should be changed to `apiFilter` and `druidFilter` respectively. And also their getters.

michael-mclawhorn:
There's a reason that API filters are considered plural while Druid filters are singular. Because API filters are treated as a list (or a list grouped into a map by dimension) while the Druid filters are a tree with a leaf node as the reference variable.

QubitPi:
Addressed by https://github.com/yahoo/fili/pull/507

PR #501, opened 2017-08-11
Make API Request processing pluggable

deepakb91:

PR #507, opened 2017-08-16
Rename filter variables and methods in DataApiRequest

mpardesh:
Addresses #497


mpardesh:
Looks like I accidentally mixed some commits here...I will fix this

archolewa:
Note that this is a breaking change (anybody who uses either of the getters in a request mapper will experience compile errors when they upgrade).

Instead, we should add new methods with the desired names, and deprecate the old methods.

Also we should add a CHANGELOG.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -1009,8 +1009,13 @@ public Granularity getGranularity() {
         return this.intervals;
     }
 
+    public Map<Dimension, Set<ApiFilter>> getApiFilters() {
+        return this.apiFilters;
+    }
+
+    @Deprecated
     public Map<Dimension, Set<ApiFilter>> getFilters() {
```
archolewa: These deprecated methods should delegate to the non-deprecated version (i.e. `getFilters()` should return `getApiFilters()`. Furthermore, we should add a Javadoc, with a `@deprecated` annotation indicating which method to use instead.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DataApiRequest.java
```diff
@@ -1020,9 +1025,28 @@ public Granularity getGranularity() {
...
+     * <p>
+     * The Druid filters are built (an expensive operation) every time this method is called. Use it judiciously.
+     *
+     * @return the Druid filter
+     *
+     * @deprecated this method has been deprecated because its name is unclear. Please use {@link #getDruidFilter()}.
+     */
+    @Deprecated
     public Filter getFilter() {
         try (TimedPhase timer = RequestLog.startTiming("BuildingDruidFilter")) {
```
archolewa: This should just delegate to `getDruidFilter`.

archolewa:
@mpardesh Just reviewed this change. Had a few small comments. Once those are addressed, I'll be willing to give my approval.

michael-mclawhorn:
There's a problem with the underlying issue.

There's a reason that API filters are considered plural while Druid filters are singular.  Because API filters are treated as a list (or a list grouped into a map by dimension) while the Druid filters are a tree with a leaf node as the reference variable.

archolewa:
@michael-mclawhorn As far as I can tell, the api filters method is still called `getApiFilters`, while the druid filter is called `getDruidFilter`. So things are still singular and plural.

michael-mclawhorn:
@archolewa Yeah, I see that.
I actually think I'd rather make a breaking change than each the deprecation cost here, but I suppose that's not the friendliest choice.  This is going to create rebase pain with https://github.com/yahoo/fili/pull/501, so I'd merge it quickly if you don't want to fix it.

PR #516, opened 2017-08-18
Part 5 of RFC for SQL Support

kevinhinterlong:


fili-sql/src/main/resources/moduleConfig.properties
```diff
@@ -8,8 +8,8 @@ moduleDependencies = fili-core
 
 # Disabling the druid dimension loader by default
 # When set to true you will also need to populate the list of dimensions using druid_dim_loader_dimensions
-bard__druid_dimensions_loader_enabled = false
-bard__druid_dim_loader_row_limit = 1000
+bard__druid_dimensions_loader_enabled = true
+bard__druid_coordinator_metadata_enabled = false
```
garyluoex: Reason for setting true and false? might want to modify comment as well.

kevinhinterlong: Right I'll update the comments. `bard__druid_dimensions_loader_enabled` tells the "druid" dimension loader which is now just a dimension loader to start the load task - so this works sql as well now.
`bard__druid_coordinator_metadata_enabled` tries to contact the druid coordinator directly which won't work if fili is only backed by sql data.

Issue #333, opened 2017-06-08
Add a Metric library metric that indicates Fili uptime


cdeszaq:
It would be good to know how long the system (the Fili app) has been up since it last restarted. One particular reason is many metrics are "lifelong" counters, and they are much more useful when you know how long the "life" is.

Also, aligning logs with the numbers reported by the metrics is much easier if you have this checkpoint.

kevinhinterlong:
I'm not sure if there's a specific place where we add in gauges/metrics but it seems pretty straightforward

```java
MetricRegistryFactory.getRegistry().register("uptime",
        (Gauge<Long>) () -> ManagementFactory.getRuntimeMXBean().getUptime()
);
```

PR #518, opened 2017-08-20
Add uptime metric

mpardesh:
Addresses #333

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@ Current
 -------
 
 ### Added:
+
 - [Add Uptime Status Metric](https://github.com/yahoo/fili/pull/518)
+    * Add a metric to show how long Fili has been running 
 
 - [Have Tables Endpoint Support (but not use) Additional Query Parameters](https://github.com/yahoo/fili/pull/437)
     * Make the availability consider the TablesApiRequest by passing it into the getLogicalTableFullView method
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Current
 -------
 
 ### Added:
+- [Add Uptime Status Metric](https://github.com/yahoo/fili/pull/518)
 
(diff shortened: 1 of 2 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -600,6 +604,13 @@ protected Ratio getRatio() {
                     }
             );
         }
+
+        if (!metrics.containsKey(UPTIME)) {
+            MetricRegistryFactory.getRegistry().register(
+                    UPTIME,
+                    (Gauge<Long>) () -> ManagementFactory.getRuntimeMXBean().getUptime()
```
tarrantzhang: `metricRegistry` is already defined as `ManagementFactory.getRuntimeMXBean()`, so you may use that.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Current
 -------
 
 ### Added:
+- [Add Uptime Status Metric](https://github.com/yahoo/fili/pull/518)
```
tarrantzhang: Just by the title it is not obvious to me what this change added. So maybe one or more bulletin points followed by explanation could do better.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -154,6 +156,8 @@
     private static final String METER_SPLITS_TOTAL_RATIO = "queries.meter.split_queries.total_ratio";
     private static final String METER_SPLITS_RATIO = "queries.meter.split_queries.ratio";
 
+    private static final String UPTIME = "uptime";
```
kevinhinterlong: Rename this to be `"jvm.uptime"` just to be clear which uptime is being reported.

Probably rename the variable name as well

PR #519, opened 2017-08-23
Decouple from model util

michael-mclawhorn:
A quick draft change to decouple extraction serialization from specific subclasses of KVS Dimension.


fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LookupDimension.java
```diff
@@ -51,6 +59,32 @@ public boolean equals(Object o) {
...
+     */
+    @JsonIgnore
+    public Optional<ExtractionFunction> getExtractionFunction() {
+
+        List<ExtractionFunction> extractionFunctions = getNamespaces().stream()
+                .map(
+                        namespace -> new LookupExtractionFunction(
+                                new NamespaceLookup(namespace),
+                                false,
+                                "Unknown " + namespace,
```
garyluoex: While we are at it, we can update this message to `__unsupportedValues` or something like it since this value will be shown to the user. `Unknown something__something` does not make much sense for the UI to show I guess.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/RegisteredLookupDimension.java
```diff
@@ -62,4 +69,31 @@ public String toString() {
...
+    @Override
+    @JsonIgnore
+    public Optional<ExtractionFunction> getExtractionFunction() {
+
+        List<ExtractionFunction> extractionFunctions = getLookups().stream()
+                .map(
+                        lookup -> new RegisteredLookupExtractionFunction(
+                                lookup,
+                                false,
+                                "Unknown " + lookup,
```
garyluoex: same

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LookupDimension.java
michael-mclawhorn: I like this comment, but I'm going to want to loop @jkusa in and make this a separate issue.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/ExtractionFunctionDimension.java
```diff
@@ -0,0 +1,21 @@
...
+ * A dimension that supports extraction functions.
+ */
+public interface ExtractionFunctionDimension extends Dimension {
+
+    /**
+     * Build an extraction function model object.
+     *
+     * @return  Take the internal namespaces and construct a model object for the extraction functions.
+     */
+    Optional<ExtractionFunction> getExtractionFunction();
```
michael-mclawhorn: No.  We don't want to change or complicate the Dimension interface if we can avoid it.

Issue #494, opened 2017-08-08
Update Response Format contract to use headers


michael-mclawhorn:
Response Format should follow the following rules:
responseFormat URI parameter and Accept header are evaluated.

URI parameter has precedence over Accept header.
* Acceptable defaults will map make application/json synonymous with json default format.
* application/vnd.api+json will be synonymous with json-api format.
* text/csv will be synonymous with the csv format.

If no responseFormat is specified in the URI then the server may choose any of the Accept types presented and no promise is made as to which will be preferred if multiple legal mappings exist.

tarrantzhang:
I think I can take this one.

QubitPi:
Addressed by https://github.com/yahoo/fili/pull/520

PR #520, opened 2017-08-23
Allow header accept field as a source of format type

tarrantzhang:
Address #494


fili-core/src/main/java/com/yahoo/bard/webservice/web/ResponseFormatResolver.java
```diff
@@ -0,0 +1,21 @@
...
+package com.yahoo.bard.webservice.web;
+
+import javax.ws.rs.container.ContainerRequestContext;
+
+/**
+ * A functional interface which allows the logic of resolving response format customizable. It has one method that
+ * takes the format name string from URI and a ContainerRequestContext object from which the header Accept field is
+ * accessible.
+ */
+public interface ResponseFormatResolver {
```
michael-mclawhorn: you might want to have this explicitly implement BiFunction<String, ContainerRequestContext, String>

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -1098,6 +1102,32 @@ protected DruidWebService buildMetadataDruidWebService(ObjectMapper mapper) {
...
+    /**
+     * Create a ResponseFormatResolver for Servlet objects.
+     * <p>
+     * Currently default types are json, jsonapi and csv types.
+     *
+     * @return A ResponseFormatResolver
+     */
+    protected ResponseFormatResolver buildResponseFormatResolver() {
+        return (format, containerRequestContext) -> {
+            Map<String, String> formatsMap = new LinkedHashMap<>();
```
michael-mclawhorn: Build the map above the return so it's not being rebuilt every time, then close over the map.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -1098,6 +1102,32 @@ protected DruidWebService buildMetadataDruidWebService(ObjectMapper mapper) {
...
+            formatsMap.put("application/json", "json");
+            formatsMap.put("application/vnd.api+json", "jsonapi");
+            formatsMap.put("text/csv", "csv");
+
+            String headerFormat = containerRequestContext.getHeaderString("Accept");
+            if (format != null || headerFormat == null) {
+                return format;
+            }
+            return formatsMap.entrySet().stream()
```
michael-mclawhorn: Why stream if we're just doing a map lookup?

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -1098,6 +1102,32 @@ protected DruidWebService buildMetadataDruidWebService(ObjectMapper mapper) {
...
+    protected ResponseFormatResolver buildResponseFormatResolver() {
+        return (format, containerRequestContext) -> {
+            Map<String, String> formatsMap = new LinkedHashMap<>();
+            formatsMap.put("application/json", "json");
+            formatsMap.put("application/vnd.api+json", "jsonapi");
+            formatsMap.put("text/csv", "csv");
+
+            String headerFormat = containerRequestContext.getHeaderString("Accept");
+            if (format != null || headerFormat == null) {
+                return format;
```
michael-mclawhorn: If format != null, does it matter which value is passed in?

Also, if there is a null format there should still be a default value, in the case of the standard fili, it's json.

tarrantzhang: I think the URI format should precedence header format right? So if URI format is not null, we will just return the URI format. Is this what you are asking about? 

As for your second question, null case is already handled in [FiliResponseWriterSelector](https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/web/FiliResponseWriterSelector.java#L44)

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
tarrantzhang: It's not a map lookup. A valid header format could be any string that contains our map keys. For example:
```
Accept : text/html, text/csv
```

fili-core/src/main/java/com/yahoo/bard/webservice/web/ResponseFormatResolver.java
tarrantzhang: I'm new to SAM, but since it's an interface, implementing another interface is not allowed in Java, right? Or did I misunderstand what you mean by "implement BiFunction<..."

tarrantzhang:
@michael-mclawhorn travis-ci has been stuck for a long time. Do you know why?

michael-mclawhorn:
I don't.  Try poking it with a fresh push and see what happens.

fili-core/src/main/java/com/yahoo/bard/webservice/web/ResponseFormatResolver.java
```diff
@@ -0,0 +1,21 @@
...
+    /**
+     * Resolve desirable format from URI and ContainerRequestContext.
+     *
+     * @param format  The format String from URI
+     * @param containerRequestContext  ContainerRequestContext object that contains request related information
+     * @return A resolved format decided by the function
+     */
+    String accept(String format, ContainerRequestContext containerRequestContext);
```
archolewa: We should either return `Optional` here, or explicitly document that if this function returns `null,` Fili will use the default format.

That way, customers will know that this method can handle a `null` format without having to dig through Fili itself.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/MetricsServlet.java
```diff
@@ -179,7 +184,14 @@ public Response getMetric(
             RequestLog.startTiming(this);
             RequestLog.record(new MetricRequest(metricName));
 
-            MetricsApiRequest apiRequest = new MetricsApiRequest(metricName, null, "", "", metricDictionary, uriInfo);
+            MetricsApiRequest apiRequest = new MetricsApiRequest(
```
archolewa: Should unwrap this.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/ResponseFormatResolverSpec.groovy
```diff
@@ -0,0 +1,48 @@
...
+
+import spock.lang.Shared
+import spock.lang.Specification
+
+import javax.ws.rs.container.ContainerRequestContext
+
+class ResponseFormatResolverSpec extends Specification {
+
+    @Shared
+    ResponseFormatResolver resolver =  [ accept: {format, containerRequestContext ->
```
archolewa: This should not be reimplemented here. If we want to test our default implementation of the Resolver (and we do) then we should pull this in out into a proper test rather than inlining it in the binder factory.

As is, this test is useless. It doesn't actually test any code that we're using in production!

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ Current
 
 ### Added:
 
+- [Evaluate format type from both URI and Accept header](https://github.com/yahoo/fili/pull/495)
+    * Add a new functional interface `ResponseFormatResolver`.
+    * Implement a concrete implementation of `ResponseFormatResolver` in `AbstractBindingFactory`.
+
```
michael-mclawhorn: These notes describe where the changes are, but we should have something in here about the nature of the change (e.g. mapping Accept headers and coalescing with URI defined headers)

fili-core/src/main/java/com/yahoo/bard/webservice/web/ResponseFormatResolver.java
michael-mclawhorn: Sorry, since it's an interface, extend, not implement.

fili-core/src/main/java/com/yahoo/bard/webservice/web/TablesApiRequest.java
```diff
@@ -115,6 +115,7 @@ public TablesApiRequest(
         );
     }
 
+
```
michael-mclawhorn: I would have expected this change to create a checkstyle error.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/ResponseFormatResolverSpec.groovy
```diff
@@ -0,0 +1,39 @@
...
+package com.yahoo.bard.webservice.web
+
+import spock.lang.Shared
+import spock.lang.Specification
+import spock.lang.Unroll
+
+import javax.inject.Inject
+import javax.ws.rs.container.ContainerRequestContext
+
+class ResponseFormatResolverSpec extends Specification {
```
michael-mclawhorn: Technically, what you're testing here is DefaultResponseFormatResolver.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/apirequest/ApiRequestSpec.groovy
```diff
@@ -97,6 +97,7 @@ class ApiRequestSpec extends Specification {
         ResponseFormatType.CSV  | new ConcreteApiRequest().generateAcceptFormat("csv")
     }
 
+
```
michael-mclawhorn: Double whitespace is probably the wrong choice here.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/JobsServletReactiveChainforResultsEndpointSpec.groovy
```diff
@@ -49,6 +50,8 @@ class JobsServletReactiveChainforResultsEndpointSpec extends Specification {
     JobsServlet mockJobServlet
     HttpResponseMaker httpResponseMaker
     ResponseWriter responseWriter
+    ResponseFormatResolver formatResolver
+
```
michael-mclawhorn: Again, double whitespace.

fili-core/src/main/java/com/yahoo/bard/webservice/web/ResponseFormatResolver.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+public interface ResponseFormatResolver {
+    /**
+     * Resolve desirable format from URI and ContainerRequestContext. In case of a null return, Fili handles null
+     * format at {@link FiliResponseWriterSelector}, which use json as default value.
+     *
+     * @param format  The format String from URI
+     * @param containerRequestContext  ContainerRequestContext object that contains request related information
+     *
+     * @return A resolved format decided by the function
```
michael-mclawhorn: The details on line 14-15 assumes things about where this will be called.  A function doesn't get to be that opinionated about it's use.

Instead add , null if none match to the @return clause.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DefaultResponseFormatResolver.java
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Constructor.
+     */
+    public DefaultResponseFormatResolver() {
+        formatsMap = new LinkedHashMap<>();
+        formatsMap.put("application/json", "json");
+        formatsMap.put("application/vnd.api+json", "jsonapi");
+        formatsMap.put("text/csv", "csv");
+    }
```
michael-mclawhorn: it's probably a good idea to pull the format names for json, jsonapi and csv out into public constant strings. They could live in this class, but I could make the case for putting them in other classes as well.

fili-core/src/main/java/com/yahoo/bard/webservice/web/DefaultResponseFormatResolver.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Constructor.
+     */
+    public DefaultResponseFormatResolver() {
+        formatsMap = new LinkedHashMap<>();
+        formatsMap.put(ACCEPT_HEADER_JSON, "json");
+        formatsMap.put(ACCEPT_HEADER_JSONAPI, "jsonapi");
+        formatsMap.put(ACCEPT_HEADER_CSV, "csv");
+    }
```
michael-mclawhorn: "it's probably a good idea to pull the format names for json, jsonapi and csv out into public constant strings. They could live in this class, but I could make the case for putting them in other classes as well."

I was actually more concerned about the values themselves.  They should be refactored with constants here, in the response builder and in the tests.

PR #521, opened 2017-08-24
Fix Lucene Cardinality in New KeyValueStores

garyluoex:


fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -191,7 +191,7 @@ public void setKeyValueStore(KeyValueStore keyValueStore) {
         this.keyValueStore = keyValueStore;
         // Check initialization for the cardinality in a keyValueStore
         if (keyValueStore.get(DimensionStoreKeyUtils.getCardinalityKey()) == null) {
-            keyValueStore.put(DimensionStoreKeyUtils.getCardinalityKey(), "0");
+            refreshCardinality();
         }
```
michael-mclawhorn: Can we get a test?

PR #522, opened 2017-08-24
Enable search provider to hot-swap index and key value store to hot-swap store location

QubitPi:


fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -318,6 +318,24 @@ private void updateDimensionRow(
     }
 
     /**
+     * Replace Lucene Index with a new index.
+     *
+     * @param newLuceneIndexPath  The path of the new index
+     */
+    public void replaceIndex(String newLuceneIndexPath) {
+        luceneIndexPath = newLuceneIndexPath;
+
```
michael-mclawhorn: Sadly, we can't do it this way because we have no way of tracking, between restarts, that the index path has changed.  You'll need to try to close the MMapDirectory, file system delete it and move the newLucenePath to the location of the original path.

QubitPi: Make sense. I will

1. lock and close the old dir
2. move the old dir to a different location
3. move the new dir to the old location
4. unlock the old dir and file system delete it

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -318,6 +318,67 @@ private void updateDimensionRow(
...
+            try {
+                oldLuceneDirectory.close();
+            } catch (IOException e) {
+                String message = ErrorMessageFormat.CANNOT_CLOSE_INDEX_DIR.format(oldLuceneDirectory);
+                LOG.error(message, e);
+            }
+
+            // move the old dir to a different location
+            try {
+                oldLuceneDirectory.rename(luceneIndexPath, luceneIndexPath + ".temp");
```
garyluoex: The behavior of rename is unclear, we need to test it before doing anything else with it.
1. Does this keep the current lucene instance available and correct?
2. Does rename changes the directory of which this is pointing to?

@michael-mclawhorn  do you know more about this? looking at the source code... this method is weird I think.

QubitPi: Yeah, I'm testing it. Looks like there is a bug from my end while calling `rename`. Investigating...

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -318,6 +323,71 @@ private void updateDimensionRow(
...
+        } finally {
+            lock.writeLock().unlock();
+        }
+
+        // delete old dir
+        try (IndexWriter indexWriter = new IndexWriter(
+                oldLuceneDirectory,
+                new IndexWriterConfig(LUCENE_ANALYZER).setRAMBufferSizeMB(BUFFER_SIZE)
+        )) {
+            indexWriter.close();
```
QubitPi: The reason I use `IndexWriter.close()` comes from 

https://stackoverflow.com/questions/12499108/how-to-delete-lucene-directory-along-with-different-segment-files

https://stackoverflow.com/questions/21901499/why-i-cannot-delete-index-directories-in-lucene-after-i-make-the-searcher-null

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -318,6 +323,71 @@ private void updateDimensionRow(
...
+            // open a new lucene index dir
+            try {
+                luceneDirectory = new MMapDirectory(Paths.get(newLuceneIndexPath));
+            } catch (IOException e) {
+                String message = ErrorMessageFormat.CANNOT_CREATE_INDEX_DIR.format(newLuceneIndexPath);
+                LOG.error(message, e);
+            }
+
+            luceneIndexPath = newLuceneIndexPath; // have lucene point to the new lucene index dir
+
```
michael-mclawhorn: This still doesn't work.  We need the new files moved to the same location as the old ones to make this resilient across restarts.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -318,6 +327,133 @@ private void updateDimensionRow(
...
+    private void moveDir(String source, String destination) {
+        try {
+            Files.move(
+                    new File(source).getAbsoluteFile().toPath(),
+                    new File(destination).getAbsoluteFile().toPath(),
+                    REPLACE_EXISTING
+            );
+        } catch (IOException e) {
+            LOG.error(ErrorMessageFormat.CANNOT_MOVE_INDEX_DIR.format(source));
+            throw new RuntimeException(e);
```
michael-mclawhorn: Why the exception we throw should be a RuntimeException, perhaps something more specific like IllegalStateException would be more descriptive.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -318,6 +327,133 @@ private void updateDimensionRow(
...
+                        throws IOException {
+                    LOG.debug("Deleting {}.", file);
+                    Files.delete(file);
+                    return FileVisitResult.CONTINUE;
+                }
+
+                @Override
+                public FileVisitResult postVisitDirectory(Path dir, IOException e) throws IOException {
+                    if (e == null) {
+                        LOG.debug("Deleting {}", dir);
```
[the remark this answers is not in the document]
michael-mclawhorn: Debug is way too chatty on a visitor.  Trace maybe.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -318,6 +327,133 @@ private void updateDimensionRow(
...
+                    LOG.debug("Moving {} to {}", file, destinationFileName);
+                    Files.move(file, destinationFileName);
+                    return FileVisitResult.CONTINUE;
+                }
+
+                @Override
+                public FileVisitResult postVisitDirectory(Path dir, IOException e) throws IOException {
+                    if (e == null) {
+                        Path destinationDirName = destinationPath.resolve(dir.getFileName());
+                        LOG.debug("Moving {} to {}", dir, destinationDirName);
```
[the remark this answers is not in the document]
michael-mclawhorn: Debug is too chatty here.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -318,6 +327,133 @@ private void updateDimensionRow(
...
+    /**
+     * Moves all files and sub-directories from one location to another.
+     * <p>
+     * Two locations must exist before calling this method.
+     *
+     * @param sourceDir  The location where files and sub-directories will be moved from
+     * @param destinationDir  The location where files and sub-directories will be moved to
+     */
+    private static void moveDirEntries(String sourceDir, String destinationDir) {
+        Path destinationPath = new File(destinationDir).getAbsoluteFile().toPath();
```
michael-mclawhorn: Why not use moveDir for both cases?

QubitPi: I wish I could. `moveDir` works when we are moving directory to a location *under the same parent*, which might not always be the case if new indexes come from a different parent

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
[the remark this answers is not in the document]
QubitPi: Changed to `trace`

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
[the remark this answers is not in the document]
QubitPi: Changed to `trace`

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
michael-mclawhorn: Huh.

QubitPi: ¯\_(ツ)_/¯

michael-mclawhorn:
Is there any reason not to use an existing library for the file io?

https://commons.apache.org/proper/commons-io/javadocs/api-2.4/org/apache/commons/io/FileUtils.html

QubitPi:
@michael-mclawhorn This is a good library for file IO. I was trying to move directories to a different parent using that actually. But the same problem happened - couldn't move to a different parent.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -318,6 +327,133 @@ private void updateDimensionRow(
...
+            try {
+                Files.createDirectory(new File(luceneIndexPath).getAbsoluteFile().toPath());
+            } catch (IOException e) {
+                LOG.error(ErrorMessageFormat.CANNOT_CREATE_INDEX_DIR.format(luceneIndexPath));
+                throw new RuntimeException(e);
+            }
+            moveDirEntries(newLuceneIndexPathString, luceneIndexPath);
+
+            // file system delete old Lucene index directory
+            deleteDir(destination);
```
michael-mclawhorn: Perhaps change the variable name destination to indicate that it's the tempDirectory to be deleted?

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ Current
 
 ### Added:
 
+- [Enable Lucene search provider to swap index](https://github.com/yahoo/fili/pull/522)
+    * This allows user to put new Lucene index at a specific location and has Fili points to the new index.
```
michael-mclawhorn: Behavior change has made this comment slightly out of date.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -318,6 +327,133 @@ private void updateDimensionRow(
...
+                public FileVisitResult visitFile(Path file, BasicFileAttributes basicFileAttributes)
+                        throws IOException {
+                    Path destinationFileName = destinationPath.resolve(file.getFileName());
+                    LOG.trace("Moving {} to {}", file, destinationFileName);
+                    Files.move(file, destinationFileName);
+                    return FileVisitResult.CONTINUE;
+                }
+
+                @Override
+                public FileVisitResult postVisitDirectory(Path dir, IOException e) throws IOException {
```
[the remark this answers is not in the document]
QubitPi:
```
parentDir
    file1
    file1
    subDir
```

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -318,6 +327,133 @@ private void updateDimensionRow(
...
+        lock.writeLock().lock();
+        try {
+            String source = luceneIndexPath;
+            String tempDir = luceneIndexPath.substring(
+                    0,
+                    luceneIndexPath.length() - (luceneIndexPath.endsWith("/") ? 1 : 0)
+            ) + "_old";
+
+            // move Lucene index directory to a different location under the same parent
+            moveDir(source, tempDir);
```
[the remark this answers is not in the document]
QubitPi: Good idea!

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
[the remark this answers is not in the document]
QubitPi: I referenced this - https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileVisitor.html

I think this method deletes all sub-directories including the directory itself

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -317,6 +323,114 @@ private void updateDimensionRow(
...
+            moveDirEntries(luceneIndexPath, tempDir);
+
+            LOG.debug("Moving all new Lucene indexes from {} to {} ...", newLuceneIndexPathString, luceneIndexPath);
+            moveDirEntries(newLuceneIndexPathString, luceneIndexPath);
+
+            LOG.debug(
+                    "Deleting {} since new Lucene indexes have been moved away from there and is now empty",
+                    newLuceneIndexPathString
+            );
+            deleteDir(newLuceneIndexPathString);
```
garyluoex: ?

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -317,6 +323,114 @@ private void updateDimensionRow(
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    /**
+     * Deletes a directory and all entries under that directory.
+     *
+     * @param path  The location of the directory that is to be deleted
+     */
+    private static void deleteDir(String path) {
```
garyluoex: There is a deleteFiles in Utils class......

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProviderManager.java
```diff
@@ -77,6 +77,22 @@ private static String getProviderPath(String providerName) {
...
+    }
+
+    /**
+     * Get the path for the lucene index files that are going to be deleted when we are cleaning up search provider.
+     *
+     * @param providersPath  Location of all providers
+     *
+     * @return the path to the lucene index files of all providers.
+     */
+    private static String getDeleteProviderPath(String providersPath) {
```
garyluoex: Why do we need this and not use the old one?

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
QubitPi: Once we move all new indexes, the directory used to contain them is empty. So I delete it

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProviderManager.java
QubitPi: Because groovy test creates new directory called `new_lucene_indexes`, this cannot be deleted using the old one during test cleanup

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
QubitPi: I forgot this works for deleting. Thanks!

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProviderSpec.groovy
```diff
@@ -98,6 +137,87 @@ class LuceneSearchProviderSpec extends SearchProviderSpec<LuceneSearchProvider>
         1 * keyValueStore.put('cardinality_key', '14')
     }
 
+    def "moveDirEntries moves all entries of a directory to a new directory, while keeping all old empty dirs"() {
+        expect:
```
tarrantzhang: Is it necessary to have both `expect` and `then` steps at the same time?

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -317,6 +325,96 @@ private void updateDimensionRow(
...
 
+    @Override
+    public void replaceIndex(String newLuceneIndexPathString) {
+        lock.writeLock().lock();
+        try {
+            Path oldLuceneIndexPath = Paths.get(luceneIndexPath);
+            String tempDir = oldLuceneIndexPath.resolveSibling(oldLuceneIndexPath.getFileName() + "_old").toString();
+
+            LOG.debug("Moving old Lucene index directory from {} to {} ...", luceneIndexPath, tempDir);
+            moveDirEntries(luceneIndexPath, tempDir);
```
tarrantzhang: Is there any logic to prevent an unexpected abort during the process moving old index to temp dir? It could result in an incomplete index. Would it be better to use a copy/delete fashion rather than move?

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProviderSpec.groovy
QubitPi: If you take a look at logic of this test, you will see that this test first makes sure `sourcePath` contains all files to be removed and `destinationPath` doesn't, which is what `expect` is all about, then the test makes sure, after `moveDirEntries`, `sourcePath` doesn't contain the files and `destinationPath` contains all the files, because all files have been moved from `sourcePath` to `destinationPath`, which is that `then` is all about.

QubitPi: If you take a look at logic of this test, you will see that this test first makes sure `sourcePath` contains all files to be removed and `destinationPath` doesn't, which is what `expect` is all about, then the test makes sure, after `moveDirEntries`, `sourcePath` doesn't contain the files and `destinationPath` contains all the files, because all files have been moved from `sourcePath` to `destinationPath`, which is that `then` is all about.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
QubitPi: `replaceIndex` itself is in copy/delete fashion.

PR #525, opened 2017-08-25
Remove the druid_coordinator_metadata feature flag.

archolewa:
-- This feature flag no longer makes any sense, because customers must
either:
    1. Write a custom `PhysicalTable` with an Availability service
    that does not rely on the DataSourceMetadataService, in which case
    the feature flag has zero impact.
    2. Turn on the feature flag, so that availability gets
    populated.

Otherwise, the availability information will never get populated, and
every data request will fail with an error about how the metadata
service doesn't know anything about the specified datasource, a message
that can be a little bit misleading.

In other words, vanilla Fili doesn't work if that feature flag isn't
turned on, so there's no point in including it.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
@@ -2,7 +2,6 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.application;
 
-import static com.yahoo.bard.webservice.config.BardFeatureFlag.DRUID_COORDINATOR_METADATA;
 import static com.yahoo.bard.webservice.config.BardFeatureFlag.DRUID_DIMENSIONS_LOADER;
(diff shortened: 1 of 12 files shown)


archolewa:
@garyluoex There isn't really any point in deprecating it when it straight up breaks things. Typically you deprecate something that still works, but may be removed at some point in the future. This just straight up doesn't work.

archolewa:
Also, I'm not sure what you mean by a "binding conflict" So far as I know, I'm not doing any binding here.

archolewa:
Oh you mean the part where we're pulling binding on the metadata webservice out of the if-block. Sure, bindings can have conflicts, but we won't be able to compile if there is, and the whole idea of a named binding is how you resolve most such conflicts (give two conflicting bindings different names).

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -137,6 +134,8 @@
 
 import javax.validation.constraints.NotNull;
 
+import rx.subjects.PublishSubject;
```
tarrantzhang: The order of this import probably should not change.

PR #489, opened 2017-08-06
Replace UI/Non-UI broker configuration with one broker

mpardesh:
Here is an attempt at replacing the UI/non-UI code to simplify the code. This feature isn't used anymore and complicates the config. I left the UI and non-UI settings in the configuration for backwards compatibility. First, we will check if `druid_broker` is set. If not, we will try to use `non_ui_druid_broker` and then `ui_druid_broker`.

PR #526, opened 2017-08-25
https://github.com/yahoo/fili/pull/489

michael-mclawhorn:
Rebasing and patching PR 489
https://github.com/yahoo/fili/pull/489


michael-mclawhorn:
Given that all I did was test fix and rebase I'm counting my :+1: from PR 489 as an approval.

PR #528, opened 2017-08-30
Fix wrong default druid url and broken getInnerMostQuery

### LINKED ISSUE #527: Bug - Not able to get the nested innermost query
Introduced with this change https://github.com/yahoo/fili/pull/485/files#diff-02ff0b920e378c6f5462c11968ce6333L65

garyluoex:
Resolves #527

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,9 @@ Current
 
 ### Changed:
 
+- [Fix wrong default druid url and broken getInnerMostQuery](https://github.com/yahoo/fili/pull/528)
+    * Comment out the wrong default druid broker url in module config, add check for validate url in `DruidClientConfigHelper `
+    * Fix broken `getInnermostQuery` method in `DruidQuery`
 
 - [Rename filter variables and methods in DataApiRequest](https://github.com/yahoo/fili/pull/507)
     * The method names `getFilter` and `getFilters` can be confusing, as well as the `filters` variable
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@ Current
 ### Changed:
 
 - [Fix wrong default druid url and broken getInnerMostQuery](https://github.com/yahoo/fili/pull/528)
-    * Comment out the wrong default druid broker url in module config, add check for validate url in `DruidClientConfigHelper `
+    * Comment out the wrong default druid broker url in module config that break old url config compatibility, add check for validate url in `DruidClientConfigHelper `
     * Fix broken `getInnermostQuery` method in `DruidQuery`
 
 - [Rename filter variables and methods in DataApiRequest](https://github.com/yahoo/fili/pull/507)
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/DruidQuery.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/DruidQuery.java
@@ -64,7 +64,7 @@ public interface DruidQuery<Q extends DruidQuery<? super Q>> {
      */
     @JsonIgnore
     default DruidQuery<?> getInnermostQuery() {
-        return getInnerQuery().map(DruidQuery.class::cast).orElse(this);
+        return getInnerQuery().isPresent() ? getInnerQuery().get().getInnermostQuery() : this;
     }
 
     /**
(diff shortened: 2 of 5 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/DruidQuery.java
```diff
@@ -64,7 +64,7 @@
      */
     @JsonIgnore
     default DruidQuery<?> getInnermostQuery() {
-        return getInnerQuery().map(DruidQuery.class::cast).orElse(this);
+        return getInnerQuery().isPresent() ? getInnerQuery().get().getInnermostQuery() : this;
```
archolewa: Why not just do

```java
   return getInnerQuery.map(DruidQuery::getInnermostQuery).orElse(this);
```

fili-core/src/main/resources/moduleConfig.properties
```diff
@@ -86,8 +86,8 @@ bard__max_results_without_filters = 10000
 bard__default_per_page = 10000
 
 # URLs for druid nodes
-bard__druid_broker = [SET ME IN APPLICATION CONFIG]
-bard__druid_coord = [SET ME IN APPLICATION CONFIG]
+# bard__druid_broker = [SET ME IN APPLICATION CONFIG]
+# bard__druid_coord = [SET ME IN APPLICATION CONFIG]
```
archolewa: I don't think we should be commenting these out. The entire purpose of having them be uncommented is that if the customer forgets to set this, the error message will tell them that the setting for this is invalid, and when the error message prints the provided setting, they'll get a hint as to where to put the proper setting.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/DruidQuery.java
garyluoex: Null pointer exception if getInnerQuery returns null right?

fili-core/src/main/resources/moduleConfig.properties
garyluoex: Then why do we even bother to be backward compatible then? This default will always be chosen over the deprecated ui and non ui settings.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/DruidQuery.java
garyluoex: Ok you are right

fili-core/src/main/resources/moduleConfig.properties
michael-mclawhorn: @archolewa  To clarify, these being set was breaking this because we were using their UNSETness to indicate backwards compatibility.  Instead they were showing as set, and blowing up unmigrated isntances.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/DruidQuery.java
garyluoex: I'm getting a type conflict using this map method, might as well keep it what it is right now since both are basically equivalent.

CHANGELOG.md
```diff
@@ -45,6 +45,9 @@ Current
 
 ### Changed:
 
+- [Fix wrong default druid url and broken getInnerMostQuery](https://github.com/yahoo/fili/pull/528)
+    * Comment out the wrong default druid broker url in module config, add check for validate url in `DruidClientConfigHelper `
```
michael-mclawhorn: Add a few more words to explain that having the default specified was causing the app to fail to do backwards compatibility.

fili-core/src/main/resources/moduleConfig.properties
archolewa: @michael-mclawhorn Ah ok that makes sense, I forgot about that.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/DruidQuery.java
archolewa: Apparently, IntelliJ is stumbling over the template types that neither Maven nor Eclim care about.

Stupid IntelliJ/complicated use of types that confuse our tools or something.

PR #532, opened 2017-09-01
Prevent NPE in test due to null instance variables in DataApiRequest

QubitPi:


fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequestImpl.java
```diff
@@ -371,14 +372,14 @@ protected Boolean isDateTimeFirstSortField(LinkedHashMap<String, SortDirection>
     public DataApiRequestImpl() {
         super();
         this.table = null;
-        this.granularity = null;
-        this.dimensions = null;
+        this.granularity = DefaultTimeGrain.DAY;
```
michael-mclawhorn: I'm not sure giving granularity a value that might appear correct is reasonable here.

It's one thing to make empty collections instead of nulls, it's a different thing to make a deliberate value.
Leave DefaultTimeGrain.DAY null, and if that breaks some things, let's see what the cost of fixing them

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/RequestQueryGranularityResolver.java
```diff
@@ -43,7 +43,7 @@ public Granularity resolveAcceptingGrain(DataApiRequest apiRequest, TemplateDrui
             return requestGranularity;
         }
 
-        if (requestGranularity.satisfiedBy(queryGrain)) {
+        if (queryGrain.satisfiedBy(requestGranularity)) {
```
michael-mclawhorn: This changes the meaning of this code!  Why are you doing this here?

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequestImpl.java
```diff
@@ -372,13 +373,13 @@ public DataApiRequestImpl() {
         super();
         this.table = null;
         this.granularity = null;
-        this.dimensions = null;
+        this.dimensions = Collections.emptySet();
```
michael-mclawhorn: For now, you can go back to setting granularity to a default, but Deprecate this method and make a note in the ChangeLog that we need to remove this method entirely, please.

fili-core/src/main/java/com/yahoo/bard/webservice/table/resolver/RequestQueryGranularityResolver.java
QubitPi: I changed it back

PR #540, opened 2017-09-22
fix constantMaker make method

asifmansoora:
The ConstantMaker make method needs to be rewritten with the `LogicalMetricInfo` class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,6 +130,9 @@ Current
 
 ### Fixed:
 
+- [Fix ConstantMaker make method with LogicalMetricInfo class](https://github.com/yahoo/fili/pull/540)
+    * The ConstantMaker make method needs to be rewritten with the LogicalMetricInfo class.
+    
 - [Slices endpoint returns druid name instead of api name](https://github.com/yahoo/fili/pull/491)
     * The slices endpoint now gives the druid name instead of the api name for dimensions.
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ Current
 
 ### Added:
 
+- [Fix ConstantMaker make method with LogicalMetricInfo class](https://github.com/yahoo/fili/pull/540)
```
garyluoex: Move to fixes instead of added.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/ConstantMaker.java
```diff
@@ -34,12 +34,12 @@ public ConstantMaker(MetricDictionary metricDictionary) {
...
     @Override
-    public LogicalMetric make(String metricName, List<String> dependentMetrics) {
+    public LogicalMetric make(LogicalMetricInfo logicalMetricInfo, List<String> dependentMetrics) {
         // Check that we have the right number of metrics
-        assertRequiredDependentMetricCount(metricName, dependentMetrics);
+        assertRequiredDependentMetricCount(logicalMetricInfo.getName(), dependentMetrics);
 
         // Actually build the metric.
-        return makeInner(metricName, dependentMetrics);
+        return makeInner(logicalMetricInfo.getName(), dependentMetrics);
```
garyluoex: Use the makeInner with logicalMetricInfo as argument instead of the string metric name version.

PR #541, opened 2017-09-22
revert the druid name and add it to fact name in slices endpoint

asifmansoora:
Reverting this PR https://github.com/yahoo/fili/pull/491 so that the `name` still points to `apiName` and added `factName` which points to `druidName`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -171,6 +171,11 @@ Current
 
 ### Removed:
 
+- [Reverted the druid name change in slices endpoint instead added to factName](https://github.com/yahoo/fili/pull/541)
+    * Reverting the PR-419(https://github.com/yahoo/fili/pull/419) so that the name still points to apiName and added factName which points to druidName.
+      `name` was not valid for cases when it is a Lookup dimension because it was pointing to the base dimension name , so reverted that change and added
+      `druidName` which is the actual druid fact name and `name` being the apiName
+    
 - [Remove custom immutable collections in favor of Guava](https://github.com/yahoo/fili/pull/479)
     * `Utils.makeImmutable(...)` was misleading and uneeded so it has been removed. Use Guava's immutable collections.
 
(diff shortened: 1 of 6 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/web/SlicesApiRequest.java
```diff
@@ -179,7 +179,8 @@ protected SlicesApiRequest() {
                     Column key = e.getKey();
                     if (key instanceof DimensionColumn) {
                         Dimension dimension = ((DimensionColumn) key).getDimension();
-                        row.put("name", table.getPhysicalColumnName(dimension.getApiName()));
+                        row.put("name", dimension.getApiName());
+                        row.put("factName", table.getPhysicalColumnName(dimension.getApiName()));
```
QubitPi: `dimension.getApiName()` to a variable since it's used twice

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/SlicesApiRequestSpec.groovy
```diff
@@ -94,9 +94,10 @@ class SlicesApiRequestSpec extends BaseDataSourceMetadataSpec {
         table.allAvailableIntervals.each {
             Map<String, Object> row = [:] as LinkedHashMap
             row["intervals"] = it.value
-            row["name"] = table.getPhysicalColumnName(it.key.name)
+            row["name"] = it.key.name
             if (it.key instanceof DimensionColumn) {
                 row["uri"] = uri + it.key.name
+                row["factName"] = table.getPhysicalColumnName(it.key.name)
```
QubitPi: Just curious, why are we putting this inside a conditional?

asifmansoora: Because `factName` was valid only for dimension rows

CHANGELOG.md
```diff
@@ -172,7 +172,9 @@ Current
 ### Removed:
 
 - [Reverted the druid name change in slices endpoint instead added to factName](https://github.com/yahoo/fili/pull/541)
-    * Reverting this PR #491 so that the name still points to apiName and added factName which points to druidName.
+    * Reverting the PR-419(https://github.com/yahoo/fili/pull/419) so that the name still points to apiName and added factName which points to druidName.
+      `name` was not valid for cases when its a Lookup dimension because it was pointing to the base dimension name , so reverted that change and added
```
QubitPi: its -> it is

PR #543, opened 2017-09-26
Enable field accessors to reference post aggregations in addition to …

michael-mclawhorn:
…aggregations.

The proximate need for this is that having metrics do not evaluate correctly on sketch metrics.  A group is going to try building a rewrite that adds synthetic sketch estimates and rewrites their havings to work around this bug.

It's not clear whether it's a better idea to wait for the sketch estimates to be fixed or to lift up this solution into Fili.  But they're going to take the hit of a first pass solution and maybe they'll contribute it back if it works well.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/postaggregation/FieldAccessorPostAggregation.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/postaggregation/FieldAccessorPostAggregation.java
@@ -5,7 +5,7 @@ package com.yahoo.bard.webservice.druid.model.postaggregation;
 import static com.yahoo.bard.webservice.druid.model.postaggregation.PostAggregation.DefaultPostAggregationType.FIELD_ACCESS;
 
 import com.yahoo.bard.webservice.data.dimension.Dimension;
-import com.yahoo.bard.webservice.druid.model.aggregation.Aggregation;
+import com.yahoo.bard.webservice.druid.model.MetricField;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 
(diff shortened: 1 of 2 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/postaggregation/FieldAccessorPostAggregation.java
```diff
@@ -24,14 +24,14 @@
...
-    private final Aggregation aggregation;
+    private final MetricField metricField;
 
     /**
      * Constructor.
      *
      * @param aggregation  Aggregation to access
      */
-    public FieldAccessorPostAggregation(@NotNull Aggregation aggregation) {
+    public FieldAccessorPostAggregation(@NotNull MetricField aggregation) {
```
QubitPi: Shall we change parameter name to `metricField`?

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/postaggregation/FieldAccessorPostAggregation.java
```diff
@@ -41,21 +41,40 @@ public FieldAccessorPostAggregation(@NotNull Aggregation aggregation) {
             throw new IllegalArgumentException(message);
         }
 
-        this.aggregation = aggregation;
+        this.metricField = aggregation;
```
QubitPi: Shall we change `aggregation` to `metricField`?

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/postaggregation/FieldAccessorPostAggregation.java
asifmansoora: @QubitPi its actually an `aggregation` that will be used here- hence the name

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/postaggregation/FieldAccessorPostAggregation.java
asifmansoora: same applies here

PR #545, opened 2017-10-03
substitute preflight method wildcard with explicit methods

colemanProjects:
While using fili, I got the error message:
`Method PUT is not allowed by Access-Control-Allow-Methods in preflight response`

After digging around I found that for preflight requests many browsers don't support a wildcard. 
https://stackoverflow.com/questions/44914330/method-put-is-not-allowed-by-access-control-allow-methods-in-preflight-response

my `PUT` requests work after applying this fix.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/ResponseCorsFilter.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/ResponseCorsFilter.java
@@ -37,7 +37,7 @@ public class ResponseCorsFilter implements ContainerResponseFilter {
         // allow all requested headers
         headers.add("Access-Control-Allow-Headers", requestedHeaders == null ? "" : requestedHeaders);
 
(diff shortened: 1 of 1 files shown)


michael-mclawhorn:
Looks like a reasonable change.  What's the motivation?  Which endpoint are you trying to PUT to?

colemanProjects:
I created a custom endpoint and needed to PUT to that one.

michael-mclawhorn:
Alright.  Add a CHANGELOG entry on this then.
Then I think it should be able to get approved.

colemanProjects:
great, will do

colemanProjects:
Changelog updated

michael-mclawhorn:
Did you mean to close it?

colemanProjects:
nope, i didn't

colemanProjects:
https://github.com/yahoo/fili/pull/546/files

michael-mclawhorn:
You could have just reopened this one

PR #546, opened 2017-10-03
substitute preflight method wildcard with explicit methods

colemanProjects:
initial pr here https://github.com/yahoo/fili/pull/545


colemanProjects:
seems like this can be merged now. I don't have write access.

PR #547, opened 2017-10-10
Improves `Dimension::getFieldByName` documentation

archolewa:
-- The `ApiFilter` constructor relies on the fact that
`Dimension::getFieldByName` throws an `IllegalArgumentException` to
perform validation. However, that is not documented on the `Dimension`
interface, which may lead implementors of `Dimension` to erroneously
assume that returning `null` is valid if a `DimensionField` is not found
(which is common for this kind of method). However, what actually
happens is that the `ApiFilter` is constructed with a `null` field,
leading to a `NullPointerException` at some point later.

-- So the documentation on `Dimension` is updated to make that explicit.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/Dimension.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/Dimension.java
@@ -75,6 +75,8 @@ public interface Dimension {
      * @param name  field name
      *
      * @return DimensionField
+     *
+     * @throws IllegalArgumentException if this dimension does not have a field with the specified name
      */
     DimensionField getFieldByName(String name);
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 Change Log
 ==========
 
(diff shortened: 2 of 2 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/Dimension.java
```diff
@@ -75,6 +75,8 @@
      * @param name  field name
      *
      * @return DimensionField
+     *
+     * @throws IllegalArgumentException if this dimension does not have a field with the specified name
```
garyluoex: This should depend on implementation right? so may be move it to key value store dimension? Or are we trying to force people to follow this contract?

archolewa: The [ApiFilter](https://github.com/yahoo/fili/blob/c930a5858291b0eb3d98bc62009144c359c9ef1b/fili-core/src/main/java/com/yahoo/bard/webservice/web/ApiFilter.java#L130) relies on this behavior to validate properly, so it very much *can't* depend on the implementation.

PR #550, opened 2017-10-13
Handle missing cardinalities graciously in case of uninitialized dime…

michael-mclawhorn:
…nsion storage.


fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/KeyValueStore.java
```diff
@@ -50,6 +50,18 @@
     String get(@NotNull String key);
 
     /**
+     * Get the value for a key from store or provide a default
```
QubitPi: period at end of sentense

michael-mclawhorn:
@QubitPi fixed

PR #551, opened 2017-10-13
Add the necessary steps for lucene replace

garyluoex:


garyluoex:
One thing I'm not sure is the lock, two write locks together, not sure if it will work

michael-mclawhorn:
We need more tests for this, but that was true before.  This fix seems right to me.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProviderSpec.groovy
```diff
@@ -191,6 +192,7 @@ class LuceneSearchProviderSpec extends SearchProviderSpec<LuceneSearchProvider>
         !Files.exists(file4)
     }
 
+    @Ignore("This test is currently not valid because the replacement index is invalid.")
```
QubitPi: This shouldn't seem to be ignored, 'cause it was working previously

QubitPi:
Given this needs to be addressed, https://github.com/yahoo/fili/issues/552, I'm OK with it

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProviderSpec.groovy
garyluoex: It no longer works, because the file we are using is fake, the method now requires a real index file

QubitPi:
@michael-mclawhorn Test will be my next urgent thing to do https://github.com/yahoo/fili/issues/552

Issue #548, opened 2017-10-11
Support dimension metadata to indicate storage strategy 


jkusa:
In order to allow clients to be notified if a dimension's values are browsable and searchable we need a way to indicate the storage strategy. 

A new dimension metadata attribute can be added at the top level called `storageStrategy` with values:

- `loaded`  --> indicates that values are loaded into fili and are browsable and searchable
- `none` --> indicates that values are NOT available
- etc

QubitPi:
[Addressed](https://github.com/yahoo/fili/pull/558)

PR #558, opened 2017-10-30
Implement dimension metadata to indicate storage strategy

QubitPi:
Address issue https://github.com/yahoo/fili/issues/548


QubitPi:
@michael-mclawhorn Sure. Next time I'll put more attention to `SearchProvider` to make implementation possibly more optimal

Issue #534, opened 2017-09-06
Remove testing constructor of *ApiRequestImpl


QubitPi:
It's not a good practice to have testing code in https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequestImpl.java . This [constructor](https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequestImpl.java#L374), along with all such constructors in ApiRequest, should be removed entirely.

QubitPi:
[Addressed](https://github.com/yahoo/fili/pull/559)

PR #559, opened 2017-10-31
Remove testing constructor of *ApiRequestImpl

QubitPi:
Addresses issue https://github.com/yahoo/fili/issues/534

* It is a better practice to separate testing code with implementation. All testing constructors of the following API requests are removed:
    - `ApiRequestImpl`
    - `DataApiRequestImpl`
    - `DimensionsApiRequestImpl`
    - `MetricsApiRequestImpl`
    - `SlicesApiRequestImpl`
    - `TablesApiRequestImpl`
* Meanwhile, construction of testing API request is delegated to testing class, e.g. `TestingDataApiRequestImpl`

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequestImpl.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequestImpl.java
@@ -24,11 +24,9 @@ import com.yahoo.bard.webservice.data.dimension.Dimension;
 import com.yahoo.bard.webservice.data.dimension.DimensionDictionary;
 import com.yahoo.bard.webservice.data.dimension.DimensionField;
 import com.yahoo.bard.webservice.data.dimension.DimensionRowNotFoundException;
-import com.yahoo.bard.webservice.data.filterbuilders.DefaultDruidFilterBuilder;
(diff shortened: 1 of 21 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequestImpl.java
```diff
@@ -294,6 +291,73 @@ public DataApiRequestImpl(
...
+     * @param sorts  Sorting info for the request
+     * @param count  Global limit for the request
+     * @param topN  Count of per-bucket limit (TopN) for the request
+     * @param asyncAfter  How long in milliseconds the user is willing to wait for a synchronous response
+     * @param timeZone  TimeZone for the request
+     * @param filterBuilder  A builder to use when building filters for the request
+     * @param havingApiGenerator  A generator to generate havings map for the request
+     * @param dateTimeSort  A dateTime sort column with its direction
+     */
+    protected DataApiRequestImpl(
```
QubitPi: This constructor already exists before this PR. The change I made was changing the method modifier from `private` to `protected` so that the testing class, which extends this `DataApiRequestImpl`, can construct testing `DataApiRequest`

PR #567, opened 2017-11-07
Fix tests with wrong time offset calculation

QubitPi:
Time-checking based tests setup time offset in a wrong way. `timeZoneId.getOffset` is fixed to take the right argument.


QubitPi:
Thanks, @asifmansoora !

QubitPi:
Thanks, @archolewa !

PR #570, opened 2017-11-08
Logical table info class

efronbs:
Pull request to get input on the current approach to Logical Table Info.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/config/DefaultLogicalTableInfo.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/config/DefaultLogicalTableInfo.java
@@ -1,0 +1,8 @@
+// Copyright 2017 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.config;
+
+public interface DefaultLogicalTableInfo {
+
+    // TODO default all the fields to simulate current name only behavior
+}
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/config/DefaultLogicalTableInfo.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/config/DefaultLogicalTableInfo.java
@@ -2,7 +2,27 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.config;
 
-public interface DefaultLogicalTableInfo {
+/**
+ * Default for LogicalTableInfo interface, everything just returns getName.
+ */
+@FunctionalInterface
+public interface DefaultLogicalTableInfo extends LogicalTableInfo {
 
-    // TODO default all the fields to simulate current name only behavior
+    /**
+     * Provides a default behavior for getLongName, which is to return the unparsed table name.
+     *
+     * @return  A string representing the human readable name of this table.
+     */
+    default String getLongName() {
+        return this.getName();
+    }
+
+    /**
+     * Provides a default behavior for getDescription, which is to return the unparsed table name.
+     *
+     * @return  A string representing the description of this table's purpose.
+     */
+    default String getDescription() {
+        return this.getName();
+    }
 }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/config/DefaultLogicalTableInfo.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/config/DefaultLogicalTableInfo.java
@@ -2,18 +2,40 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.config;
 
+import org.joda.time.ReadablePeriod;
+import org.joda.time.Years;
+
+import java.util.Optional;
+
 /**
  * Default for LogicalTableInfo interface, everything just returns getName.
  */
-@FunctionalInterface
-public interface DefaultLogicalTableInfo extends LogicalTableInfo {
+public class DefaultLogicalTableInfo implements LogicalTableInfo {
+
+    public static final String DEFAULT_CATEGORY = "General";
+    public static final ReadablePeriod DEFAULT_RETENTION = Years.ONE;
+
+    String name;
+
+    /**
+     * Constructor.
+     *
+     * @param name  The name to default for other name-like values.
+     */
+    public DefaultLogicalTableInfo(String name) {
+        this.name = name;
+    }
+    @Override
+    public String getName() {
+        return name;
+    }
 
     /**
      * Provides a default behavior for getLongName, which is to return the unparsed table name.
      *
      * @return  A string representing the human readable name of this table.
      */
-    default String getLongName() {
+    public String getLongName() {
         return this.getName();
     }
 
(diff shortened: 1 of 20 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/config/LogicalTableInfo.java
```diff
@@ -0,0 +1,32 @@
...
+     */
+    String getLongName();
+
+    /**
+     * A description of this logical table. Meant to help users understand the purpose of this table, not used for
+     * internal logic
+     *
+     * @return  A
+     */
+    String getDescription();
```
michael-mclawhorn: I think we actually want to lift up retention to this interface.

fili-core/src/main/java/com/yahoo/bard/webservice/config/DefaultLogicalTableInfo.java
```diff
@@ -0,0 +1,28 @@
...
+    default String getLongName() {
+        return this.getName();
+    }
+
+    /**
+     * Provides a default behavior for getDescription, which is to return the unparsed table name.
+     *
+     * @return  A string representing the description of this table's purpose.
+     */
+    default String getDescription() {
```
michael-mclawhorn: Add default category == LogicalTable.DEFAULT_CATEGORY
and default retention == LogicalTable.DEFAULT_RETENTION

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
```diff
@@ -160,7 +161,11 @@ public void loadLogicalTableWithGranularities(
         // For every legal grain
         for (Granularity grain : validGrains) {
             // Build the logical table
-            LogicalTable logicalTable = new LogicalTable(logicalTableName, grain, nameGroup, metricDictionary);
+            LogicalTable logicalTable = new LogicalTable(
+                    (DefaultLogicalTableInfo) () -> logicalTableName,
```
michael-mclawhorn: Does this work?  This doesn't look like it works.  The fact that this doesn't break a test worries me.

Ok, this almost certainly isn't doing what you think it's doing.

fili-core/src/main/java/com/yahoo/bard/webservice/config/DefaultLogicalTableInfo.java
```diff
@@ -0,0 +1,28 @@
+// Copyright 2017 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.config;
+
+/**
+ * Default for LogicalTableInfo interface, everything just returns getName.
+ */
+@FunctionalInterface
+public interface DefaultLogicalTableInfo extends LogicalTableInfo {
+
```
michael-mclawhorn: I think perhaps we should go ahead and make this a class that constructs off String name.

michael-mclawhorn:
basically going in the right direction.

fili-core/src/main/java/com/yahoo/bard/webservice/config/DefaultLogicalTableInfo.java
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Constructor.
+     *
+     * @param name  The name to default for other name-like values.
+     */
+    public DefaultLogicalTableInfo(String name) {
+        this.name = name;
+    }
+    @Override
```
QubitPi: Empty line

fili-core/src/main/java/com/yahoo/bard/webservice/config/DefaultLogicalTableInfo.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Default for LogicalTableInfo interface, everything just returns getName.
+ */
+public class DefaultLogicalTableInfo implements LogicalTableInfo {
+
+    public static final String DEFAULT_CATEGORY = "General";
+    public static final ReadablePeriod DEFAULT_RETENTION = Years.ONE;
+
+    String name;
```
QubitPi: `private final`

fili-core/src/main/java/com/yahoo/bard/webservice/config/DefaultLogicalTableInfo.java
```diff
@@ -0,0 +1,60 @@
...
+        return this.getName();
+    }
+
+    /**
+     * Provides a default behavior for getDescription, which is to return the unparsed table name.
+     *
+     * @return  A string representing the description of this table's purpose.
+     */
+    public String getDescription() {
+        return this.getName();
```
QubitPi: Is name supposed to be description?

fili-core/src/main/java/com/yahoo/bard/webservice/config/DefaultLogicalTableInfo.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public String getCategory() {
+        return DEFAULT_CATEGORY;
+    }
+
+    @Override
+    public Optional<ReadablePeriod> getRetention() {
+        return Optional.of(DEFAULT_RETENTION);
```
QubitPi: Can we make `DEFAULT_RETENTION` itself an optional?

fili-core/src/main/java/com/yahoo/bard/webservice/table/LogicalTable.java
```diff
@@ -43,16 +43,44 @@
...
+    }
+
+    /**
+     * Constructor
+     *  A constructor that takes in a logical table info, and uses that information.
+     *
+     * @param logicalTableInfo  Hello.
+     * @param granularity  Hello.
+     * @param tableGroup  Hello.
+     * @param metricDictionary  Hello.
```
QubitPi: Need to change description

fili-core/src/main/java/com/yahoo/bard/webservice/table/LogicalTable.java
```diff
@@ -43,16 +43,44 @@
...
             TableGroup tableGroup,
             MetricDictionary metricDictionary
     ) {
-        this(name, DEFAULT_CATEGORY, name, granularity, DEFAULT_RETENTION, name, tableGroup, metricDictionary);
+        this(new DefaultLogicalTableInfo(name), granularity, tableGroup, metricDictionary);
+    }
+
+    /**
+     * Constructor
+     *  A constructor that takes in a logical table info, and uses that information.
```
QubitPi: This line can be removed (Or make it more descriptive)

fili-core/src/main/java/com/yahoo/bard/webservice/config/DefaultLogicalTableInfo.java
michael-mclawhorn: This: http://blog.joda.org/2015/08/java-se-8-optional-pragmatic-approach.html is the guidance we're used for use of Optional.  Since this is essentially returning a constant, this may break the overarching concern about holding Optionals as internal representations.

michael-mclawhorn: The top line relevant rule, we've generally followed, is to not hold Optionals in internal state, but to use it primarily as return values where nullable state being handled is mandatory.

Issue #549, opened 2017-10-13
Uninitialized dimension storage leads to exception when processing metadata


michael-mclawhorn:
Uninitialized dimension storage causes a null based error when lucene search provider attempts to retrieve cardinality.  This totally disables metadata endpoints that provide cardinality.

QubitPi:
Doesn't seem to be an issue now, hence closing.

PR #571, opened 2017-11-09
Check initialization of dimension storage

QubitPi:
Address issue https://github.com/yahoo/fili/issues/549

Uninitialized dimension storage causes a null based error when search provider attempts to retrieve cardinality. A default method is added to search provider interface that forces a null check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Current
 
 ### Added:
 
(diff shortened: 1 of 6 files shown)


michael-mclawhorn:
I'm not 100% sure this is the solution I was thinking of for this problem.  It seems like null checking everything is a recipe for code bloat.

QubitPi:
Doesn't seem to be an issue now, hence closing.

Issue #556, opened 2017-10-26
Check for null dimension when building selector filter


michael-mclawhorn:
https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/SelectorFilter.java

Dimension should be non-null and exception checked

This not being checked creates uncaught until runtime errors.

QubitPi:
Doesn't seem to be an issue now, hence closing.

PR #572, opened 2017-11-09
Check for null dimension when building selector filter

QubitPi:
Address issue https://github.com/yahoo/fili/issues/556

Dimension is checked for null.


michael-mclawhorn:
Recommend closing this.

QubitPi:
Doesn't seem to be an issue now, hence closing.

PR #573, opened 2017-11-10
Fix dsc class scanner spec error - ignore

michael-mclawhorn:


fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/ApiFilters.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * ApiFilters specializes the type of the ApiFilterMap.
+ */
+public class ApiFilters extends LinkedHashMap<Dimension, Set<ApiFilter>> {
+
+    /**
+     * Constructor.
+     */
+    public ApiFilters() {
```
garyluoex: Does this constructor need to call super() or is it not needed?

garyluoex: It should work, good to merge

PR #574, opened 2017-11-10
Expose RequestLog's LogInfo objects.

archolewa:
-- Some customers want to be able to customize how the RequestLog is
logged. By exposing the LogInfo objects and the durations, we
effectively give customers full control over how they want to log
each component of the RequestLog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,10 @@ Current
 
 ### Changed:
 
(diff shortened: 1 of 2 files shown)


QubitPi:
@archolewa Build will pass after a rebase

michael-mclawhorn:
A small CHANGELOG would not be inappropriate.

archolewa:
@michael-mclawhorn Addressed your comment.

PR #578, opened 2017-11-15
Display corrected case on StorageStrategy serialization

QubitPi:
The default serialization of enum is `name()` which is final and thus cannot be overridden. An API method is added to return the API name of a storage strategy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,11 @@ Current
 
 
 ### Changed:
+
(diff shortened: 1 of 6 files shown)


michael-mclawhorn:
Is the contract 'loaded' or 'Loaded'?  I thought it was the former.  If the issue is the serialization, why  aren't we adding a serialization annotation to the new method?

QubitPi:
Thanks, @michael-mclawhorn !

QubitPi:
Thanks, @garyluoex !

PR #580, opened 2017-11-17
Fix having with scoped metrics bug

archolewa:
-- Before, the ApiHavingGenerator relied on the globally scoped metric
dictionary to resolve having metrics. However, that meant that Fili
failed to recognize any metrics in the having clause that appear only
in a metric dictionary of a lower scope (i.e. a metric dictionary scoped
to table name).

-- We fix the bug by introducing a new ApiHavingGenerator that
effectively builds a temporary metric dictionary from the set of
requested metrics, and then using *those* to resolve the having clause.

-- We then use that new generator as the default generator in order to
ensure that the two features (scoped metric dictionaries and having
clauses) actually work together by default.

-- Furthermore, while writing an end-to-end test reproducing the bug, we
discovered that the default table configuration strategy had the same
problem: the tableLoader resolves its metrics from the globally scoped
metric dictionary, not from any lower scope.

-- To address this problem, we introduce some new table generating
functions in the `BaseTableLoader` that effectively allow the customer
to provide a different metric dictionary for use when building each
table. The original table generating methods are kept, because most
people don't need scoped metric dictionaries, and shouldn't have to
wrestle with their complexity when building their application.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
@@ -101,6 +101,7 @@ import com.yahoo.bard.webservice.web.SlicesApiRequest;
 import com.yahoo.bard.webservice.web.TablesApiRequest;
 import com.yahoo.bard.webservice.web.apirequest.DefaultHavingApiGenerator;
 import com.yahoo.bard.webservice.web.apirequest.HavingGenerator;
+import com.yahoo.bard.webservice.web.apirequest.PerRequestDictionaryHavingGenerator;
 import com.yahoo.bard.webservice.web.handlers.workflow.DruidWorkflow;
 import com.yahoo.bard.webservice.web.handlers.workflow.RequestWorkflowProvider;
 import com.yahoo.bard.webservice.web.util.QueryWeightUtil;
@@ -635,7 +636,7 @@ public abstract class AbstractBinderFactory implements BinderFactory {
      * @return An object to generate having maps from having string.
      */
     protected HavingGenerator buildHavingGenerator(ConfigurationLoader loader) {
-        return new DefaultHavingApiGenerator(loader);
+        return new PerRequestDictionaryHavingGenerator(new DefaultHavingApiGenerator(loader.getMetricDictionary()));
     }
 
     /**
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
@@ -116,7 +116,7 @@ public abstract class BaseTableLoader implements TableLoader {
     }
 
     /**
-     * Load several logical tables into the logicalDictionary.
+     * Load several logical tables into the logicalDictionary, all with the globally-scoped metric dictionary.
      * <p>
      * Note: This builds the logical tables as well.
      *
@@ -129,12 +129,42 @@ public abstract class BaseTableLoader implements TableLoader {
             Set<? extends Granularity> validGrains,
             ResourceDictionaries dictionaries
     ) {
-        // For every logical table name, group pair
+        loadLogicalTablesWithGranularities(
+                nameGroupMap,
+                validGrains,
+                dictionaries.getLogicalDictionary(),
+                nameGroupMap.keySet().stream()
+                        .collect(Collectors.toMap(Function.identity(), i -> dictionaries.getMetricDictionary()))
+        );
+    }
+
+    /**
+     * Load several logical tables into the logicalDictionary, each with their own scoped metric dictionary.
+     * <p>
+     * Note: This builds the logical tables as well.
+     *
+     * @param nameGroupMap  A map of logical table name to table group information
+     * @param validGrains  The accepted grains for the logical table
+     * @param tableDictionary  The logical table dictionary to be populated
+     * @param scopedMetrics  A mapping from table name to the scoped MetricDictionary to use for that table
+     */
+    public void loadLogicalTablesWithGranularities(
+            Map<String, TableGroup> nameGroupMap,
+            Set<? extends Granularity> validGrains,
+            LogicalTableDictionary tableDictionary,
+            Map<String, MetricDictionary> scopedMetrics
+    ) {
         for (Map.Entry<String, TableGroup> entry : nameGroupMap.entrySet()) {
             String logicalTableName = entry.getKey();
             TableGroup group = entry.getValue();
 
-            loadLogicalTableWithGranularities(logicalTableName, group, validGrains, dictionaries);
+            loadLogicalTableWithGranularities(
+                    logicalTableName,
+                    group,
+                    validGrains,
+                    tableDictionary,
+                    scopedMetrics.get(logicalTableName)
+            );
         }
     }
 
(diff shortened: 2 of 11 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/BaseTableLoader.java
```diff
@@ -154,16 +184,44 @@ public void loadLogicalTableWithGranularities(
...
         // For every legal grain
         for (Granularity grain : validGrains) {
             // Build the logical table
-            LogicalTable logicalTable = new LogicalTable(logicalTableName, grain, nameGroup, metricDictionary);
-
+            LogicalTable logicalTable = new LogicalTable(
+                    logicalTableName,
+                    grain,
+                    nameGroup,
+                    metricDictionary
```
QubitPi: arguments can be unwrapped

fili-core/src/main/java/com/yahoo/bard/webservice/web/ApiHaving.java
```diff
@@ -58,7 +58,11 @@
      *
      * @throws BadHavingException  when having pattern is not matched or when any of its properties are not valid.
      */
-    public ApiHaving(@NotNull String havingQuery, MetricDictionary metricDictionary) throws BadHavingException {
+    public ApiHaving(
+            @NotNull String havingQuery,
+            Map<String, LogicalMetric> metricDictionary
```
QubitPi: can be unwrapped

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DefaultHavingApiGenerator.java
```diff
@@ -24,16 +25,14 @@
...
-
 /**
  * Generates having objects based on the having query in the api request.
  */
 public class DefaultHavingApiGenerator implements HavingGenerator {
     private static final Logger LOG = LoggerFactory.getLogger(DefaultHavingApiGenerator.class);
     private static final String COMMA_AFTER_BRACKET_PATTERN = "(?<=]),";
 
-    private final MetricDictionary metricDictionary;
+    private final Map<String, LogicalMetric> metricDictionary;
```
QubitPi: Why are we changing type here?

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DefaultHavingApiGenerator.java
```diff
@@ -44,6 +43,15 @@ public DefaultHavingApiGenerator(ConfigurationLoader loader) {
         this.metricDictionary = loader.getMetricDictionary();
     }
 
+    /**
+     * Constructor.
+     *
+     * @param metricDictionary  Dictionary of metrics for generating APIs
+     */
+    public DefaultHavingApiGenerator(Map<String, LogicalMetric> metricDictionary) {
```
QubitPi: I assume the other constructor above this constructs a global metric dictionary, right? If it's the case, it's a good idea to add a paragraph of description to clarify that(and this constructor as well)

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/ScopedMetricHavingDataServletSpec.groovy
```diff
@@ -0,0 +1,99 @@
+// Copyright 2016 Yahoo Inc.
```
QubitPi: 2017

fili-core/src/test/java/com/yahoo/bard/webservice/data/config/metric/TestMetricLoader.java
```diff
@@ -103,8 +104,11 @@ public void loadMetricDictionary(MetricDictionary metricDictionary) {
         );
 
         metrics.stream().map(MetricInstance::make).forEach(metricDictionary::add);
+        metricDictionary.getScope("shapes").add(
+                new MetricInstance(A_SCOPED_WIDTH, longSumMaker, WIDTH).make()
```
QubitPi: This constructor has been deprecated. Could we use a different one?

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -635,7 +636,7 @@ protected DruidFilterBuilder buildDruidFilterBuilder() {
      * @return An object to generate having maps from having string.
      */
     protected HavingGenerator buildHavingGenerator(ConfigurationLoader loader) {
-        return new DefaultHavingApiGenerator(loader);
+        return new PerRequestDictionaryHavingGenerator(new DefaultHavingApiGenerator(loader.getMetricDictionary()));
```
QubitPi: Looks like we don't need to pass `loader` anymore, but only `MetricDictionary` instead

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/ScopedMetricHavingDataServletSpec.groovy
```diff
@@ -0,0 +1,99 @@
...
+    @Override
+    String getTarget() {
+        return "data/shapes/week/color"
+    }
+
+    @Override
+    Map<String, List<String>> getQueryParams() {
+        [
+                "metrics" : ["scopedWidth"],
+                "dateTime": ["2014-06-02%2F2014-06-09"],
```
QubitPi: Could we decode `%2F` to make it more readable :) ?

ghost (commented):
I'm little curious. Isn't global dictionary supposed to contain all valid scoped dictionary entries?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/ScopedMetricHavingDataServletSpec.groovy
```diff
@@ -0,0 +1,99 @@
+// Copyright 2016 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.web.endpoints
+
+class ScopedMetricHavingDataServletSpec extends BaseDataServletComponentSpec {
```
QubitPi: I'm little curious, how does this Spec endup making the changes to `ExpectedTablesFullViewEndpointSpec.groovy` Spec above?

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
archolewa: Two reasons not to:

1. Just because *this* `HavingGenerator` doesn't need all the configuration, doesn't mean other people's custom `HavingGenerator won't.

2. Changing `buildHavingGenerator` would break any customers that override it, and we don't want to make backwards incompatible changes if we don't have to.

fili-core/src/main/java/com/yahoo/bard/webservice/web/ApiHaving.java
archolewa: No we can't. The exception sends it over 120 characters.

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DefaultHavingApiGenerator.java
archolewa: A very rather pedantic reason that came out of me being a bit anal retentive. Throughout the codebase, the `MetricDictionary` has a very specific (though unenforced) meaning: The collection of *all* configured metrics at any given scope. However, in order to fix the bug without complicating the `HavingGenerator` interface, we needed to pass in a subset of the configured metrics: those that were requested by the query. I didn't like referring to that as a `MetricDictionary`, because I felt it violated the principal of least surprise. So we loosened the requirements down to a map of metrics instead.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/ScopedMetricHavingDataServletSpec.groovy
archolewa: I needed to add a scoped metric to our metric configuration in order to reproduce the bug. The `ExpectedTablesFullViewEndpointSpec` prints all the metrics for each table. So I needed to update the shapes table's output with the new metric.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
QubitPi: make sense

ghost (approved):

garyluoex (approved):

michael-mclawhorn (approved):

PR #583, opened 2017-11-18
change comment from 'form' to 'from'

colemanProjects:


cdeszaq:
@QubitPi, Can this PR merge? It looks like the spot you linked isn't valid (or at least doesn't seem to have to same problem) any more.

QubitPi:
@cdeszaq Sure. Thanks for updating!

Issue #587, opened 2017-11-21
MetricInstance should have a getter for LogicalMetricInfo


QubitPi:
There are 3 instance variables inside `MetricInstance` class, two of which have getters. The one without getter, `LogicalMetricInfo`, should have one, as well, so that subclass can access it without creating a duplicate `LogicalMetricInfo` inside their own.

QubitPi:
Addressed by https://github.com/yahoo/fili/pull/588

PR #588, opened 2017-11-21
Add getter for LogicalMetricInfo in MetricInstance

QubitPi:
https://github.com/yahoo/fili/issues/587: There are 3 instance variables inside `MetricInstance` class, two of which have getters. The one without getter, `LogicalMetricInfo`, should have one, as well, so that subclass can access it without creating a duplicate `LogicalMetricInfo` inside their own.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ Current
 
 ### Added:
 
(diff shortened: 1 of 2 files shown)


QubitPi:
Thanks, Gary!

QubitPi:
Thanks, @asifmansoora !

PR #593, opened 2017-12-01
Metric type field annotation

efronbs:


QubitPi:
@efronbs Did you pass the build locally?

Issue #273, opened 2017-05-12
lucene_hits_per_page is ignored


archolewa:
The paramater [lucene_hits_per_page](https://github.com/yahoo/fili/blob/816deafb0edf5eee88b1f60f403cc0289907c602/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProviderManager.java#L22) is initialized, but then ignored.

The paramater [max_results_without_filter](https://github.com/yahoo/fili/blob/872379bb346a444579ad7e76eb3306eef74731a6/fili-core/src/main/java/com/yahoo/bard/webservice/web/util/PaginationParameters.java#L26) is used instead.

We should decide how to handle this. Two possibilities are:

1. Allow one to override the other with some sort of priority for lucene-backed dimensions.

2. Remove `lucene_hits_page` completely. This will generate an error for services that use the parameter when deploying, but that's better than silently ignoring it.

archolewa:
`max_results_without_filter` is also incorrectly named, because the limit applies even when filters are applied. It's actually being used as the maximum size of a single page of results.

cdeszaq:
I'm a fan of option 2, and I would even go so far as to look for it and log a warning, since we'd rather people move off of it. And I agree, renaming the thing so that it's config name matches what it's used for would be good.

PR #600, opened 2017-12-15
Remove unused param & fix naming

QubitPi:
Address https://github.com/yahoo/fili/issues/273, @archolewa

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/util/PaginationParameters.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/util/PaginationParameters.java
@@ -25,17 +25,17 @@ public class PaginationParameters {
     private static final String FIRST = "first";
     private static final String LAST = "last";
 
-    private static final int DEFAULT_MAX_RESULTS_WITHOUT_FILTERS = 10000;
-    private static final int MAX_RESULTS_WITHOUT_FILTER = SYSTEM_CONFIG.getIntProperty(
-            SYSTEM_CONFIG.getPackageVariableName("max_results_without_filters"),
-            DEFAULT_MAX_RESULTS_WITHOUT_FILTERS
+    private static final int DEFAULT_MAX_RESULTS_PER_PAGE = 10000;
+    private static final int MAX_RESULTS_PER_PAGE = SYSTEM_CONFIG.getIntProperty(
+            SYSTEM_CONFIG.getPackageVariableName("max_results_per_page"),
+            DEFAULT_MAX_RESULTS_PER_PAGE
     );
 
     /**
      * The common pagination parameter object used by non-paginating method in search providers.
      */
     public static final PaginationParameters EVERYTHING_IN_ONE_PAGE = new PaginationParameters(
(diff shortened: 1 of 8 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/web/ratelimit/DefaultRateLimiter.java
```diff
@@ -122,8 +122,8 @@ protected AtomicInteger getCount(String userName) {
      * @param initialCount  Initial count that we're incrementing and checking against the limit
      * @param requestLimit  Limit to check the incremented initial count against
      *
-     * @return True if the incremented count is <= the request limit, false if it's gone over and the request limit
```
QubitPi: `<=` doesn't pass checkstyle.

QubitPi: `<=` doesn't pass checkstyle

fili-core/src/main/resources/moduleConfig.properties
```diff
@@ -78,8 +78,8 @@ bard__lucene_index_path = [SET ME IN APPLICATION CONFIG]
 # Lucene search timeout in milliseconds
 bard__lucene_search_timeout_ms = 600000
 
-# setting for maximum allowed results without any filters - used for /dim/values endpoint
-bard__max_results_without_filters = 10000
+# setting for maximum allowed results per page - used for /dim/values endpoint
+bard__max_results_per_page = 10000
```
cdeszaq: These config name changes need to be called out in the changelog I think.

fili-core/src/main/java/com/yahoo/bard/webservice/web/util/PaginationParameters.java
```diff
@@ -25,17 +25,17 @@
     private static final String FIRST = "first";
     private static final String LAST = "last";
 
-    private static final int DEFAULT_MAX_RESULTS_WITHOUT_FILTERS = 10000;
-    private static final int MAX_RESULTS_WITHOUT_FILTER = SYSTEM_CONFIG.getIntProperty(
-            SYSTEM_CONFIG.getPackageVariableName("max_results_without_filters"),
-            DEFAULT_MAX_RESULTS_WITHOUT_FILTERS
+    private static final int DEFAULT_MAX_RESULTS_PER_PAGE = 10000;
+    private static final int MAX_RESULTS_PER_PAGE = SYSTEM_CONFIG.getIntProperty(
+            SYSTEM_CONFIG.getPackageVariableName("max_results_per_page"),
```
cdeszaq: Changing this config param name is backwards incompatible. Can we continue to use the old param, logging a warning if it's still used, but respect it, while looking for the new one if it's there?

The order would be something like:

```groovy
MAX_RESULTS_PER_PAGE = max_results_without_filter 
    ? logErrorAndReturnMaxResultsWoFilterValue
    : max_results_per_page;
```

michael-mclawhorn:
@QubitPi Either resolve @cdeszaq 's compatibility concerns or close, please

Issue #323, opened 2017-06-08
Misleading or Incorrect Error Message on Unconfigured Time Grain


cdeszaq:
A query like this:

    http://example.com/v1/data/myTable/hour?metrics=aMetric&dateTime=2014/2015

where the `myTable` table has been configured to have the `day`, `week`, and `month` grains, but not the `hour` grain,

Returns the following misleading error message:

    Table name 'myTable' is missing

Note that we are asking for a time grain of "hour," even though we only support time grains of day, week, and month. And in fact, the error message is actually incorrect, since `myTable` _is_ a table, and isn't missing! Instead, we should be returning an error something like:

    Time grain 'hour' is not supported. We only support time grains: day, week, and month

PR #601, opened 2017-12-16
Fix handling of partially-matched LogicalTable lookup

QubitPi:
Address issue https://github.com/yahoo/fili/issues/323

The misleading message mentioned in https://github.com/yahoo/fili/issues/323 happens because at https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequestImpl.java#L190 , a `tableId` has a valid table name but has an **invalid** time grain. This makes `get()` return `null`. The correct behavior is by scanning table name and give an error it it doesn't exist in dictionary. Then look up by key `TableIdentifier` and give the second error if we don't find the identifier.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/ApiRequestImpl.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/ApiRequestImpl.java
@@ -13,7 +13,7 @@ import static com.yahoo.bard.webservice.web.ErrorMessageFormat.INVALID_ASYNC_AFT
 import static com.yahoo.bard.webservice.web.ErrorMessageFormat.INVALID_INTERVAL_GRANULARITY;
 import static com.yahoo.bard.webservice.web.ErrorMessageFormat.INVALID_TIME_ZONE;
 import static com.yahoo.bard.webservice.web.ErrorMessageFormat.METRICS_NOT_IN_TABLE;
-import static com.yahoo.bard.webservice.web.ErrorMessageFormat.TABLE_GRANULARITY_MISMATCH;
+import static com.yahoo.bard.webservice.web.ErrorMessageFormat.TABLE_UNDEFINED;
 import static com.yahoo.bard.webservice.web.ErrorMessageFormat.TIME_ALIGNMENT;
 import static com.yahoo.bard.webservice.web.ErrorMessageFormat.UNKNOWN_GRANULARITY;
 
@@ -54,6 +54,8 @@ import org.joda.time.DateTimeZone;
 import org.joda.time.Duration;
 import org.joda.time.Interval;
 import org.joda.time.Period;
+import org.joda.time.PeriodType;
+import org.joda.time.ReadablePeriod;
 import org.joda.time.format.DateTimeFormatter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -656,17 +658,32 @@ public abstract class ApiRequestImpl implements ApiRequest {
             Granularity granularity,
             LogicalTableDictionary logicalTableDictionary
     ) throws BadApiRequestException {
-        LogicalTable generated = logicalTableDictionary.get(new TableIdentifier(tableName, granularity));
-
-        // check if requested logical table grain pair exists
-        if (generated == null) {
-            String msg = TABLE_GRANULARITY_MISMATCH.logFormat(granularity, tableName);
-            LOG.error(msg);
-            throw new BadApiRequestException(msg);
+        // No matching table name is found for this table
+        if (!logicalTableDictionary.getNames().contains(tableName)) {
+            LOG.debug(TABLE_UNDEFINED.logFormat(tableName));
+            throw new BadApiRequestException(TABLE_UNDEFINED.format(tableName));
         }
 
-        LOG.trace("Generated logical table: {} with granularity {}", generated, granularity);
-        return generated;
+        LogicalTable table = logicalTableDictionary.get(new TableIdentifier(tableName, granularity));
+        // No matching granularity is found for this table
+        if (table == null) {
+            List<ReadablePeriod> granularities = logicalTableDictionary.getGranularities(tableName);
+            String message = ErrorMessageFormat.GRANULARITY_NOT_SUPPORTED.format(
+                    granularity.getName(),
+                    tableName,
+                    logicalTableDictionary.getGranularities(tableName).isEmpty()
+                            ? "no granularities"
+                            : granularities.stream()
+                            .map(ReadablePeriod::getPeriodType)
+                            .map(PeriodType::getName)
+                            .collect(Collectors.joining(", "))
+            );
+            LOG.error(message);
+            throw new IllegalArgumentException(message);
+        }
+
+        LOG.trace("Generated logical table: {} with granularity {}", table, granularity);
+        return table;
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


fili-core/src/test/groovy/com/yahoo/bard/webservice/web/TablesApiRequestImplSpec.groovy
```diff
@@ -124,10 +124,10 @@ class TablesApiRequestImplSpec extends Specification {
...
-        name     | grain     | dictionary      | exception              | reason
-        "pets"   | "day"     | emptyDictionary | BadApiRequestException | ".*Logical Table Dictionary is empty.*"
-        "beasts" | "day"     | fullDictionary  | BadApiRequestException | ".*Table name.*does not exist.*"
-        "pets"   | "century" | fullDictionary  | BadApiRequestException | ".*not a valid granularity.*"
-        "pets"   | "hour"    | fullDictionary  | BadApiRequestException | "Invalid pair of granularity .* and table.*"
+        name     | grain     | dictionary      | exception                | reason
+        "pets"   | "day"     | emptyDictionary | BadApiRequestException   | ErrorMessageFormat.EMPTY_DICTIONARY.logFormat("Logical Table")
+        "beasts" | "day"     | fullDictionary  | BadApiRequestException   | ErrorMessageFormat.TABLE_UNDEFINED.logFormat("beasts")
+        "pets"   | "century" | fullDictionary  | BadApiRequestException   | ErrorMessageFormat.UNKNOWN_GRANULARITY.logFormat("century")
+        "pets"   | "hour"    | fullDictionary  | IllegalArgumentException | ErrorMessageFormat.GRANULARITY_NOT_SUPPORTED.format("PT1H", "pets", ".*")
```
QubitPi: This test cases has properly tested `get`

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/TablesApiRequestImplSpec.groovy
```diff
@@ -124,10 +124,10 @@ class TablesApiRequestImplSpec extends Specification {
...
         where:
-        name     | grain     | dictionary      | exception              | reason
-        "pets"   | "day"     | emptyDictionary | BadApiRequestException | ".*Logical Table Dictionary is empty.*"
-        "beasts" | "day"     | fullDictionary  | BadApiRequestException | ".*Table name.*does not exist.*"
-        "pets"   | "century" | fullDictionary  | BadApiRequestException | ".*not a valid granularity.*"
-        "pets"   | "hour"    | fullDictionary  | BadApiRequestException | "Invalid pair of granularity .* and table.*"
+        name     | grain     | dictionary      | exception                | reason
+        "pets"   | "day"     | emptyDictionary | BadApiRequestException   | ErrorMessageFormat.EMPTY_DICTIONARY.logFormat("Logical Table")
+        "beasts" | "day"     | fullDictionary  | BadApiRequestException   | ErrorMessageFormat.TABLE_UNDEFINED.logFormat("beasts")
+        "pets"   | "century" | fullDictionary  | BadApiRequestException   | ErrorMessageFormat.UNKNOWN_GRANULARITY.logFormat("century")
```
QubitPi: Error message(`reason`) is changed to use `ErrorMessageFormat` formatting, which makes the code more maintainable.

fili-core/src/main/java/com/yahoo/bard/webservice/web/ratelimit/DefaultRateLimiter.java
```diff
@@ -122,8 +122,8 @@ protected AtomicInteger getCount(String userName) {
      * @param initialCount  Initial count that we're incrementing and checking against the limit
      * @param requestLimit  Limit to check the incremented initial count against
      *
-     * @return True if the incremented count is <= the request limit, false if it's gone over and the request limit
```
QubitPi: `<=` doesn't pass checkstyle

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/ApiRequestImpl.java
```diff
@@ -660,9 +660,8 @@ protected LogicalTable generateTable(
 
         // check if requested logical table grain pair exists
         if (generated == null) {
-            String msg = TABLE_GRANULARITY_MISMATCH.logFormat(granularity, tableName);
-            LOG.error(msg);
-            throw new BadApiRequestException(msg);
+            LOG.debug(TABLE_UNDEFINED.logFormat(tableName));
+            throw new BadApiRequestException(TABLE_UNDEFINED.format(tableName));
```
QubitPi: `TABLE_GRANULARITY_MISMATCH ` is changed to `TABLE_UNDEFINED ` type of error message, because `logicalTableDictionary.get()` above has already identified the table-granularity-mismatch

Also, this `TABLE_GRANULARITY_MISMATCH ` is misleading, because having `generated == null` doesn't necessarily mean it's a mismatch; it could also be that table with this granularity does not exist in logicalTableDictionary at all.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/apirequest/ApiRequestImplSpec.groovy
```diff
@@ -52,7 +50,7 @@ class ApiRequestImplSpec extends Specification {
 
         then: "we get a BadApiRequestException"
         BadApiRequestException exception = thrown()
-        exception.message == TABLE_GRANULARITY_MISMATCH.logFormat(granularity, nonExistingTableName)
+        exception.message == ErrorMessageFormat.TABLE_UNDEFINED.logFormat(nonExistingTableName)
```
QubitPi: Because of the change from `TABLE_GRANULARITY_MISMATCH` to `TABLE_UNDEFINED` mentioned above

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequestImpl.java
```diff
@@ -184,14 +182,7 @@ public DataApiRequestImpl(
...
 
-        TableIdentifier tableId = new TableIdentifier(tableName, this.granularity);
-
-        // Logical table must be in the logical table dictionary
-        this.table = bardConfigResources.getLogicalTableDictionary().get(tableId);
-        if (this.table == null) {
-            LOG.debug(TABLE_UNDEFINED.logFormat(tableName));
-            throw new BadApiRequestException(TABLE_UNDEFINED.format(tableName));
-        }
+        this.table = generateTable(tableName, this.granularity, bardConfigResources.getLogicalTableDictionary());
```
QubitPi: `ApiRequestImpl` has the same logic to generate `LogicalTable`. So the logic is moved to there

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/apirequest/ApiRequestImplSpec.groovy
```diff
@@ -28,6 +26,7 @@ class ApiRequestImplSpec extends Specification {
         String tableName = "tableName"
         Granularity granularity = Mock(Granularity)
         LogicalTableDictionary logicalTableDictionary = Mock(LogicalTableDictionary)
+        logicalTableDictionary.getNames() >> [tableName]
```
QubitPi: Need to do a tableName scan before `LogicalTable` lookup to make sure tableName exist.

michael-mclawhorn:
I'd like to suggest that instead we enhance `LogicalTableDictionary`.  It already has finders by Metric and Dimension, we could add a findByName.  I think that would move the responsibility onto the dictionary rather than into the client code.

Issue #603, opened 2017-12-19
Add DataApiRequest Factory in place of direct construction


michael-mclawhorn:
Preprocessing config and request parameters requires a lifecycle before constructing the DataApiRequest.
Also, custom implementations of DataApiRequest cannot be injected without a factory.

PR #604, opened 2017-12-19
DataApiRequestFactory Addresses: https://github.com/yahoo/fili/issues/603

michael-mclawhorn:
This change enables
* pre processing of arguments used to build DataApiRequest
* custom DataApiRequestImpl injection


fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -632,6 +636,17 @@ protected DruidFilterBuilder buildDruidFilterBuilder() {
         return new DefaultDruidFilterBuilder();
     }
 
+    /**
+     * Creates an factory that constructs DataApiRequests
+     * .
+     * Constructs a {@link DefaultDataApiRequestFactory} by default.
+     *
+     * @return An object to build Druid filters from API filters
```
archolewa: This return clause isn't accurate.

PR #606, opened 2017-12-22
Annotate Functional Interface

QubitPi:
Add `@FunctionalInterface` annotation to all functional interfaces.

It is not mandatory to use it, but it is best practice to use it with functional interface to avoid addition of extra methods accidentally. If the interface is annotated with `@FunctionalInterface` annotation and we try to have more than one abstract method, it throws compiler error.

The major benefit of functional interface is that we can use lambda expressions to instantiate them and avoid using bulky anonymous class implementation. Since there is only one abstract function in the functional interfaces, there is no confusion in applying the lambda expression to the method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ Current
 
 ### Added:
 
+- [Annotate Functional Interface](https://github.com/yahoo/fili/pull/606)
+    * Add `@FunctionalInterface` annotation to all functional interfaces.
+
 - [Implement LookupLoadTask](https://github.com/yahoo/fili/pull/620)
     * Add capability for Fili to check load statuses of Druid lookups.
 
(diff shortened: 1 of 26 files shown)


QubitPi:
@tarrantzhang Thank you! I've added to them all except for `ConfigPhysicalTable`, `TaggedDimensionField`, and `ExtractionFunctionDimension`. They are actually not functional interface

cdeszaq:
@QubitPi I think this just needs a little bit of manual intervention to resolve the conflicts and it should be ready to merge.

QubitPi:
@cdeszaq I don't see conflicts. Could you point me to the conflict location?

QubitPi:
@cdeszaq I think all conflicts are resolved 😃

Issue #608, opened 2017-12-30
Lift required override on deprecated method in MetricLoader


QubitPi:
Downstream projects that want to implement [`loadMetricDictionary`](https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/MetricLoader.java#L29) get compiler error unless they are forced to implement the [deprecated version](https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/MetricLoader.java#L21). They should be able to implement the new version without worrying about the old version.

PR #609, opened 2017-12-30
Lift required override on deprecated method in MetricLoader

QubitPi:
Address issue https://github.com/yahoo/fili/issues/608

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,6 +122,10 @@ Current
 
 ### Changed:
 
+- [Lift required override on deprecated method in MetricLoader](https://github.com/yahoo/fili/pull/609)
+    * Add default implementation to deprecated `loadMetricDictionary` in `MetricLoader` so that downstream projects are
+      able to implement the new version without worrying about the deprecated version.
+
 - [Added DataApiRequestFactory layer](https://github.com/yahoo/fili/issues/603)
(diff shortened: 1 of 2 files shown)


archolewa:
I'm a little late to the party, but I think it's a really bad idea to "deprecate" a method and then have it throw an unsupported operation exception. If a method now makes *such little sense* that we can't support it, then we should just delete it. All else being equal, compile errors >>>>>>> runtime errors. 

Especially since this is a trivially easy thing to fix for people using the old method: add an extra parameter to the method signature. The implementation doesn't need to be touched at all.

archolewa:
*Especially* since the other implementation is a default!

Which means that if someone adds a new interface, their class will compile, and then explode at runtime by default.

archolewa:
Frankly, I think we should just make the "new" method no longer default and delete the old method if we don't want to make people provide an implementation of the deprecated method.

QubitPi:
According to doc, deprecated method can only be removed after a new release.

I'm OK if you open a new PR or roll back this.

archolewa:
Eh, I don't think I care enough to go reverting code. This is fairly minor, and our codebase isn't widely used.

I do think we should clean this up for the next minor rev though, which we're probably past due for (current version: 0.9.118).

PR #611, opened 2018-01-10
Implement TimeFormatExtractionFunction

QubitPi:
Enable [`TimeFormatExtractionFunction`](http://druid.io/docs/0.10.1/querying/dimensionspecs.html#time-format-extraction-function) in Fili so that API users could interact with Druid using `TimeFormatExtractionFunction` through Fili.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Current
 
 ### Added:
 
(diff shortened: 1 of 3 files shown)


QubitPi:
Thanks @michael-mclawhorn and @garyluoex . I've added the tests in `TimeFormatExtractionFunctionSpec`

Issue #612, opened 2018-01-12
show=none is not being respected


michael-mclawhorn:
Show clause = none should suppress the dimension from being expressed at all.  Instead it's retaining the PK.

QubitPi:
Addressed by https://github.com/yahoo/fili/pull/613

PR #613, opened 2018-01-12
Corrected behavior of show=none to support silent dimension columns.

michael-mclawhorn:
Addresses: https://github.com/yahoo/fili/issues/612


michael-mclawhorn:
@archolewa

PR #618, opened 2018-01-30
Parameterized metric design

michael-mclawhorn:
The proposal for parameterized metrics.
Please review.
This proposal may close as soon at Monday, Feb 5th, 2018 if all feedback has been successfully responded to.

### DIFF
--- a/rfc/parameterizedMetrics.md
+++ b/rfc/parameterizedMetrics.md
@@ -25,17 +25,19 @@ The structure of a parameterized metric:
 
 ```
 metricName(parameterName=value)
+
+metricName [ '(' parameterName=value [ ',' parameterName=value ]*  ')' ]?
 ```
 
 metricName -> an api metric name which should appear as a resource at the `/v1/metrics/metricName` endpoint.
-
-Balanced outer parentheses are the signal that a metric is to be processed as parameterized metric.  Imbalanced parentheses or unescaped internal parentheses will probably result in request errors.
 
 parameterName -> an identifier for a parameter supported by this metric.  Identifier names should follow the same format rules as api resource identifiers.
 
 value -> a value to be used for creating a modified version of baseMetric.  URL encoding may be required if some of the characters in the value contain special characters.  Escaping values containing ampersand is mandatory and it may also be necessary to encode values containing comma or parentheses depending on implementation.
 
-If multiple parameter names are used for this dimension they should be joined by commas with no intermediate spaces (similar to the filter value syntax)
+Values containing characters delimiters such as comma or parantheses must have them nested in other scoping characters such as quotes, square brackets or parantheses.
+
+If multiple parameter names are used for this metric they should be joined by commas with no intermediate spaces (similar to the filters value syntax)
 
 i.e.
 
@@ -43,28 +45,88 @@ i.e.
 metricName(parameterName=value,otherParameter=otherValue)
 ```
 
+
 Response schema
 ---------------
 
 Unless aliasing is supported and used (see below), responses will use the full string of the metric expression as the column name for values resulting from parametrized metric evaluation.
 
+Example query:
+```json
+...&amp;metrics=revenue,netRevenue,revenue(currency=USD)
+```
+
+Example response:
+```json
+{
+  "rows": [
+    {
+      "dateTime": "2018-02-01 00:00:00.000",
+      "mediaType|id": "-2",
+      "currency|id": "AUD",
+      "revenue": "100",
+      "revenue|currency": "AUD",
+      "netRevenue": "50",
+      "netRevenue|currency": "CAD",
+      "revenue(currency=USD)": "120",
+      "revenue(currency=USD)|currency": "USD"
+    }
+}
+```
 
 Reserved parameter names
 ------------------------
 
-Two optional features which may be supported in some fili implementations are 'filter' and 'as'.  These parameter names should be treated as reserved when configuring metrics.
+Two optional features which may be supported in some fili implementations are 'filters' and 'as'.  These parameter names should be treated as reserved when configuring metrics.
 
 Dynamic aggregation filters, if supported, use the parameter name `filters`. Values will be parsed according to the same rules as the entries in the `filters` query parameter.  The expected behavior of the filters parameter is that rows in the base fact will not be aggregated unless the included filters is true.  This is potentially costly in performance and global filters should always be preferred to dynamic filters, if possible.
 
 The `as` parameter name, if supported, will create an alias for the dynamic metric which will be used in the schema of the response as well as in the having clause of the request.
 
-Example:
+
+Aliasing with 'as'
+------------------
+
+Aliasing metrics provides a shorthand for the response and for the having clause, giving a parameterized metric expression an alternate, readable, name.
+
+Example request:
 
 ```
 ?metrics=users(filters=dim1|id-in['a','b'],as=DimOneInAOrB)&having=DimOneInAOrB-gt[50]
 ```
 
 This would create a variation of the standard `users` aggregation which only counted users from records where the dimension `dim1` had values "a" or "b", and would display result rows only which for which this metric had at least 50 users.
+
+Example response:
+```json
+{
+  "rows": [
+    {
+      "dateTime": "2018-02-01 00:00:00.000",
+      "DimOneInAOrB": "100"
+    }
+}
+```
+
+Aliasing does not work WITHIN the metric clause itself.  Parameterized metrics can be defined in other blocks, but aliases cannot be parameterized themselves. 
+
+Illegal:
+```
+?metrics=metric1(filters=dim1|id-in['a','b'],as=aliasName),aliasName
+```
+
+Also illegal:
+```
+?metrics=metric1(as=aliasName),&amp;having=aliasName(filters=dim1|id-in['a'])
+```
+
+Duplicate Metrics
+-----------------
+
+Currently metrics with identical meanings and representations within an API Request produce a 400 error of the form 'Duplicate metric(s) are not allowed in the API request.'  Parameterized metrics may allow the creation of metrics with the same meaning but different definitions. (Such as the same metric expression with distinct aliases or two terms with parameters in a different order but the same pairs).  These may or may not be considered 'duplicates' as an implementation detail.  
+
+Regardless, the actual string label used to define a metric, and not a 'normalized' form should be preserved in the response schema.
+
 
 Parameterized Metric Metadata
 =============================
@@ -102,7 +164,7 @@ Recommended types to be implemented:
 
 `identifier`: The parameter value will be parsed as an identifier string.
 
-`filter`: The parameter accepts values parsed identically to the filters query parameter.
+`filters`: The parameter accepts values parsed identically to the filters query parameter.
 
 `dimension`: The parameter accepts values from a dimension.
 
@@ -116,9 +178,54 @@ Proposed future type extensions:
 
 `interval`: The parameter accepts intervals as used within dateTime query parameter.
 
+Parameter Identifiers
+---------------------
+
+Parameters need to have the same constraints as resource names such as those at the metrics endpoints. By default any java variable identifier style (non numeric alpha followed by alphanumeric, '_' if necessary).  Mix case is encouraged but not mandatory. 
+
+Parameter keywords are presumed to be case sensitive, although fili implementations can soften this restriction.
+
 Reserved Parameter Types
 ------------------------
 
-The `filters` reserved parameter should have an implicit type of `filter`
+The `filters` reserved parameter should have an implicit type of `filters`
 
 The `as` reserved parameter should have an implicit type of `identifier`
+
+Parameter Metadata
+------------------
+
+The dimension parameter type: 
+
+```json
+{
+  type: dimension, 
+  dimensionName: [DIMENSION_NAME],
+  defaultValue: [DEFAULT_VALUE] 
+}
+```
+
+**dimensionName**:  Mandatory.  This must be a dimension available at /v1/data/dimensions/DIMENSION_NAME
+
+This dimension does not need to be attached to any tables in the system.
+
+**defaultValue**:  Optional.  "" if not default value is defined.
+
+This value should usually be defined with the same behavior as the parameter being omitted.
+    
+Error Handling
+==============
+
+Based on the current requirements, support for anything but the simplest set of parameters will require a grammar.  The following errors should be generally expected:
+
+**"Unparseable Metrics"**:  If the grammar or tokenizing logic cannot process the metrics expression at all.
+
+**"Invalid Metric Name"**: If the metric name before the parameters cannot be associated with a base metric name.
+
+**"Unsupported Parameter"**: If the parameter name in a parameter list doesn't exist on that metric.
+
+**"Invalid Parameter Value"**: If the parameter value doesn't match the parsing rules for the type of the parameter.  
+
+**"Duplicate Metric"**: If an 'identical' metric is defined on the metrics list.
+
+**"Unexpected Having Metric"**: If a metric described in the having expression isn't represented in the metrics section.
--- a/rfc/parameterizedMetrics.md
+++ b/rfc/parameterizedMetrics.md
@@ -52,7 +52,7 @@ Response schema
 Unless aliasing is supported and used (see below), responses will use the full string of the metric expression as the column name for values resulting from parametrized metric evaluation.
 
 Example query:
-```json
+```
 ...&metrics=revenue,netRevenue,revenue(currency=USD)
 ```
 
@@ -70,7 +70,9 @@ Example response:
       "netRevenue|currency": "CAD",
       "revenue(currency=USD)": "120",
       "revenue(currency=USD)|currency": "USD"
+      
     }
+    ]
 }
 ```
 
@@ -100,11 +102,12 @@ This would create a variation of the standard `users` aggregation which only cou
 Example response:
 ```json
 {
-  "rows": [
-    {
-      "dateTime": "2018-02-01 00:00:00.000",
-      "DimOneInAOrB": "100"
-    }
+    "rows": [
+        {
+            "dateTime": "2018-02-01 00:00:00.000",
+            "DimOneInAOrB": "100"
+        }
+    ]
 }
 ```
 
@@ -133,15 +136,15 @@ Parameterized Metric Metadata
 
 The metrics endpoint (and the corresponding serialization in the fullView format on the tables endpoint) will now contain an optional field for parameters.
 
-```
+```json
 { 
-    name:"grossRevenue",
-    type:"money"
-    parameters: {
-        currency: {  
-            type: dimension, 
-            dimensionName: "currencyUnits",
-            defaultValue:"billing" 
+    "name": "grossRevenue",
+    "type": "money",
+    "parameters": {
+        "currency": {  
+            "type": "dimension", 
+            "dimensionName": "currencyUnits",
+            "defaultValue":"billing" 
         }
     }
 }
@@ -199,9 +202,9 @@ The dimension parameter type:
 
 ```json
 {
-  type: dimension, 
-  dimensionName: [DIMENSION_NAME],
-  defaultValue: [DEFAULT_VALUE] 
+  "type": "dimension", 
+  "dimensionName": "[DIMENSION_NAME]",
+  "defaultValue": "[DEFAULT_VALUE]" 
 }
 ```
 
--- a/rfc/parameterizedMetrics.md
+++ b/rfc/parameterizedMetrics.md
@@ -29,11 +29,11 @@ metricName(parameterName=value)
 metricName [ '(' parameterName=value [ ',' parameterName=value ]*  ')' ]?
 ```
 
-metricName -> an api metric name which should appear as a resource at the `/v1/metrics/metricName` endpoint.
+**metricName** -> an api metric name which should appear as a resource at the `/v1/metrics/metricName` endpoint.
 
-parameterName -> an identifier for a parameter supported by this metric.  Identifier names should follow the same format rules as api resource identifiers.
+**parameterName** -> an identifier for a parameter supported by this metric.  Identifier names should follow the same format rules as api resource identifiers.
 
-value -> a value to be used for creating a modified version of baseMetric.  URL encoding may be required if some of the characters in the value contain special characters.  Escaping values containing ampersand is mandatory and it may also be necessary to encode values containing comma or parentheses depending on implementation.
+**value** -> a value to be used for creating a modified version of baseMetric.  URL encoding may be required if some of the characters in the value contain special characters.  Escaping values containing ampersand is mandatory and it may also be necessary to encode values containing comma or parentheses depending on implementation.
 
 Values containing characters delimiters such as comma or parantheses must have them nested in other scoping characters such as quotes, square brackets or parantheses.
 
@@ -53,7 +53,7 @@ Unless aliasing is supported and used (see below), responses will use the full s
 
 Example query:
 ```json
-...&amp;metrics=revenue,netRevenue,revenue(currency=USD)
+...&metrics=revenue,netRevenue,revenue(currency=USD)
 ```
 
 Example response:
@@ -81,7 +81,7 @@ Two optional features which may be supported in some fili implementations are 'f
 
 Dynamic aggregation filters, if supported, use the parameter name `filters`. Values will be parsed according to the same rules as the entries in the `filters` query parameter.  The expected behavior of the filters parameter is that rows in the base fact will not be aggregated unless the included filters is true.  This is potentially costly in performance and global filters should always be preferred to dynamic filters, if possible.
 
-The `as` parameter name, if supported, will create an alias for the dynamic metric which will be used in the schema of the response as well as in the having clause of the request.
+The `as` parameter name, if supported, will create an alias for the dynamic metric which will be used in the schema of the response as well as in the having and sort clauses of the request.
 
 
 Aliasing with 'as'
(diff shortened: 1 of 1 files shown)


rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,124 @@
...
+Standard metrics in Fili are configured at load time.  However if reporting requires many different similar caluclations, such as adjusting precision or converting values of different units into a shared aggregate, or applying custom metric filters, it may be impractical or even impossible to configure all the interesting report metrics.
+
+With parameterized metrics, open-ended variations may be supported for a given conceptual metric calculation, simplifying configuration and possibly unlocking otherwise impractical freedom in metric calculation.
+
+
+The structure of a parameterized metric:
+----------------------------------------
+
+```
+metricName(parameterName=value)
```
archolewa: We should make it clear here that this is considered the name of the metric, unless aliased, i.e. this exact same string needs to be used in the having clause.

Furthermore, is this allowed:

`metricName(parameterName=value,AS=metricName)`.

If it is, then we should sit down and think about the rules of aliasing. Consider:

`mertricName(parameterName=value,AS=metricName),metricName(paramaterName=value2,AS=metricName2)`.

Which `metricName` should the second metric expression refer to? The first expression, or the base metric?

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,124 @@
...
+
+Unless aliasing is supported and used (see below), responses will use the full string of the metric expression as the column name for values resulting from parametrized metric evaluation.
+
+
+Reserved parameter names
+------------------------
+
+Two optional features which may be supported in some fili implementations are 'filter' and 'AS'.  These parameter names should be treated as reserved when configuring metrics.
+
+Dynamic aggregation filters, if supported, use the parameter name `filter`. Values will be parsed according to the same rules as the entries in the `filters` query parameter.  The expected behavior of the filter parameter is that rows in the base fact will not be aggregated unless the included filter is true.  This is potentially costly in performance and global filters should always be preferred to dynamic filters, if possible.
```
archolewa: We should be consistent with our reserved keywords, either all lowercase or all uppercase. I personally vote for all lowercase, because that's slightly easier to type.

archolewa: I also think we should use `filters` rather than `filter`, for the sake of consistency.

michael-mclawhorn: the `filters` clause takes a list of filter expressions.  A single `filter` is what the filter clause would take.  Or at least that's what I was originally thinking.

I suppose there's no reason it couldn't take a list of filters.  I'll update accordingly.

rfc/parameterizedMetrics.md
michael-mclawhorn: How soon an alias becomes available is an interesting implementation problem.

It would be easiest to say that aliasing simply won't work between metrics.

rfc/parameterizedMetrics.md
michael-mclawhorn: This actually suggests a larger pattern that could be followed.  Specifically, we could have other query limiting terms be reserved.  reserve 'dateTime' and provide a default time based filter on aggregation.

michael-mclawhorn: Actually, there's a noisome problem, which is that the filters clause supports internal commas.  Which will conflict with the commas used to separate fields within the parameter list.  I'm not sure how to readably resolve that.

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,124 @@
+Parameterized Metric Support
+============================
+
+Parameterized metrics are a subset of dynamic request metrics.  Specifically they focus on modifying the calculation of metrics by altering values used in the calculation of existing metrics or mapping their values at aggregation time.
+
+The expected contract of parameterized metrics is as follows: Given a base metric with api name `metricName`, it can be suffixed with a procedure-call list of keyword parameter names and values.  These values will be used in the calculation of the resulting new metric.
+
```
QubitPi: When we talk about `new metric`, are we constraining result rows with parameters or transforming to a different metric? If it's the latter, do we need to configure metric dictionary or metric maker differently?

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,124 @@
...
+
+Reserved parameter names
+------------------------
+
+Two optional features which may be supported in some fili implementations are 'filter' and 'AS'.  These parameter names should be treated as reserved when configuring metrics.
+
+Dynamic aggregation filters, if supported, use the parameter name `filter`. Values will be parsed according to the same rules as the entries in the `filters` query parameter.  The expected behavior of the filter parameter is that rows in the base fact will not be aggregated unless the included filter is true.  This is potentially costly in performance and global filters should always be preferred to dynamic filters, if possible.
+
+The `AS` parameter name, if supported, will create an alias for the dynamic metric which will be used in the schema of the response as well as in the having clause of the request.
```
QubitPi: Is `AS` essentially the same as dimension lookups?

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,124 @@
...
+
+Example:
+
+```
+?metrics=users(filter,dim1|id-in['a','b'],AS=DimOneInAOrB)&having=DimOneInAOrB-gt[50]
+```
+
+This would create a variation of the standard `users` aggregation which only counted users from records where the dimension `dim1` had values "a" or "b", and would display result rows only which for which this metric had at least 50 users.
```
QubitPi: Will the response show metric as `DimOneInAOrB` instead of `users`?

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+}
+```
+
+Parameters will be a JSON object whose keys are parameter names supported on this metric.
+
+The values of a parameter entry will contain a mandatory field 'type'.
+
+In this example, the `dimension` supports fields `dimensionName` and `defaultValue`.  Dimension name is the name of a dimension whose values are legal for this parameter.
```
QubitPi: Is `dimension` a String or a [DimensionSpec](http://druid.io/docs/latest/querying/dimensionspecs.html)?

cythrawll:
After talking to @michael-mclawhorn, I was suggesting we flesh out the metadata that comes with each parameter type a bit more.  (We know for example `type: dimension` will have corresponding `dimensionName` key, what kind of meta data do we expect for each of the other types).  Other than that 👍

rfc/parameterizedMetrics.md
jkusa: Speaking of case sensitivity, sounds like you are suggesting parameter names are case sensitive. Can we call that out.

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,124 @@
...
+parameterName -> an identifier for a parameter supported by this metric.  Identifier names should follow the same format rules as api resource identifiers.
+
+value -> a value to be used for creating a modified version of baseMetric.  URL encoding may be required if some of the characters in the value contain special characters.  Escaping values containing ampersand is mandatory and it may also be necessary to encode values containing comma or parentheses depending on implementation.
+
+If multiple parameter names are used for this dimension they should be joined by commas with no intermediate spaces (similar to the filter value syntax)
+
+i.e.
+
+```
+metricName(parameterName=value,otherParameter=otherValue)
```
jkusa: Is there a reason white space cannot be the delimiter?

`metricName(parameterName=value otherParameter=otherValue)`

I think it might read better and you don't have to worry about it as a value.

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,124 @@
...
+`date`: The parameter accepts values that parse as the DateTime part of the interval used within dateTime query parameter.
+
+`interval`: The parameter accepts intervals as used within dateTime query parameter.
+
+Reserved Parameter Types
+------------------------
+
+The `filter` reserved parameter should have an implicit type of `filter`
+
+The `AS` reserved parameter should have an implicit type of `identifier`
```
jkusa: Can we have a section on error handling? Incorrect param name, bad value (type), syntax error, etc.

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,124 @@
...
+Two optional features which may be supported in some fili implementations are 'filter' and 'AS'.  These parameter names should be treated as reserved when configuring metrics.
+
+Dynamic aggregation filters, if supported, use the parameter name `filter`. Values will be parsed according to the same rules as the entries in the `filters` query parameter.  The expected behavior of the filter parameter is that rows in the base fact will not be aggregated unless the included filter is true.  This is potentially costly in performance and global filters should always be preferred to dynamic filters, if possible.
+
+The `AS` parameter name, if supported, will create an alias for the dynamic metric which will be used in the schema of the response as well as in the having clause of the request.
+
+Example:
+
+```
+?metrics=users(filter,dim1|id-in['a','b'],AS=DimOneInAOrB)&having=DimOneInAOrB-gt[50]
```
cythrawll: syntax error. should be `filter=` not `filter,`

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,124 @@
...
+i.e.
+
+```
+metricName(parameterName=value,otherParameter=otherValue)
+```
+
+Response schema
+---------------
+
+Unless aliasing is supported and used (see below), responses will use the full string of the metric expression as the column name for values resulting from parametrized metric evaluation.
```
jkusa: Can we have an example response?

rfc/parameterizedMetrics.md
jkusa: How will duplicate metrics with the same param values be handled?

rfc/parameterizedMetrics.md
jkusa: Will the order of multiple key-value pairs be normalized across multiple metrics in the response? Or will the response order will be based on the request order?

rfc/parameterizedMetrics.md
michael-mclawhorn: We're building a new Logical Metric instance using a transformed TDQ and result set mapper (and name)

michael-mclawhorn: Although I'm trying to restrict this document to API concerns for the most part.

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,124 @@
...
+The structure of a parameterized metric:
+----------------------------------------
+
+```
+metricName(parameterName=value)
+```
+
+metricName -> an api metric name which should appear as a resource at the `/v1/metrics/metricName` endpoint.
+
+Balanced outer parentheses are the signal that a metric is to be processed as parameterized metric.  Imbalanced parentheses or unescaped internal parentheses will probably result in request errors.
```
will-lauer: This seems like a meaningless statement to me. There is (or should be) a grammar. Anything not conforming is an error. Balancing of parens is not the signal that the expression is good.

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,124 @@
...
+
+parameterName -> an identifier for a parameter supported by this metric.  Identifier names should follow the same format rules as api resource identifiers.
+
+value -> a value to be used for creating a modified version of baseMetric.  URL encoding may be required if some of the characters in the value contain special characters.  Escaping values containing ampersand is mandatory and it may also be necessary to encode values containing comma or parentheses depending on implementation.
```
will-lauer: I suspect that quoting will be necessary as well, independent of URL encoding. For example, it should be perfectly reasonable to have the metric `pageViews(as="my pageviews (some cool description, and other stuff)", filter="somefilter")`.

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,124 @@
...
+
+If multiple parameter names are used for this dimension they should be joined by commas with no intermediate spaces (similar to the filter value syntax)
```
will-lauer: Using the term `dimension` here is slightly confusing since we aren't talking about dimensions. I would just say `If multiple parameters are present, they should be separated by commas with no additional white space`.

rfc/parameterizedMetrics.md
will-lauer: I would expect it to be handled exactly the same as a query requesting "pageviews,pageviews". If that doesn't error, neither should this. A more interesting question is "pageviews(as=pv1),pageviews(as=pv2)", which I think should be successful, even though its the same value being generated.

rfc/parameterizedMetrics.md
will-lauer: For purposes of _validation_, we should consider `foo(a=x,b=y)` to be the same as `foo(b=y,a=x)`. For display purposes, we should consider these unique, and preserve the ordering of parameters that was requested by the user. That is, the metric name is an exact string match. But users should *NOT* assume that metric order itself is preserved in the response.

rfc/parameterizedMetrics.md
will-lauer: No, this is the equivalent of AS in a SQL statement. That is `select foo as a, bar as b from mytable` returns two columns, `a` and `b`. There is no dimension lookup here. Its just a simple renaming.

rfc/parameterizedMetrics.md
will-lauer: This example highlights the need for quoting of arguments. I suspect we might want to make quotes mandatory to make the grammar sane. I think we need the following to be a valid version of the above query:
```
?metrics=users(filter="dim1|id-in['a','b']",as=DimOneInAOrB)&having=DimOneInAOrB-gt[50]
```

rfc/parameterizedMetrics.md
will-lauer: yes

rfc/parameterizedMetrics.md
will-lauer: I think singular (filter) makes more sense here that plural (filters), even though the filter expression can be composed of multiple components that are AND'd together. Logically, I like druid's model, where filter is always singular.

rfc/parameterizedMetrics.md
jkusa: I am fine with returning `400` as it does today:

`Duplicate metric(s) are not allowed in the API request..`

Let's just call it out that the same is true with param values.

jkusa: I hope this RFC can focus on the parameterized metric spec.  Details for `as`, `filter(s)` and other built in params can be their own RFC.

rfc/parameterizedMetrics.md
michael-mclawhorn: should have been 'metric' not 'dimension'

rfc/parameterizedMetrics.md
michael-mclawhorn: singular filter aligns with a tree expression.  If we ever support trees of constraints then we can support singlular filter.

rfc/parameterizedMetrics.md
michael-mclawhorn: The filter parser that would be needed for this will create a nesting scope around '[ ]' that would disambiguate the ','.

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,124 @@
...
+Parameterized Metric Metadata
+=============================
+
+The metrics endpoint (and the corresponding serialization in the fullView format on the tables endpoint) will now contain an optional field for parameters.
+
+```
+{ 
+    name:"grossRevenue",
+    type:"money"
+    parameters: {
```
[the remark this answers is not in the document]
michael-mclawhorn: we were going to use an object to create a clear key uniqueness.

rfc/parameterizedMetrics.md
michael-mclawhorn: in this case it's an API dimension, I'll clarify that.

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,231 @@
...
+```
+
+Reserved parameter names
+------------------------
+
+Two optional features which may be supported in some fili implementations are 'filters' and 'as'.  These parameter names should be treated as reserved when configuring metrics.
+
+Dynamic aggregation filters, if supported, use the parameter name `filters`. Values will be parsed according to the same rules as the entries in the `filters` query parameter.  The expected behavior of the filters parameter is that rows in the base fact will not be aggregated unless the included filters is true.  This is potentially costly in performance and global filters should always be preferred to dynamic filters, if possible.
+
+The `as` parameter name, if supported, will create an alias for the dynamic metric which will be used in the schema of the response as well as in the having clause of the request.
```
jkusa: and `sort`?

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,231 @@
...
+Values containing characters delimiters such as comma or parantheses must have them nested in other scoping characters such as quotes, square brackets or parantheses.
+
+If multiple parameter names are used for this metric they should be joined by commas with no intermediate spaces (similar to the filters value syntax)
+
+i.e.
+
+```
+metricName(parameterName=value,otherParameter=otherValue)
+```
+
```
jkusa: Can we call out that all params where a metric is specified supports this new syntax?

 - `metrics`
 - `having`
- `sort`

michael-mclawhorn:
@archolewa @QubitPi  please rereview as you're able.

rfc/parameterizedMetrics.md
michael-mclawhorn: done, below

QubitPi:
@michael-mclawhorn Looks perfect to me 👍

michael-mclawhorn:
@archolewa Did I answer all your concerns?  If so, please approve.

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,236 @@
...
+metricName [ '(' parameterName=value [ ',' parameterName=value ]*  ')' ]?
+```
+
+**metricName** -> an api metric name which should appear as a resource at the `/v1/metrics/metricName` endpoint.
+
+**parameterName** -> an identifier for a parameter supported by this metric.  Identifier names should follow the same format rules as api resource identifiers.
+
+**value** -> a value to be used for creating a modified version of baseMetric.  URL encoding may be required if some of the characters in the value contain special characters.  Escaping values containing ampersand is mandatory and it may also be necessary to encode values containing comma or parentheses depending on implementation.
+
+Values containing characters delimiters such as comma or parantheses must have them nested in other scoping characters such as quotes, square brackets or parantheses.
```
cdeszaq: I think this is getting at "what is a valid character for a value", which is a good one to address. It also raises the question of "how to I scope my value?" In a complex case of "passing a metric to another metric", it could get messy if there are layers of parameterization. For example, how would this pseudo-metric be validly represented?

```
metricA(
    parameterOfTypeMetric=metricB(
        parameterOfTypeInt=1,
        parameterOfTypeBool=false
    ),
    parameterOfTypeMetric=metricC,
    parameterOfTypeTimeGrain=day,
    parameterOfTypeKVMap={
        key1=value1,
        key2=value2
    }
)
```

If we want parens to be the "value bound", we should indicate that in the spec. In which case, anything inside parens is treated as the "value" of the param and can't be confused by the parser. That would turn our example into this:

```
metricA(
    parameterOfTypeMetric=(
        metricB(
            parameterOfTypeInt=1,
            parameterOfTypeBool=false
        )
    ),
    parameterOfTypeMetric=metricC,
    parameterOfTypeTimeGrain=day,
    parameterOfTypeKVMap=(
        {
            key1=value1,
            key2=value2
        }
    )
)

// Compressed
metricA(parameterOfTypeMetric=(metricB(parameterOfTypeInt=1,parameterOfTypeBool=false)),parameterOfTypeMetric=metricC,parameterOfTypeTimeGrain=day,parameterOfTypeKVMap=({key1=value1,key2=value2}))
```

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,236 @@
...
+    {
+      "dateTime": "2018-02-01 00:00:00.000",
+      "mediaType|id": "-2",
+      "currency|id": "AUD",
+      "revenue": "100",
+      "revenue|currency": "AUD",
+      "netRevenue": "50",
+      "netRevenue|currency": "CAD",
+      "revenue(currency=USD)": "120",
+      "revenue(currency=USD)|currency": "USD"
```
cdeszaq: It's not clear to me where these response fields that _look_ like dimension field values describing the metric are coming from. Are these coming from a dimension clause that isn't shown in the example? Or are they multiple columns coming from a single metric?

I'm guessing there's a lot of behavior bundled into the Dimension parameter type that isn't part of this spec, and/or I'm simply forgetting the ability for metrics to explode into multiple columns in a response (similar to how Dimensions can).

Is that new behavior, and part of this spec? Existing behavior that I'm forgetting about? New behavior _separate_ from this spec?

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,236 @@
...
+      
+    }
+    ]
+}
+```
+
+Reserved parameter names
```
cdeszaq: Are there any other parameter names we want to cordon off as "reserved words" for future use? It's much harder to add new parameters defined into the system later. To put it a different way: Are there any other parameters that we can foresee applying to _all_ metrics?

Or perhaps we want to reserve a "prefix" as a namespace for future core additions?

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,236 @@
...
+}
+```
+
+Parameters will be a JSON object whose keys are parameter names supported on this metric.
+
+The values of a parameter entry will contain a mandatory field 'type'.
+
+In this example, the `dimension` supports fields `dimensionName` and `defaultValue`.  Dimension name is the name of a dimension whose values are legal for this parameter.  
+
+The `defaultValue` field is an optimization for UI clients to provide a 'recommended' value for the parameter, if appropriate.  `defaultValue`, if used, is recommended to be equivalent to not supplying a value for this parameter at all.
```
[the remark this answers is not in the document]
cdeszaq: As a follow-on to this, are parameters always considered optional? Is there such a thing as a metric that _requires_ a parameter to be specified?

I can see the case for not allowing required parameters right now and allowing for them later, but it wasn't clear to me based on this spec, so I wanted to ask.

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,236 @@
...
+
+Proposed future type extensions:
+
+`enum`: The parameter accepts values from an included list of values.
+
+`number`: The parameter accepts simple (signed) numeric values.
+
+`date`: The parameter accepts values that parse as the DateTime part of the interval used within dateTime query parameter.
+
+`interval`: The parameter accepts intervals as used within dateTime query parameter.
```
cdeszaq: Time grain would be another highly-likely type that might be worth calling out.

cdeszaq:
Sorry for the late review on this. Love the idea and direction, just some details to cover.

rfc/parameterizedMetrics.md
michael-mclawhorn: I wasn't anticipating passing metrics to metrics in this fashion.
It definitely complicates parsing rules.

rfc/parameterizedMetrics.md
michael-mclawhorn: I think these cases need to be considered when evaluating parameterized metrics.
If we define a syntax that doesn't handle the very next two things we'll want to do with it, it's not a good spec.

rfc/parameterizedMetrics.md
michael-mclawhorn: Metrics having multiple columns has been implemented in at least one custom Fili.
It looks like no one has pulled a reference implementation up into Fili, but it was the driving use case for needing to support a parameterized metric with navi.

The sense of the behavior is that currency binds a dimension to the evaluation of the metric, in this case, a units dimension.  I should get that into an issue/rfc for general dissection. 

Much earlier discussions around complex metrics imagined them as json blobs inside the json response, but this doesn't play nicely with CSV reporting.  Multicolumn with field operators does make nice tables and so that's what was built, with an especial focus on [navi](https://github.com/yahoo/navi) integration.

Whether expanded metrics are supported or not, the name of the metric in resulting json needs to be defined.

rfc/parameterizedMetrics.md
michael-mclawhorn: I like this.  So 'core:X' would be reserved and 'X' would be the default alias for that namespaced term?  I like that. It makes things flexible and could let us strip the filter and as cases out of this spec.

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,236 @@
...
+
+Also illegal:
+```
+?metrics=metric1(as=aliasName),&amp;having=aliasName(filters=dim1|id-in['a'])
+```
+
+Duplicate Metrics
+-----------------
+
+Currently metrics with identical meanings and representations within an API Request produce a 400 error of the form 'Duplicate metric(s) are not allowed in the API request.'  Parameterized metrics may allow the creation of metrics with the same meaning but different definitions. (Such as the same metric expression with distinct aliases or two terms with parameters in a different order but the same pairs).  These may or may not be considered 'duplicates' as an implementation detail.
```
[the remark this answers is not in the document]
michael-mclawhorn: So, 'by default metrics distinct by uri string representation will be considered as distinct metrics for validation purposes'?

rfc/parameterizedMetrics.md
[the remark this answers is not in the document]
michael-mclawhorn: My thinking had been to make required parameters not a thing by any default mechanism and instead have all parameters have a default value.

rfc/parameterizedMetrics.md
```diff
@@ -0,0 +1,236 @@
...
+  "dimensionName": "[DIMENSION_NAME]",
+  "defaultValue": "[DEFAULT_VALUE]" 
+}
+```
+
+**dimensionName**:  Mandatory.  This must be a dimension available at /v1/data/dimensions/DIMENSION_NAME
+
+This dimension does not need to be attached to any tables in the system.
+
+**defaultValue**:  Optional.  "" if not default value is defined.
```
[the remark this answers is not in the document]
michael-mclawhorn: I'll address this, but the idea was tthat:
"metric(param1=defaultValue)" to have exactly the same meaning as "metric()" and as "metric"
'as' would be an exception to '""' is the meaning if no default is defined, since the actual default is the fully qualified metric expression.

PR #628, opened 2018-02-15
RFC - Enable PhysicalTable-Specific Time Limit in PartitionCompositeTable

QubitPi:


rfc/unionPartitionAvailability.md
```diff
@@ -0,0 +1,50 @@
+# Enable union of partition availability
+
+This RFC implements [Github Issue 627](https://github.com/yahoo/fili/issues/627).
+
+## Goals
+
+Enable query to multiple datasources with non-intersecting data availabilities to see an union view of availabilities, 
+instead of an intersection.
+
```
michael-mclawhorn: Not exactly.

Partition availability already has a concept of 'determine participating tables' and 'intersect the availability of participating tables'.  The behavior we should be thinking of is to have the underlying tables limiting the times in which they participate.

So if DS1 in included for filter set X
and DS 2 is also included for filter set X

If DS2 is defined as starting on interval 2018/FUTURE
then the availability is

Interval: THE PAST/2018
Intersect: all tables in range (DS1)
Interval 2018/The Future
Intersect: all tables in range (DS1, DS2)

michael-mclawhorn: So in the case of say all of YAMP and BRX together:

2016/2017
INTERSECT: BR-US, YAMP-US
INTERSECT: 2017/2018
BR-US, YAMP-US, YAMP-EMEA, YAMP-AIPAC, YAMP-AUNZ

rfc/unionPartitionAvailability.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+When we ask for data of M in 2017, it is legitimate to expect data from DS1 to be returned. But current behavior is
+[returning a response with no data and missing interval of 2017 notified to client](https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java#L92).
+
+What needs to be fixed is that, instead of returning no data, data of 2017 from DS1 shall be returned.
+
+## Strategy
+1. Turn [PartitionAvailability](https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/table/availability/PartitionAvailability.java)
+to an abstract class.
```
michael-mclawhorn: Partition availability should probably not be an abstract class.  I think this should probably be a subclass of Partition, or a sibling class, refactoring as necessary.

rfc/unionPartitionAvailability.md
QubitPi: @michael-mclawhorn If I understand you correctly, we are essentially filtering table columns and then intersecting, instead of unioning all availabilies, right?

QubitPi:
Based on @michael-mclawhorn 's idea, I will change the design in the following way: 

Do not change the interface of PhysicalTable, but try to change the PartitionCompositeTable so that each participating partition has a notion of "starting time" for availability. The decision on returning "missing interval" then becomes:

if there is no data in this interval && this interval is *before* the "starting time" -> don't mark this as missing-data interval
if there is no data in this interval && this interval is *after* the "starting time" -> mark this as missing-data interval

michael-mclawhorn:
I'd suggest making this a subclass of Partition Physical Table simply because the contract will be sufficiently more complex than the current one..

Also, remember get availability doesn't know which intervals are being requested.  So the starting point might be the union of all available intervals and then 'subtracting' intervals which are missing per your described contract.

QubitPi:
@efronbs My understanding is that `DataSourceMetadataLoadTask` shall auto-detect the datasource lose and update segment metadata info and, thus, availability info. 

But you are welcome to add other enhancement if you have.

PR #632, opened 2018-02-20
Minified version of tables full view test

efronbs:
When the `ExpectedTablesFullViewEndpointSpec` test fails, generating the error logs will often cause an `OutOfMemoryException` and all error logs (including the power assert) will be lost. This makes debugging issues that this test catches really difficult, because you can't see the differences between the expected and actual responses.

This PR adds a minified version of the `ExpectedTablesFullViewEndpointSpec`. Hopefully, when the `ExpectedTablesFullViewEndpointSpec` test fails, the minified version will also fail, and the error logs will NOT be lost, making changes to the TablesServlet easier to debug.


efronbs:
While I may have not done this properly, I have tried increasing the JVM heap size locally which does not seem to have any effect.

I have tried printing an error message, which does not work (I assume because heap space is exceeded you can't allocated space for a buffer in the same process), so it is likely that logging the power assert will not work either, though I will try that and if it works I will implement that change. 

This change is meant for usability and not to harden our testing suite, so if there are no other reasonable options I think adding the second test is better than weakening our current test.

PR #633, opened 2018-02-24
Replace Old gmaven with New gmavenplus

QubitPi:
**gmaven** is a [very old plugin](https://mvnrepository.com/artifact/org.codehaus.gmaven/gmaven-plugin) that has stopped its development about 7 years ago.

**gmavenplus**  is a rewrite of gmaven and is [now being actively maintained](https://mvnrepository.com/artifact/org.codehaus.gmavenplus/gmavenplus-plugin).

This PR switch from using gmaven to adapting gmavenplus.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,6 +84,7 @@
         <version.lucene>6.6.0</version.lucene>
         <version.jackson>2.6.2</version.jackson>
         <version.groovy>2.4.5</version.groovy>
+        <version.gmavenplus.plugin>1.6</version.gmavenplus.plugin>
         <version.guava>21.0</version.guava>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -84,6 +84,7 @@
         <version.lucene>6.6.0</version.lucene>
         <version.jackson>2.6.2</version.jackson>
         <version.groovy>2.4.5</version.groovy>
+        <version.gmavenplus.plugin>1.5</version.gmavenplus.plugin>
```
cdeszaq: Why not v1.6?

QubitPi: @cdeszaq Updated

Issue #637, opened 2018-03-01
Generalize forKey Method


QubitPi:
In, `EnumUtils.forKey()` method, key type can really be generalized so that we don't need 2 separate methods for String and int keys

PR #638, opened 2018-03-01
Generalize forKey Method

QubitPi:
https://github.com/yahoo/fili/issues/637

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/util/EnumUtils.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/util/EnumUtils.java
@@ -19,6 +19,7 @@ public class EnumUtils {
      * @param enumeration  Enum we are getting value by key for (needed due to type erasure)
      *
      * @return The enum value for the key
+     *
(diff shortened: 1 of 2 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/util/EnumUtils.java
```diff
@@ -9,6 +9,27 @@
...
+     * @param <K>  Type of the key to get the enum value for
+     * @param key  Key to get the enum value for
+     * @param mapping  Mapping to use to look up the value by key
+     * @param enumeration  Enum we are getting value by key for (needed due to type erasure)
+     *
+     * @return The enum value for the key
+     *
+     * @throws IllegalArgumentException if this enum type has no constant with the specified name
+     */
+    public static <T extends Enum<T>, K> T forKey(K key, Map<K, T> mapping, Class<T> enumeration) {
```
cdeszaq: I'd love to see a test for more than just `String` as the key type. Perhaps I missed it, but I didn't see any tests that cover the non-String case.

Issue #640, opened 2018-03-02
Add Missing perPage Param


QubitPi:
Pagination Links on the First Page Are Missing perPage Param

PR #641, opened 2018-03-02
Add Missing perPage Param

QubitPi:
This PR fixes https://github.com/yahoo/fili/issues/640


fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/JobsServlet.java
```diff
@@ -541,7 +541,9 @@ protected ResponseContext addPaginationInfoToResponseContext(
                 .map(pair -> Utils.withRight(pair, pair.getRight().getAsInt()))
                 .map(pair -> Utils.withRight(
                         pair,
-                        uriInfo.getRequestUriBuilder().replaceQueryParam("page", pair.getRight())
+                        uriInfo.getRequestUriBuilder()
+                                .replaceQueryParam("page", pair.getRight())
+                                .replaceQueryParam("perPage", pages.getPerPage())
```
michael-mclawhorn: You replaced "page" and "perPage" everywhere but here.

Issue #657, opened 2018-03-27
Partial intervals which have been deserialized may be more generic than SimplifiedIntervalList and fail downstream.


QubitPi:
Some downstream projects generated partial intervals as `ArrayList`, which cannot be cased to `SimplifiedIntervalList` in places like [`getPartialIntervalsWithDefault`](https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/PartialDataRequestHandler.java#L113). The result is a casting exception which crashes downstream applications.

Although https://github.com/yahoo/fili/pull/658 is a reasonable alternative to cleaning up all the source of the type precision loss. This does raise the question of whether this needs to be an SIL in all these interfaces, but this is good enough for now.

**Edit - This problem happens in `getVolatileIntervalsWithDefault` as well which is fixed in https://github.com/yahoo/fili/pull/678**

michael-mclawhorn:
This is particularly being detecting in HttpResponseMaker after async triggered serialization and deserialization.

QubitPi:
Addressed by https://github.com/yahoo/fili/pull/658

PR #658, opened 2018-03-27
Avoid casting to generate SimplifiedIntervalList

QubitPi:
Addresses https://github.com/yahoo/fili/issues/657


fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/PartialDataRequestHandler.java
```diff
@@ -111,9 +114,11 @@ public boolean handleRequest(
      * @return the missing intervals from the request or an empty list
      */
     public static SimplifiedIntervalList getPartialIntervalsWithDefault(Map<String, Serializable> context) {
-        return (SimplifiedIntervalList) context.computeIfAbsent(
-                MISSING_INTERVALS_CONTEXT_KEY.getName(),
-                (ignored) -> new SimplifiedIntervalList()
+        return new SimplifiedIntervalList(
```
michael-mclawhorn: Need a comment here to the effect of 'A serialization issue may result in the context key being a list but not a Simplified Interval List'

And a javadoc with  'fix' section, referencing a link to the issue.

QubitPi: @michael-mclawhorn Thanks. I've update this issue: https://github.com/yahoo/fili/issues/657

QubitPi:
@michael-mclawhorn Thank you~

Issue #627, opened 2018-02-15
Physical Table needs a "start instance" of availability


QubitPi:
When a Druid instance has datasources with segments of different intervals, a query that touches union-ed datasources can return no data with missing intervals, because these datasources have different data availability alignment.

PR #660, opened 2018-03-30
Implement PhysicalTable-Specific Time Limit in PartitionCompositeTable

QubitPi:
Resolves https://github.com/yahoo/fili/issues/627
Implements https://github.com/yahoo/fili/pull/628


fili-core/src/main/java/com/yahoo/bard/webservice/util/SimplifiedIntervalList.java
```diff
@@ -369,4 +370,22 @@ public SimplifiedIntervalList subtract(SimplifiedIntervalList that) {
...
     }
+
+    /**
+     * Returns an Optional of inner interval with the oldest starting instance.
+     *
+     * @return the Optional of inner interval with the oldest starting instance
+     */
+    public Optional<Interval> getStart() {
+        return stream().min(IntervalStartComparator.INSTANCE);
+    }
```
michael-mclawhorn: simplified interval list guarantees that the start of the earliest interval will be the earliest start.

fili-core/src/main/java/com/yahoo/bard/webservice/util/SimplifiedIntervalList.java
```diff
@@ -369,4 +370,22 @@ public SimplifiedIntervalList subtract(SimplifiedIntervalList that) {
...
+
+    /**
+     * Returns an Optional of inner interval with the latest ending instance.
+     *
+     * @return the Optional of inner interval with the latest ending instance
+     */
+    public Optional<Interval> getEnd() {
+        return stream().max(IntervalEndComparator.INSTANCE);
+    }
```
michael-mclawhorn: similarly SIL guaranatees that the end of the last interval in the list is the end of the range of time.

QubitPi:
@michael-mclawhorn I think we can close it

PR #662, opened 2018-04-04
Deliver more readable health check message

QubitPi:


QubitPi:
@archolewa Thanks for the comments!

PR #663, opened 2018-04-04
Adds a factory for building `ResponseProcessor`.

archolewa:
-- We would like to do a little bit of Druid error analysis to extract
some more illuminating error codes out of some of Druid's 500's. The
error callback is controlled by the `ResponseProcessor`.

-- Unfortunately, the `ResponseProcessor` can't be injected directly
because it depends on the `DataApiRequest`, which is built directly in
each request. So instead we wrap the creation of `ResponseProcessor` in
a factory that can be injected.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
@@ -73,8 +73,8 @@ import com.yahoo.bard.webservice.druid.util.FieldConverters;
 import com.yahoo.bard.webservice.druid.util.SketchFieldConverter;
 import com.yahoo.bard.webservice.metadata.DataSourceMetadataLoadTask;
 import com.yahoo.bard.webservice.metadata.DataSourceMetadataService;
-import com.yahoo.bard.webservice.metadata.RegisteredLookupMetadataLoadTask;
 import com.yahoo.bard.webservice.metadata.QuerySigningService;
+import com.yahoo.bard.webservice.metadata.RegisteredLookupMetadataLoadTask;
 import com.yahoo.bard.webservice.metadata.RequestedIntervalsFunction;
 import com.yahoo.bard.webservice.metadata.SegmentIntervalsHashIdGenerator;
 import com.yahoo.bard.webservice.table.LogicalTableDictionary;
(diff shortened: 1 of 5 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -1200,6 +1204,19 @@ protected ResponseFormatResolver buildResponseFormatResolver() {
...
+     * Returns the class to bind to {@link ResponseProcessorFactory}.
+     * <p>
+     * The ResponseProcessorFactory allows us to inject a custom {@link
+     * com.yahoo.bard.webservice.web.responseprocessors.ResponseProcessor} despite the fact that these processors depend
+     * on objects that are built uniquely for each request.
+     *
+     * @return A class that implements {@link ResponseProcessorFactory}.
+     */
+    protected Class<? extends ResponseProcessorFactory> buildResponseProcessorFactory() {
+        return ResultSetResponseProcessorFactory.class;
```
QubitPi: Prob. not a blocker for this PR, but what if other `ResponseProcessor` implementations also need custom injection?

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/DataServlet.java
```diff
@@ -172,7 +176,8 @@ public DataServlet(
             BroadcastChannel<String> preResponseStoredNotifications,
             HttpResponseMaker httpResponseMaker,
             ResponseFormatResolver formatResolver,
-            DataApiRequestFactory dataApiRequestFactory
+            DataApiRequestFactory dataApiRequestFactory,
+            ResponseProcessorFactory responseProcessorFactory
```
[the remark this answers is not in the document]
archolewa: We haven't been doing that for the DataServlet. And we could, but seeing as how
we always add something to this constructor whenever we need to add something
new to inject, `DataServlet` would *very* quickly become *very* noisy.

Also, although it is technically possible to override the `DataServlet`, I think
that's sufficiently weird and rare enough that having hordes of giant 
deprecated constructors running around would do more harm than good.

Especially since this is an easy enough thing to fix.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
archolewa: Erm, that's the whole idea of this PR? If somebody wants to inject their
own custom `ResponseProcessor` then they:

1. Define their `CustomResponseProcessor`.
2. Define their `CustomResponseProcessorFactory`.
3. Override this method and return `CustomResponseProcessorFactory.class`

archolewa: Do you mean if they have non-empty constructors? Then they'd need to make their
constructor injectable so that HK2 can handle it.

QubitPi: Make sense

Issue #668, opened 2018-04-17
Remove deprecated code


michael-mclawhorn:

michael-mclawhorn:
There are a lot of 0.9 and 0.8 deprecations hanging around.

PR #669, opened 2018-04-17
Clearing deprecations

michael-mclawhorn:
https://github.com/yahoo/fili/issues/668


QubitPi:
@michael-mclawhorn Can we merge this?

cdeszaq:
:shipit: Though since these are breaking changes, the major version should be revved.

cdeszaq:
I think I just didn’t catch that we were intentionally holding this for the next major rev. Perhaps it was called out somewhere that it was the plan, but I wanted to be sure it didn’t get missed.

Regardless, these are good changes. :shipit:

PR #675, opened 2018-04-23
Add tests to Utils

QubitPi:
https://github.com/yahoo/fili/issues/700

### DIFF
--- a/fili-core/src/test/groovy/com/yahoo/bard/webservice/util/UtilsSpec.groovy
+++ b/fili-core/src/test/groovy/com/yahoo/bard/webservice/util/UtilsSpec.groovy
@@ -1,0 +1,80 @@
+// Copyright 2018 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.util
+
+import spock.lang.Specification
+
+import java.nio.file.Files
+import java.nio.file.Paths
+
+class UtilsSpec extends Specification {
+
+    private static final TEMP_DIR = "target/tmp/util/"
+
+    def cleanup() {
+        Files.deleteIfExists(Paths.get(TEMP_DIR))
+    }
+
+    def "Get by sub-type returns all instances of that sub-type"() {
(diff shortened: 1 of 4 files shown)


fili-core/src/test/groovy/com/yahoo/bard/webservice/util/UtilsSpec.groovy
```diff
@@ -0,0 +1,80 @@
...
+        and: "getting by HashSet returns both HashSet and LinkedHashSet instances"
+        Utils.getSubsetByType(set, HashSet) == set as LinkedHashSet
+    }
+
+    def "File path without parent dir generates the parent"() {
+        given: "a file path without a parent directory"
+        String path = TEMP_DIR + "data.txt"
+
+        // make sure parent directory does not exist yet
+        assert ! new File(TEMP_DIR).exists()
```
cdeszaq: (minor stylistic) You can make this an `expect` block and and turn the comment into a block comment so you don't need the `assert`

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/UtilsSpec.groovy
```diff
@@ -0,0 +1,80 @@
...
+        String subDir = TEMP_DIR + "subDir/"
+        String subFile = TEMP_DIR + "subDir/data.txt"
+
+        and: "create the directory contents"
+        Files.createFile(Paths.get(file1))
+        Files.createFile(Paths.get(file2))
+        Files.createDirectory(Paths.get(subDir))
+        Files.createFile(Paths.get(subFile))
+
+        // make sure all files and sub-dir exist
```
cdeszaq: ditto using an `expect` block after a `given` or `setup` block to assert preconditions for a test.

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/UtilsSpec.groovy
QubitPi: I like that. Thanks!

QubitPi:
Thank you, @cdeszaq , for the review and approval!

Issue #687, opened 2018-05-07
Enable Druid In-filter in Fili


QubitPi:
A problem with `DefaultDruidFilterBuilder` is that if a `contains` filter resolves to, for example, 10000 matched dimension rows, there will be a giant Druid query with 10000 filters. It has been our experience that such kind of query eventually times out on the Druid side.

QubitPi:
Problem solved by https://github.com/yahoo/fili/pull/688

PR #688, opened 2018-05-07
Enable Druid In-filter in Fili

QubitPi:
Proboem
=======
https://github.com/yahoo/fili/issues/687

Solution
======
* `DefaultDruidFilterBuilder` is renamed to `DruidOrFilterBuilder`
* Implement `DruidInFilterBuilder` which turns list of selector filter generated by `DruidOrFilterBuilder` into a single Druid in-filter. The in-filter resolves the [timeout issue](https://github.com/yahoo/fili/issues/687). The in-filter could substantially shorten druid query, making the query more sharable and readable.
* Fili now uses the `DruidInFilterBuilder` as the "Default" Druid query filter builder, in stead of the old `DruidOrFilterBuilder`, because the new default filter is terser and packs the query payload tighter, and in proportion to the number of filters being applied.


fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidInFilterBuilder.java
```diff
@@ -0,0 +1,80 @@
...
+            Filter inFilter = new InFilter(
+                    dimension,
+                    ((OrFilter) disjunction).getFields().stream()
+                            .map(subFilter -> (SelectorFilter) subFilter)
+                            .map(SelectorFilter::getValue)
+                            .distinct()
+                            .collect(Collectors.toList())
+            );
+
+            inFilters.add(normalizedFilter == filter ? inFilter : new NotFilter(inFilter));
```
michael-mclawhorn: We can do better than this.
This will result in several overlapping in filters for adjacent filters.  Instead we should collect all the positive filters into one collection and all the negative filters into another collection and build up to two Druid In Filters.

This also suggests we could deduplicate productively in the OrFilterBuilder in the same way.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -175,6 +175,18 @@
...
     private static final String DRUID_HEADER_SUPPLIER_CLASS = "druid_header_supplier_class";
 
+    /**
+     * Druid filter builder related.
+     */
+    private static final String FILTER_BUILDER_INIT_ERROR = "Unable to create Druid filter builder";
+    private static final String DRUID_FILTER_BUILDER_KEY = "druid_filter_builder";
+    private static final String DEFAULT_DRUID_FILTER_BUILDER =
+            "com.yahoo.bard.webservice.data.filterbuilders.DruidInFilterBuilder";
+    private static final String DRUID_FILTER_BUILDER = SYSTEM_CONFIG.getStringProperty(
```
archolewa: Why are we using configuration for this? Why not just do the same thing we 
always do for injectable classes: If the customer wants a different
implementation, override the appropriate getter in their binder factory.

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidInFilterBuilder.java
```diff
@@ -0,0 +1,80 @@
...
+            throws DimensionRowNotFoundException {
+        LOG.trace("Building dimension filter using dimension: {} \n\n and set of filter: {}", dimension, filters);
+
+        List<Filter> inFilters = new ArrayList<>();
+        for (ApiFilter filter : filters) {
+            ApiFilter normalizedFilter = filter;
+            if (normalizedFilter.getOperation().equals(FilterOperation.notin)) {
+                normalizedFilter = filter.withOperation(FilterOperation.in);
+            }
+            Filter disjunction = new OrFilter(buildSelectorFilters(
```
archolewa: Why go through all these loops of building a disjunction of selector filters
and then building an inFilter from that?

Why not just get the DimensionRows, and exact the ids from each row, and build
the inFilter directly? Something like:

```java
    List<String> ids = getFilteredDimensionRows(
            dimension,
            Collections.singleton(normalizedFilter)
    )
            .stream()
            .map(DimensionRow::getId)
            .collect(Collectors.toList());
    InFilter filter = new InFilter(dimension, ids);
```

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/InFilter.java
```diff
@@ -32,7 +33,7 @@
      */
     public InFilter(Dimension dimension, @NotNull List<String> values) {
         super(dimension, DefaultFilterType.IN);
-        this.values = values;
+        this.values = new ArrayList<>(values);
```
QubitPi: Use defensive copy

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidInFilterBuilder.java
```diff
@@ -48,28 +47,32 @@ protected Filter buildDimensionFilter(Dimension dimension, Set<ApiFilter> filter
...
-            Filter inFilter = new InFilter(
-                    dimension,
-                    ((OrFilter) disjunction).getFields().stream()
-                            .map(subFilter -> (SelectorFilter) subFilter)
-                            .map(SelectorFilter::getValue)
-                            .distinct()
-                            .collect(Collectors.toList())
-            );
+            List<String> values = getFilteredDimensionRows(dimension, Collections.singleton(normalizedFilter)).stream()
+                    .map(row -> row.get(dimension.getKey()))
```
archolewa: DimensionRow has a function that does the same thing: `getKeyValue`. So you can just write:

```java
.map(DimensionRow::getKeyValue)
```

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidInFilterBuilder.java
```diff
@@ -0,0 +1,83 @@
...
+    @Override
+    protected Filter buildDimensionFilter(Dimension dimension, Set<ApiFilter> filters)
+            throws DimensionRowNotFoundException {
+        LOG.trace("Building dimension filter using dimension: {} \n\n and set of filter: {}", dimension, filters);
+
+        List<Filter> inFilters = new ArrayList<>(); // A list with at most 2 in-filters(positive & negative)
+        Set<String> positiveInValues = new HashSet<>(); // contains values for positive in-filter
+        Set<String> negativeInValues = new HashSet<>(); // contains values for negative in-filter
+
+        for (ApiFilter filter : filters) {
```
archolewa: Taking another look at this, it looks like all of these filters are from
the same dimension.

So why are we resolving each filter individually? It seems to me like this would
be extremely inefficient in many realword cases (especially if you have 
autogenerated filters, like filters for row security).

Also, it's wrong.

For example, suppose we have the filters:

```java
    dimension|name-contains[e],dimension|id-in[1,2,3]
```

Furthermore, suppose this dimension has 10,000+ rows, and every name is of the
form `e-somethingorother`.

In the example above, our end result should be 
`in(1, 2, 3)`. However, we never actually trim any results out of 
`positiveInValues`. So instead the `inFilter` will contain 10,000+ values, even
though it should only have 3.

However, even if we fix that bug, this is still does not scale at all.
This filter shouldn't actually be that bad, assuming whatever is
backing your SearchProvider (i.e. Lucene) is at all smart about how it handles
`in` vs.  `contains`, which Lucene probably is considering its maturity.
The way this is currently implemented, we're not taking advantage of 
*any* intelligence that may or may not exist in the SearchProvider 
implementation, because we're executing each filter clause individually. 

And even if the SearchProvider is dumb as a box of rocks, this is still not 
an ideal implementation, because we're allocating a set with 10,000+ 
values, copying 10,000 numbers into it, and then later trimming it down to 3,
instead of just getting 3 values back from the SearchProvider.

What we really should be doing I think, is just sending all the filters down to
the SearchProvider at once if we have at least one positive filter.

The only time we should be doing anything remotely like this is if all the
filters are negative. In that case, we would have to send each filter down
individually, because we have no way of encoding top-level ors, so we can't
just pull out the common negative and send down positive versions of all the
filters.

So if we have:

```java
    dimension|name-contains[e],dimension|id-in[1,2,3],dimension|id-notin[1]
```

Then we should just send this whole set down to the SearchProvider at once.

If we have the filter:

```java
    dimension|id-notin[1,2,3],dimension|name-notin[e-1234],dimension|id-notin[5]
```
Then yes, we'll need to send each filter down individually, because this is 
equivalent to

```
NOT(dimension|id-in[1,2,3] OR dimension|name-in[e-1234] OR dimension|id-in[5])
```

and we have no way of encoding the OR's.

fili-core/src/test/groovy/com/yahoo/bard/webservice/filterbuilders/DruidInFilterBuilderSpec.groovy
```diff
@@ -0,0 +1,156 @@
...
+import com.yahoo.bard.webservice.druid.model.filter.Filter
+import com.yahoo.bard.webservice.druid.model.filter.InFilter
+import com.yahoo.bard.webservice.druid.model.filter.NotFilter
+import com.yahoo.bard.webservice.web.ApiFilter
+
+import spock.lang.Shared
+import spock.lang.Specification
+import spock.lang.Unroll
+
+class DruidInFilterBuilderSpec extends Specification {
```
archolewa: It looks like we're only testing single filter clauses at time. We really 
should have some tests with multiple filter clauses. If we had, the tests would
have caught the bug I pointed out above.

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidInFilterBuilder.java
michael-mclawhorn: To specify the application Andrew's point, the algorithm we should be following is:
a) group collect positive and negative constraints
b) let the search provider resolve all the positive constraints, but a Druid in-filter out of those.
c) let the search provider resolve all the negative constraints, but a druid in-filter out of those and encapsulate it in a not filter
d) And those results together.

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/ConjunctionDruidFilterBuilder.java
```diff
@@ -88,6 +96,80 @@ public Filter buildFilters(Map<Dimension, Set<ApiFilter>> filterMap) throws Dime
...
+     * @return a stream of the negated filters
+     */
+    protected static Stream<ApiFilter> negateNegativeFilters(Collection<ApiFilter> negativeFilters) {
+        return negativeFilters.stream()
+                // TODO - refactor this and next map when more than 1 not* FilterOperations are supported.
+                .peek(filter -> {
+                    if (!FilterOperation.notin.equals(filter.getOperation())) {
+                        String message = String.format(NON_NEGATIVE_FILTER_ERROR_FORMAT, filter);
+                        LOG.error(message);
+                        throw new IllegalArgumentException(message);
```
michael-mclawhorn: What are we supposed to do with this error if it occurs?

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ pull request if there was one.
 
 ### Added:
 
+- [Enable Druid In-filter in Fili](https://github.com/yahoo/fili/pull/688)
+    * `DefaultDruidFilterBuilder` is renamed to `DruidOrFilterBuilder`
+    * Implement `DruidInFilterBuilder` which turns list of selector filter generated by `DruidOrFilterBuilder` into a
+      Druid in-filter. The in-filter could substantially shorten druid query, making the query more sharable and
+      readable.
```
cdeszaq: Why is "sharable and readable" desired? Since the Druid query is system-facing (ie. it intended for consumption by Druid, not a human), what impact does an in-filter have on that?

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ pull request if there was one.
...
+    * Fili now uses the `DruidInFilterBuilder` as the "Default" Druid query filter builder, in stead of the old
+      `DruidOrFilterBuilder`
```
cdeszaq: Why change the default? Is it faster? Less memory? Easier to reason about? At least a _little_ of the reason in the changelog would be good.

This is a good description of the change, but is it always the case that an IN filter is more performant than an OR filter? Should it _always_ be the default, or do we need to be a little smarter than that in order to not make queries with only a small number of filters slower?

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/ConjunctionDruidFilterBuilder.java
```diff
@@ -88,6 +96,80 @@ public Filter buildFilters(Map<Dimension, Set<ApiFilter>> filterMap) throws Dime
...
+
+        return new ImmutablePair<>(positiveFilters, negativeFilters);
+    }
+
+    /**
+     * Negates a collection of negative filters, i.e. {@link FilterOperation#notin} {@code =>}
+     * {@link FilterOperation#in}, and returns a stream of the negated filters.
+     * <p>
+     * This method throws {@link IllegalArgumentException} if any one of the filter collection passed in is not a
+     * negative filter, i.e. {@link FilterOperation#notin}.
```
cdeszaq: Would it simplify this method's contract if it negated any filters passed in? So, if it gets a negative filter, it turns it into the positive filter, and if it _does_ get a positive filter, it flips it to a negative filter?

It may make the body of the method a little bit more complex, but it seems like it would greatly simplify the checking that the caller has to do to call this safely.

Alternatively, can we lean on the type system to require the collection being passed in are all negative filters?

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/ConjunctionDruidFilterBuilder.java
```diff
@@ -88,6 +96,80 @@ public Filter buildFilters(Map<Dimension, Set<ApiFilter>> filterMap) throws Dime
...
+     * Splits a set of {@link ApiFilter}s into two groups of positive and negative filters.
+     * <p>
+     * An {@link ApiFilter} is defined to be positive if its filter operation is one of
+     * <ul>
+     *     <li> {@link FilterOperation#in}
+     *     <li> {@link FilterOperation#startswith}
+     *     <li> {@link FilterOperation#contains}
+     *     <li> {@link FilterOperation#eq}
+     * </ul>
+     * An {@link ApiFilter} is defined to be negative if its filter operation is {@link FilterOperation#notin}.
```
cdeszaq: Can this split (positive/negative) be pushed to the type system to enforce?

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/ConjunctionDruidFilterBuilder.java
```diff
@@ -88,6 +96,80 @@ public Filter buildFilters(Map<Dimension, Set<ApiFilter>> filterMap) throws Dime
...
+     *
+     * @param filters  A set of API filters that are to be splitted
+     *
+     * @return a pair whose left is positive filter and whose right is negative filers
+     */
+    protected static Pair<Set<ApiFilter>, Set<ApiFilter>> splitApiFilters(Set<ApiFilter> filters) {
+        Set<ApiFilter> positiveFilters = new HashSet<>();
+        Set<ApiFilter> negativeFilters = new HashSet<>();
```
cdeszaq: Since we have Guava's ImmutableSet available to us, can we use it's builder in order to end up with these being immutable w/o having to wrap them?

If not, please wrap them into unmodifiable sets before making the pair.

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/ConjunctionDruidFilterBuilder.java
```diff
@@ -88,6 +96,80 @@ public Filter buildFilters(Map<Dimension, Set<ApiFilter>> filterMap) throws Dime
...
+
+        for (ApiFilter filter : filters) {
+            if (FilterOperation.notin.equals(filter.getOperation())) {
```
cdeszaq: (style) If you wanted to do this in a streamy way, you could use a grouping collector I believe.

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidInFilterBuilder.java
```diff
@@ -0,0 +1,99 @@
...
+ */
+public class DruidInFilterBuilder extends ConjunctionDruidFilterBuilder {
+    private static final Logger LOG = LoggerFactory.getLogger(DruidInFilterBuilder.class);
+
+    @Override
+    protected Filter buildDimensionFilter(Dimension dimension, Set<ApiFilter> filters)
+            throws DimensionRowNotFoundException {
+        LOG.trace("Building dimension filter using dimension: {} \n\n and set of filter: {}", dimension, filters);
+
+        List<Filter> inFilters = new ArrayList<>(); // A set with at most two in-filters(positive & negative)
```
cdeszaq: This isn't used until _much_ later in the class. It's usually a good idea to define variables just before they are used so that they are easier to trace through a method.

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidInFilterBuilder.java
```diff
@@ -0,0 +1,99 @@
...
+
+        // split ApiFilters into two groups: positive filters & negative filters
+        Pair<Set<ApiFilter>, Set<ApiFilter>> splittedFilters = splitApiFilters(filters);
+        Set<ApiFilter> positiveFilters = splittedFilters.getLeft();
+        Set<ApiFilter> negativeFilters = negateNegativeFilters(splittedFilters.getRight()).collect(Collectors.toSet());
```
cdeszaq: Since the next time you access `negativeFilters` you stream it, why collect it into a `Set` here?

CHANGELOG.md
michael-mclawhorn: I believe that it's always TERSER.  I'm pretty sure it unpacks to the same logic in Druid historical nodes.  But this packs the query payload tighter, and in proportion to the number of filters being applied.

CHANGELOG.md
archolewa: Debugging mostly. If there's a bug (or you're trying to do some analysis of the 
logs for whatever reason), and it's not clear if it's in Druid or
the API, then you may need to send your resident Druid expert the Druid query
that's being sent down.

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/ConjunctionDruidFilterBuilder.java
archolewa: Sounds like a lot of specification overhead for minimal gain.

Converting between types requires *so much ceremony*.

cdeszaq: I don't disagree that Java makes type conversion painful some times, but I always like to let the compiler do the checking and thinking when possible via the type system instead of embedding what is essentially type information into system logic / rules.

I trust types and the compiler more than my own code ;-)

But, if it's not worth it and we think this logic is easy to understand and hard to screw up, then I agree that the gain from the compiler keeping it correct is not worth it.

CHANGELOG.md
cdeszaq: Fair enough. Does it make sense to add that bit of "why do we care about this?" to the change log?

CHANGELOG.md
cdeszaq: Including this "why" in the commit message or the changelog would be good so that someone else doesn't come along and ask the same question. Or at least when they do, we can point them somewhere to answer the question.

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/ConjunctionDruidFilterBuilder.java
cdeszaq: I was mostly hoping to move from a runtime failure detection to compile-time, but I see your point about this encapsulating the complexity. However, because it throws an illegal argument, it's actually pushing the complexity of "am I calling this with a valid set of filters" to the caller, so the caller is still forced to deal with the complexity of "make sure you're sending in the right thing". 

Having the type system enforce that makes it a little more fool-proof for the caller (even if local) because the compiler will blow up if they try to do it wrong, rather than having to build out tests to check it. Which is more complex / harder to maintain, having to have tests for all of the "scenarios" for which this method needs to look for, or only having what the type system allows?

If we've got the rich tests around the complexity of these rules, and tests ensuring the caller is handling everything correctly, I'm less concerned, but otherwise I still lean towards "harder to screw up".

QubitPi: I've added additional test - https://github.com/yahoo/fili/pull/688/files#diff-dee9ab9980853004178878638e51ebe8R108

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidInFilterBuilder.java
```diff
@@ -0,0 +1,99 @@
...
+public class DruidInFilterBuilder extends ConjunctionDruidFilterBuilder {
+    private static final Logger LOG = LoggerFactory.getLogger(DruidInFilterBuilder.class);
+
+    @Override
+    protected Filter buildDimensionFilter(Dimension dimension, Set<ApiFilter> filters)
+            throws DimensionRowNotFoundException {
+        LOG.trace("Building dimension filter using dimension: {} \n\n and set of filter: {}", dimension, filters);
+
+        // split ApiFilters into two groups: positive filters & negative filters
+        Pair<Set<ApiFilter>, Set<ApiFilter>> splittedFilters = splitApiFilters(filters);
```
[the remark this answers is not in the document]
QubitPi: `positiveAndNegativeSplitFilters` is nice!

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidInFilterBuilder.java
```diff
@@ -0,0 +1,99 @@
...
+        Set<ApiFilter> positiveFilters = splittedFilters.getLeft();
+        Set<ApiFilter> negativeFilters = negateNegativeFilters(splittedFilters.getRight());
+
+        // search for matched values of the positive filter by sending all of the filters down to search provider once
+        List<String> inValues = positiveFilters.isEmpty()
```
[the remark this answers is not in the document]
QubitPi: I have tried this approach but the code doesn't seem to become much simpler:

```java
public class DruidInFilterBuilder extends ConjunctionDruidFilterBuilder {
    private static final Logger LOG = LoggerFactory.getLogger(DruidInFilterBuilder.class);

    @Override
    protected Filter buildDimensionFilter(Dimension dimension, Set<ApiFilter> filters)
            throws DimensionRowNotFoundException {
        LOG.trace("Building dimension filter using dimension: {} \n\n and set of filter: {}", dimension, filters);

        // split ApiFilters into two groups: positive filters & negative filters
        Pair<Set<ApiFilter>, Set<ApiFilter>> positiveAndNegativeSplitFilters = splitApiFilters(filters);

        Optional<Filter> optionalInFilter = buildInFilter(dimension, positiveAndNegativeSplitFilters.getLeft());
        Optional<Filter> optionalNotInFilter = buildNotInFilter(
                dimension,
                negateNegativeFilters(positiveAndNegativeSplitFilters.getRight())
        );

        Filter newFilter = null;
        if (optionalInFilter.isPresent() && optionalNotInFilter.isPresent()) {
            newFilter = new AndFilter(Arrays.asList(optionalInFilter.get(), optionalNotInFilter.get()));
        } else if (optionalInFilter.isPresent()) {
            newFilter = optionalInFilter.get();
        } else if (optionalNotInFilter.isPresent()) {
            newFilter = optionalNotInFilter.get();
        } else {
            String msg = ErrorMessageFormat.DIMENSION_ROWS_NOT_FOUND.format(dimension.getApiName(), filters);
            LOG.debug(msg);
            throw new DimensionRowNotFoundException(msg);
        }

        LOG.trace("Filter: {}", newFilter);
        return newFilter;
    }

    /**
     * Constructs an {@link InFilter} in an Optional that filters on a set of specified positive API filters.
     *
     * @param dimension  Dimension for the filters
     * @param positiveFilters  All positive filters belonging to that dimension
     *
     * @return the Optional {@code InFilter} that filters on the set of specified positive API filters
     *
     * @throws DimensionRowNotFoundException if the filters filter out all dimension rows
     */
    private static Optional<Filter> buildInFilter(
            Dimension dimension,
            Set<ApiFilter> positiveFilters
    ) throws DimensionRowNotFoundException {
        List<String> inValues = getInValues(dimension, positiveFilters);
        return inValues.isEmpty()
                ? Optional.of(new InFilter(dimension, inValues))
                : Optional.empty();
    }

    /**
     * Returns all matching values of positive filters by sending all of the filters down to search
     * provider once.
     *
     * @param dimension  Dimension for the filters
     * @param positiveFilters  All positive filters belonging to that dimension
     *
     * @return all matching values of positive filters
     *
     * @throws DimensionRowNotFoundException if the filters filter out all dimension rows
     */
    private static List<String> getInValues(
            Dimension dimension,
            Set<ApiFilter> positiveFilters
    ) throws DimensionRowNotFoundException {
        return positiveFilters.isEmpty()
                ? Collections.emptyList()
                : getFilteredDimensionRowValues(dimension, positiveFilters);
    }

    /**
     * Constructs an {@link NotFilter} in an Optional that filters on a set of specified negative API filters.
     *
     * @param dimension  Dimension for the filters
     * @param negativeFilters  All negative filters belonging to that dimension
     *
     * @return the Optional {@code NotFilter} that filters on the set of specified negative API filters
     *
     * @throws DimensionRowNotFoundException if the filters filter out all dimension rows
     */
    private static Optional<Filter> buildNotInFilter(
            Dimension dimension,
            Set<ApiFilter> negativeFilters
    ) throws DimensionRowNotFoundException {
        List<String> notInValues = getNotInValues(dimension, negativeFilters);
        return notInValues.isEmpty()
                ? Optional.of(new NotFilter(new InFilter(dimension, notInValues)))
                : Optional.empty();
    }

    /**
     * Returns all matching values of negative filters by sending each filter down to search provider
     * one-by-one.
     *
     * @param dimension  Dimension for the filters
     * @param negativeFilters  All negative filters belonging to that dimension
     *
     * @return all matching values of negative filters
     *
     * @throws DimensionRowNotFoundException if the filters filter out all dimension rows
     */
    private static List<String> getNotInValues(
            Dimension dimension,
            Set<ApiFilter> negativeFilters
    ) throws DimensionRowNotFoundException {
        return negativeFilters.stream()
                .map(apiFilter -> {
                    try {
                        return getFilteredDimensionRowValues(dimension, Collections.singleton(apiFilter));
                    } catch (DimensionRowNotFoundException exception) {
                        throw new RuntimeException(exception);
                    }
                })
                .flatMap(List::stream)
                .distinct()
                .collect(Collectors.toList());
    }
}
```

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidInFilterBuilder.java
```diff
@@ -0,0 +1,99 @@
...
+                ? Collections.emptyList()
+                : getFilteredDimensionRowValues(dimension, positiveFilters);
+
+        // search for matched values of the negative filter by sending each filter down to search provider one-by-one
+        List<String> notInValues = negativeFilters.stream()
+                .map(apiFilter -> {
+                    try {
+                        return getFilteredDimensionRowValues(dimension, Collections.singleton(apiFilter));
+                    } catch (DimensionRowNotFoundException exception) {
+                        throw new RuntimeException(exception);
```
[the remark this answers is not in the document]
QubitPi: DimensionRowNotFoundException could not be thrown again in stream. Maybe we could use `IllegalStateException`?

CHANGELOG.md
```diff
@@ -7,14 +7,23 @@ pull request if there was one.
 
 ### Added:
 
+- [Enable Druid In-filter in Fili](https://github.com/yahoo/fili/pull/688)
```
efronbs: test

fili-core/src/test/groovy/com/yahoo/bard/webservice/filterbuilders/ConjunctionDruidFilterBuilderSpec.groovy
```diff
@@ -69,6 +73,48 @@ class ConjunctionDruidFilterBuilderSpec extends Specification {
...
+        expect:
+        filterBuilder.negateNegativeFilters([ apiFilters[negativeFilter] ]) == [ apiFilters[negated] ] as Set
+
+        where:
+        negativeFilter     || negated
+        "ageIdNotin56"     || "ageIdIn56"
+        "ageDescNotin1429" || "ageDescIn1429"
+    }
+
+    def "Negating positive filter throws error"() {
```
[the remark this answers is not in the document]
michael-mclawhorn: The behavior of the implementation is to force the filters positive.  It doesn't have to care whether they were originally negative.  We can make error handling for that or we can just treat this as a 'toUpperCase' stye method where the original truth doesn't matter.

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidInFilterBuilder.java
```diff
@@ -0,0 +1,99 @@
...
+
+        // split ApiFilters into two groups: positive filters & negative filters
+        Pair<Set<ApiFilter>, Set<ApiFilter>> positiveAndNegativeSplitFilters = splitApiFilters(filters);
+        Set<ApiFilter> positiveFilters = positiveAndNegativeSplitFilters.getLeft();
+        Set<ApiFilter> negativeFilters = negateNegativeFilters(positiveAndNegativeSplitFilters.getRight());
+
+        // search for matched values of the positive filter by sending all of the filters down to search provider once
+        List<String> inValues = positiveFilters.isEmpty()
+                ? Collections.emptyList()
+                : getFilteredDimensionRowValues(dimension, positiveFilters);
```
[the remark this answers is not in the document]
QubitPi: We are sending all of the filters down to search provider once. Lucene search provider guarantees the uniqueness of the matched values.

fili-core/src/test/groovy/com/yahoo/bard/webservice/filterbuilders/ConjunctionDruidFilterBuilderSpec.groovy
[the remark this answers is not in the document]
QubitPi: @archolewa provides a very good explanations about why we want to treat pos. & neg. filters separately

Issue #691, opened 2018-05-08
Query interval of "P1D/current" with "all" granularity results in a Bad Request (400)


QubitPi:
"P1D/current" should be a valid interval in this case, because Druid allows such query

400 originated from https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/ApiRequestImpl.java#L552

PR #692, opened 2018-05-09
Legitimate request having interval of 'P1D/current' with 'all' grain

QubitPi:
Addresses https://github.com/yahoo/fili/issues/691

The root cause is the fact that `AllGranularity` doesn't support `roundFloor()`(shown in picture below), which is why current code avoid the case of parsing interval under "all" granularity.

A costly and not-necessarily the right option is to have `AllGranularity` implement `TimeGrain`.
But a much better solution would be simply allowing such interval string during parsing, which this PR chooses

```
                              +---------------+
                              |               |
                          +--->  Granularity  <----+
                          |   |               |    |
roundFloor()  +-----+     |   +---------------+    |
                    |     |                        |
                    |     |                        |
                    +-----+-----+         +--------+-------+
                    | TimeGrain |         | AllGranularity |
                    +-----^-----+         +----------------+
                          |
                          |
                 +--------+--------+
                 |ZoneLessTimeGrain|
                 +--------^--------+
                          |
                          |
                          |
                  +-------+--------+
                  |DefaultTimeGrain|
                  +----------------+

```

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/ApiRequestImpl.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/ApiRequestImpl.java
@@ -549,12 +549,16 @@ public abstract class ApiRequestImpl implements ApiRequest {
             String dateText,
             DateTimeFormatter timeFormatter
     ) throws BadApiRequestException {
-        //If granularity is all and dateText is macro, then throw an exception
         TimeMacros macro = TimeMacros.forName(dateText);
         if (macro != null) {
             if (granularity instanceof AllGranularity) {
-                LOG.debug(INVALID_INTERVAL_GRANULARITY.logFormat(macro, dateText));
-                throw new BadApiRequestException(INVALID_INTERVAL_GRANULARITY.format(macro, dateText));
+                //If granularity is all and dateText is a macro of TimeMacros.NEXT, then throw an exception
+                if (TimeMacros.NEXT.equals(macro)) {
+                    LOG.debug(INVALID_INTERVAL_GRANULARITY.logFormat(macro, dateText));
+                    throw new BadApiRequestException(INVALID_INTERVAL_GRANULARITY.format(macro, dateText));
+                } else {
+                    return DateTime.now();
+                }
             }
             return macro.getDateTime(now, (TimeGrain) granularity);
         }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/ApiRequestImpl.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/ApiRequestImpl.java
@@ -552,12 +552,13 @@ public abstract class ApiRequestImpl implements ApiRequest {
         TimeMacros macro = TimeMacros.forName(dateText);
         if (macro != null) {
             if (granularity instanceof AllGranularity) {
-                //If granularity is all and dateText is a macro of TimeMacros.NEXT, then throw an exception
-                if (TimeMacros.NEXT.equals(macro)) {
+                if (TimeMacros.CURRENT.equals(macro)) {
+                    // Granularity is "all" and dateText is a macro of TimeMacros.CURRENT, this valid
+                    return DateTime.now();
+                } else {
+                    // Granularity is "all" and dateText is not a macro of TimeMacros.CURRENT, this not valid
                     LOG.debug(INVALID_INTERVAL_GRANULARITY.logFormat(macro, dateText));
                     throw new BadApiRequestException(INVALID_INTERVAL_GRANULARITY.format(macro, dateText));
-                } else {
-                    return DateTime.now();
                 }
             }
             return macro.getDateTime(now, (TimeGrain) granularity);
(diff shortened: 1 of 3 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/ApiRequestImpl.java
```diff
@@ -549,12 +549,17 @@ public static DateTime getAsDateTime(
...
     ) throws BadApiRequestException {
-        //If granularity is all and dateText is macro, then throw an exception
         TimeMacros macro = TimeMacros.forName(dateText);
         if (macro != null) {
             if (granularity instanceof AllGranularity) {
-                LOG.debug(INVALID_INTERVAL_GRANULARITY.logFormat(macro, dateText));
-                throw new BadApiRequestException(INVALID_INTERVAL_GRANULARITY.format(macro, dateText));
+                if (TimeMacros.CURRENT.equals(macro)) {
+                    // Granularity is "all" and dateText is a macro of TimeMacros.CURRENT, this valid
+                    return DateTime.now();
```
archolewa: This makes the semantics of `CURRENT` weird. Because Fili uses closed-open 
intervals, *technically* CURRENT shouldn't include the last bucket (i.e. if 
I run `P1D/Current` on an hourly then I should get results starting 24 hours
ago, to the start of the curent hour). However, there is only one bucket in the
All grain, so you shouldn't be getting anything!

The sane query (based on our current semantics) would be `P1D/NEXT`, since that
includes the current partial bucket.

I'm not totally opposed to giving the All grain its own rules, because All is
so weird, but this does make me wonder if our users understand what `CURRENT`
and `NEXT` actually mean, and therefore if we need to come up with some clearer
aliases. 

The other question is whether we should just give `CURRENT` and `NEXT` the same
semantics for the All grain. I think I'd prefer that over having `CURRENT` be 
defined and `NEXT` be undefined. Especially since All has weird rules anyway.

@michael-mclawhorn What do you think?

michael-mclawhorn: @QubitPi is responding to a user request from customer who doesn't know 'current' means.  He specifically expects current to mean 'start of the current day' which is a bias from only using day bounded queries and day bounded loading.  Given that druid doesn't HAVE a concept of current, it's not true that the above query is 'supported by druid'.

Maybe I'm too close to this, but I don't think they're unclear.  I think the confusion here comes from the navi-ui (https://github.com/yahoo/navi) hiding the relative complexities of time by making the time grain an ordinary dimension whose absence is treated as 'all'.

Now, this is TOTALLY REASONABLE from a reporting-modeling standpoint, but it doesn't play as nicely with Fili's self-appointed crusade to abstract away the physical slice optimizations. 

As I see it, Fili has taken the opinion that, with the exception of that magic hour when one day becomes the next day (which has caused real confusion when 'next day' flipped in UTC (server time) and demos were being done in California at the end of the day and it flipped mid-demo.) the construction of a daily query shouldn't be impacted by the underlying loading structure unless the users want it to be.

This is a long winded way of saying that Fili could loosen it's rigidity on time.  But if the underlying problem is that Navi is obfuscating the meaning of 'current', adding macros like 'NOW'(this millisecond?) 'TODAY' (now-roundfloor on Day) , 'THISHOUR' (now, roundfloor on Hour) or even 'TODAY+1D' or 'TODAY+1'  (now+ 1 dayfield roundfloor on Day) etc which would be non-polymorphic with respect to grain won't help the actually impacted users if Navi isn't prepared to take advantage of them.

By and large direct-API users are better educated as to the meaning of the macros.  And I'm happy to allow more, richer macros if we have an api stakeholder.

ALL THAT SAID, the semantics of Current on grain all are VERY problematic.  If Navi is constructing queries that allow current with the all grain, I'd say it's a Navi bug.

The possible meanings of CURRENT in the all granularity:
Current -> now (round to current millisecond)  (Next, then undefined)
alternatively, the intuition that the reporting user had was that current meant the current day.  In the context of this example one /might/ infer that P1D means that current should be rounded as a day.  But what if the interval was P1D1H or P3D?

@jkusa I think this if Navi is sending the current macro when no time grain is selected this might be better thought of as a navi bug.  If Navi would like to have some different semantics around macros, that's something I'd be open to.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/DataApiRequestIntervalsSpec.groovy
```diff
@@ -134,6 +135,60 @@ class DataApiRequestIntervalsSpec extends Specification {
...
+
+        where:
+        numPastDays | _
+        1           | _
+        2           | _
+        3           | _
+    }
+
+    @Unroll
+    def "current/P#numPastDays D with 'all' granularity generates an interval of #numPastDays days for the future" () {
```
cdeszaq: I'm inclined to go with "these changes don't make sense", but if they _do_ fit the established semantics, the tests for the newly valid syntax _should_ be able to be slotted into the existing tests. If not, why?

QubitPi:
Thanks @archolewa @michael-mclawhorn @cdeszaq !

Issue #693, opened 2018-05-10
Class re-organization


QubitPi:
`Granularity` interface exists in `druid.query` package, but all implementations of `Granularity` exist in a different package.

Similarly , `*ApiRequest` interfaces and their impls are also in separate packages.

QubitPi:
Addressed by https://github.com/yahoo/fili/pull/694

PR #694, opened 2018-05-10
Class re-organization

QubitPi:
Addresses https://github.com/yahoo/fili/issues/693


michael-mclawhorn:
I can't argue that it doesn't make more sense there.
Needs a changelog.

QubitPi:
Thanks, @michael-mclawhorn !
Thank you for the approval, @cdeszaq !

Issue #695, opened 2018-05-14
Javadoc for AbstractDruidQuery is off


QubitPi:

QubitPi:
Addressed by https://github.com/yahoo/fili/pull/696

PR #696, opened 2018-05-14
Fix Javadoc of AbstractDruidQuery

ghost:
Address issue https://github.com/yahoo/fili/issues/695


QubitPi:
@michael-mclawhorn Thanks!

PR #697, opened 2018-05-16
Latest time macro

deepakb91:


fili-core/src/test/groovy/com/yahoo/bard/webservice/util/TableUtilsSpec.groovy
```diff
@@ -153,4 +154,58 @@ class TableUtilsSpec extends  Specification {
...
+        AvailabilityTestingUtils.TestAvailability availability2 = new AvailabilityTestingUtils.TestAvailability(
+                [Mock(DataSourceName)] as Set,
+                ["availability": [interval2] as Set]
+        )
+
+        PhysicalTableSchema physicalTableSchema = Mock(PhysicalTableSchema)
+        physicalTableSchema.getPhysicalColumnName(_ as String) >> ""
+        physicalTableSchema.getColumns() >> Collections.emptySet()
+
+        SimplifiedIntervalList simplifiedIntervalList1 = Mock(SimplifiedIntervalList)
```
tarrantzhang: Instead of creating a Mock of `SimplifiedIntervalList`, should just create two new `SimplifiedIntervalList`
```
        SimplifiedIntervalList simplifiedIntervalList1 = new SimplifiedIntervalList([interval1] as Set)
        SimplifiedIntervalList simplifiedIntervalList2 = new SimplifiedIntervalList([interval2] as Set)
```

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/TableUtilsSpec.groovy
```diff
@@ -153,4 +154,58 @@ class TableUtilsSpec extends  Specification {
...
+        Map<Column, SimplifiedIntervalList> intervalList1 = new HashMap<>()
+        intervalList1.put(column1, simplifiedIntervalList1)
+
+        Map<Column, SimplifiedIntervalList> intervalList2 = new HashMap<>()
+        intervalList2.put(column2, simplifiedIntervalList2)
+
+        ConfigPhysicalTable configPhysicalTable1 = Mock(ConfigPhysicalTable)
+        ConfigPhysicalTable configPhysicalTable2 = Mock(ConfigPhysicalTable)
+        configPhysicalTable1.getAvailability() >> availability1
+        configPhysicalTable1.getAvailableIntervals() >> intervalList1
```
tarrantzhang: Instead of `getAvailableIntervals()`, should use `getAllAvailableIntervals()` as it was used [here](https://github.com/yahoo/fili/blob/468268340368ff2c498e072e804c89c82eb0ec02/fili-core/src/main/java/com/yahoo/bard/webservice/util/TableUtils.java#L109)

michael-mclawhorn:
The README comments on this PR were inaccurate and inadequate.

Michael McLawhorn
Senior Software Development Engineer
o: (217) 255-4215
 

    On Wednesday, May 16, 2018, 5:37:25 PM CDT, deepakb91 <notifications@github.com> wrote:  
 
 
Merged #697.

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequestImpl.java
```diff
@@ -294,10 +295,16 @@ public DataApiRequestImpl(
...
         }
-
         DateTimeFormatter dateTimeFormatter = generateDateTimeFormatter(timeZone);
 
-        this.intervals = generateIntervals(intervals, this.granularity, dateTimeFormatter);
+        if (BardFeatureFlag.CURRENT_MACRO_USES_LATEST.isOn()) {
+            DateTime firstUnavailableInstant = TableUtils.logicalTableAvailability(getTable()).getLast().getEnd();
+            DateTime adjustedNow =  firstUnavailableInstant.isBeforeNow() ? firstUnavailableInstant : new DateTime();
+
+            this.intervals = generateIntervals(adjustedNow, intervals, this.granularity, dateTimeFormatter);
```
cdeszaq: Where's the test for this? And the documentation for what the feature flag is doing?

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequestImpl.java
```diff
@@ -294,10 +295,16 @@ public DataApiRequestImpl(
             LOG.debug(TABLE_UNDEFINED.logFormat(tableName));
             throw new BadApiRequestException(TABLE_UNDEFINED.format(tableName));
         }
-
         DateTimeFormatter dateTimeFormatter = generateDateTimeFormatter(timeZone);
 
-        this.intervals = generateIntervals(intervals, this.granularity, dateTimeFormatter);
+        if (BardFeatureFlag.CURRENT_MACRO_USES_LATEST.isOn()) {
```
cdeszaq: Why isn't this able to be handled by swapping the `MacroCalculationStrategy` for the `CURRENT` macro? There's already an extension point around macro behavior, so why doesn't it work for this use-case?

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/ApiRequestImpl.java
```diff
@@ -385,6 +385,27 @@ protected void validateMetrics(Set<LogicalMetric> logicalMetrics, LogicalTable t
             String apiIntervalQuery,
             Granularity granularity,
             DateTimeFormatter dateTimeFormatter
+    ) throws BadApiRequestException {
+        return generateIntervals(new DateTime(), apiIntervalQuery, granularity, dateTimeFormatter);
+    }
+
+
+    /**
+     * Extracts the set of intervals from the api request.
```
cdeszaq: How is this version of the method different from the one that doesn't take "now"? Why would the caller choose to use one over the other?

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequestImpl.java
```diff
@@ -294,10 +295,16 @@ public DataApiRequestImpl(
...
+            DateTime firstUnavailableInstant = TableUtils.logicalTableAvailability(getTable()).getLast().getEnd();
```
cdeszaq: How is this correct if it doesn't take in query constraints?

cdeszaq:
@michael-mclawhorn Yes, a bit broader discussion would be good. Time is hard, and a lot of deep thought, sweat, blood, and tears when into Fili's treatment of it, so I would hate for that to be lost or made less reliable.

As for those groups:  
I do still have admin access (yay for personal accounts!), and they appear to still be wide open to the public for posting (no account needed), so there shouldn't be any problem there for the [fili-users](https://groups.google.com/forum/#!forum/fili-users) and [fili-developers](https://groups.google.com/forum/#!forum/fili-developers) forums. I see that you've created a 2nd account on one of them with a new email address, so anyone should be able to do so as well.

There's also the [Core Team](https://groups.google.com/forum/#!forum/fili-core-team) group, which is visible to members-only and does require approval when requesting to join, but no one has requested.

As folks join the groups with any new email addresses[1], I'll give them the same permissions they had before if I recognize who their email likely ties to, so we should be able to re-constitute the group membership. (I've already given your 2nd account ownership privileges, so you should be able to manage whichever forum that was). Once everyone who cares is re-populated, we can clean up accounts and names so everyone doesn't have a 2 after their name ;-)

There's also Gitter if more real-time conversation is useful, though for something like time and macros which can be quite complex, something a bit more long-form is probably useful.

[1]: I recommend using personal accounts for this reason... work contacts are efemeral

PR #707, opened 2018-05-30
Dimension restructure RFC PR.

michael-mclawhorn:
PR to illustrate the proposed dimension restructure RFC

https://github.com/yahoo/fili/issues/713

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionRow.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionRow.java
@@ -4,6 +4,7 @@ package com.yahoo.bard.webservice.data.dimension;
 
 import com.yahoo.bard.webservice.util.StreamUtils;
 
+import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
@@ -15,6 +16,21 @@ import javax.validation.constraints.NotNull;
 public class DimensionRow extends LinkedHashMap<DimensionField, String> implements Comparable<DimensionRow> {
 
     private final String keyValue;
+
+
+    /**
+     * Build a dimension row with a key field value and a map of field values.
+     *
+     * @param key  the keyField for this row
+     * @param dimensionValue  the value for the key
+     */
+    public DimensionRow(@NotNull DimensionField key, String dimensionValue) {
+        super(Collections.singletonMap(key, dimensionValue));
+        if (key == null) {
+            throw new IllegalArgumentException("Missing key " + key);
+        }
+        this.keyValue = dimensionValue;
+    }
 
     /**
      * Build a dimension row with a key field value and a map of field values.
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/MapStoreManager.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/MapStoreManager.java
@@ -11,7 +11,7 @@ import java.util.Map;
 public class MapStoreManager {
 
     // Hold singleton instances by name
-    private static final Map<String, MapStore> MAP_STORES = new HashMap<>();
+    protected static final Map<String, MapStore> MAP_STORES = new HashMap<>();
 
     /**
      * Factory for singleton instances by name.
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/rfc_dimension/Dimension.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/rfc_dimension/Dimension.java
@@ -1,0 +1,56 @@
+// Copyright 2017 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.data.dimension.rfc_dimension;
+
+import com.yahoo.bard.webservice.druid.serializers.DimensionToDefaultDimensionSpec;
+
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+
+/**
+ * Dimension interface.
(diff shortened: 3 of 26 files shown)


rfc/DimensionRestructure.md
```diff
@@ -0,0 +1,46 @@
+Dimension Restructure RFC
+-------------------------
+
+Dimensions in 0.9 and before have several responsibilities.
+They serve as a config entity for table schemas and representing metadata.  They are used as the hook for query
+serialization.  They serve as a client for dimension value (domain) loading.  They act as a gateway to search
+dimension values.
+
+This refactor is oriented toward simplifying and structuring Dimensions.  The existing Dimension interface will
+continue to provide the core access point for query and schema planning.
```
QubitPi: Good words on dimension. Could we have some intro. on domain?

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/rfc_dimension/Dimension.java
```diff
@@ -0,0 +1,56 @@
...
+     * @return The dimension schema class
+     */
+    DimensionDescriptor getDimensionDescriptor();
+
+    /**
+     * Encapsulates DimensionRows and cardinality
+     *
+     * @return The dimension row storage associated with this dimension.
+     */
+    IndexedDomain getDomain();
```
archolewa: Should this be an `IndexedDomain`? I was under the impression that `Dimension` would only work with `Domain` (of course, implementations could require a more specific type) to allow for dimensions whose domains can't be indexed (i.e. because they're too big).

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/rfc_dimension/Domain.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Get a dimension row given an id.
+     *
+     * @see {@link Dimension#findDimensionRowByKeyValue(String)}
+     *
+     * @param keyValue  key value
+     *
+     * @return a dimension row - returns the first one found if there are multiple, or null if no matching row is found
+     */
+    DimensionRow findDimensionRowByKeyValue(String keyValue);
```
archolewa: Seems to me this should be made optional, unless we have strong backwards compatibility reason not to. The "returns a thing except maybe null" seems exactly what Optional is meant for.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/rfc_dimension/Domain.java
```diff
@@ -0,0 +1,77 @@
...
+    StorageStrategy getStorageStrategy();
+
+    /**
+     * Get the cardinality of the dimension, if reportable.
+     *
+     * @see {@link Dimension#getCardinality()}
+     *
+     * @return cardinality, if measurable, otherwise Optional.empty()
+     */
+    Optional<Integer> getCardinality();
```
archolewa: Should be `OptionalInt`?

archolewa: Personally, I don't care either way. `Optional<Integer>` has more capabilities (it has `map`, and similar methods, where as `OptionalInt` is just a glorified `null` check), though `OptionalInt` does save on autoboxing. The document does say `OptionalInt` though.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionRow.java
```diff
@@ -16,6 +17,21 @@
...
+    /**
+     * Build a dimension row with a key field value and a map of field values.
+     *
+     * @param key  the keyField for this row
+     * @param dimensionValue  the value for the key
+     */
+    public DimensionRow(@NotNull DimensionField key, String dimensionValue) {
+        super(Collections.singletonMap(key, dimensionValue));
+        if (key == null) {
+            throw new IllegalArgumentException("Missing key " + key);
```
QubitPi: `key` will always be `null`. We could change message to `String.format("Missing key for dimension value '%s'", dimensionValue)`

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionRow.java
```diff
@@ -16,6 +17,21 @@
 
     private final String keyValue;
 
+
+    /**
+     * Build a dimension row with a key field value and a map of field values.
```
QubitPi: Looks like this constructor only builds a singleton map. Should we say instead something like "Build a dimension row with a key field value and a **field value**"

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/MapStoreManager.java
```diff
@@ -11,7 +11,7 @@
 public class MapStoreManager {
 
     // Hold singleton instances by name
-    private static final Map<String, MapStore> MAP_STORES = new HashMap<>();
+    protected static final Map<String, MapStore> MAP_STORES = new HashMap<>();
```
QubitPi: Why do we make it `protected`?

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/rfc_dimension/DomainSchema.java
```diff
@@ -0,0 +1,48 @@
+// Copyright 2017 Yahoo Inc.
```
QubitPi: 2018

michael-mclawhorn:
This will definitely have to wait until 0.11 at the earliest

Issue #708, opened 2018-06-01
Luthier - external configuration loading for fili


michael-mclawhorn:
This prototype feature will be to build up a file-based configuration for Fili.

Proposed feature roll out stages:

Phase One:
-----------
1 - Create a 'Luthier' submodule that is a sample application based on the Wikipedia autowiring
2 - Create `Definition` classes for Metrics, Makers, Dimensions, Tables (Logical and Physical)
3 - Create a definition reader to read from JSON to create definitions
4 - Create a loader to produce dictionary entries from the definition sources
5 - Enhance luthier example to autowire code to output definition files as a CLI rather than launching directly from autowire

Phase Two:
-----------
1 - Enhance definition support to extended maker types, metric types, complex metrics
2 - Create Lua library to produce calculated metrics and extended makers

Phase Three:
------------
1 - Enhance definition support to include extended dimension types (registered lookup support)
2- Enhance definition support to include extensible Table types

michael-mclawhorn:
{
  "fieldSets": {
    "default": [
      {
        name: "id",
        description: "Dimension ID",
        tags: [
          "primaryKey"
        ]
      },
      {
        name: "description",
        description: "Dimension description"
      }
    ],
    "carFields": [
      {
        name: "id",
        description: "Dimension ID",
        tags: [
          "primaryKey"
        ]
      },
      {
        name: "make",
        description: "Manufacturer"
      },
      {
        name: "model",
        description: "Model of a car"
      }
      }
    ]
  }
  "dimensions": [
    {
      "name": "dimension1",
      "description": "a dimension implicitly using default fields by leaving the fields parameter out"
    },
    {
      "name": "cars",
      "description": "a dimension explicitly using the car dimension set by name",
      "fields": "carfields"
    },
    {
      "name": "customLocationDimension",
      "description": "a dimension explicitly using a custom field definition",
      "fields": [
        {
          name: "id",
          description: "Geolocation Code",
          tags: [
            "primaryKey",
            "geoType"
          ]
        },
        {
          name: "county",
          description: "County of location"
        },
        {
          name: "state",
          description: "State of location"
        },
        {
          name: "country",
          description: "Country of location"
        }
      ]
    }
  ]
}

michael-mclawhorn:
`{
  "fieldSets": {
    "default": [
      {
        name: "id",
        description: "Dimension ID",
        tags: [
          "primaryKey"
        ]
      },
      {
        name: "description",
        description: "Dimension description"
      }
[...]

PR #719, opened 2018-06-19
Luthier-draft-1

ForestCold:
https://github.com/yahoo/fili/issues/708


luthier/src/main/java/com/yahoo/luthier/webservice/data/config/dimension/DefaultDimensionTemplate.java
```diff
@@ -0,0 +1,151 @@
...
+        }
+        return this.fields.getFieldList().stream()
+                .map(DimensionFieldInfoTemplate::build)
+                .collect(Collectors.toCollection(LinkedHashSet::new));
+    }
+
+    @Override
+    public DimensionName build() {
+        return this::getApiName;
+    }
```
efronbs: Ok, so I think this should NOT be returning a dimension name, but should instead be returning a `DimensionConfig`. 
For this default implementation you should probably be constructing a `DefaultKeyValueStoreDimensionConfig`. That being said, I am not sure how to handle the key value store nor the search provider that are passed into this.

@michael-mclawhorn any suggestions?

luthier/src/main/java/com/yahoo/luthier/webservice/data/config/dimension/DimensionTemplate.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    LinkedHashSet<DimensionField> getFields(Map<String,
+            LinkedHashSet<DimensionFieldInfoTemplate>> fieldDictionary);
+
+    /**
+     * Build a dimensionName instance.
+     *
+     * @return a dimensionName instance
+     */
+    DimensionName build();
```
efronbs: Change this to a `DimensionConfig`

luthier/src/main/java/com/yahoo/luthier/webservice/data/config/dimension/ExternalDimensionsLoader.java
```diff
@@ -0,0 +1,130 @@
...
+                        .map(
+                                dimensionName -> new DefaultKeyValueStoreDimensionConfig(
+                                        dimensionName.build(),
+                                        dimensionName.getApiName(),
+                                        dimensionName.getDescription(),
+                                        dimensionName.getLongName(),
+                                        dimensionName.getCategory(),
+                                        dimensionName.getFields(dimensionConfig.getFieldSets()),
+                                        getDefaultKeyValueStore(dimensionName),
+                                        getDefaultSearchProvider(dimensionName)
```
efronbs: once build is changed, I think you should just be able to call build here instead of creating the DimensionConfig

luthier/src/main/java/com/yahoo/luthier/webservice/data/config/metric/MetricMakerDictionary.java
```diff
@@ -0,0 +1,213 @@
...
+        if ("DimensionDictionary".equals(paramType)) {
+            return dimensionDictionary;
+        }
+        if ("ZonelessTimeGrain".equals(paramType)) {
+            return DefaultTimeGrain.valueOf(maker.getParams().get(paramName));
+        }
+        if ("ArithmeticPostAggregationFunction".equals(paramType)) {
+            return ArithmeticPostAggregation
+                    .ArithmeticPostAggregationFunction
+                    .valueOf(maker.getParams().get(paramName));
```
efronbs: for enum types you should not be doing a string matching. The selected candidate constructor should already have the enum type that you matched to. You should be able to use the enum type from the candidate constructor along with the param value from the config to get the actual enum value to pass to your candidate constructor.

luthier/src/main/java/com/yahoo/luthier/webservice/data/config/metric/MetricMakerDictionary.java
```diff
@@ -0,0 +1,213 @@
...
+        }
+        if ("boolean".equals(paramType)) {
+            return Boolean.parseBoolean(maker.getParams().get(paramName));
+        }
+        if ("int".equals(paramType)) {
+            return Integer.parseInt(maker.getParams().get(paramName));
+        }
+        if ("double".equals(paramType)) {
+            return Double.parseDouble(maker.getParams().get(paramName));
+        }
```
efronbs: Similarly, I'm sure you shouldn't have to do string matching here. There is probably a way to generalize this, which you should go try to figure out. 

We cannot have this generalized because then we are restricting the types of parameters you can pass to metric makers by what is hard coded here. It is very reasonable for somebody to want to pass a string to a metric maker, or even a map or list of strings, which you cannot do here. My instinct is for the template to handle resolving the type, but that may not be right.

luthier/src/main/java/com/yahoo/luthier/webservice/data/config/metric/MetricMakerDictionary.java
```diff
@@ -0,0 +1,213 @@
...
+    public MetricMaker findByName(String metricMakerName) {
+        return nameToMetricMaker.get(metricMakerName.toLowerCase(Locale.ENGLISH));
+    }
+
+    /**
+     * Get all metric makers available in metric maker dictionary.
+     *
+     * @return a set of metric makers
+     */
+    public Set<MetricMaker> findAll() {
```
efronbs: rename to `getAll()`

luthier/src/main/java/com/yahoo/luthier/webservice/data/config/metric/MetricMakerDictionary.java
```diff
@@ -0,0 +1,213 @@
...
+     * @param metricMakers collection of metric makers to add
+     * @return <tt>true</tt> if the dictionary changed as a result of the call
+     */
+    public boolean addAll(Collection<MetricMaker> metricMakers) {
+        boolean flag = false;
+        for (MetricMaker metricMaker : metricMakers) {
+            flag = add(metricMaker.toString(), metricMaker) || flag;
+        }
+        return flag;
+    }
```
efronbs: you might want to short circuit here and throw an error. 

@michael-mclawhorn what do you think?

config/src/utils/metricUtils.lua
```diff
@@ -0,0 +1,186 @@
...
+
+-------------------------------------------------------------------------------
+-- Makers
+-------------------------------------------------------------------------------
+
+--- Parse a group of config makers and add them into a table.
+--
+-- @param makers  A group of makers
+-- @param t  The table for storing makers
+function M.add_makers(makers, t)
```
archolewa: It looks like we support either a table or a list.

Frankly, I think we should get rid of the list approach completely. First off,
we don't want to give people a bunch of different shapes our data could take.
That just makes it harder for them to learn our configuration language, and it
makes it harder for them to do their own operations on their configuration.

Our data structures should take a single shape as much as possible, unless that
single shape inhibits clarity too much, and should be as simple as possible
(though no simpler :D).

Second, I just don't see any advantages that the list approach has over the 
table approach, except for brevity. But that brevity is obtained by sacrificing
clarity, so it's not worth it.

Our goal with this configuration language should be that it's dead simple to 
use, read, and write Lua code to manipulate.

config/src/tables.lua
```diff
@@ -0,0 +1,71 @@
...
+    physicalTables = {},
+    logicalTables = {}
+}
+
+-------------------------------------------------------------------------------
+-- Default
+-------------------------------------------------------------------------------
+
+DEFAULT = {
+    ALL_DIM = {
```
archolewa: Shouldn't we be deriving these from the `metrics` and `dimensions` modules?

Keeping these up to date with those two modules would be a pain.

config/src/tables.lua
```diff
@@ -0,0 +1,71 @@
...
+        "bigThetaSketch"
+    }
+}
+
+-------------------------------------------------------------------------------
+-- Physical Tables
+-------------------------------------------------------------------------------
+
+-- table name = {description, metrics, dimensions, granuality}
+PHYSICALTABLES = {
```
archolewa: This should be a table, not a list.

config/src/tables.lua
```diff
@@ -0,0 +1,71 @@
...
+    wikiticker = {nil, {"added", "delta", "deleted"}, DEFAULT.ALL_DIM, "HOUR"}
+}
+
+-------------------------------------------------------------------------------
+-- Logical Tables
+-------------------------------------------------------------------------------
+
+-- table name = {description, metrics, dimensions, granuality, physicaltable}
+LOGICALTABLES = {
```
archolewa: This shoudl be a table, not a list.

config/src/utils/jsonParser.lua
```diff
@@ -0,0 +1,45 @@
...
+
+local JsonTable = {}
+
+--- Save a lua table to a json file.
+--
+-- @param filepath  A file for saving json
+-- @param table A lua table to be saved as json file
+-- @return the lua table
+JsonTable.save = function(filepath, table)
+    local file = io.open(filepath, "w")
```
archolewa: When `io.open` fails the first parameter is `nil`, and it *also* returns a
second parameter, which is some sort of error message. At the minimum, we
should print out that error message if we fail to open the file.

Though we really should use a
[logging framework](https://github.com/rxi/log.lua), and log it.

fili-generic-example/src/main/resources/applicationConfig.properties
```diff
@@ -8,9 +8,9 @@ moduleDependencies = fili
 
 bard__version = Local Mode
 bard__resource_binder=com.yahoo.wiki.webservice.application.GenericBinderFactory
-bard__druid_broker=http://localhost:8082/druid/v2
+bard__druid_broker=http://attemptcontempt.corp.ne1.yahoo.com:8082/druid/v2
 
-bard__druid_coord=http://localhost:8081/druid/coordinator/v1
+bard__druid_coord=http://attemptcontempt.corp.ne1.yahoo.com:8081/druid/coordinator/v1
```
michael-mclawhorn: These need to be changed/removed

fili-wikipedia-example/src/main/resources/applicationConfig.properties
```diff
@@ -9,9 +9,9 @@ moduleDependencies = fili
 bard__version = Local Mode
 bard__resource_binder=com.yahoo.wiki.webservice.application.WikiBinderFactory
 
-bard__non_ui_druid_broker=http://localhost:8082/druid/v2
-bard__ui_druid_broker=http://localhost:8082/druid/v2
-bard__druid_coord=http://localhost:8081/druid/coordinator/v1
+bard__non_ui_druid_broker=http://attemptcontempt.corp.ne1.yahoo.com:8082/druid/v2
+bard__ui_druid_broker=http://attemptcontempt.corp.ne1.yahoo.com:8082/druid/v2
+bard__druid_coord=http://attemptcontempt.corp.ne1.yahoo.com:8081/druid/coordinator/v1
```
michael-mclawhorn: Ditto

michael-mclawhorn:
Luthier design attempt two about to land.

PR #722, opened 2018-06-21
Minor Improvements on Druid package

QubitPi:


fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/dimension/extractionfunction/RegularExpressionExtractionFunction.java
```diff
@@ -9,7 +9,18 @@
...
+ * <pre>
+ * {@code
+ *     {
+ *         "type": "regex",
+ *         "expr": <regular_expression>
+ *     }
+ * }
+ * </pre>
+ * <b>Note that the {@code index}, {@code replaceMissingValue}, and {@code replaceMissingValueWith} are using default
+ * values and are not included in the serialization<b>.
```
michael-mclawhorn: This is erroring on the missing close b tag.

PR #724, opened 2018-06-21
Inject customizable extraction functions to RegisteredLookupDimension

QubitPi:
Addresses issue https://github.com/yahoo/fili/issues/723

Problem
======

We have a business requirement to return different default value for different registered lookup. To be more specific, here is an example of serialized registered lookup extraction function

```json
{
    "type" : "registeredLookup",
    "lookup" : "<lookup name>",
    "retainMissingValue" : false,
    "replaceMissingValueWith" : "<default lookup value>",
    "injective" : false,
    "optimize" : true
}
```

API supports lookups with user-defined lookup name as in `"lookup" : "<lookup name>"`. It is, however, not possible for downstream projects to specify the rest of attributes. We need a different default lookup value as in ` "replaceMissingValueWith" : "<default lookup value>"` for each lookup extraction function.

Solution
======

The cause of this issue is in `DefaultRegisteredLookupDimensionConfig`, we only pass in a list of lookup names. If, instead, we pass in a list of `RegisteredLookupExtractionFunction`, the downstream projects could freely define all of the attributes for a registered lookup extraction function and then configure them in. 

This PR essentially changes `List<String> lookups` to `List<ExtractionFunction> registeredLookupExtractionFns` in `DefaultRegisteredLookupDimensionConfig`.

@michael-mclawhorn This might have impact on https://github.com/yahoo/fili/pull/719

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/RegisteredLookupDimensionConfig.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/RegisteredLookupDimensionConfig.java
@@ -18,9 +18,9 @@ public interface RegisteredLookupDimensionConfig extends DimensionConfig {
     }
 
     /**
-     * Returns a list of extraction function for the lookup dimension values.
+     * Returns a list of extraction functions for the lookup dimension values.
      *
-     * @return the list of extraction function for the lookup dimension values
+     * @return the list of extraction functions for the lookup dimension values
      */
     List<ExtractionFunction> getRegisteredLookupExtractionFns();
 }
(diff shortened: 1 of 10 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/config/dimension/RegisteredLookupDimensionConfig.java
```diff
@@ -17,9 +18,9 @@ default Class getType() {
     }
 
     /**
-     * Returns a list of lookups used to configure the Lookup dimension.
+     * Returns a list of extraction function for the lookup dimension values.
```
michael-mclawhorn: (sp)  extraction functions

QubitPi:
Thank you for the review and approvals, @michael-mclawhorn @efronbs !

PR #739, opened 2018-07-11
fix resource retrieval error message to print info.

efronbs:


fili-core/src/main/java/com/yahoo/bard/webservice/web/ErrorMessageFormat.java
```diff
@@ -154,7 +154,7 @@
     ),
     LOG_UNABLE_TO_DESERIALIZE("Unable to deserialize results for job %s from %s"),
 
-    RESOURCE_RETRIEVAL_FAILURE("Unable to retrieve the resource for given resource name: %s."),
+    RESOURCE_RETRIEVAL_FAILURE("Unable to retrieve the resource for given resource name: %s. info: %s"),
```
michael-mclawhorn: Since the sentence ends with the period, Info ought to be capitalized.  Or you could end with a comma or semicolon.

Issue #740, opened 2018-07-12
Making Custom Sketch operations possible in PostAggreations


deepakb91:
1. The `asSketchEstimate(SketchAggregation candidate)` of `ThetaSketchFieldConverter` class currently takes `SketchAggregation` as the parameter and returns `SketchAggregation` as a `ThetaSketchEstimatePostAggregation `. 
2. Rather than taking `SketchAggregation` as a parameter, we can make two versions of `asSketchEstimate` with one taking `Aggregation` as a parameter and the other taking `PostAggregation` as a parameter so that only the metrics which are part of `Aggregation` are marked as field access whereas the ones which are part of `PostAggregation` are not. 
3. This will help in doing some custom sketch operations in the PostAggregations.

QubitPi:
Resolved by https://github.com/yahoo/fili/pull/741

PR #741, opened 2018-07-12
making custom sketch operation possible in PostAggregations

deepakb91:
This PR addresses the issue [740](https://github.com/yahoo/fili/issues/740).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,9 +30,9 @@ pull request if there was one.
 
 ### Changed:
 
-- [Making Custom Sketch operations possible in PostAggreations](https://github.com/yahoo/fili/issues/740)
-    * Removed `FuzzySetPostAggregation asSketchEstimate(SketchAggregation candidate)` method from `FieldConverters` interface.
-    * Added `ThetaSketchEstimatePostAggregation asSketchEstimate(Aggregation candidate)` and `ThetaSketchEstimatePostAggregation asSketchEstimate(PostAggregation postAggregation)` to the class `ThetaSketchFieldConverter`.  
+- [Making Custom Sketch operations possible in PostAggreations](https://github.com/yahoo/fili/issues/740)    
+    * Added `PostAggregation` and `Aggregation` instance check in `asSketchEstimate(MetricField field)` method of the class `ThetaSketchFieldConverter`.
+    * `FieldAccessorPostAggregation` is called only for `Aggregation` and not for `PostAggregation`. 
 
 - [Let DimensionApiRequestMapper throw RequestValidationException instead of BadApiRequestException](https://github.com/yahoo/fili/pull/715)
     * `DimensionApiRequestMapper.apply()` is made to obey the interface contract by throwing
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,10 @@ pull request if there was one.
 
 ### Changed:
 
+- [Making Custom Sketch operations possible in PostAggreations](https://github.com/yahoo/fili/issues/740)
+    * Removed `FuzzySetPostAggregation asSketchEstimate(SketchAggregation candidate)` method from `FieldConverters` interface.
+    * Added `ThetaSketchEstimatePostAggregation asSketchEstimate(Aggregation candidate)` and `ThetaSketchEstimatePostAggregation asSketchEstimate(PostAggregation postAggregation)` to the class `ThetaSketchFieldConverter`.  
+
 - [Let DimensionApiRequestMapper throw RequestValidationException instead of BadApiRequestException](https://github.com/yahoo/fili/pull/715)
     * `DimensionApiRequestMapper.apply()` is made to obey the interface contract by throwing
       `RequestValidationException` instead of ``BadApiRequestException``
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/util/SketchFieldConverter.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/util/SketchFieldConverter.java
@@ -35,13 +35,7 @@ public class SketchFieldConverter implements FieldConverters {
         return new SketchMergeAggregation(candidate.getName(), candidate.getFieldName(), candidate.getSize());
     }
 
-    /**
-     * Get the candidate PostAggregation as a SketchEstimatePostAggregation.
-     *
-     * @param candidate  PostAggregation to "convert"
-     *
-     * @return The PostAggregation as a SketchEstimatePostAggregation
-     */
+    @Override
     public SketchEstimatePostAggregation asSketchEstimate(SketchAggregation candidate) {
         String name = candidate.getName() + "_estimate";
         PostAggregation field = new FieldAccessorPostAggregation(candidate);
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/util/SketchFieldConverter.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/util/SketchFieldConverter.java
@@ -35,7 +35,13 @@ public class SketchFieldConverter implements FieldConverters {
         return new SketchMergeAggregation(candidate.getName(), candidate.getFieldName(), candidate.getSize());
     }
 
-    @Override
+    /**
+     * Get the candidate PostAggregation as a SketchEstimatePostAggregation.
+     *
+     * @param candidate  PostAggregation to "convert"
+     *
+     * @return The PostAggregation as a SketchEstimatePostAggregation
+     */
     public SketchEstimatePostAggregation asSketchEstimate(SketchAggregation candidate) {
         String name = candidate.getName() + "_estimate";
         PostAggregation field = new FieldAccessorPostAggregation(candidate);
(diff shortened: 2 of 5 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/druid/util/ThetaSketchFieldConverter.java
```diff
@@ -30,21 +31,45 @@ public  ThetaSketchAggregation asInnerSketch(SketchAggregation candidate) {
...
         if (field.isSketch()) {
             if (field instanceof PostAggregation) {
-                PostAggregation pa = (PostAggregation) field;
-                return new ThetaSketchEstimatePostAggregation(field.getName() + "_estimate", pa);
+                return asSketchEstimate((PostAggregation) field);
+            }
+            else if (field instanceof Aggregation) {
+                return asSketchEstimate((Aggregation) field);
+            } else {
+                String message = "Given metric field " + field.toString() + " is neither a type of " +
```
QubitPi: Consider `String.format()` instead of string concat for readability

fili-core/src/main/java/com/yahoo/bard/webservice/druid/util/ThetaSketchFieldConverter.java
```diff
@@ -30,21 +31,45 @@ public  ThetaSketchAggregation asInnerSketch(SketchAggregation candidate) {
...
+     * Get the PostAggregation as a SketchEstimatePostAggregation.
+     *
+     * @param postAggregation  PostAggregation to "convert"
+     *
+     * @return The PostAggregation as a SketchEstimatePostAggregation
+     */
+    public ThetaSketchEstimatePostAggregation asSketchEstimate(PostAggregation postAggregation) {
+        String name = postAggregation.getName() + "_estimate";
+        return new ThetaSketchEstimatePostAggregation(name, postAggregation);
+    }
```
QubitPi: I'm little concerned about these three overloadings:

`asSketchEstimate(Aggregation candidate)`
`asSketchEstimate(PostAggregation postAggregation)`
`asSketchEstimate(MetricField field)`

Given that both `Aggregation` and `PostAggregation` implements `MetricField`, consider this case:

```java
List<MetricField> fields = [new xyzAggregation(...), new xyzPostAggregation(...), new xyzMetricField(...)]
for (MetricField field : fields) {
    thetaSketchFieldConverter.asSketchEstimate(field)
}
```

I suspect the case above will end up calling `asSketchEstimate(MetricField field)` for all 3 iterations although `xyzPostAggregation` should really call `asSketchEstimate(PostAggregation postAggregation)` and `xyzAggregation` should really call `asSketchEstimate(Aggregation candidate)`

The choice of which overloading to invoke is made at compile time. For all three iterations of the loop, the compile-time type of the parameter is the same: `MetricField`. The runtime type is different in each iteration, but this does not affect the choice of overloading. Because the compile-time type of the parameter is `MetricField`, the only applicable overloading is the third one, `asSketchEstimate(MetricField field)`, and this overloading is invoked in each iteration of the loop.

__One solution is not exporting two overloadings with the same number of parameters.__

fili-core/src/main/java/com/yahoo/bard/webservice/druid/util/SketchFieldConverter.java
```diff
@@ -35,7 +35,13 @@ public SketchMergeAggregation asInnerSketch(SketchAggregation candidate) {
...
     }
 
-    @Override
+    /**
+     * Get the candidate PostAggregation as a SketchEstimatePostAggregation.
+     *
+     * @param candidate  PostAggregation to "convert"
+     *
+     * @return The PostAggregation as a SketchEstimatePostAggregation
+     */
```
QubitPi: If we deprecate the interface method, please put `@Override` back 😃

CHANGELOG.md
```diff
@@ -26,6 +26,10 @@ pull request if there was one.
 
 ### Changed:
 
+- [Making Custom Sketch operations possible in PostAggreations](https://github.com/yahoo/fili/issues/740)
+    * Removed `FuzzySetPostAggregation asSketchEstimate(SketchAggregation candidate)` method from `FieldConverters` interface.
+    * Added `ThetaSketchEstimatePostAggregation asSketchEstimate(Aggregation candidate)` and `ThetaSketchEstimatePostAggregation asSketchEstimate(PostAggregation postAggregation)` to the class `ThetaSketchFieldConverter`.
```
QubitPi: I assume each line in changelog file should also not be longer than 120 chars. @michael-mclawhorn What do you think?

fili-core/src/main/java/com/yahoo/bard/webservice/druid/util/ThetaSketchFieldConverter.java
```diff
@@ -30,21 +31,45 @@ public  ThetaSketchAggregation asInnerSketch(SketchAggregation candidate) {
...
 
+    /**
+     * Get the PostAggregation as a SketchEstimatePostAggregation.
+     *
+     * @param postAggregation  PostAggregation to "convert"
+     *
+     * @return The PostAggregation as a SketchEstimatePostAggregation
+     */
+    public ThetaSketchEstimatePostAggregation asSketchEstimate(PostAggregation postAggregation) {
+        String name = postAggregation.getName() + "_estimate";
```
QubitPi: Consider `String.format()` instead of string concat for readability.

QubitPi: This line can be inlined with the next `return` statement.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/util/ThetaSketchFieldConverter.java
```diff
@@ -30,21 +31,45 @@ public  ThetaSketchAggregation asInnerSketch(SketchAggregation candidate) {
...
+        return new ThetaSketchEstimatePostAggregation(name, postAggregation);
+    }
+
     @Override
     public FuzzySetPostAggregation asSketchEstimate(MetricField field) {
         if (field.isSketch()) {
```
QubitPi: This is my personal preference and you can decide whether to take it or not:

I prefer reverting the condition like this

```java
if (!field.isSketch()) {
    String message = "Given metric field " + field.toString() + " isn't sketch";
    LOG.error(message);
    throw new IllegalArgumentException(message);
}

if (field instanceof PostAggregation) {
    return asSketchEstimate((PostAggregation) field);
}
else if (field instanceof Aggregation) {
    return asSketchEstimate((Aggregation) field);
} else {
    String message = "Given metric field " + field.toString() + " is neither a type of " +
            "PostAggregation nor Aggregation";
    LOG.error(message);
    throw new IllegalArgumentException(message);
}
```

This moves more complicated logic out of if branch, making the code easier to read.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/util/ThetaSketchFieldConverter.java
```diff
@@ -30,21 +31,45 @@ public  ThetaSketchAggregation asInnerSketch(SketchAggregation candidate) {
         return new ThetaSketchAggregation(candidate.getName(), candidate.getFieldName(), candidate.getSize());
     }
 
-   @Override
```
QubitPi: Same

fili-core/src/main/java/com/yahoo/bard/webservice/druid/util/ThetaSketchFieldConverter.java
```diff
@@ -39,15 +40,23 @@ public ThetaSketchEstimatePostAggregation asSketchEstimate(SketchAggregation can
...
+        if (!field.isSketch()) {
+            String message = String.format("Given metric field %s isn't sketch", field.toString());
+            LOG.error(message);
+            throw new IllegalArgumentException(message);
+        }
+
+        if (field instanceof PostAggregation) {
+            return new ThetaSketchEstimatePostAggregation(field.getName() + "_estimate", (PostAggregation) field);
+        }
+        else if (field instanceof Aggregation) {
```
QubitPi: `else` should be bump to prev. line

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/config/metric/makers/MetricMakerSpec.groovy
```diff
@@ -200,14 +228,26 @@ class MetricMakerSpec extends Specification {
...
+
+    @Unroll
+    def "asSketchEstimate() throws IllegalArgumentException for sketch type which is neither a type of PostAggregation nor Aggregation"() {
+        when:
+        CONVERTER.asSketchEstimate(squareMetricWithSketch)
+
+        then:
+        final IllegalArgumentException exception = thrown()
+        exception.message == 'Given metric field ' + squareMetricWithSketch + ' is neither a type of PostAggregation nor Aggregation'
+
```
QubitPi: Remove extra line

PR #742, opened 2018-07-13
Add injection point for handling exceptions

archolewa:
-- Before we hard-coded how we handled exceptions in the Servlets.
For example, by default we return a 400 - Bad Request if we get an
Exception. However, this is very inflexible. For example, suppose
customers have some customizations (like their own SearchProviders) that
may generate errors the DataServlet doesn't know to look for, but aren't
the fault of the user. Then, Fili currently returns a 400, even though
it should probably be some sort of 5xx.

So we add an injection point to all servlets that allows customers to
provide their own logic for handling DataServlet errors.


fili-core/src/main/java/com/yahoo/bard/webservice/exception/ExceptionHandler.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.core.UriInfo;
+
+/**
+ * Allows a customer to inject custom code for handling Exceptions in Fili.
+ *
+ * For example, customers may want to return 500's for some custom exceptions, or return
+ * an empty set when dimension rows are not found.
+ */
+public interface ExceptionHandler {
```
QubitPi: Where is this interface used?

fili-core/src/main/java/com/yahoo/bard/webservice/exception/DataExceptionHandler.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.core.UriInfo;
+
+/**
+ * Allows a customer to inject custom code for handling Exceptions in Fili's DataServlet.
+ *
+ * For example, customers may want to return 500's for some custom exceptions, or return
+ * an empty set when dimension rows are not found.
+ */
+public interface DataExceptionHandler {
```
QubitPi: Why do we make `DataExceptionHandler`  and `MetadataExceptionHandler` separate?

fili-core/src/main/java/com/yahoo/bard/webservice/exception/ExceptionHandler.java
michael-mclawhorn: This is orphaned?

fili-core/src/main/java/com/yahoo/bard/webservice/exception/MetadataExceptionHandler.java
```diff
@@ -0,0 +1,41 @@
...
+     * during request construction
+     * @param uriInfo  The URI of the request
+     * @param requestContext  HTTP request context for the request
+     * @param metadataEntityName  The name of the metadata entity (dimension, metric, etc) being accessed
+     *
+     * @return The Response to send to the user
+     */
+    Response handleThrowable(
+            Throwable e,
+            Optional<ApiRequest> request,
```
michael-mclawhorn: Should this be ? extends ApiRequest allowing the subclasses to be typesafer?

fili-core/src/main/java/com/yahoo/bard/webservice/exception/ExceptionHandler.java
archolewa: This is vestigial, and should be removed.

fili-core/src/main/java/com/yahoo/bard/webservice/exception/DataExceptionHandler.java
archolewa: Because of how the two servlets work. The DataServlet uses asynchronous responses, so we need to pass in the asyncResponse for this method to resume.

The metadata servlets meanwhile use synchronous responses (if you look at their `get*` methods, you'll see  that they return a `Response`), so we need to do the same thing in the Metadata interface.

Kinda hard to put both of those behaviors into one interface.

fili-core/src/main/java/com/yahoo/bard/webservice/exception/MetadataExceptionHandler.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.ws.rs.core.UriInfo;
+
+/**
+ * Like {@link DataExceptionHandler}, but handles the metadata endpoints.
+ *
+ * The data and metadata endpoints have very different response models (asynchronous vs. synchronous respectively). As
+ * a result, there is not a clean way to unify them into a single exception handler. So they each
+ * get their own
+ */
+public interface MetadataExceptionHandler {
```
QubitPi: Maybe we can have a marker interface `ExceptionHandler` whose child interfaces are `MetadataExceptionHandler ` and `DataExceptionHandler`

fili-core/src/main/java/com/yahoo/bard/webservice/exception/FiliDataExceptionHandler.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.Optional;
+
+import javax.ws.rs.container.AsyncResponse;
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.core.UriInfo;
+
+/**
+ * The default implementation of {@link DataExceptionHandler}.
+ *
```
QubitPi: `<p>`

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/DimensionsServlet.java
```diff
@@ -152,19 +152,17 @@ public Response getAllDimensions(
...
             LOG.debug("Dimensions Endpoint Response: {}", response.getEntity());
-            responseSender = () -> response;
-        } catch (RequestValidationException e) {
-            LOG.debug(e.getMessage(), e);
-            responseSender = () -> Response.status(e.getStatus()).entity(e.getErrorHttpMsg()).build();
-        } catch (Error | Exception e) {
-            String msg = ErrorMessageFormat.REQUEST_PROCESSING_EXCEPTION.format(e.getMessage());
-            LOG.error(msg, e);
-            responseSender = () -> Response.status(Status.BAD_REQUEST).entity(msg).build();
+            return response;
```
QubitPi: 👍 I used to have no idea what responseSender is for

fili-core/src/main/java/com/yahoo/bard/webservice/exception/MetadataExceptionHandler.java
archolewa: That sounds like unnecessary infrastructure to me. Since both interfaces 
have such different signatures, we'll still need to either use the 
`(Meta)DataExceptionHandler` in the appropriate Servlet, or we'll have to cast.
So I don't know what such a marker interface buys us.

fili-core/src/main/java/com/yahoo/bard/webservice/exception/FiliJobsExceptionHandler.java
```diff
@@ -41,10 +41,9 @@ public FiliJobsExceptionHandler(ObjectMappersSuite objectMappers) {
     @Override
     public Response handleThrowable(
             Throwable e,
-            Optional<ApiRequest> request,
+            Optional<? extends ApiRequest> request,
```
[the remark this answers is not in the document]
archolewa: I'm not sure what you're asking. So far as I know, you can't make a type more specific when implementing an interface method. Are you suggesting providing a separate interface for each metadata endpoint?

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/DataServlet.java
```diff
@@ -451,17 +452,18 @@ public void getData(
...
+                e,
+                asyncResponse,
+                Optional.ofNullable(apiRequest),
+                containerRequestContext,
+                uriInfo,
+                writer
+            );
+            // Generally, it's expected that implementations of `ExceptionHandler` will resume
+            // the response in every case. This exists so that if someone writes a buggy handler
+            // that fails to resume the response, they at least get *something* back.
```
michael-mclawhorn: The 'just in case' asynch response resume should check if the response is already done and if it /isn't/ send a message indicating that an error was mishandled.  (A 500 error, probably, also passing along the error that was badly handled)

michael-mclawhorn:
This thing may want for a quick test, maybe directly on dataservlet.getData(), initialize with a bunch of Mocks, throw an NPE early in processing the request, and then have an error handler for NPE catch and respond to it correctly.

archolewa:
@michael-mclawhorn I'm hesitant to do a direct test on `getData` like that, because that will be a *very* brittle test. Every time somebody adds a new injection point, that test will break even if it has nothing to do error handling, because we'll need to build a `DataServlet`. Same if/when we add additional query parameters, and modify `getData`. Meanwhile, we already have the [`ErrorDataServletSpec`](https://github.com/yahoo/fili/blob/4b354af34bcc359a159c85db7ac75c141d10e8da/fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/ErrorDataServletSpec.groovy) which forces and tests a variety of error situations in a more stable fashion. 

The only thing the `ErrorServlet` isn't going to test directly (though it does exercise) is the failsafe response if someone misses a step, so that's the only thing that could benefit from such a test.

I'm really hesitant to add a brittle test to test something that doesn't happen in the default implementation. 

That being said, I have no problem adding unit tests for the exception handlers.

archolewa:
Added some tests, as requested by @michael-mclawhorn. Note that I very explicitly did *not* try to merge the tests under a shared specification, despite them having some shared tests. The implementations are all completely separate, and I think the shared implementation are a coincidence, and not because of any inherent relationship. Also, these days I like keeping my tests self-contained (navigating Specification trees is annoying) if I can, and the shared tests are simple enough that I don't think we got a whole lot out of trying to share them anyway.

PR #743, opened 2018-07-16
Change BasePhysicalTable to defer getAvailableIntervals() to availability

efronbs:
Issue: https://github.com/yahoo/fili/issues/744


michael-mclawhorn:
Needs and issue and a changelog.

PR #745, opened 2018-07-17
Change BaseCompositePhysicalTable to a concrete class.

efronbs:
Issue https://github.com/yahoo/fili/issues/746

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,9 @@ pull request if there was one.
 
 ### Changed:
 
+- [Change `BaseCompositePhysicalTable` into a concrete class]()
+    * Currently `BaseCompositePhysicalTable` is an abstract class, though it has all of the functionality for a simple composite physical table. Changed to a concrete class to allow for simple composite table behavior with requiring an extension.  
+
 - [Change availability behavior on BasePhysicalTable](https://github.com/yahoo/fili/pull/743)
     * Currently `BasePhysicalTable` overrides `getAvailableIntervals(constraint)` and `getAllAvailableIntervals()`, and defers this behavior to its availability. This PR changes `BasePhysicalTable` to also override `getAvailableIntervals()` and defer to its availability.  
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ pull request if there was one.
 
 ### Changed:
 
-- [Change `BaseCompositePhysicalTable` into a concrete class and deprecate now unneeded classes](https://github.com/yahoo/fili/pull/745)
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -30,6 +30,9 @@ pull request if there was one.
 
 ### Changed:
 
+- [Change `BaseCompositePhysicalTable` into a concrete class]()
```
QubitPi: Issue link missing

fili-core/src/main/java/com/yahoo/bard/webservice/table/BaseCompositePhysicalTable.java
```diff
@@ -19,7 +19,7 @@
 /**
  * An implementation of BasePhysicalTable that contains multiple tables.
  */
-public abstract class BaseCompositePhysicalTable extends BasePhysicalTable {
+public class BaseCompositePhysicalTable extends BasePhysicalTable {
```
QubitPi: If we make this concrete, then we should deprecate `MetricUnionCompositeTable` and `PartitionCompositeTable` as well because all they were doing were constructing `MetricUnionAvailability` and `PartitionAvailability`, respectively.

The following is not a blocker to this PR:

1. `MetricUnionCompositeTableDefinition` doesn't construct `MetricUnionCompositeTable` anymore since it's been deprecated. Instead, the definition should construct `BaseCompositePhysicalTable`
2. Need a builder method for `MetricUnionAvailability` that builds `MetricUnionAvailability` directly
3. Same tasks(1 & 2) for `DimensionListPartitionTableDefinition`(`PartitionCompositeTable`)

CHANGELOG.md
```diff
@@ -30,7 +30,7 @@ pull request if there was one.
 
 ### Changed:
 
-- [Change `BaseCompositePhysicalTable` into a concrete class]()
+- [Change `BaseCompositePhysicalTable` into a concrete class](https://github.com/yahoo/fili/pull/745)
```
QubitPi: Need an entry to Deprecated section

CHANGELOG.md
```diff
@@ -30,6 +30,10 @@ pull request if there was one.
 
 ### Changed:
 
+- [Change `BaseCompositePhysicalTable` into a concrete class](https://github.com/yahoo/fili/pull/745)
+    * Currently `BaseCompositePhysicalTable` is an abstract class, though it has all of the functionality for a simple composite physical table. Changed to a concrete class to allow for simple composite table behavior with requiring an extension.
+    * Two simple implementations of `BaseCompositePhysicalTable`, `PartitionCompositeTable` and `MetricUnionCompositeTable` are now deprecated. Instead, the availabilities for these tables should be created directly and passed in to `BaseCompositePhysicalTable`.
```
QubitPi: The 2nd * can be deleted

PR #757, opened 2018-08-09
Adding withDoFork to LookBackQuery

### LINKED ISSUE #756: lookback query does not bump the query id inside withIntervals
`withIntervals` method of `LookbackQuery` currently calls the `withDataSource` method which passes `false` for the `doFork` parameter. This will not increment the query id and therefore causes inconsistency in the logging.

deepakb91:
Fixes #756

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/LookbackQuery.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/LookbackQuery.java
@@ -266,20 +266,7 @@ public class LookbackQuery extends AbstractDruidAggregationQuery<LookbackQuery> 
 
     @Override
     public LookbackQuery withIntervals(Collection<Interval> intervals) {
-        return new LookbackQuery(
-                new QueryDataSource(getInnerQueryUnchecked().withIntervals(intervals)),
-                granularity,
-                filter,
-                aggregations,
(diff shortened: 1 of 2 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/LookbackQuery.java
```diff
@@ -266,7 +266,20 @@ public LookbackQuery withFilter(Filter filter) {
...
+                aggregations,
+                postAggregations,
+                intervals,
+                context,
+                true,
+                lookbackOffsets,
+                lookbackPrefixes,
+                having,
+                limitSpec
+        );
```
QubitPi: I believe a `wither` is supposed to mutate a single attribute of an object.

Could we have a new `wither` for `doFork`? You can chain the two withers together for your purpose

ghost (changes requested):

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/LookbackQuery.java
deepakb91: Added a new `wither` and modified the change log

ghost (approved):
Change log entry title needs update

PR #759, opened 2018-08-10
`AsyncWebServiceRequestHandler` routes to service.

archolewa:
-- There have been some rumblings of a need to support being able to
choose between one of several different druid clusters. For example,
this may be used to detect bad data and automatically route to a
different cluster with better data. Or it may be used to put two
different druid clusters behind the same API, and route to one or the
other depending on the shape of the query.

-- This PR implements this by introducing a `DruidWebServiceSelector` to
the `AsyncWebServiceRequestHandler`. The `DruidWebServiceSelector` is a
SAM that takes the druid query, api request, and request context and returns
the `DruidWebservice` to send the query to.


michael-mclawhorn:
Needs ChangeLog, Needs Issue.  
Seems fine otherwise.

michael-mclawhorn:
@archolewa Still breaking

archolewa:
Provides the minimal hooks needed to implement: https://github.com/yahoo/fili/issues/798

fili-core/src/main/java/com/yahoo/bard/webservice/web/handlers/AsyncWebServiceRequestHandler.java
```diff
@@ -32,8 +44,7 @@
      * @param mapper  The mapper for all JSON processing
      */
     public AsyncWebServiceRequestHandler(DruidWebService druidWebService, ObjectMapper mapper) {
-        super(mapper);
-        this.druidWebService = druidWebService;
+        this((i1, i2, i3) -> druidWebService, mapper);
```
michael-mclawhorn: Maybe touch the @param druidWebservice  comment to say 'A single webservice for handling all requests' to draw distinction from the selector constructor format.

PR #761, opened 2018-08-16
Ignore empty post-aggs serialization

QubitPi:
Address issue https://github.com/yahoo/fili/issues/762

If API request does not need any post aggs, then serialized query will not include it in this PR


michael-mclawhorn:
Does this matter?  I mean, as a change, it's fine.  But it seems pretty fine grain.

michael-mclawhorn:
To elaborate, this is potentially going to break a bunch of tests.

QubitPi:
I personally dislike the empty serialization but we can close it if we really don't want it

PR #766, opened 2018-08-24
Wrapping DruidInFilterBuilder as default filter builder under a feature flag

### LINKED ISSUE #765: Making InFilter has the default druid filter breaks downstream projects
`buildDruidFilterBuilder` method in `AbstractBinderFactory` class constructs a `DruidInFilterBuilder` by default. This breaks the downstream projects if they do not handle InFilters explicitly.

deepakb91:
Fixed #765


michael-mclawhorn:
We will want to switch default state on this property in the next major release, so make sure that you schedule fixing support for the 'in' filter.

QubitPi:
@michael-mclawhorn I'll fix it

PR #776, opened 2018-09-20
Api request cleanups

michael-mclawhorn:


fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/ApiRequestImpl.java
```diff
@@ -680,12 +600,11 @@ public ResponseFormatType getFormat() {
...
 
     @Override
-    public long getAsyncAfter() {
+    public Long getAsyncAfter() {
         return asyncAfter;
     }
 
-    @Override
-    public PaginationParameters getDefaultPagination() {
+    public static PaginationParameters getDefaultPagination() {
```
archolewa: This is dangerous. We're not using `getDefaultPagination` anywhere, let alone overriding, so I don't care that much, but we'd probably be better off deprecating this method, adding a static version, and using the static version, rather than just making this method static.

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequest.java
```diff
@@ -169,7 +171,9 @@
      * @param dimensionDictionary  DimensionDictionary
      *
      * @return Set of filter objects.
+     * @deprecated Use getFilterGenerator.generate instead
```
archolewa: There doesn't appear to be a `getFilterGenerator` method on this interface. Is that intentional?

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DimensionsApiRequestImpl.java
```diff
@@ -167,7 +167,7 @@ private DimensionsApiRequestImpl(
         for (String apiFilter : apiFilters) {
             ApiFilter newFilter;
             try {
-                newFilter = ApiFilterGenerator.build(apiFilter, dimensionDictionary);
+                newFilter = generateFilter(apiFilter, dimensionDictionary);
```
archolewa: I'm confused here. It looks like in the DataApiRequestImpl, we're using the
FilterGenerator to parse the *entire* filterString in one shot. However, in
the DimensionsApiRequestImpl, we're first splitting the filters and then passing
each one in one at a time. Why the difference? Why can't we just pass in the
`filterQuery` string to the filter generator in one shot?

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/binders/FilterBinders.java
```diff
@@ -28,13 +43,87 @@
 import javax.validation.constraints.NotNull;
 
 /**
- * Factory for ApiFilter objects. Has methods to build an ApiFilter object out of the filter clause in a Fili Api
- * request, and producing an ApiFilter by unioning the values of two provided ApiFilters.
+ * This utility class captures default implementations for binding and validating API models for filtering requests.
  */
-public class ApiFilterGenerator {
-    private static final Logger LOG = LoggerFactory.getLogger(ApiFilterGenerator.class);
+public class FilterBinders {
```
archolewa: This rename is dangerous (for example, people may be using the old class to
generate filters). The old class should be deprecated, instead of removed
completely.

michael-mclawhorn: Darn.  I let Ben merge at that location, knowing I'd want to move it in the very next check in.  I hadn't counted on people rebasing in between.

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/ApiRequestImpl.java
michael-mclawhorn: the default pagination is particular to the use case.
In this case the ApiRequests have a configuration driven Default Pagination.

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequest.java
michael-mclawhorn: No, but I'm taking care of it in the next commit.

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DimensionsApiRequestImpl.java
michael-mclawhorn: This is a case where two different refactors were done at different times.  I'll align them.

michael-mclawhorn: Nope.  The filter builder in DimensionApiRequest doesn't bind using a logical table.  The one in DataApiRequest does.

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/ApiRequestImpl.java
michael-mclawhorn: Moving all calls to the method to references to the static instance variable.  There are configurable default paginations and they should probably be centralized.  But not by making this static.  Definitely not by making this static.

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DimensionsApiRequestImpl.java
archolewa: Hmm. Ok, I'm willing to pass on this for now, then. It'd be nice if we could find some way of aligning the interfaces for generating filters (reduces cognitive overhead in my opinion), but it's not worth thinking more deeply about now.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/InFilter.java
```diff
@@ -52,8 +53,8 @@ public InFilter withValues(List<String> values) {
      *
      * @return The list of values to filter on
      */
-    public List<String> getValues() {
-        return Collections.unmodifiableList(values);
+    public Set<String> getValues() {
```
archolewa: Should we have this return a `TreeSet` instead? Do we want to make a consistent
ordering based on `compareTo` a contract on the function, or do we want this to
just be a convenient implementation detail for testing?

fili-core/src/main/java/com/yahoo/bard/webservice/web/FilteredThetaSketchMetricsHelper.java
```diff
@@ -439,11 +466,12 @@ public String generateMetricName(String filterString) {
 
     @Override
     public Set<Dimension> gatherFilterDimensions(Filter filter, Set<Dimension> dimensions) {
-
         if (filter instanceof SelectorFilter) {
             dimensions.add(((SelectorFilter) filter).getDimension());
+        } else if (filter instanceof InFilter) {
```
archolewa: Blegh. Seems to me like this is something that we should pull onto a method
on the Filter interface. A few seconds of poking suggests this could be gnarly
though, so I'm not going to block on it.

But we're probably going to have this problem come up again.

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/ThetaSketchIntersectionReportingResources.groovy
```diff
@@ -176,6 +178,10 @@ class ThetaSketchIntersectionReportingResources extends Specification {
         ArrayNode metricJsonObjArray = mapper.readTree("[{\"filter\":{\"AND\":\"country|id-in[US,IN],property|id-in[114,125]\"},\"name\":\"foo\"},{\"filter\":{},\"name\":\"pageviews\"}]")
         JsonNode jsonobject = metricJsonObjArray.get(0)
         filterObj = jsonobject.get("filter")
+        //InFilter inFilter1 = new InFilter(propertyDim, Lists.asList("114", "125"))
```
archolewa: Why comment this out? Why not just remove it?

archolewa:
A few small comments, but nothing blocking.

michael-mclawhorn:
Actually that approval wiping feature is not turned on for this repo.  Oath, Inc. did turn it on for the internal ones.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/InFilter.java
michael-mclawhorn: I talked to Atul who works on druid and he can't think of a reason that ordering these values all the time wouldn't be a reasonable contract.  So I'll make this change.

fili-core/src/main/java/com/yahoo/bard/webservice/web/FilteredThetaSketchMetricsHelper.java
michael-mclawhorn: The Helper class itself is the first code smell.  The if else chain is the second. Yeah.

PR #777, opened 2018-09-20
DimensionServlet refactor

tarrantzhang:
Extract getting pagination of dimension rows into a protected method from `/<dimensionsname>/values` endpoint. This allows child class to override this logic.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/DimensionsServlet.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/DimensionsServlet.java
@@ -286,13 +286,7 @@ public class DimensionsServlet extends EndpointServlet {
                     .getPaginationParameters()
                     .orElse(apiRequest.getDefaultPagination());
 
-            Pagination<DimensionRow> pagedRows = apiRequest.getFilters().isEmpty() ?
-                    searchProvider.findAllDimensionRowsPaged(paginationParameters) :
-                    searchProvider.findFilteredDimensionRowsPaged(
-                            apiRequest.getFilters(),
-                            paginationParameters
-                    );
-
+            Pagination<DimensionRow> pagedRows = getPagedRows(apiRequest, searchProvider, paginationParameters);
             Response.ResponseBuilder builder = Response.status(Response.Status.OK);
 
             Stream<Map<String, String>> rows = ResponsePaginator.paginate(builder, pagedRows, uriInfo)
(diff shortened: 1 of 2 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/DimensionsServlet.java
```diff
@@ -332,6 +326,27 @@ public Response getDimensionRows(
...
 
+    /***
+     * Get the pagination of dimension rows.
+     *
+     * @param apiRequest The apiRequest
+     * @param searchProvider The searchProvider
+     * @param paginationParameters The pagination parameters
+     * @return pagination of dimensionRow
+     */
+    protected Pagination<DimensionRow> getPagedRows(
```
QubitPi: For public `getDimensionRows()` which invokes this method, the documentation should indicate this overridable method is invoked. By convention, a method that invokes overridable methods contains a description of these invocations at the end of its documentation comment. The description begins with the phrase “This implementation.” This phrase should not be taken to indicate that the behavior may change from release to release. It connotes that the description concerns the inner workings of the method.

QubitPi: @michael-mclawhorn Do you think it makes more sense to move this method to the skeletal implementation which is `EndpointServlet`? Other servlet might find it useful in the future.

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/DimensionsServlet.java
```diff
@@ -332,6 +326,27 @@ public Response getDimensionRows(
         }
     }
 
+    /***
+     * Get the pagination of dimension rows.
```
QubitPi: Description is good.

Detailed description should start in a different paragraph saying "if there is filter in API request, then... otherwise, ..."

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/DimensionsServlet.java
```diff
@@ -332,6 +326,27 @@ public Response getDimensionRows(
...
+     *
+     * @param apiRequest The apiRequest
+     * @param searchProvider The searchProvider
+     * @param paginationParameters The pagination parameters
```
QubitPi: Fili has two spaces between para name and description

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/DimensionsServlet.java
```diff
@@ -332,6 +326,27 @@ public Response getDimensionRows(
...
+     * @return pagination of dimensionRow
```
QubitPi: separate param and return with an empty line

michael-mclawhorn:
One quick piece of paperwork, could you create an issue for this change and then switch the CHANGELOG url to refer to it?

fili-core/src/main/java/com/yahoo/bard/webservice/web/endpoints/DimensionsServlet.java
```diff
@@ -332,6 +330,32 @@ public Response getDimensionRows(
         }
     }
 
+    /***
+     * Get the pagination of dimension rows.
+     * <p>
+     * if there is filter in API request, then apply the filter to the querying dimension rows, otherwise returns all
+     * dimension rows.
+     * </p>
```
QubitPi: Remove `</p>`

PR #779, opened 2018-09-21
Moving tests and internal code more completely onto LogicalMetricInfo…

michael-mclawhorn:
… based metric config.


michael-mclawhorn:
Sure.  I had thought we could do it before that, but I don't remember when we first declared those deprecations.

PR #790, opened 2018-09-26
Organize imports

wcekan:
Simple cleanup to make all imports conform to existing standard.


fili-core/src/test/java/com/yahoo/bard/webservice/application/JerseyTestBinder.java
```diff
@@ -54,6 +52,7 @@
 import java.util.Map;
 import java.util.Set;
 
+import javax.ws.rs.client.Invocation.Builder;
```
QubitPi: Why do we make it a non-static import?

wcekan: That apparently was a mistake.  `Builder` is a class, not a static field so this is correct.

QubitPi: I assume static class and static field are all static members. But this is not a big deal anyway

PR #797, opened 2018-10-01
Enable Druid Bound dimension filter in Fili

RVRSKumar:
Enabling Druid Bound Filter in FILI

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
@@ -8,7 +8,6 @@ import com.yahoo.bard.webservice.util.EnumUtils;
 
 import javax.validation.constraints.Null;
 import java.util.Objects;
-import java.util.Optional;
 
 /**
  * Bound filters supports filtering on ranges of dimension values.
@@ -33,6 +32,11 @@ public class BoundFilter extends DimensionalFilter<BoundFilter> {
             this.orderingType = EnumUtils.enumJsonName(this);
         }
 
+        /**
+         * Get the JSON representation of this class.
+         *
+         * @return the JSON representation.
+         */
         @JsonValue
         public String toJson() {
             return orderingType;
@@ -90,7 +94,7 @@ public class BoundFilter extends DimensionalFilter<BoundFilter> {
     public Ordering getOrdering() {
         return this.ordering;
     }
-    
+
     //CHECKSTYLE:OFF
     @Override
     public BoundFilter withDimension(Dimension dimension) {
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
@@ -1,0 +1,175 @@
+// Copyright 2018 Oath Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.druid.model.filter;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+import com.yahoo.bard.webservice.data.dimension.Dimension;
+import com.yahoo.bard.webservice.util.EnumUtils;
+
+import javax.validation.constraints.Null;
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * Bound filters supports filtering on ranges of dimension values.
+ */
+public class BoundFilter extends DimensionalFilter<BoundFilter> {
+
+    /**
+     * Ordering specified for the range filtering.
+     */
+    public enum Ordering {
+        LEXICOGRAPHIC,
+        ALPHANUMERIC,
+        NUMERIC,
+        STRLEN;
+
+        final String orderingType;
+
+        /**
+         * Constructor.
+         */
+        Ordering() {
+            this.orderingType = EnumUtils.enumJsonName(this);
+        }
+
+        @JsonValue
+        public String toJson() {
+            return orderingType;
+        }
+    }
+
+    private final String lower;
+    private final String upper;
+    private final Boolean lowerStrict;
+    private final Boolean upperStrict;
+    private final Ordering ordering;
+
+    /**
+     * Constructor.
+     *
+     * @param dimension The druid dimension to be filtered
+     * @param lower The lower bound of the dimension value to be filtered (Optional)
+     * @param upper The upper bound of the dimension value to be filtered (Optional)
+     * @param lowerStrict Boolean to enable/ disable strict filtering for lower bounds (Optional)
+     * @param upperStrict Boolean to enable/ disable strict filtering for upper bounds (Optional)
+     * @param ordering The Ordering to be applied for the dimension filtering
+     */
+    public BoundFilter (
+            Dimension dimension,
+            @Null String lower,
+            @Null String upper,
+            @Null Boolean lowerStrict,
+            @Null Boolean upperStrict,
+            @Null Ordering ordering
+    ) {
+        super(dimension, DefaultFilterType.BOUND);
+        this.lower = lower;
+        this.upper = upper;
+        this.lowerStrict = lowerStrict;
+        this.upperStrict = upperStrict;
+        this.ordering = ordering;
+    }
+
+    public String getLower() {
+        return this.lower;
+    }
+
+    public String getUpper() {
+        return this.upper;
+    }
+
+    public Boolean isLowerStrict() {
+        return this.lowerStrict;
+    }
+
+    public Boolean isUpperStrict() {
+        return this.upperStrict;
+    }
+
+    public Ordering getOrdering() {
+        return this.ordering;
+    }
+    
+    //CHECKSTYLE:OFF
+    @Override
+    public BoundFilter withDimension(Dimension dimension) {
+        return new BoundFilter(dimension, null, null, false, false, null);
+    }
+
+    public BoundFilter withLowerBound(String lower) {
+        return new BoundFilter(getDimension(), lower, null, false, false, null);
+    }
+
+    public BoundFilter withUpperBound(String upper) {
+        return new BoundFilter(getDimension(), null, upper, false, false, null);
+    }
+
+    public BoundFilter withLowerAndUpperBound(String lower, String upper) {
+        return new BoundFilter(getDimension(), lower, upper, false, false, null);
+    }
+
+    public BoundFilter withLowerBoundStrict(String lower, Boolean lowerStrict) {
+        return new BoundFilter(getDimension(), lower, null, lowerStrict, false, null);
+    }
+
+    public BoundFilter withUpperBoundStrict(String upper, Boolean upperStrict) {
+        return new BoundFilter(getDimension(), null, upper, false, upperStrict, null);
+    }
+
+    public BoundFilter withLowerBoundStrictUpperBound(String lower, String upper, Boolean lowerStrict) {
+        return new BoundFilter(getDimension(), lower, upper, lowerStrict, null, null);
+    }
+
+    public BoundFilter withLowerBoundUpperBoundStrict(String lower, String upper, Boolean upperStrict) {
+        return new BoundFilter(getDimension(), lower, upper, null, upperStrict, null);
+    }
+
+    public BoundFilter withLowerBoundStrictUpperBoundStrict(String lower, String upper, Boolean lowerStrict, Boolean upperStrict) {
+        return new BoundFilter(getDimension(), lower, upper, lowerStrict, upperStrict, null);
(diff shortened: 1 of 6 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
```diff
@@ -0,0 +1,179 @@
...
+     * Constructor.
+     *
+     * @param dimension The druid dimension to be filtered
+     * @param lower The lower bound of the dimension value to be filtered (Optional)
+     * @param upper The upper bound of the dimension value to be filtered (Optional)
+     * @param lowerStrict Boolean to enable/ disable strict filtering for lower bounds (Optional)
+     * @param upperStrict Boolean to enable/ disable strict filtering for upper bounds (Optional)
+     * @param ordering The Ordering to be applied for the dimension filtering
+     */
+    public BoundFilter (
```
QubitPi: Given that there are optional parameters in this constructor, we should have a separate constructor that accept only required Filter fields and defaults all the optional ones by calling this constructor. That will avoid API callers from doing some "confusing" calls like `BoundFilter(..., false, false, ...)`

Lets put default values for optional parameters in accordance with the default-spec in http://druid.io/docs/latest/querying/filters.html

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
```diff
@@ -0,0 +1,179 @@
...
+            Dimension dimension,
+            @Null String lower,
+            @Null String upper,
+            @Null Boolean lowerStrict,
+            @Null Boolean upperStrict,
```
QubitPi: We prefer prefer primitives to boxed primitives for the following reasons:

* Primitives are generally more time- and space-efficient than boxed primitives
* It avoids the mis-use of `==` operator on boxed primitives
* When primitives and boxed primitives are mixed in a single operation, the boxed primitive is auto-unboxed, resulting in the potential `NullPointerException` runtime error
* Instantiating boxed primitives is slower

The legitimate use cases for boxed primitives in Fili is when

1. they are used as elements, keys, and values in collections, because you cannot put primitives in collections.
2. they are used as type parameters in parameterized types
3. making reflective method invocations

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
```diff
@@ -0,0 +1,179 @@
...
+
+    public Boolean isLowerStrict() {
+        return this.lowerStrict;
+    }
+
+    public Boolean isUpperStrict() {
+        return this.upperStrict;
+    }
+
+    public Ordering getOrdering() {
```
QubitPi: hmm...No Javadoc should trigger checkstyle to fail... @michael-mclawhorn We are doing something wrong in our checkstyle config

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
```diff
@@ -0,0 +1,179 @@
...
+
+    public BoundFilter withLowerBoundStrictUpperBoundOrdering(String lower, String upper, Boolean lowerStrict, Ordering ordering) {
+        return new BoundFilter(getDimension(), lower, upper, lowerStrict, null, ordering);
+    }
+
+    public BoundFilter withLowerBoundUpperBoundStrictOrdering(String lower, String upper, Boolean upperStrict, Ordering ordering) {
+        return new BoundFilter(getDimension(), lower, upper, null, upperStrict, ordering);
+    }
+
+    public BoundFilter withLowerBoundStrictUpperBoundStrictOrdering(String lower, String upper, Boolean lowerStrict, Boolean upperStrict, Ordering ordering) {
```
QubitPi: I'm impressed by this parameter explosion... If we must go like this, could we use some sort of builder pattern?

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
RVRSKumar: http://checkstyle.sourceforge.net/config_javadoc.html#JavadocMethod

Currently for getter methods, we do not need to provide java docs because of this property `allowMissingPropertyJavadoc`.

https://github.com/yahoo/fili/blob/master/checkstyle-style.xml#L186

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
RVRSKumar: Updated

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
```diff
@@ -0,0 +1,179 @@
...
+     * @param dimension The druid dimension to be filtered
+     * @param lower The lower bound of the dimension value to be filtered (Optional)
+     * @param upper The upper bound of the dimension value to be filtered (Optional)
+     * @param lowerStrict Boolean to enable/ disable strict filtering for lower bounds (Optional)
+     * @param upperStrict Boolean to enable/ disable strict filtering for upper bounds (Optional)
+     * @param ordering The Ordering to be applied for the dimension filtering
+     */
+    public BoundFilter (
+            Dimension dimension,
+            @Null String lower,
```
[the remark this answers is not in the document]
RVRSKumar: Updated

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
```diff
@@ -0,0 +1,179 @@
...
+
+/**
+ * Bound filters supports filtering on ranges of dimension values.
+ */
+public class BoundFilter extends DimensionalFilter<BoundFilter> {
+
+    /**
+     * Ordering specified for the range filtering.
+     */
+    public enum Ordering {
```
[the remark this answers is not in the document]
RVRSKumar: Good catch! Made it into a separate ENUM implementation

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
```diff
@@ -0,0 +1,179 @@
+// Copyright 2018 Oath Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.druid.model.filter;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+import com.yahoo.bard.webservice.data.dimension.Dimension;
+import com.yahoo.bard.webservice.util.EnumUtils;
+
+import javax.validation.constraints.Null;
+import java.util.Objects;
```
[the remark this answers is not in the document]
RVRSKumar: Done

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
```diff
@@ -0,0 +1,179 @@
+// Copyright 2018 Oath Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.druid.model.filter;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+import com.yahoo.bard.webservice.data.dimension.Dimension;
+import com.yahoo.bard.webservice.util.EnumUtils;
```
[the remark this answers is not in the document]
RVRSKumar: Done

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
michael-mclawhorn: Let's do a constructor that takes only dimension, upper and lower

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
michael-mclawhorn: Yeah, I missed this too, but we have a lot of simple getters without javadoc in our codebase.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    public BoundFilter withOrdering(Ordering ordering) {
+        return new BoundFilter(getDimension(), getLower(), getUpper(), isLowerStrict(), isUpperStrict(), ordering);
+    }
+    //CHECKSTYLE:ON
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(super.hashCode(), DefaultFilterType.BOUND);
```
michael-mclawhorn: need to include all the fields in the hashcode.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
michael-mclawhorn: need to include all the fields in the equals.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/filter/BoundFilterSpec.groovy
```diff
@@ -0,0 +1,126 @@
...
+        dataSource.getQuery() >> Optional.empty()
+        physicalTable.getPhysicalColumnName(_) >> "foo"
+        druidQuery = Mock(DruidAggregationQuery)
+        dimension.getExtractionFunction() >> Optional.of(extractionFunction)
+
+        druidQuery.getDataSource() >> dataSource
+
+    }
+
+    def "Checking serialization of null values"() {
```
michael-mclawhorn: Should be more of a sentence.
"Null values omitted when serialized."

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/filter/BoundFilterSpec.groovy
```diff
@@ -0,0 +1,126 @@
...
+        given:
+        BoundFilter filter = new BoundFilter(dimension, null, null, null, null, null);
+        druidQuery.getFilter() >> filter
+        String serializedFilter = objectMapper.writeValueAsString(druidQuery)
+
+        expect:
+        objectMapper.readTree(serializedFilter).get("filter") == objectMapper.readTree(expectedSerialization)
+    }
+
+    def "Checking the functioning of Bound Filter"() {
```
michael-mclawhorn: "Sample values serialize correctly" perhaps?

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/filter/BoundFilterSpec.groovy
```diff
@@ -0,0 +1,126 @@
...
+                Ordering.NUMERIC
+        )
+        druidQuery.getFilter() >> filter
+        String serializedFilter = objectMapper.writeValueAsString(druidQuery)
+
+        expect:
+        objectMapper.readTree(serializedFilter).get("filter") == objectMapper.readTree(expectedSerializationBoundFilter)
+    }
+
+    def "Checking the functioning of nestWith()"() {
```
michael-mclawhorn: With methods make updating copies.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.Objects;
+
+/**
+ * Bound filters supports filtering on ranges of dimension values.
+ */
+public class BoundFilter extends DimensionalFilter<BoundFilter> {
+
+    private final String lower;
+    private final String upper;
+    private final Boolean lowerStrict;
```
michael-mclawhorn: javadoc the meaning of the strict fields.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/filter/BoundFilter.java
```diff
@@ -0,0 +1,105 @@
...
+    private final Boolean upperStrict;
+    private final Ordering ordering;
```
michael-mclawhorn: javadoc that default ordering on the remote system is lexicographic.

PR #799, opened 2018-10-04
Version bump PR

michael-mclawhorn:
The 0.11 release
Lots of cleanup in this release.  Not a ton of major new features.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,56 @@ All notable changes to Fili will be documented here. Changes are accumulated as 
 major version. Each change has a link to the pull request that makes the change and to the issue that triggered the
 pull request if there was one.
 
+Current
+-------
+
+### Changed:
+
+### Deprecated:
+
+### Removed:
+
+### Fixed:
+
+### Known Issues:
+
+### Added:
+
+v0.10.48 - 2018/10/04
+=====
+
+0.10 Highlights
+-------------
+
+### Extensibility improvements
+
+A number of classes which were either not extensible or very difficult to extend have been restructured.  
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,56 @@ All notable changes to Fili will be documented here. Changes are accumulated as
...
+A lot of tech debt has been paid down in the form of removing deprecated code and moving code off deprecated methods.
+Old sketch support has been retired in favor of the official community open source version: 
+(https://datasketches.github.io/docs/Theta/ThetaSketchFramework.html) 
+
+Some deprecations have been removed because the migration path off of those deprecated methods seemed less useful
+than simply supporting the older (simpler) contract.
+
+### General cleanup
+
+Some packages have been rationalized together or apart.  Some concrete classes became interfaces and vice-versa.
```
archolewa: I know we don't really have the tools right now to make this easy, but I do think that in the future under the highlights, we should explicitly list everything we expect to be a breaking change (basically, anytime a `@PublicApi` gets changed).

I know we have a `Changed` section, but that includes *all* changes, but not just changes that break the public API, so it's noisy.

michael-mclawhorn: Maybe we should have a section for RFCs with migration impacts?

archolewa: I think a 'known breaking changes' will be sufficient. Or call it 'changed PublicApi', to make it easier to cross-reference with the annotation. RFCs with migration impacts would be fine so long as it doesn't add too much process (we should make it as easy as possible to document breaking changes, even small ones).

PR #807, opened 2018-10-12
Add api bound filter

RVRSKumar:
- Changing the contract of `ApiFilter`
- Constructing a `DruidBoundFilter` class

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidBoundFilterBuilder.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidBoundFilterBuilder.java
@@ -1,0 +1,123 @@
+// Copyright 2018 Oath Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.data.filterbuilders;
+
+import static com.yahoo.bard.webservice.web.ErrorMessageFormat.FILTER_OPERATOR_INVALID;
+import static com.yahoo.bard.webservice.web.ErrorMessageFormat.TOO_MANY_DRUID_FILTERS;
+
+import com.yahoo.bard.webservice.config.SystemConfig;
+import com.yahoo.bard.webservice.config.SystemConfigProvider;
+import com.yahoo.bard.webservice.data.dimension.Dimension;
+import com.yahoo.bard.webservice.data.dimension.DimensionRowNotFoundException;
+import com.yahoo.bard.webservice.druid.model.filter.AndFilter;
+import com.yahoo.bard.webservice.druid.model.filter.BoundFilter;
+import com.yahoo.bard.webservice.druid.model.filter.Filter;
+import com.yahoo.bard.webservice.exception.TooManyDruidFiltersException;
+import com.yahoo.bard.webservice.web.ApiFilter;
+import com.yahoo.bard.webservice.web.DefaultFilterOperation;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * A DruidBoundFilterBuilder builds a Druid Bound Filter for supported dimensions, which can be sent directly to Druid.
+ * <p>
+ *     Druid Bound Filter currently supports the following API filter operations:
+ *          <li>{@link DefaultFilterOperation#gt}</li>
+ *          <li><{@link DefaultFilterOperation#lt}/li>
+ *          <li>{@link DefaultFilterOperation#gte}</li>
+ *          <li>{@link DefaultFilterOperation#lte}</li>
+ *          <li>{@link DefaultFilterOperation#between}</li>
+ * </p>
+ * <p>
+ *     So the filter:
+ *     {@code startDate|id-gt[2018-10-10]}
+ *     is translated into
+ *     {@code DruidBoundFilterBuilder(startDate,2018-10-10,null)}
+ *     which is serialized as following to druid
+ *     <pre><code>
+ *         {
+ *             "type": "bound",
+ *             "dimension": "startDate",
+ *             "lower": "2018-10-10",
+ *             "lowerStrict": false,
+ *             "upper": null ,
+ *             "upperStrict": false,
+ *             "ordering": null }
+ *         }
+ *     </code></pre>
+ * </p>
+ */
+public class DruidBoundFilterBuilder implements DruidFilterBuilder {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ConjunctionDruidFilterBuilder.class);
+    private static final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance();
+
+    private static final int DEFAULT_MAX_NUM_DRUID_FILTERS = 10000;
+    private static final int MAX_NUM_DRUID_FILTERS = SYSTEM_CONFIG.getIntProperty(
+            SYSTEM_CONFIG.getPackageVariableName("max_num_druid_filters"),
+            DEFAULT_MAX_NUM_DRUID_FILTERS
+    );
+
+    @Override
+    public Filter buildFilters(final Map<Dimension, Set<ApiFilter>> filterMap) throws DimensionRowNotFoundException {
+        LOG.trace("Building Druid Bound Filters using filter map: {}", filterMap);
+
+        List<Filter> filters =
+                    filterMap.values()
+                            .stream()
+                            .flatMap(Set::stream)
+                            .map(filter -> buildDruidBoundFilters(filter))
+                            .collect(Collectors.toList())
+        ;
+
+        if (filters.size() == 1) {
+            return filters.get(0);
+        }
+
+        AndFilter boundFilter = new AndFilter(filters);
+
+        if (boundFilter.getFields().size() > MAX_NUM_DRUID_FILTERS) {
+            LOG.error(TOO_MANY_DRUID_FILTERS.logFormat());
+            throw new TooManyDruidFiltersException(TOO_MANY_DRUID_FILTERS.format());
+        }
+
+        LOG.trace("Filter: {}", boundFilter);
+        return boundFilter;
+    }
+
+    /**
+     * Builds a Druid Bound Filter based on the operation type.
+     *
+     * @param filter APIFilter associated to a given dimension
+     *
+     * @return A Druid Bound Filter based on the filter operation type
+     */
+    public Filter buildDruidBoundFilters(ApiFilter filter) {
+        List<String> values = filter.getValuesList();
+        Dimension dimension = filter.getDimension();
+        DefaultFilterOperation filterOperation = (DefaultFilterOperation) filter.getOperation();
+        switch (filterOperation) {
+            case lte:
+                return BoundFilter.buildUpperBoundFilter(dimension, values.get(0), true);
+            case gte:
+                return BoundFilter.buildLowerBoundFilter(dimension, values.get(0), true);
+            case gt:
+                return BoundFilter.buildLowerBoundFilter(dimension, values.get(0), false);
+            case lt:
+                return BoundFilter.buildUpperBoundFilter(dimension, values.get(0), false);
+            case between:
+                return BoundFilter
+                        .buildUpperBoundFilter(dimension, values.get(1), false)
+                        .withLowerBound(values.get(0));
+            default:
+                LOG.error(FILTER_OPERATOR_INVALID.logFormat(filterOperation.getName()));
+                return null;
+        }
+    }
+}
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/util/StreamUtils.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/util/StreamUtils.java
@@ -199,4 +199,18 @@ public class StreamUtils {
     public static <T> Set<T> setMerge(Set<T> a, Set<T> b) {
         return Stream.concat(a.stream(), b.stream()).collect(Collectors.toSet());
     }
+
+    /**
+     * Merge two sets without modifying either, preserving order.
+     * This method implements {@link BinaryOperator} for use in stream reductions.
+     *
+     * @param a One input set
+     * @param b Another input set
+     * @param <T> The type of the underlying sets
+     *
+     * @return A new set containing the values of the original sets
+     */
+    public static <T> Set<T> orderedSetMerge(Set<T> a, Set<T> b) {
+        return Stream.concat(a.stream(), b.stream()).collect(Collectors.toCollection(LinkedHashSet::new));
+    }
 }
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/ApiFilter.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/ApiFilter.java
@@ -1,4 +1,4 @@
-// Copyright 2016 Yahoo Inc.
+// Copyright 2018 Oath Inc.
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.web;
 
@@ -6,7 +6,11 @@ import com.yahoo.bard.webservice.data.dimension.Dimension;
 import com.yahoo.bard.webservice.data.dimension.DimensionDictionary;
 import com.yahoo.bard.webservice.data.dimension.DimensionField;
 
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 
@@ -20,7 +24,7 @@ public class ApiFilter {
     private final Dimension dimension;
     private final DimensionField dimensionField;
     private final FilterOperation operation;
-    private final Set<String> values;
+    private final List<String> values;
 
     /**
      * Constructor.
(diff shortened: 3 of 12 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/web/ApiFilter.java
```diff
@@ -30,16 +30,34 @@
      * @param operation  Operation the filter operates with
      * @param values  The values the filter uses when operating
      */
+    @Deprecated
```
archolewa: Should include some JavaDoc briefly explaining why this was deprecated, and 
what alternative people should use.

fili-core/src/main/java/com/yahoo/bard/webservice/web/ApiFilter.java
```diff
@@ -105,7 +123,19 @@ public ApiFilter withOperation(FilterOperation operation) {
...
+    }
+
+    /**
+     * Creates a new ApiFilter based on this ApiFilter, except with the new set of filter values replacing this
+     * ApiFilter's filter values.
+     *
+     * @param values The new set of values to use in the new ApiFilter.
+     * @return the new ApiFilter object.
+     */
+    @Deprecated
```
archolewa: [ditto]https://github.com/yahoo/fili/pull/807/files#r225196611)

archolewa:
:+1: on my comments, though I'm not comfortable formally approving the code changes themselves, since I don't have any context.

fili-core/src/main/java/com/yahoo/bard/webservice/web/ApiFilter.java
```diff
@@ -105,7 +126,22 @@ public ApiFilter withOperation(FilterOperation operation) {
...
+     * ApiFilter's filter values.
+     *
+     * @param values The new set of values to use in the new ApiFilter.
+     * @return the new ApiFilter object.
+     *
+     * @deprecated the order of insertion of filter values to be preserved, use
+     * {@link #withValues(LinkedHashSet)} method instead
+     */
+    @Deprecated
+    public ApiFilter withValues(Collection<String> values) {
```
QubitPi: [ditto](https://github.com/yahoo/fili/pull/807/files#diff-da59bb50f2ea4852c2fc8cadad5b7be6R58)

This seems to be the time when we must violate the guidelines. In this case, we should make sure the two behave exactly the same. One way to do this is to have `withValues(Collection<String>)` call `withValues(LinkedHashSet<String>)`

fili-core/src/main/java/com/yahoo/bard/webservice/web/ApiFilter.java
```diff
@@ -105,7 +126,22 @@ public ApiFilter withOperation(FilterOperation operation) {
...
+    public ApiFilter withValues(LinkedHashSet<String> values) {
+        return new ApiFilter(dimension, dimensionField, operation, values);
+    }
+
+    /**
+     * Creates a new ApiFilter based on this ApiFilter, except with the new set of filter values replacing this
+     * ApiFilter's filter values.
+     *
+     * @param values The new set of values to use in the new ApiFilter.
+     * @return the new ApiFilter object.
```
QubitPi: There should be an empty line between `@patam` and `@return`

ghost (changes requested):

fili-core/src/main/java/com/yahoo/bard/webservice/web/ApiFilter.java
```diff
@@ -29,17 +29,38 @@
...
+     * @param dimension  Dimension the filter operates on
+     * @param dimensionField  Dimension Field the filter operates on
+     * @param operation  Operation the filter operates with
+     * @param values  The values the filter uses when operating (The order that we specify the filters is important)
+     */
+    public ApiFilter(
+            @NotNull Dimension dimension,
+            DimensionField dimensionField,
+            FilterOperation operation,
+            LinkedHashSet<String> values
```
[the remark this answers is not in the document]
michael-mclawhorn: I think you misunderstand what will happen at compile time.

If the caller is typed as a linked hash set, it'll call the new constructor.  If it's typed more generally it'll hit the deprecated constructor.

This deprecation is intended to warn callers of the constructor that they should be narrowing their contract and if they're using set or list to expect it to fail at some point in the future.  Anything that has a compile time value other than LinkedHashSet will be getting deprecation warnings.

michael-mclawhorn (changes requested):
I have +1ed the changes that Jiaqi requested that I agree with.  There's no point in me approving since changes remove approvals.  But otherwise, I'm happy.

fili-core/src/main/java/com/yahoo/bard/webservice/web/ApiFilter.java
RVRSKumar: Refactored the code

fili-core/src/main/java/com/yahoo/bard/webservice/web/DefaultFilterOperation.java
```diff
@@ -10,7 +10,12 @@
     notin,
     startswith,
     contains,
-    eq
+    eq,
+    lt,
+    lte,
+    gt,
+    gte,
+    between
```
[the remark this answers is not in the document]
RVRSKumar: lt, lte, gt & gte are straight forward. So will include a comment for between alone

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidBoundFilterBuilder.java
```diff
@@ -0,0 +1,122 @@
...
+    private static final int MAX_NUM_DRUID_FILTERS = SYSTEM_CONFIG.getIntProperty(
+            SYSTEM_CONFIG.getPackageVariableName("max_num_druid_filters"),
+            DEFAULT_MAX_NUM_DRUID_FILTERS
+    );
+
+    @Override
+    public Filter buildFilters(final Map<Dimension, Set<ApiFilter>> filterMap) throws DimensionRowNotFoundException {
+        LOG.trace("Building Druid Bound Filters using filter map: {}", filterMap);
+
+        // Assuming there is just going to be a single dimension for BOUND filter operations.
```
QubitPi: We should bump this up to class level Javadoc and add a log warn or throw an exception if there are multiple dimensions

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidBoundFilterBuilder.java
```diff
@@ -0,0 +1,122 @@
...
+    private static final SystemConfig SYSTEM_CONFIG = SystemConfigProvider.getInstance();
+
+    private static final int DEFAULT_MAX_NUM_DRUID_FILTERS = 10000;
+    private static final int MAX_NUM_DRUID_FILTERS = SYSTEM_CONFIG.getIntProperty(
+            SYSTEM_CONFIG.getPackageVariableName("max_num_druid_filters"),
+            DEFAULT_MAX_NUM_DRUID_FILTERS
+    );
+
+    @Override
+    public Filter buildFilters(final Map<Dimension, Set<ApiFilter>> filterMap) throws DimensionRowNotFoundException {
```
QubitPi: I think we can get rid of `throws DimensionRowNotFoundException`

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidBoundFilterBuilder.java
```diff
@@ -0,0 +1,122 @@
...
+        LOG.trace("Filter: {}", boundFilter);
+        return boundFilter;
+    }
+
+    /**
+     * Builds a Druid Bound Filter based on the operation type.
+     *
+     * @param filter APIFilter associated to a given dimension
+     * @param dimension Dimension for the filters
+     * @return A Druid Bound Filter based on the filter operation type
```
QubitPi: Empty line between `return` and `param`

ghost (changes requested):

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidBoundFilterBuilder.java
RVRSKumar: No it is not possible, Override signature should have it

RVRSKumar: Else it complaints :(

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidBoundFilterBuilder.java
RVRSKumar: No we are collecting Dimensions for other filter builders and not for DruidBoundFilter, this is based on Design

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidBoundFilterBuilder.java
```diff
@@ -0,0 +1,122 @@
...
+            case lte:
+                return new BoundFilter(dimension, null, values.get(0));
+            case gte:
+                return new BoundFilter(dimension, values.get(0), null);
+            case gt:
+                return new BoundFilter(dimension, values.get(0), null).withLowerBoundStrict(true);
+            case lt:
+                return new BoundFilter(dimension, null, values.get(0)).withUpperBoundStrict(true);
+            case between:
+                return new BoundFilter(dimension, values.get(0), values.get(1)).withUpperBoundStrict(true);
```
QubitPi: It is hard to read what all constructor calls above do and how they are different among each other. I would suggest adding static factory methods to BoundFilter to make it much more readable.

fili-core/src/main/java/com/yahoo/bard/webservice/util/StreamUtils.java
```diff
@@ -199,4 +199,18 @@
...
+     * Merge two sets without modifying either, preserving order.
+     * This method implements {@link BinaryOperator} for use in stream reductions.
+     *
+     * @param a  One input set
+     * @param b  Another input set
+     * @param <T>  The type of the underlying sets
+     *
+     * @return  A new set containing the values of the original sets
+     */
+    public static <T> Set<T> orderedSetMerge(Set<T> a, Set<T> b) {
```
QubitPi: We need a test for it

fili-core/src/main/java/com/yahoo/bard/webservice/util/StreamUtils.java
```diff
@@ -199,4 +199,18 @@
     public static <T> Set<T> setMerge(Set<T> a, Set<T> b) {
         return Stream.concat(a.stream(), b.stream()).collect(Collectors.toSet());
     }
+
+    /**
+     * Merge two sets without modifying either, preserving order.
+     * This method implements {@link BinaryOperator} for use in stream reductions.
```
QubitPi: I don't see how `BinaryOperator` come into play

fili-core/src/main/java/com/yahoo/bard/webservice/util/StreamUtils.java
```diff
@@ -199,4 +199,18 @@
...
+     *
+     * @param a  One input set
+     * @param b  Another input set
+     * @param <T>  The type of the underlying sets
+     *
+     * @return  A new set containing the values of the original sets
```
QubitPi: single space between tag and description. We don't capitalize the first letter in `return` description

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidBoundFilterBuilder.java
RVRSKumar: Addressed

ghost (changes requested):

fili-core/src/main/java/com/yahoo/bard/webservice/web/ApiFilter.java
```diff
@@ -142,6 +149,15 @@ public FilterOperation getOperation() {
...
     public Set<String> getValues() {
+        return new LinkedHashSet<>(this.values);
+    }
+
+    /**
+     * Getter for this ApiFilter's set of filter values.
+     *
+     * @return the list of values
+     */
+    public List<String> getValuesList() {
```
QubitPi: I don't know why we have 2 getters. Could we simply remove this?

ghost (changes requested):

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidBoundFilterBuilder.java
RVRSKumar: Could you be more elaborate? And give a sample static factory method?

fili-core/src/test/groovy/com/yahoo/bard/webservice/filterbuilders/DruidBoundFilterBuilderSpec.groovy
```diff
@@ -0,0 +1,81 @@
...
+            assert filters.containsAll(druidExpected) && druidExpected.containsAll(filters)
+        }
+
+        where:
+        apiName = resources.d7.apiName
+        apiList                               | filterType | druidList
+        ["startDateIdLt"]                     | BOUND      | ["startDateIdLt"]
+        ["startDateIdLt", "startDateIdGte"]   | AND        | ["startDateIdLt", "startDateIdGte"]
+        ["startDateIdBetween"]                | BOUND      | ["startDateIdBetween"]
+    }
```
QubitPi: If we throw exception on multi-dimension values, we should also add a test for it

QubitPi: Nees a test for `buildDruidBoundFilters` as well

ghost (changes requested):

fili-core/src/main/java/com/yahoo/bard/webservice/web/ApiFilter.java
RVRSKumar: No @archolewa wants backward compatibility for existing getter methods

fili-core/src/test/groovy/com/yahoo/bard/webservice/filterbuilders/DruidBoundFilterBuilderSpec.groovy
RVRSKumar: I dont understand we need this

RVRSKumar: `buildFilters` tests this implicitly. Followed the similar architecture in DruidInFilterBuilder

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidBoundFilterBuilder.java
QubitPi: Sure.

In the case of `lte`, if you add a static factory method like this below

```java
    /**
     * 
     * @param dimension
     * @param upper
     * @return
     */
    public static BoundFilter lessThanOrEqualBound(Dimension dimension, String upper) {
        return new BoundFilter(dimension, null, upper);
    }
```

Then you could replace line https://github.com/yahoo/fili/pull/807/files/b48fea135e0d7429b6417859e80d911e77a1ba5f#diff-bf47a08b799024e899dd3860f487eb38R108 with

``` java
return BoundFilter.lessThanOrEqualBound(dimension, values.get(0));
```

This makes it obvious for read that we are creating a upper-bounded filter and we don't need to pass that mysterious `null` argument

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidBoundFilterBuilder.java
michael-mclawhorn: I'm not sure why we're making this assumption instead of just looping over entries.

fili-core/src/main/java/com/yahoo/bard/webservice/util/StreamUtils.java
michael-mclawhorn: I'm getting to the point where if the tools don't enforce it where we might want to consider dropping this style requirement.  If the tools don't support it, it's frustrating for engineers to police it.

fili-core/src/main/java/com/yahoo/bard/webservice/data/filterbuilders/DruidBoundFilterBuilder.java
michael-mclawhorn: I don't dislike this suggestion.
I think if we were going to do this, I'd use:
BoundFilter.buildUpper(value, boolean excludeEquals)
BoundFilter.buildLower(value, boolean excludeEquals)
Then the only chain we'd need to do was fo between and we could just use the full constructor potentially for that one (with no nulls passed)

michael-mclawhorn (changes requested):
I think we can't assume a single dimension in the builder.  Instead we should loop over each of the ApiFilters, resolving the dimension from that ApiFilter instance.

RVRSKumar:
@michael-mclawhorn I thought we agreed on the fact that we don't have to collection the `ApiFilter` belonging to a dimension, similar to `SearProviderBacked`. Instead, we said we would `AND` Multiple druid filters to Druid.

ghost (approved):

fili-core/src/test/groovy/com/yahoo/bard/webservice/filterbuilders/DruidBoundFilterBuilderSpec.groovy
michael-mclawhorn: It does, but we're not testing every operator.  It would be easier to test every operator by feeding them into a test on buildDruidBoundFilters.

michael-mclawhorn (changes requested):
The only change I'd make it to make sure that each of the five paths through the switch are tested.

Then it's a 👍

michael-mclawhorn (approved):

Issue #814, opened 2018-10-30
Create less DruidFilterBuilders


QubitPi:
DruidFilterBuilder, like `DruidInFilterBuilder` are stateless. A single instance can be shared, which shall reduce repeated objects being created.

PR #815, opened 2018-10-30
Create less DruidFilterBuilders

QubitPi:
Issue - https://github.com/yahoo/fili/issues/814


michael-mclawhorn:
I don't object to this change, but aren't we building a single filter builder across the whole system via the binder?

Issue #822, opened 2018-10-31
Datasources should be able to have concrete start and end dates


efronbs:
currently there are no start and end dates associated with individual datasources. As such, any intervals that a datasource does not have are considered missing. However, there are situations where a datasource only holds data for a concrete amount of time. Including that datasource in a query with other datasources that have data after the ended datasource should not mark the newer data as missing. 

The solution is to add expected start and end dates to availabilities. This allows other classes to ignore missing intervals that are outside of the expected range.

PR #823, opened 2018-10-31
Add expected start and end date to availability

efronbs:
issue: https://github.com/yahoo/fili/issues/822

PLEASE READ:
This PR changes the algorithm for how `PartitionAvailability` reports its available intervals. For all existing cases the output should be the same. The tests are the most important part of this PR. Please spend most of your time reviewing the new tests. The two new tests in `PartitonAvailabilitySpec` are the most important tests.

The first picture is the tests for finding missing intervals, and the second picture is for testing `getAvailabilities(constraint)` with start and end dates. The tests are in the same order as in the pictures.

Visualization of first block of tests in PartitionAvailabilitySpec, in order

Visualization of second block of tests in PartitionAvailabilitySpec, in order


fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/BaseCompositeAvailabilitySpec.groovy
```diff
@@ -0,0 +1,250 @@
...
+
+        where:
+        testExpectedStart_1 |   testExpectedStart_2 |   testExpectedStart_3 |   desc
+        earliestStart       |   middleStart         |   null                |   "one"
+        earliestStart       |   null                |   null                |   "two"
+        null                |   null                |   null                |   "all"
+    }
+
+    @Unroll
+    def "If #desc availabilities have no end date, the composite has no end date"() {
```
michael-mclawhorn: This test doesn't DO anything with it's parameters.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/BaseCompositeAvailabilitySpec.groovy
```diff
@@ -0,0 +1,250 @@
...
+        testExpectedEnd_1   |   testExpectedEnd_2   |   testExpectedEnd_3   |   desc
+        earliestEnd         |   middleEnd           |   null                |   "one"
+        earliestEnd         |   null                |   null                |   "two"
+        null                |   null                |   null                |   "all"
+    }
+
+    @Unroll
+    def "earliest start is at position #pos in the availabilities stream"() {
+        given:
+        expectedStart_1 = Optional.ofNullable((DateTime) testExpectedStart_1)
```
michael-mclawhorn: This test also seems to not do anything with the arguments.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/BaseCompositeAvailabilitySpec.groovy
efronbs: A lot of the logic is happening in the setup function. `compositeAvailability` is being constructed out of 3 different Availability mocks. The parameters are setting the availability of those 3 mocks, and I am checking the behavior of the composite availability that wraps them.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/BaseCompositeAvailabilitySpec.groovy
efronbs: same as above

PR #824, opened 2018-10-31
Jackson databind vulnerability

michael-mclawhorn:


fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -739,7 +739,7 @@ private Query getFilterQuery(Set<ApiFilter> filters) {
         } finally {
             lock.readLock().unlock();
         }
-        return new SinglePagePagination<>(
+        return new SinglePagePagination<DimensionRow>(
```
QubitPi: Diamond operator seems to be fine (I see method signature already specified type)

fili-core/src/main/java/com/yahoo/bard/webservice/util/AllPagesPagination.java
```diff
@@ -43,7 +43,7 @@ public AllPagesPagination(Collection<T> entireCollection, PaginationParameters p
         this.collectionSize = entireCollection.size();
         this.pageToFetch = paginationParameters.getPage(entireCollection.size());
         this.countPerPage = paginationParameters.getPerPage();
-        this.lastPage = (collectionSize > countPerPage) ? (collectionSize - 1) / countPerPage + 1 : 1;
+        this.lastPage = (collectionSize > countPerPage) ? (int) ((collectionSize - 1) / countPerPage + 1) : 1;
```
archolewa: We should make `lastPage` `long` as well. What happens if someone asks for a lot of dimension rows, with a page size of 1?

CHANGELOG.md
```diff
@@ -40,6 +40,20 @@ Current
 - [Filter Code now intersects security constraints instead of unioning with requests](https://github.com/yahoo/fili/issues/812)
     * Switched to ensure security and request filters don't merge but instead intersect
 
+- [Bump Jackson version to patch vulnerability](https://github.com/yahoo/fili/issues/770)
+    * Bumped dependency version to 2.9.5 
+    * Made serialization order more specific in several classes
+    * Fixed bad format in error message
+    * Moved tests off of serialization of `SimplifiedIntervalList`.  That's turning out to be hard to solve.
+
+- [Bump lucene version to patch vulnerability](https://github.com/yahoo/fili/issues/819)
```
archolewa: I feel like we should explicitly call out that this forces a breaking change on Pagination, since anybody who has their own search providers has to work with them. In particular that the number of results, and the last page are now (or should be) longs.

fili-core/src/main/java/com/yahoo/bard/webservice/util/AllPagesPagination.java
michael-mclawhorn: Reverting to a error if we get greater than maxint since that doesn't seem to be compatible with our scoredoc collector that we're using.

CHANGELOG.md
michael-mclawhorn: Reverting the breaking change in favor of an error on largeints and no interface changes.

michael-mclawhorn:
@archolewa I've reverted all the interface impacting changes.

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
```diff
@@ -700,7 +702,14 @@ private Query getFilterQuery(Set<ApiFilter> filters) {
                         perPage
                 );
                 hits = hitDocs.scoreDocs;
-                documentCount = hitDocs.totalHits;
+                if (hitDocs.totalHits > Integer.MAX_VALUE) {
+                    String message = String.format(TOO_MANY_DOCUMENTS, hitDocs.totalHits);
+                    IllegalStateException illegalStateException = new IllegalStateException(message);
+                    LOG.error(illegalStateException.getMessage(), illegalStateException);
+                    throw illegalStateException;
```
archolewa: Why not use the same exception for when `perPage` exceeds `maxResults`: a `RowLimitReachedException`? Seems to me like this is a similar kind of problem (we're getting back more results than we can handle). For reference: https://github.com/yahoo/fili/pull/824/files#diff-0e7dd2ec8a518d234151c865cd2b6dacR764

michael-mclawhorn:
@archolewa Done

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/LuceneSearchProvider.java
michael-mclawhorn: I fixed this and then accidentally reverted it.

PR #828, opened 2018-11-02
Configurable limit on filename truncation.

michael-mclawhorn:
Preserve file extensions.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/util/ResponseUtils.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/util/ResponseUtils.java
@@ -2,6 +2,9 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.web.util;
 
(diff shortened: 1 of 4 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/web/util/ResponseUtils.java
```diff
@@ -43,8 +49,12 @@ public String getCsvContentDispositionValue(ContainerRequestContext containerReq
             interval = "_" + interval.replace("/", "_").replace(",", "__");
         }
 
-        String rawFilePath = uriPath + interval + ".csv";
-        String truncatedFilePath = rawFilePath.substring(0, Math.min(rawFilePath.length(), MAX_EXCEL_FILE_PATH_LENGTH));
-        return "attachment; filename=" + truncatedFilePath;
+        String extension = ".csv";
+        String filePath = uriPath + interval;
+        filePath = (maxFileLength >= 0 && filePath.length() > maxFileLength) ?
```
efronbs: max length should be at least 1, not 0

PR #829, opened 2018-11-07
Adds pointers to alternative of deprecated feature flags.

archolewa:


fili-core/src/main/java/com/yahoo/bard/webservice/config/BardFeatureFlag.java
```diff
@@ -9,9 +9,9 @@
 
     CURRENT_MACRO_USES_LATEST("current_macro_uses_latest"),
     PARTIAL_DATA("partial_data_enabled"),
-    // Use CacheFeatureFlag::TTL instead.
+    /** Use {@link CacheFeatureFlag::TTL} instead. */
```
kevinhinterlong: `::` -> `#`

Issue #821, opened 2018-10-31
AycnhHttpClient vulnerability


michael-mclawhorn:
https://github.com/yahoo/fili/network/alert/pom.xml/org.asynchttpclient:async-http-client/open

PR #830, opened 2018-11-08
upgrade netty and async http client

efronbs:


michael-mclawhorn:
https://github.com/yahoo/fili/issues/821

PR #834, opened 2018-11-09
Fix snapshot version

michael-mclawhorn:
I know someone did this earlier and I closed it. 
My bad, I thought the build system might resolve this.


kevinhinterlong:
I think we also need to update the version here ([according to this](https://github.com/yahoo/fili/blob/master/docs/contributing/cutting-stable-release.md#start-next-development-line))
https://github.com/yahoo/fili/blob/5e88f569a6e6a965ea5257b81d21d697d66cdda3/pom.xml#L73

Issue #709, opened 2018-06-05
Support attachment filename in Fili requests


michael-mclawhorn:
Fili currently only sets download file names in the csv use case and doesn't allow any user specification of attachment format.

We should, via the format=FOO parameter support attachment filename specification.

Proposed contract change:
for jsonapi, json and csv (but not debug) allow parameterized arguments (similar to the parameterized metrics rfc) in the form of:
json[filename=foo].

Where the presence of filename will trigger/override the adding of the header:
"Content-Disposition: attachment; filename="foo""

format: csv by default will continue to add Content-Disposition with the generated filename, json, by default will continue to use the http default ('inline') behavior.

cdeszaq:
I'm all for handling "download vs. not" uniformly across resources, but there are a few concerns I have with this:

1. What is the difference between a JSON response w/o a `Content-Disposition` header and one _with_ the header? My understanding is that the presence of the header triggers the user agent (browser) to download the response as a file, rather than try to display it in the viewport. 

   As such, would it make sense for the toggle to be "download or not" rather than "which filename would you like me to tell you this response is, and if I don't see a file name I won't tell the user agent to download this"?

2. Why does the client get to determine the file name? 

   Not only would that mess with caching of the response, but it's also the server that is determining what's _in_ the file and is best suited to give it an unambiguous name. 

   Lastly, it's also _super_ easy for the recipient of the file to change the name (though perhaps not the client application), so is this added complexity of accepting a user-specified file name worth the complexity? I'm guessing it is, but I've not seen (or don't recall) a convincing argument for it.

3. This seems to be very much in line with the existing [HTTP](https://tools.ietf.org/html/rfc7231#section-5.3.2) [`Accept`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) header, which exists to allow optimistic [content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation) [negotiation](https://restfulapi.net/content-negotiation/) between the client and the server, essentially providing a pathway for the client to ask for the resource in a particular format. With the general notion that query parameters override existing headers, we should make sure that any change we make moves towards allowing the `format` query param to behave the same way as the `Accept` header.

   I'm not asking for the full "`format` is just an override for `Accept` and we do the same thing for both inputs" change here, but only that we take the long-term view of where we want to be (ie. use the HTTP spec where possible) into account here.

[...]

PR #835, opened 2018-11-15
Add filename parameter to data queries

efronbs:
Issue: https://github.com/yahoo/fili/issues/709

This PR only exposes the filename parameter to data queries, though `getDownloadFilename()` was added to the `ApiRequest` interface. If you think more query types should have this parameter exposed please comment.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/HttpResponseMaker.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/HttpResponseMaker.java
@@ -33,6 +33,7 @@ import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 import javax.inject.Inject;
@@ -192,7 +193,8 @@ public class HttpResponseMaker {
         return buildAndAddResponseHeaders(
                 rspBuilder,
                 responseFormatType,
-                containerRequestContext
+                containerRequestContext,
+                apiRequest.getDownloadFilename().orElse(null)
         );
     }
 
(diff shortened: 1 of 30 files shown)


efronbs:
I want to manually test the new parameter before merging

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/ApiRequestImpl.java
```diff
@@ -87,6 +87,7 @@
     protected final ResponseFormatType format;
     protected final Optional<PaginationParameters> paginationParameters;
     protected final long asyncAfter;
+    protected final Optional<String> downloadFilename;
```
michael-mclawhorn: Don't store as Optional.  Instead store as potentially null and unpack to optional on the way out.

efronbs: why not store as optional?

fili-core/src/main/java/com/yahoo/bard/webservice/data/HttpResponseMaker.java
```diff
@@ -267,16 +265,35 @@ protected ResponseBuilder buildAndAddResponseHeaders(
...
+
+    /**
+     * Builds the headers for the response and gives them to the provided response builder to be added to the response.
+     *
+     * @param rspBuilder  ResponseBuilder that handles adding the headers to the response
+     * @param responseFormatType  The type of the response
+     * @param containerRequestContext  The request context
+     * @param downloadFilename  The filename the response should be downloaded as. Null or empty indicates the response
+     * should not be downloaded and instead rendered by the browser
+     * @return the response builder that has had the headers added
```
michael-mclawhorn: vertical whitespace between @params and @return

PR #838, opened 2018-11-19
Better expose concrete start and end date 

efronbs:
issue: https://github.com/yahoo/fili/issues/822

Customers that depend on physical table definitions to construct tables and availabilities have a difficult time using the expected start and end date functionality. This PR adds some constructors to make the feature more usable.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/ConcretePhysicalTableDefinition.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/ConcretePhysicalTableDefinition.java
@@ -23,7 +23,7 @@ import java.util.Set;
 public class ConcretePhysicalTableDefinition extends PhysicalTableDefinition {
 
     /**
-     * Define a physical table using a zoned time grain.
+     * Define a physical table using a zoned time grain. Defaults to table with no expected start nor end dates.
      *
      * @param name  The table name
      * @param timeGrain  The zoned time grain
@@ -46,9 +46,9 @@ public class ConcretePhysicalTableDefinition extends PhysicalTableDefinition {
      * @param timeGrain  The zoned time grain
      * @param metricNames  The Set of metric names on the table
      * @param dimensionConfigs  The dimension configurations
-     * @param expectedStartDate  The expected start date of the datasource the constructed table will represent. Empty
+     * @param expectedStartDate  The expected start date of the datasource the constructed table will represent. Null
      * indicates there is NO expected start date
-     * @param expectedEndDate  The expected end date of the datasource the constructed table will represent. Empty
+     * @param expectedEndDate  The expected end date of the datasource the constructed table will represent. Null
      * indicates there is NO expected end date
      */
     public ConcretePhysicalTableDefinition(
@@ -63,7 +63,8 @@ public class ConcretePhysicalTableDefinition extends PhysicalTableDefinition {
     }
 
     /**
-     * Define a physical table with provided logical to physical column name mappings.
+     * Define a physical table with provided logical to physical column name mappings. Defaults to config with no
+     * expected start nor end dates.
      *
      * @param name  The table name
      * @param timeGrain  The zoned time grain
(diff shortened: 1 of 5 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/ConcretePhysicalTableDefinition.java
```diff
@@ -34,7 +36,30 @@ public ConcretePhysicalTableDefinition(
...
+     * @param expectedEndDate  The expected end date of the datasource the constructed table will represent. Null
+     * indicates there is NO expected end date
+     */
+    public ConcretePhysicalTableDefinition(
+            TableName name,
+            ZonedTimeGrain timeGrain,
+            Set<FieldName> metricNames,
+            Set<? extends DimensionConfig> dimensionConfigs,
+            DateTime expectedStartDate,
+            DateTime expectedEndDate
```
QubitPi: Looks like the two dates are optional. Since constructor arguments are pilling up with optional arguments, shall we make use builder pattern for  `ConcretePhysicalTableDefinition` and `PhysicalTableDefinition`?

efronbs: We can't eliminate the old constructors until we move to 0.12

As for not being optional, my reasoning is that these are config objects, so it seems reasonable to maintain the nulls and let the actual model object decide what to do with the nulls. In this case it is wrap them in optionals, but still.

efronbs: wait for 0.12 because we can't break public method and constructor contracts without a version bump. Just something that was agreed to after similar changes caused a lot of pain for downstream projects.

michael-mclawhorn: Adding a builder pattern makes sense, feel free to spawn an issue.  We should look at how comprehensively we want to do that.
Obviously not in this PR.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/ConcretePhysicalTableDefinition.java
```diff
@@ -53,7 +78,48 @@ public ConcretePhysicalTableDefinition(
             Set<? extends DimensionConfig> dimensionConfigs,
             Map<String, String> logicalToPhysicalNames
     ) {
-        super(name, timeGrain, metricNames, dimensionConfigs, logicalToPhysicalNames);
+        super(
+                name,
```
michael-mclawhorn: Make a note in the documentation for this method that it defaults the table to having no configured start or end date.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/ConcretePhysicalTableDefinition.java
```diff
@@ -53,7 +78,48 @@ public ConcretePhysicalTableDefinition(
...
+     *
+     * @param name  The table name
+     * @param timeGrain  The zoned time grain
+     * @param metricNames  The Set of metric names on the table
+     * @param dimensionConfigs  The dimension configurations
+     * @param logicalToPhysicalNames  A map from logical column names to physical column names
+     * @param expectedStartDate  The expected start date of the datasource the constructed table will represent. Empty
+     * indicates there is NO expected start date
+     * @param expectedEndDate  The expected end date of the datasource the constructed table will represent. Empty
+     * indicates there is NO expected end date
```
michael-mclawhorn: Documentation fix.  Empty-> null since we're not using Optionals.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table/PhysicalTableDefinition.java
```diff
@@ -75,12 +102,47 @@ protected PhysicalTableDefinition(
             Set<FieldName> metricNames,
             Set<? extends DimensionConfig> dimensionConfigs,
             Map<String, String> logicalToPhysicalNames
+    ) {
```
michael-mclawhorn: Same documentation change to indicated the defaulting.

Issue #839, opened 2018-11-29
Filtered aggregations nest incorrectly


efronbs:
Currently `FilteredAggregation` nests incorrectly. It simply makes the filtered aggregation the inner query, and makes the outer query the aggregation it wraps without the filter. Some downstream customers have found this behavior generates erroneous nested queries. Specifically, if the wrapped aggregation relies on a fact dimension that is not grouped on, that dimension will not be exposed by the inner query making the outer query attempt to reference a dimension that doesn't exist in its context. 

The fix is the call nest on the wrapped aggregation, and apply the filter to the inner query of the wrapped aggregation. the outer query of the wrapped aggregation simply becomes the outer query on the nesting that is returned (after naming has been correctly set of course)

PR #840, opened 2018-11-29
fixed nesting behavior for filtered queries

efronbs:
issue: https://github.com/yahoo/fili/issues/839

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/aggregation/FilteredAggregation.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/aggregation/FilteredAggregation.java
@@ -59,6 +59,7 @@ public class FilteredAggregation extends Aggregation {
     }
 
     /**
+     * TODO maybe rewrite to reflect logic change
      * Splits an Aggregation for 2-pass aggregation into an inner filtered aggregation &amp; outer aggregation. The
      * outer aggregation is obtained by unwrapping the inner filtered aggregation and getting just the aggregation.
      * The outer aggregation fieldName will reference the inner aggregation name. The inner aggregation is unmodified.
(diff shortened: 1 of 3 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/aggregation/FilteredAggregation.java
```diff
@@ -59,6 +59,7 @@ public FilteredAggregation(@NotNull String name, String fieldName, Aggregation a
     }
 
     /**
+     * TODO maybe rewrite to reflect logic change
      * Splits an Aggregation for 2-pass aggregation into an inner filtered aggregation &amp; outer aggregation. The
      * outer aggregation is obtained by unwrapping the inner filtered aggregation and getting just the aggregation.
      * The outer aggregation fieldName will reference the inner aggregation name. The inner aggregation is unmodified.
```
michael-mclawhorn: Yeah, rewrite:

Filtered aggregation delegates nesting rules to the aggregation being filtered.  Then the filter is applied to the inner aggregation which is both probably efficient on the size of the aggregation performed, and also means that the filtering dimension doesn't need to be added to the grouping expression of the nested query.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/aggregation/FilteredAggregationSpec.groovy
```diff
@@ -165,6 +167,29 @@ class FilteredAggregationSpec extends Specification{
         filteredAgg.getAggregation().getName() == "FOO_NO_BAR-114_127"
     }
 
+    def "test nesting pushes filter to bottom"() {
```
michael-mclawhorn: Lovely test.  Four stars.

PR #848, opened 2018-12-12
Allow Optional nesting in Aggregation

### LINKED ISSUE #847: Aggregation nesting breaks downstream project
We have `Averager` which extends `Aggregation`. `nest` method in `Aggregation`  should not be nesting `Averager` to the inner query. Currently, `nest` method in `Aggregation` class returns a `Pair` of `Aggregation`. The fix would be to make this `nest` method return a `Pair` of `Optional<Aggregation>` that would allow down stream projects to have control of nesting.

deepakb91:
Fixes #847

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/TemplateDruidQuery.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/TemplateDruidQuery.java
@@ -155,9 +155,9 @@ public class TemplateDruidQuery implements DruidAggregationQuery<TemplateDruidQu
         LinkedHashSet<Aggregation> innerAggregations = new LinkedHashSet<>();
         LinkedHashSet<Aggregation> outerAggregations = new LinkedHashSet<>();
         for (Aggregation agg : aggregations) {
-            Pair<Aggregation, Aggregation> split = agg.nest();
-            innerAggregations.add(split.getRight());
-            outerAggregations.add(split.getLeft());
+            Pair<Optional<Aggregation>, Optional<Aggregation>> split = agg.nest();
+            innerAggregations.add(split.getRight().get());
+            outerAggregations.add(split.getLeft().get());
         }
 
         // Create the inner query.
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/TemplateDruidQuery.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/TemplateDruidQuery.java
@@ -156,8 +156,8 @@ public class TemplateDruidQuery implements DruidAggregationQuery<TemplateDruidQu
         LinkedHashSet<Aggregation> outerAggregations = new LinkedHashSet<>();
         for (Aggregation agg : aggregations) {
             Pair<Optional<Aggregation>, Optional<Aggregation>> split = agg.nest();
-            innerAggregations.add(split.getRight().get());
-            outerAggregations.add(split.getLeft().get());
+            split.getRight().map(innerAggregations::add);
+            split.getLeft().map(outerAggregations::add);
         }
 
         // Create the inner query.
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/aggregation/Aggregation.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/aggregation/Aggregation.java
@@ -98,9 +98,7 @@ public abstract class Aggregation implements MetricField {
         String nestingName = this.name;
         Aggregation outer = this.withFieldName(nestingName);
         Aggregation inner = this.withName(nestingName);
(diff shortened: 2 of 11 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/TemplateDruidQuery.java
```diff
@@ -155,9 +155,9 @@ public TemplateDruidQuery nest() {
         LinkedHashSet<Aggregation> innerAggregations = new LinkedHashSet<>();
         LinkedHashSet<Aggregation> outerAggregations = new LinkedHashSet<>();
         for (Aggregation agg : aggregations) {
-            Pair<Aggregation, Aggregation> split = agg.nest();
-            innerAggregations.add(split.getRight());
-            outerAggregations.add(split.getLeft());
+            Pair<Optional<Aggregation>, Optional<Aggregation>> split = agg.nest();
+            innerAggregations.add(split.getRight().get());
+            outerAggregations.add(split.getLeft().get());
```
michael-mclawhorn: Nope.
You need to do something like:
split.getLeft().stream().each(innerAggregations::add)
or
if( ! split.getLeft().isEmpty() ) { innerAggregations.add(split.getLeft()) }

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/aggregation/Aggregation.java
```diff
@@ -93,11 +94,13 @@ public String getFieldName() {
...
      * @return A pair where pair.left is the outer aggregation and pair.right is the inner.
      */
-    public Pair<Aggregation, Aggregation> nest() {
+    public Pair<Optional<Aggregation>, Optional<Aggregation>> nest() {
         String nestingName = this.name;
         Aggregation outer = this.withFieldName(nestingName);
         Aggregation inner = this.withName(nestingName);
-        return new ImmutablePair<>(outer, inner);
+        Optional<Aggregation> outerOptional = Optional.of(outer);
+        Optional<Aggregation> innerOptional = Optional.of(inner);
```
michael-mclawhorn: you can do this inline in the return to save variable declaration.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/aggregation/CardinalityAggregation.java
```diff
@@ -115,7 +116,7 @@ public CardinalityAggregation withByRow(boolean byRow)  {
      * @return Always throws an exception
      */
     @Override
-    public Pair<Aggregation, Aggregation> nest() {
+    public Pair<Optional<Aggregation>, Optional<Aggregation>> nest() {
```
michael-mclawhorn: Actually they could now with an Optional.empty() going into the inner query.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/aggregation/CountAggregation.java
```diff
@@ -48,11 +50,13 @@ public Aggregation withFieldName(String fieldName) {
...
      */
     @Override
-    public Pair<Aggregation, Aggregation> nest() {
+    public Pair<Optional<Aggregation>, Optional<Aggregation>> nest() {
         String nestingName = getName();
         Aggregation inner = withName(nestingName);
         Aggregation outer = new LongSumAggregation(getName(), nestingName);
-        return new ImmutablePair<>(outer, inner);
+        Optional<Aggregation> outerOptional = Optional.of(outer);
+        Optional<Aggregation> innerOptional = Optional.of(inner);
```
michael-mclawhorn: can inline on the return

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/aggregation/FilteredAggregation.java
```diff
@@ -67,12 +68,14 @@ public FilteredAggregation(@NotNull String name, String fieldName, Aggregation a
...
      */
     @Override
-    public Pair<Aggregation, Aggregation> nest() {
-        Pair<Aggregation, Aggregation> wrappedAggNested = this.getAggregation().nest();
-        Aggregation inner = this.withAggregation(wrappedAggNested.getRight());
-        Aggregation outer = wrappedAggNested.getLeft();
+    public Pair<Optional<Aggregation>, Optional<Aggregation>> nest() {
+        Pair<Optional<Aggregation>, Optional<Aggregation>> wrappedAggNested = this.getAggregation().nest();
+        Aggregation inner = this.withAggregation(wrappedAggNested.getRight().get());
+        Aggregation outer = wrappedAggNested.getLeft().get();
```
michael-mclawhorn: You definitely don't want to do it this way.
You should handle the wrapped nest being Optional.empty by making the filtered aggregation return optional empty if there's no nested agg to filter.

PR #857, opened 2019-01-31
Added copy with modification to DimensionRow and build a ResultSetMap…

michael-mclawhorn:
…per able to transform dimension rows from a ResultSet.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionRow.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionRow.java
@@ -79,12 +79,12 @@ public class DimensionRow extends LinkedHashMap<DimensionField, String> implemen
     }
 
     /**
-     * Create a copy of a dimension row, and then transform it using a mapping function.
+     * Copies a DimensionRow and transforms its fields using the specified mapper.
      *
-     * @param row  The row to be copied.
-     * @param mapper  The transform function.
+     * @param row The row to be transformed
+     * @param mapper  A function that takes a DimensionField and its current value, and returns the field's new value.
      *
-     * @return  A copy of this row with changes.
+     * @return A copy of the DimensionRow with its fields transformed by the specified function
      */
     public static DimensionRow copyWithReplace(DimensionRow row,  BiFunction<DimensionField, String, String> mapper) {
         DimensionRow newRow = new DimensionRow(row);
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionRow.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionRow.java
@@ -6,6 +6,7 @@ import com.yahoo.bard.webservice.util.StreamUtils;
 
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.function.BiFunction;
 
 import javax.validation.constraints.NotNull;
 
@@ -14,7 +15,19 @@ import javax.validation.constraints.NotNull;
  */
 public class DimensionRow extends LinkedHashMap<DimensionField, String> implements Comparable<DimensionRow> {
 
+    private final DimensionField key;
     private final String keyValue;
+
+    /**
+     * Build a copy of a dimension row.
+     *
+     * @param row  the dimension row to be copied
+     */
+    private DimensionRow(@NotNull DimensionRow row) {
+        super(row);
+        this.key = row.key;
+        this.keyValue = row.keyValue;
+    }
 
     /**
      * Build a dimension row with a key field value and a map of field values.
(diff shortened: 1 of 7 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionRow.java
```diff
@@ -63,4 +77,18 @@ public int compareTo(DimensionRow that) {
         return entrySet().stream()
                 .collect(StreamUtils.toLinkedMap(entry -> entry.getKey().getName(), Map.Entry::getValue));
     }
+
+    /**
+     * Create a copy of a dimension row, and then transform it using a mapping function.
```
archolewa: It isn't really clear from the documentation that when we mean "transform" it we mean, apply a mapping to each dimension field's value. Something like:

```
Copies a DimensionRow and transforms its fields using the specified mapper.

@param row The row to be transformed
@param mapper  A function that takes a DimensionField and its current value, and returns the field's new value.

@return A copy of the DimensionRow with its fields transformed by the specified function
```

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/mappers/MemoizingDimensionMappingResultSetMapper.java
```diff
@@ -0,0 +1,142 @@
...
+ * A ResultSetMapper that transforms a ResultSet by transforming its {@link DimensionRow}s.
+ *
+ * If memoization is used, transformed rows will be cached.  The cost of memoization is proportional to the hashcode
+ * cost of the DimensionRow entries and the value is proportional to the number of repeated rows and the cost of
+ * transforming them.  In general, the expectation is for high redundancy of dimension rows per result set, so
+ * memoization is on by default.
+ */
+public class MemoizingDimensionMappingResultSetMapper extends ResultSetMapper {
+
+    private final BiFunction<DimensionColumn, DimensionRow, Boolean> columnRowMatcher;
```
archolewa: `java.util.function` has a `BiPredicate`: https://docs.oracle.com/javase/8/docs/api/java/util/function/BiPredicate.html that you could use. This should simplify this type some, and make it clearer what this is doing.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/DimensionRowSpec.groovy
```diff
@@ -26,4 +29,31 @@ class DimensionRowSpec extends Specification {
...
+
+    def "missing key throws error"() {
+        when:
+        DimensionRow testRow = new DimensionRow(missingKeyField, [(keyField): keyValue, (nonKeyField): nonKeyValue])
+
+        then:
+        thrown(IllegalArgumentException)
+    }
+
+    def "copyWithReplace copies and replaces"() {
```
archolewa: The name implies that it copies and replaces, but I don't see any validation that it actually copies (i.e. leaves the original unmodified). Should we include an additional assertion that `testRow != expectedRow`?

fili-core/src/test/groovy/com/yahoo/bard/webservice/web/endpoints/DimensionsServletComponentSpec.groovy
```diff
@@ -321,7 +321,7 @@ class DimensionsServletComponentSpec extends Specification {
         result.head().getHeaderString("Content-Disposition") == "attachment; filename=dimensions-shape-values.csv"
         result.head().getHeaderString("Content-Type") == "text/csv; charset=utf-8"
         Map<String, String> actualLinks = GroovyTestUtils.splitHeaderLinks(result.head().getHeaderString("Link"))
-        actualLinks.every {name, actualLink -> GroovyTestUtils.compareURL(actualLink, expectedLinks[name])}
+        actualLinks.every {name, actualLink -> GroovyTestUtils.compareURL(actualLink, expectedLinks [name])}
```
archolewa: Any particular reason for this change?

/NonBlocker

michael-mclawhorn: it actually got in accidentally.
I'll remove it.

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/dimension/DimensionRowSpec.groovy
michael-mclawhorn: done

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/mappers/MemoizingDimensionMappingResultSetMapper.java
michael-mclawhorn: done

fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/DimensionRow.java
michael-mclawhorn: indeed, clearer.

PR #859, opened 2019-02-10
Unlikely argument type TableName for String

wcekan:
Unlikely argument type `TableName` for `containsKey` on a `Map<String,ConfigPhysicalTable>`
	/fili-core/src/main/java/com/yahoo/bard/webservice/data/config/table	line 263


michael-mclawhorn:
A bugfix comment in the CHANGELOG would not be inappropriate.

PR #894, opened 2019-04-24
Added support for transforming comma list values into a boolean value.

michael-mclawhorn:
First part of implementing tag support.

Create an extraction function to transform comma list values to Yes/No

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/dimension/extractionfunction/TagExtractionFunctionFactory.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/dimension/extractionfunction/TagExtractionFunctionFactory.java
@@ -1,0 +1,88 @@
+// Copyright 2019 Verizon Media Group
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.druid.model.dimension.extractionfunction;
+
+import java.util.HashMap;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Tag Extraction Function handles a case where a comma delimited list of values is being filtered for the existence
+ * of a particular tag value.
+ *
+ *  Given a target value 11, it should match:
+ *  11
+ *  11,12
+ *  10,11
+ *  10,11,12
+ *
+ *  and not match on values where 11 is only a substring:
+ *  e.g.
+ *  111
+ */
+public class TagExtractionFunctionFactory {
+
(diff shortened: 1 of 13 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/dimension/extractionfunction/TagExtractionFunctionFactory.java
```diff
@@ -0,0 +1,88 @@
+// Copyright 2019 Verizon Media Group
```
will-lauer: My understanding was that this was not the correct copyright to use yet for our code. You may want to check this with Gil before committing it.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/dimension/extractionfunction/TagExtractionFunctionFactorySpec.groovy
```diff
@@ -0,0 +1,66 @@
...
+        String patternString = String.format(TagExtractionFunctionFactory.DEFAULT_TAG_REG_EX_FORMAT, expression);
+        Pattern.compile(patternString).matcher(value).matches() == expected
+
+        where:
+        value      | expression | expected
+        "11,12,13" | "11"       | true
+        "11"       | "11"       | true
+        "10,11"    | "11"       | true
+        "111"      | "11"       | false
+        ""         | ""         | true
```
will-lauer: This last case looks incorrect. We should NOT find 11 when the initial value is empty!

michael-mclawhorn: This test case changes the value being sought to emptystring.

michael-mclawhorn: On further discussion having a null string tag is probably unreasonable.

fili-core/src/test/groovy/com/yahoo/bard/webservice/druid/model/dimension/extractionfunction/TagExtractionFunctionFactorySpec.groovy
```diff
@@ -61,6 +73,5 @@ class TagExtractionFunctionFactorySpec extends Specification {
         "11"       | "11"       | true
         "10,11"    | "11"       | true
         "111"      | "11"       | false
-        ""         | ""         | true
```
will-lauer: I think its worth having a test of `"" | "11" | false`

PR #895, opened 2019-04-25
Refactor to allow additional hooks when building ApiFilters

michael-mclawhorn:


fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/binders/FilterBinders.java
```diff
@@ -50,7 +52,47 @@
     protected static final Pattern API_FILTER_PATTERN =
             Pattern.compile("([^\\|]+)\\|([^-]+)-([^\\[]+)\\[([^\\]]+)\\]?");
 
-    public static final FilterBinders INSTANCE = new FilterBinders();
+    private static FilterBinders instance = new FilterBinders();
+
+    private final FilterFactory filterFactory;
+
+    /**
+     * Constructor.
```
will-lauer: Comment is incorrect. THis is NOT a constructor.

PR #897, opened 2019-04-26
Fixed corrupted getDefaultDimensionFields after show

wcekan:
Merges duplicates without changing either set.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequestImpl.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequestImpl.java
@@ -1315,10 +1315,8 @@ public class DataApiRequestImpl extends ApiRequestImpl implements DataApiRequest
                     .collect(Collectors.toMap(
                             pathSegment -> dimensionDictionary.findByApiName(pathSegment.getPath()),
                             pathSegment -> bindShowClause(pathSegment, dimensionDictionary),
-                            (LinkedHashSet<DimensionField> e, LinkedHashSet<DimensionField> i) -> {
-                                e.addAll(i);
-                                return e;
-                            },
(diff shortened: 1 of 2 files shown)


wcekan:
@archolewa @michael-mclawhorn please review

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequestImpl.java
```diff
@@ -1315,10 +1315,8 @@ protected void validateTopN(String topNRequest,  int topN, LinkedHashSet<OrderBy
                     .collect(Collectors.toMap(
                             pathSegment -> dimensionDictionary.findByApiName(pathSegment.getPath()),
                             pathSegment -> bindShowClause(pathSegment, dimensionDictionary),
-                            (LinkedHashSet<DimensionField> e, LinkedHashSet<DimensionField> i) -> {
-                                e.addAll(i);
-                                return e;
-                            },
+                            (LinkedHashSet<DimensionField> e, LinkedHashSet<DimensionField> i) ->
+                            (LinkedHashSet<DimensionField>) StreamUtils.orderedSetMerge(e, i),
```
michael-mclawhorn: Oooh nasty.  What a sinister little bug.

archolewa: This should be indented eight spaces, since it's inside a lambda expression.

Also, instead of casting this to a `LinkedHashSet`, I think we should update the signature of `orderedSetMerge` to return a `LinkedHashSet`. Should be a safe change, because `LinkedHashSet` is a subclass of `Set`.

wcekan: Or perhaps made generic, so it returns the same type as the arguments.

Issue #902, opened 2019-05-01
Add api filters to logical table


efronbs:
Currently LogicalTable functions as a view on a group of physical tables. This works by controlling the set of dimensions and metrics and individual logical table supports. Even if the underlying physical tables can support a large set of dimensions and metrics, a logical table will only report on them if they are in its schema. However, changing the schema of a logical table only hides columns, and can't hide or exclude rows that are not desirable on the logical table. 

Adding ApiFilters to the logical table allows the logical table to hide rows, which expands its functionality as a view on the underlying data store.

PR #903, opened 2019-05-01
Add ApiFilters to logical table

efronbs:
Issue: https://github.com/yahoo/fili/issues/902


michael-mclawhorn:
I think you should make the getter to be an Optional<ApiFilters> and not deprecate the constructor that doesn't have api filters on it. 
Otherwise looks fine.

fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/ApiFilters.java
```diff
@@ -28,4 +29,41 @@ public ApiFilters() {
...
+     *      dim2: [apiFilter3]
+     * }
+     * ApiFilters f2 = {
+     *      dim2: [apiFilter3, apiFilter4]
+     *      dim3: [apiFilter5]
+     * }
+     *
+     * result = {
+     *      dim1: [apiFilter1, apiFilter2]
+     *      dim2: [apiFilter3, apiFilter4]
```
michael-mclawhorn: This seems wrong.  If I'm filtering to apiFilter3, the narrower filter not the broader filter should apply.  You can do this by just anding them together rather than merging them.

michael-mclawhorn:
What's your rationale for building LogicalTable ApiFilters implementation onto TablesApiRequestImpl but not onto DataApiRequestImpl?

fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidQueryBuilder.java
```diff
@@ -132,7 +133,14 @@ public DruidQueryBuilder(
         // Resolve the table from the the group, the combined dimensions in request, and template time grain
         QueryPlanningConstraint constraint = new QueryPlanningConstraint(request, template);
         ConstrainedTable table = resolver.resolve(group.getPhysicalTables(), constraint).withConstraint(constraint);
-        Filter filter = druidFilterBuilder.buildFilters(request.getApiFilters());
+
+        // combine the filters on the requested logical table with the api query filters
+        ApiFilters apiFilters = ApiFilters.merge(
+                logicalTable.getFilters(),
+                request.getApiFilters()
+        );
```
michael-mclawhorn: This should probably be append, not merge.

fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/ApiFilters.java
efronbs: I'm not sure I totally understand. in this situation I meant for apiFilter3 in both f1 and f2 to be the exact same filter. I think the example you just gave would be trying to merge the filters

dim1|id-in[val1]
and 
dim1|id-in[val2,val3]

I was under the impression that both of those filters should be used and that the druid filter builder would handle merging the logic. In that case the result be something like:

{ dim1: [ dim1|id-in[val1], dim1|id-in[val2,val3] ] as Set } as Map

fili-core/src/main/java/com/yahoo/bard/webservice/data/DruidQueryBuilder.java
efronbs: What happens if they both have filters on the same dimension? I don't think the right option is ever to remove one of the filters. for example, in the case:

logical table fitler: `dim|id-notin[val1,val2]
request filter: `dim|id-in[val3]`

Those filters are not mutually exclusive. There is at least not an obvious way remove either. I think they need to be merged, but I'm not exactly sure what the merging algorithm should be. Again, I thought handing the naively merged ApiFilters off to the DruidFilterBuilder would handle  the logic for us.

efronbs:
Responses to other reviews:

> I think you should make the getter to be an Optional and not deprecate the constructor that doesn't have api filters on it.

Is there a reason to return an optional instead of an empty ApiFilters object? I will not deprecate the other constructors though.

> What's your rationale for building LogicalTable ApiFilters implementation onto TablesApiRequestImpl but not onto DataApiRequestImpl?

This is something I'm not sure of. Originally I only put the filter merge on the `DruidQueryBuilder`. I thought it made more sense to keep the `DataApiRequestImpl` strictly a representation of the api request, and account for the filters on the logical table where appropriate (i.e. when building the data query). It also means that extensions of DataApiRequest that manipulate the ApiFilters object don't have to worry about merging the LogicalTable filters with the request filters. 

However, when I was trying to make changes to the `TablesApiRequestImpl` I didn't really see any other appropriate place to merge the filters.

Ultimately, I'm not really sure if it would make sense to always just merge logical table filters at the *ApiRequestImpl level or to try to keep that merge separate.

fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/ApiFilters.java
efronbs: I guess that's more of a union than a merge though.

PR #904, opened 2019-05-03
Tweak ApiFilters on LogicalTable to be realized as an Optional<ApiFilt…

michael-mclawhorn:
…ers>


fili-core/src/test/groovy/com/yahoo/bard/webservice/data/DruidQueryBuilderSpec.groovy
```diff
@@ -582,6 +584,6 @@ public class DruidQueryBuilderSpec extends Specification {
         builder.buildQuery(apiRequest, resources.simpleTemplateQuery)
 
         then:
-        1 * dfb.buildFilters({(ApiFilters) it == expectedApiFilters}) >> { ApiFilters filterMap -> resources.druidFilterBuilder.buildFilters(filterMap)}
+        1 * dfb.buildFilters(_) >> { ApiFilters filterMap -> resources.druidFilterBuilder.buildFilters(filterMap)}
```
efronbs: I don't think the interaction inside of the buildFilters should be removed. That was how the spec was testing that the api filters were being merged properly, and I don't see any other places a check for that was added

efronbs: I was trying to use the last type of interaction in this example:
http://spockframework.org/spock/docs/1.3/interaction_based_testing.html#_argument_constraints

PR #906, opened 2019-05-13
remove unnecessary else

wcekan:
The unnecessary else after return/throw tends to be harder to read.  
However, it makes no functional change to the code.


wcekan:
@will-lauer Yes.  My IDE is unhappy.

michael-mclawhorn:
I'm on the fence about whether this should have a changelog.  It /shouldn't/ have any impact, but without a changelog entry it's harder to find where the changes got made.

I guess I'm ok without it.

PR #907, opened 2019-05-14
Luthier design.

archolewa:


archolewa:
The branch name is wrong, closing in favor of #914 which has the correct branch name, as well as the most up to date changes.

PR #909, opened 2019-05-21
Fixed druid dimension loader issue.

michael-mclawhorn:
Should resolve issue with stuck in time druid dimension loader.


fili-core/src/test/groovy/com/yahoo/bard/webservice/application/DruidDimensionValueLoaderSpec.groovy
```diff
@@ -0,0 +1,58 @@
+package com.yahoo.bard.webservice.application
```
QubitPi: Header

PR #910, opened 2019-05-22
format some markdown lists in the design doc

gab-umich:
It looks nicer on the default github markdown viewer.


gab-umich:
This is shadowed by a newer PR, which is made against the correctly named branch.

PR #911, opened 2019-05-23
Adds key-specific expiration to DataCache.

archolewa:
-- We have a few customers who would like to be able to set different
expiration times for different values in Fili's response cache. This
PR adds a new method to the DataCache interface
`boolean set(String key, T value, int expiration)` that allows people
to set the expiration for the specified key at cache insertion time.
The default implementation drops the expiration, so customers who have
their own DataCache implementation will be unaffected.

-- We also implmenet the method in `MemDataCache` and delegate the
old `set(String key, T value)` to it, using the configured
`EXPIRATION` as the expiration time.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/cache/DataCache.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/cache/DataCache.java
@@ -2,6 +2,8 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.data.cache;
 
+import org.joda.time.DateTime;
+
 import java.io.Serializable;
 
 /**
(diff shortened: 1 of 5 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/cache/DataCache.java
```diff
@@ -33,6 +33,27 @@
...
+     *
+     * By default, this method throws away the expiration.
+     *
+     *
+     * @param key  the key under which this object should be added.
+     * @param value  the object to store
+     * @param expiration The expiration time, it may either be Unix time (number of seconds since January 1, 1970, as
+     * a 32-bit value), or a number of seconds starting from current time. In the latter case, this number of seconds
+     * may not exceed 60*60*24*30 (number of seconds in 30 days); if the number sent by a client is larger than that,
+     * the server will consider it to be real Unix time value rather than an offset from current time.
```
michael-mclawhorn: Shouldn't the system have an opinion on whether this is a seconds remaining or a time of expiration?

If we need to support the expire-at-time use case perhaps we should have it take a Date instead for specificity.

michael-mclawhorn:
sonarcloud would love a test on the change

fili-core/src/main/java/com/yahoo/bard/webservice/data/cache/DataCache.java
archolewa: Yeah, I wasn't sure how much meaning we wanted to impose at the interface level, so I basically took what we have in memcache and threw it here as a starting point. I'm ok with making it a Date.

PR #914, opened 2019-05-29
Luthier design

archolewa:
**Issue #, if available:**
#708 

**Description of changes:**

Implements Luthier, an external lua-based configuration for Fili. Read the `luthier-design.md` included in this PR for details.

I confirm that this contribution is made under the terms of your choice and that I have the authority necessary to make this contribution on behalf of its copyright owner.


luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/FactoryPark.java
```diff
@@ -14,12 +14,14 @@
  */
 public class FactoryPark<T> {
 
-    public static final String FACTORY_KEY = "type";
+    static final String ENTITY_TYPE = "factory park";
```
michael-mclawhorn: this probably shouldn't be called an 'entity type'.
Other places in our design 'Entity' is an instance of a 'Concept'.

PR #916, opened 2019-05-30
Various fixes Centering the dimensionConfig issues

gab-umich:
**Issue #, if available:**

**Description of changes:**
Briefings:
Removed the prebuilt JSONs from the repo, added to .gitignore to encourage rebuilds for users; 
Fixed the Dimension JSON format (detail in commit msg);
Removed the an extra element "desc" in the dimension table, we have "description" already;
Partially Beautified all output JSON. 

I confirm that this contribution is made under the terms of your choice and that I have the authority necessary to make this contribution on behalf of its copyright owner.


luthier/config/src/config.lua
```diff
@@ -26,6 +26,10 @@ local dimensionConfig = dimensionUtils.build_dimensions_config(dimensions)
 local metricConfig = metricsUtils.build_metric_config(metrics)
 local tableConfig = tableUtils.build_table_config(tables)
 
+-- make the directory ../external, which can be automated in 
+-- a script since creating a dir in Lua is awkard, in future.
+-- can be circumvented by using the LuaFileSystem module
+os.execute("mkdir -p ../external/")
```
archolewa: This basically locks Fili into being run on a Unixy system, which isn't great.
On the other hand, I *also* don't want to force people to download a bunch of modules
to use the Lua config. With that
mind, I'm wondering if it's better to actually create the empty `external` directory, add
it to git, but ignore its contents. Unfortunately, it looks like doing so requires
a bit of an ugly hack. You need to basically put an empty file (like say `.keep`)
and force git to version control it despite `external` being ignored.

Not ideal, and kinda ugly, but I personally think that's better than imposing Unix
on people running Fili, or requiring people to install luarocks so they can
install a file system module. The other option of course, is that we don't have
an external directory at all. Just dump the JSON files into the `luthier/config`
directory. @mclawhor Thoughts?

luthier/config/src/lib/json.lua
```diff
@@ -22,6 +22,9 @@
 -- SOFTWARE.
 --
 
+-- Modified by Gabriel on May 29, 2019 to create a better formatting when encoding 
+-- JSON. Slack: @gluo; GitHub: @gab-umich
```
archolewa: I wouldn't keep any of these changes. There's no point
in trying to make the JSON "pretty." In the normal course of things it's only consumed
by the computer, and if a human needs to look at it, there are a lot of tools available,
like `python -m json.tool` or `jq` to prettify it. Generating pretty JSON is unnecessary
complexity.

luthier/config/src/utils/dimensionUtils.lua
```diff
@@ -55,7 +55,7 @@ function M.build_dimensions_config(dimensions)
         dim_copy.apiName = dim_copy.apiName or name
         dim_copy.longName = dim_copy.longName or name
         dim_copy.description = dim_copy.description or name
-        table.insert(configuration, dim_copy)
+        configuration[ dim_copy.apiName ] = dim_copy
```
archolewa: In Lua, you don't typically put spaces in table lookups. So this
should be:
```lua
configuration[dim_copy.apiName] = dim_copy
```

luthier/config/src/config.lua
gab-umich: I would say that this directory should be created in build time as part of the set up process. 
Is there any platform independent way of creating an empty dir as part of the maven instruction?

luthier/config/src/lib/json.lua
gab-umich: I agree. Modularizing JSON formatting is vastly better.

luthier/config/src/utils/dimensionUtils.lua
gab-umich: Can I make this change here in the PR or should I redo it on my branch and submit a new PR?

luthier/config/src/config.lua
archolewa: This might give some ideas: https://stackoverflow.com/questions/4085973/create-directory-when-needed-in-maven

luthier/config/src/utils/dimensionUtils.lua
archolewa: The expected workflow is that you add additional commits to this PR to address people's comments. So the flow is:

1. Add PR.
2. People --stare in awe at your incredible code-- make comments.
3. Add commits to the PR that address those commits.
4. If reviews approve, merge the PR, else go to 2.

luthier/config/src/lib/json.lua
```diff
@@ -398,4 +398,4 @@ function json.decode(str)
 end
 
 
-return json
+return json
```
archolewa: Looks like you removed a trailing newline at the end of this file. That should probably be put back.

PR #919, opened 2019-06-04
Resource node test and various small typo fixes

gab-umich:
**Issue #, if available:**

**Description of changes:**

- ResourceNodeSupplier is fully tested. 
  - This PR also tweaks the Lua script to generate necessary testing JSON for further uses. 
- Test for FactoryPark is set up
- Test for LuthierIndustrialPark is set up.
> (btw I really have mixed feelings about the naming here. Either the prefix of Luthier- should be before every component [which imho is an overkill], or we just go with using IndustrialPark)
- Some typo spot on the way of testing :D


.gitignore
```diff
@@ -18,6 +18,7 @@ out
 # make sure the external config files are built 
 # on user's side each time, instead of using the ones from repo
 luthier/config/external
```
archolewa: If we're not going to be using this directory anymore, we can
also remove it from the .gitignore. That's not big deal though.

.gitignore
```diff
@@ -18,6 +18,7 @@ out
...
+luthier/src/test/resource
```
archolewa: We shouldn't ignore the resources directory. That may include
some configuration files that we *do* want to source control. Instead,
we should ignore just the JSON files in there. So something like:
`luthier/src/test/resources/`

Also, it should be `resources` not `resource`. Note the plural.

luthier/config/src/config.lua
```diff
@@ -30,7 +30,14 @@ local tableConfig = tableUtils.build_table_config(tables)
 -- a script since creating a dir in Lua is awkard, in future.
 -- can be circumvented by using the LuaFileSystem module
 os.execute("mkdir -p ../external/")
+os.execute("mkdir -p ../../src/test/resource/")
```
archolewa: Should be `resources` not `resource` (note the plural). Also, I would
pull the string "../../src/test/resources/" into a local variable. That way
you can do:
```lua
os.execute("mkdir -p " . resourcesDirectory)
parser.save(resourcesDirectory . "DimensionConfig.json")
...
```
Makes it easier to change the resourcesDirectory if we need to,
and makes things a bit more readable in my opinion.

luthier/config/src/config.lua
```diff
@@ -30,7 +30,14 @@ local tableConfig = tableUtils.build_table_config(tables)
...
 parser.save("../external/DimensionConfig.json", dimensionConfig)
 parser.save("../external/MetricConfig.json", metricConfig)
 parser.save("../external/TableConfig.json", tableConfig)
 -- parser.save("../external/MakerConfig.json", require("makers"))
```
archolewa: Why aren't we deleting all these lines? There's no point in
putting the JSON in two places.

luthier/config/src/config.lua
```diff
@@ -30,7 +30,14 @@ local tableConfig = tableUtils.build_table_config(tables)
...
+
+-- add to the test/resource
+parser.save("../../src/test/resource/DimensionConfig.json", dimensionConfig)
+parser.save("../../src/test/resource/MetricConfig.json", metricConfig)
+parser.save("../../src/test/resource/TableConfig.json", tableConfig)
+-- parser.save("../../src/test/resource/MakerConfig.json", require("makers"))
```
archolewa: Why is this commented out?

luthier/config/src/dimensions.lua
```diff
@@ -69,7 +69,8 @@ local field = dimensionUtils.field
 local FIELDSETS = {
     default = { pk "ID", field "DESC" },
     country = { pk "ID", field("DESC", "COUNTY", "STATE")},
-    page = { pk "ID", field "DESC" }
+    page = { pk "ID", field "DESC" },
+    test = { pk "TEST_PK", field ("TEST_FIELD_1", "TEST_FIELD_2", "TEST_FIELD_3") }
```
archolewa: There shouldn't be a space between `field` and `("TEST_FIELD_1", ...)`. It's a
function call.

luthier/config/src/dimensions.lua
```diff
@@ -121,6 +122,15 @@ local searchProviders = dimensionUtils.searchProviders
 local keyValueStores = dimensionUtils.keyValueStores
 
 return {
+    testDimension = {
+        longName = "a longName for testing",
+        description = "a description for testing",
+        fields = FIELDSETS.test,
+        category = "a category for testing",
+        type = "TestDimensionFactory",  -- type
```
archolewa: Would remove the `--type` comment at the end of the line. It's redundant.

luthier/config/src/dimensions.lua
```diff
@@ -121,6 +122,15 @@ local searchProviders = dimensionUtils.searchProviders
...
+        searchProvider=searchProviders.noop,
+        keyValueStore=searchProviders.noop
```
archolewa: Should be `keyValueStores.noop`

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -34,7 +34,7 @@
      * @param resourceDictionaries  The dictionaries to initialize the industrial park with.
      * @param dimensionFactories The map of factories for creating dimensions from external config
      */
-    protected LuthierIndustrialPark(
+    public LuthierIndustrialPark(
```
archolewa: Why make this `public`? The intent is that people use LuthierIndustrialPark.Builder to
build an instance, rather than building one with the constructor. For one thing, the Builder makes it
a lot easier to add custom factories without having to also pull in all the local stuff. If you want
to add your own dimensionFactory, it's as easy as:
```java
LuthierIndustrialPark park = new LuthierIndustrialPark.Builder()
.withDimensionFactory("testDimensionFactory", new TestDimensionFactory<>())
.build();
```

No need for making the constructor public.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/FactoryParkTest.groovy
```diff
@@ -0,0 +1,14 @@
+package com.yahoo.bard.webservice.config.luthier
+
+import spock.lang.Specification
+
+class FactoryParkTest extends Specification {
+    void setup() {
+    }
+
+    void cleanup() {
+    }
```
archolewa: Neither one of these methods need to be defined unless you plan on using them.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/FactoryParkTest.groovy
```diff
@@ -0,0 +1,14 @@
...
+
+    def "BuildEntity"() {
```
archolewa: We typically use much longer and more descriptive names, because Spock/Groovy make that easy.
Something like
```groovy
"A FactoryPark<Dimension> correctly builds a Dimension from JSON."() {

```

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialParkTest.groovy
```diff
@@ -0,0 +1,25 @@
...
+import com.yahoo.bard.webservice.data.config.ResourceDictionaries
+import com.yahoo.bard.webservice.data.dimension.Dimension
+import spock.lang.Specification
+
+class LuthierIndustrialParkTest extends Specification {
+    LuthierIndustrialPark industrialPark
+    void setup() {
+        ResourceDictionaries resourceDictionaries = new ResourceDictionaries()
+        Map<String, Factory<Dimension>> dimensionFactoriesMap = new HashMap<>()
+        industrialPark = new LuthierIndustrialPark(resourceDictionaries, dimensionFactoriesMap)
```
archolewa: I would use the Builder rather than building a LuthierIndustrialPark with the constructor:
```groovy
industrialPark = new LuthierIndustrialPark.Builder()
.withDimensionFactory("test", new TestDimensionFactory())
.build()
```

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/ResourceNodeSupplierTest.groovy
```diff
@@ -0,0 +1,52 @@
+package com.yahoo.bard.webservice.config.luthier
+
+import com.fasterxml.jackson.databind.node.ArrayNode
+import com.fasterxml.jackson.databind.node.ObjectNode
+import spock.lang.Specification
+
+class ResourceNodeSupplierTest extends Specification {
+    ResourceNodeSupplier defaultResourceNodeSupplier
```
archolewa: Since we're only using this in one test ("Node Existence"), we can define it in a
`given` block in that test. No reason to make it a class field.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/ResourceNodeSupplierTest.groovy
```diff
@@ -0,0 +1,52 @@
...
+    void setup() {
+        defaultResourceNodeSupplier = new ResourceNodeSupplier(null)
+        testResourceNodeSupplier = new ResourceNodeSupplier("DimensionConfig.json")
+    }
+
+    void cleanup() {
+
+    }
+
+    def "Node Existence"() {
```
archolewa: This can be named better. Something like:
```groovy
def "ResourceNodeSupplier returns a node if exists, and null if it doesn't.
```
Though that does make me wonder if maybe we should have the `ResourceNodeSupplier::get`
return an `Optional<ObjectNode>` if we expect `null` to be a well-defined return.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/ResourceNodeSupplierTest.groovy
```diff
@@ -0,0 +1,52 @@
...
+        when:
+        ObjectNode defaultNode = defaultResourceNodeSupplier.get()
+        ObjectNode dimensionNode = testResourceNodeSupplier.get()
+        then:
+        defaultNode == null
+        dimensionNode != null
```
archolewa: We typically indent when and then blocks. Makes things more readable, even if it
isn't syntactically required. So something like:
when:
ObjectNode defaultNode = ...
then:
defaultNode == null

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/ResourceNodeSupplierTest.groovy
```diff
@@ -0,0 +1,52 @@
...
+        fields.size() == 4
+        fields.get(0).get("name").textValue() == "TEST_PK"
+        fields.get(0).get("tags").get(0).textValue() == "primaryKey"
+        fields.get(1).get("name").textValue() == "TEST_FIELD_1"
+        fields.get(2).get("name").textValue() == "TEST_FIELD_2"
+        fields.get(3).get("name").textValue() == "TEST_FIELD_3"
+        category == "a category for testing"
+        description == "a description for testing"
+        searchProvider == "com.yahoo.bard.webservice.data.dimension.impl.NoOpSearchProvider"
+        keyValueStore == "com.yahoo.bard.webservice.data.dimension.impl.NoOpSearchProvider"
```
archolewa: This is wrong. The keyValueStore should be returning a `NoOpKeyValueStore`.

archolewa:
I would rather put `Luthier` in front of everything. What seems like overkill to you is consistency to me. Plus, it makes it obvious what parts of Fili are associated with Luthier, which could make it easier for developers to dive into the plumbing behind Luthier even if they aren't super familiar with the codebase.

.gitignore
gab-umich: What I changed it to is that we ignore the current *Config.json files one by one, but not the entire dir. 
This way we are a little bit more explicit when we think some config files should always be rebuilt, by adding that to the .gitignore list.

.gitignore
gab-umich: Fixed. 
What I changed it to is that we ignore the current *Config.json files one by one, but not the entire dir.
This way we are a little bit more explicit when we think some config files should always be rebuilt, by adding that to the .gitignore list.

luthier/config/src/config.lua
gab-umich: That is a good suggestion.

luthier/config/src/config.lua
gab-umich: Done.

luthier/config/src/config.lua
gab-umich: This is commented out because we don't have the requirement "maker.lua" yet. 
We can uncomment that when we get there.

luthier/config/src/dimensions.lua
gab-umich: I was not aware of that. Thanks for pointing it out.

luthier/config/src/dimensions.lua
gab-umich: Fixed.

luthier/config/src/dimensions.lua
gab-umich: Fixed.

gab-umich: Actually there is no keyValueStore.noop
I will use .memory instead. This prompts another change to make because all the other dimensions are wrongly populated previously.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/FactoryParkTest.groovy
gab-umich: Removed.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/FactoryParkTest.groovy
gab-umich: Addressed.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/ResourceNodeSupplierTest.groovy
gab-umich: Changed.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/ResourceNodeSupplierTest.groovy
gab-umich: Done.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/ResourceNodeSupplierTest.groovy
gab-umich: Now fixed with the previous typo.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/ResourceNodeSupplierTest.groovy
gab-umich: Changed to a more current version and added another test to address the current behavior. 

Right now it throws an error in the case that { the file does not exist, the file is malformed}, instead of returning a null. 

I agree with the optional wrapper, but I am not sure if `null` will be actually meaningful in practice. (The only reasonable case to return a `null` is when `ResourceNodeSupplier( null )` is used to initialize the resourceNodeSupplier, thus the getter function see that the `resourceName == null`.) 

I suggest we should consider the optional if we plan to make a setter available for `resourceName`, which might be exposed to `null` value.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: The problem here is that the 
`LuthierBinderFactory::getConfigurationLoader()` was using a ctor directly. I have changed the implementation there to use a Builder as well. 
To make it really work, I think we should make the Builder a **static** nested class of LuthierIndustrialPark, since we want it to be called before an instance of it's parent is initialized. 

The current ctor for Builder requires a preexisting ResourceDictionaries instance. 
To make a `Builder()` ctor is surprisingly non-trivial, because we cannot just set its resourceDictionaries as a `new ResourceDictionaries`, as that can be un-initialized. (The initializer is protected in the LuthierBinderFactory)

Since this is more intertwined with the design behind `LuthierREsourceDictionaries`, I think we should go and talk about this face-to-face sometime before resolving this issue.

gab-umich:
> I would rather put `Luthier` in front of everything. What seems like overkill to you is consistency to me. Plus, it makes it obvious what parts of Fili are associated with Luthier, which could make it easier for developers to dive into the plumbing behind Luthier even if they aren't super familiar with the codebase.

I see your concern. Let's refactor this in a seperate PR.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
archolewa: The `Builder` should be `static` yes. 

I'm not sure I understand why creating a constructor for the Builder is non-trivial. What do you mean a `new ResourceDictionaries()` can be "unitialized"? If we don't want to build the `ResourceDictionaries` in the `Builder` constructor, we can pass it in:

```java
 LuthierIndustrialPark.Builder builder = new LuthierIndustrialPark.Builder(new LuthierResourceDictionaries());
```

or
```java
 LuthierResourceDictionaries resourceDictionaries = new LuthierResourceDictionaries();
  // prepolulate resourceDictionaries
 LuthierIndustrialPark.Builder builder = new LuthierIndustrialPark.Builder(resourceDictionaries);
```

I don't see what the problem is with any of this.

Now, maybe my typing was too general. The `LuthierIndustrialPark.Builder` should probably take and/or build a `LuthierResourceDictionaries` object. I'd forgotten that needed to exist when I was whipping up the `Builder` logic.

gab-umich: The previous comment was made as I was not aware of a particular Java syntax. Now I have modified it so it no longer has such problem.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialParkTest.groovy
gab-umich: Fixed.

.gitignore
archolewa: It doesn't look like we're actually ignoring any of the *Config.json files, and they've all been added to version control
Is that intentional?

luthier/src/main/java/com/yahoo/bard/webservice/application/LuthierBinderFactory.java
```diff
@@ -25,7 +25,9 @@
     protected ConfigurationLoader getConfigurationLoader() {
         LuthierResourceDictionaries resourceDictionaries = new LuthierResourceDictionaries();
         initializeDictionaries(resourceDictionaries);
-        return new LuthierIndustrialPark(resourceDictionaries, getDimensionFactories());
+        LuthierIndustrialPark park = new LuthierIndustrialPark.Builder(resourceDictionaries)
+                .withDimensionFactories(getDimensionFactories()).build();
+        return park;
```
archolewa: No need for the `park` local variable. We can just return `new LuthierIndustrialPark.Builder()...` directly.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -94,22 +94,25 @@ public LogicalTableDictionary getLogicalTableDictionary() {
...
-    public class Builder {
+    public static class Builder {
     
         private Map<String, Factory<Dimension>> dimensionFactories;
         private final ResourceDictionaries resourceDictionaries;
         public Builder(ResourceDictionaries resourceDictionaries) {
             this.resourceDictionaries = resourceDictionaries;
             dimensionFactories = getDefaultDimensionFactories();
         }
+        public Builder() {
```
archolewa: There should be an blank line between constructors:
```java
public Builder(ResourceDictionaries resourceDictionaries) {
...


public Builder() {
...

```

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultDimensionFactory.java
```diff
@@ -0,0 +1,118 @@
...
+import com.yahoo.bard.webservice.data.dimension.DimensionRow;
+import com.yahoo.bard.webservice.data.dimension.SearchProvider;
+import com.yahoo.bard.webservice.data.dimension.metadata.StorageStrategy;
+import org.joda.time.DateTime;
+
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+
+public class DefaultDimensionFactory implements Factory<Dimension> {
```
archolewa: We should probably call this `KeyValueStoreDimensionFactory`, because that's what the "default" dimension type is.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultDimensionFactory.java
```diff
@@ -0,0 +1,118 @@
...
+
+    @Override
+    public Dimension build(String name, ObjectNode configTable, LuthierIndustrialPark resourceFactories) {
+        return new Dimension() {
```
archolewa: Why are we building an anonymous inner class here? Why not an instance of `KeyValueStoreDimension`?

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/ResourceNodeSupplierTest.groovy
```diff
@@ -0,0 +1,92 @@
...
+            ArrayNode fields = node.get("fields")
+            String category = node.get("category").textValue()
+            String description = node.get("description").textValue()
+            String searchProvider = node.get("searchProvider").textValue()
+            String keyValueStore = node.get("keyValueStore").textValue()
+        then:
+            apiName == "testDimension"
+            longName == "a longName for testing"
+            fields.size() == 4
+            fields.get(0).get("name").textValue() == "TEST_PK"
```
archolewa: Might be a bit better to use a for loop for this:
```groovy
List expectedFieldNames = ["primaryKey", "TEST_FIELD_1", "TEST_FIELD_2", "TEST_FIELD_3"]
for (int i = 0; i < fields.size(); i++) {
fielsds.get(i).get("name").textValue() == expectedFieldNames[i]

```
A bit more concise and scales better if we decide to add additional fields or whatever.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/ResourceNodeSupplierTest.groovy
```diff
@@ -0,0 +1,92 @@
...
+                // guarantees that at least one
+                // {
+                //    "tags": [
+                //       "primaryKey"
+                //    ],
+                //    "name": ...
+                // }
+                // exists in the fields value
+                it.get("fields").any {
+                    it.has("tags")
```
archolewa: This doesn't seem to be doing anything. Also, rather than using an if-statement, I think it'd
be clearer if you used `filter`:
```groovy
it.get("fields")
.filter {it.has("tags")
.any {it.get("tags").any {it.textValue() == "primaryKey"
```
`any` returns false I believe if you pass it an empty collection, thought that should be verified.

luthier/src/main/java/com/yahoo/bard/webservice/application/LuthierBinderFactory.java
gab-umich: Fixed.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: Done.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultDimensionFactory.java
gab-umich: Done.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/ResourceNodeSupplierTest.groovy
gab-umich: Addressed.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/ResourceNodeSupplierTest.groovy
gab-umich: The previous idea was that at least one of the field need to satisfy both (has "tags") and (content of "tags" contains primaryKey. Without the first one an empty "fields" will trivially pass.)

I agree to switch to the filter, and sense the design is changed (now any field has a "tags" list, albeit might be empty), the new logic will be different by a bit. It will be live shortly.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/ResourceNodeSupplierTest.groovy
gab-umich: Done.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/ResourceNodeSupplierTest.groovy
gab-umich: I will actually not check primaryKey at all in this new commit, since that is not a syntax level check the ResourceNodeSupplier should test for. 

I will implement it right below the layer it is immediately dependent on.

luthier/config/src/dimensions.lua
```diff
@@ -121,209 +122,218 @@ local searchProviders = dimensionUtils.searchProviders
 local keyValueStores = dimensionUtils.keyValueStores
 
 return {
+    testDimension = {
+        longName = "a longName for testing",
+        description = "a description for testing",
+        fields = FIELDSETS.test,
+        category = "a category for testing",
+        type = "KeyValueStoreDimensionFactory",
```
archolewa: Personally, I would call this `KeyValueStoreDimension`, rather than `KeyValueStoreDimensionFactory`. Yes,
we're using the `KeyValueStoreDimensionFactory` to build a `KeyValueStoreDimension`, and we're using this
as a key into a map of factories, but customers don't care
about that. They only care about what kind of dimension this is, and it's a `KeyValueStoreDimension`.

luthier/config/src/utils/dimensionUtils.lua
```diff
@@ -14,6 +14,7 @@ local M = {}
 
 local misc = require 'utils/misc'
 
+local default_type = "KeyValueStoreDimensionFactory"
```
archolewa: Same here. I would make this `KeyValueStoreDimension` rather than `KeyValueStoreDimensionFactory`.

luthier/pom.xml
```diff
@@ -34,5 +34,11 @@
             <artifactId>fili-core</artifactId>
             <type>test-jar</type>
         </dependency>
+        <dependency>
```
archolewa: Why are we adding this dependency?

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -4,8 +4,10 @@
 
 import com.yahoo.bard.webservice.data.config.ConfigurationLoader;
 import com.yahoo.bard.webservice.data.config.ResourceDictionaries;
-import com.yahoo.bard.webservice.data.dimension.Dimension;
-import com.yahoo.bard.webservice.data.dimension.DimensionDictionary;
+import com.yahoo.bard.webservice.data.dimension.*;
```
archolewa: We don't use star imports in Fili. All imports should be explicitly imported. To keep IntelliJ from automatically collapsing your imports into stars: https://stackoverflow.com/questions/3587071/disable-intellij-starred-package-imports

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -58,6 +60,7 @@ protected LuthierIndustrialPark(
      *
      * @return the dimension instance corresponding to this name.
      */
+    // TODO: Should this really be a package protected method?
```
archolewa: No. It should be a public method. It was just slapped together very quickly so that you'd have something to work with.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/KeyValueStoreDimensionFactory.java
```diff
@@ -0,0 +1,67 @@
+package com.yahoo.bard.webservice.config.luthier.factories;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.yahoo.bard.webservice.config.luthier.Factory;
+import com.yahoo.bard.webservice.config.luthier.LuthierIndustrialPark;
+import com.yahoo.bard.webservice.data.config.LuthierDimensionField;
+import com.yahoo.bard.webservice.data.dimension.*;
```
archolewa: Shouldn't be using star imports.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/KeyValueStoreDimensionFactory.java
```diff
@@ -0,0 +1,67 @@
...
+        SearchProvider searchProvider = resourceFactories.getSearchProvider(
+                configTable.get("searchProvider").textValue()
+        );
+        LinkedHashSet<DimensionField> dimensionFields = new LinkedHashSet<>();
+        for(JsonNode node : configTable.get("fields")) {
+            List<String> tags = new ArrayList<>();
+            for (final JsonNode strNode : node.get("tags")) {
+                tags.add( strNode.textValue() );
+            }
+            dimensionFields.add( new LuthierDimensionField( EnumUtils.camelCase( node.get("name").textValue() ),
```
archolewa: The wrapping here needs to be cleaned up. Each argument should be on its own line, double indented with the
closing parenthesis indented to the same level as `dimensionFields`.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/KeyValueStoreDimensionFactory.java
```diff
@@ -0,0 +1,67 @@
...
+                                                    "Error: currently there is no description",
+                                                            tags) );            // TODO: Magic values!
+        }
+        boolean isAggregatable = true;                                          // TODO: Magic values!
+        LinkedHashSet<DimensionField> defaultDimensionFields = dimensionFields;
+
```
archolewa: Don't forget we need to include a list in the Lua config for default dimension fields.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/factories/KeyValueStoreDimensionFactorySpec.groovy
```diff
@@ -0,0 +1,70 @@
...
+                assert dimensionFields[i].getDescription() == "Error: currently there is no description"
+            }
+            key == dimensionFields[0]
+
+            // type correctness of searchProvider
+            // we cannnot check keyValueStore directly
+            searchProvider.getClass() == NoOpSearchProvider
+    }
+
+    def "When new dimension's name match with the previous dimension, use a reference to the previous one"() {
```
archolewa: This test shouldn't be passing. The fact that it does means there is some
state leaking between tests (i.e the resource dictionary cache). This is bad.
We don't want state leaking between tests, because it imposes an implicit
ordering on our tests. An implicit ordering on tests can lead to sporadic test
failures. For example, this test will fail if it's executed before the test
above it, and there is *no guarantee* on the order the test framework executes
tests. So we need to make a few changes:

1. We need to add a cleanup method that clears the resource dictionaries
between tests.

2. We need to update this test so that it queries industrial park for the same
dimension twice. That way it's standalone.

luthier/pom.xml
gab-umich: That was an oversight, my IDE auto-added that in a commit without my notice.

luthier/config/src/dimensions.lua
gab-umich: Changed.

luthier/config/src/utils/dimensionUtils.lua
gab-umich: Changed.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/factories/KeyValueStoreDimensionFactorySpec.groovy
archolewa: Disregard this comment. I missed the setup method when I was reviewing this test, and incorrectly thought the `testDimension` was being initialized in the first test.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/KeyValueStoreDimensionFactory.java
```diff
@@ -0,0 +1,73 @@
...
+            }
+            dimensionFields.add(
+                    new LuthierDimensionField(
+                            EnumUtils.camelCase( node.get("name").textValue() ),
+                            "Error: currently there is no description",             // TODO: Magic values!
+                            tags)
+            );
+        }
+        boolean isAggregatable = true;                                              // TODO: Magic values!
+        LinkedHashSet<DimensionField> defaultDimensionFields = dimensionFields;     // TODO: include this in Lua configs
```
archolewa: We should address these magic values as a part of this PR, since it's part of adding support for configuring dimensions. It should be easy enough:
two new fields for dimensions: a boolean `isAggregatable`, which defaults to `true`, and a list of strings `dimensionFields`, which defaults to the
empty list.

gab-umich: Sure, I was planning to resolve the rest of the Magic values in the next PR (as much as possible). 
I also fixed the `description`'s default value. Added tests for all of the three, which is pushed as well. Please feel free to give it a final look.

luthier/config/src/dimensions.lua
```diff
@@ -117,213 +122,226 @@ local FIELDSETS = {
...
-
 return {
+    testDimension = {
+        longName = "a longName for testing",
+        description = "a description for testing",
+        fields = FIELDSETS.test,
+        category = "a category for testing",
+        type = "KeyValueStoreDimension",
+        isAggregatable = false,
+        defaultFields = {
```
archolewa: Keep in mind that the `defaultFields` is a subset of `fields` that are shown in
the response when the user doesn't say otherwise. Therefore, this should be a list of
strings that contains just the *names* of the fields to default to. So:
```lua
defaultFields = {"TEST_PK", "TEST_FIELD_1", "TEST_FIELD_2"
```
Note that the defaultFields are all already defined in `fields`.

luthier/config/src/utils/dimensionUtils.lua
```diff
@@ -52,10 +55,13 @@ function M.build_dimensions_config(dimensions)
...
     for name, dimension in pairs(dimensions) do
         local dim_copy = misc.shallow_copy(dimension)
-        dim_copy.apiName = dim_copy.apiName or name
         dim_copy.longName = dim_copy.longName or name
         dim_copy.description = dim_copy.description or name
-        configuration[dim_copy.apiName] = dim_copy
+        dim_copy.type = dim_copy.type or default_type
+        dim_copy.category = dim_copy.category or default_category
+        dim_copy.defaultFields = dim_copy.defaultFields or {}
+        if dim_copy.isAggregatable == nil then dim_copy.isAggregatable = true end  -- defaults isAggregatable to true
```
archolewa: Personally, I would wrap the if statement:
```lua
if dim_copy.isAggregatable == nil then
dim_copy.isAggregatable = true
end
```
It's clearer in my opinion: it highlights the important part: the assignment.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/KeyValueStoreDimensionFactory.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+
+public class KeyValueStoreDimensionFactory implements Factory<Dimension> {
+
+
+    /**
+     * @param fieldsNode the JsonNode object that points to the content of "fields" key
+     *                   or "defaultFields" key
```
archolewa: We should not be using the same function for `defaultFields` as `fields`. `defaultFields` is a list of
strings, *not* a list of tables containing field definitions. Parsing the `defaultFields` should be as
simple as iterating over `defaultFields`, and looking up each one in a `Map<String, DimensionField>` of the
actual dimension fields. Note that this means you'll probably want to parse the `fields` into a `LinkedHashMap`,
use it to hydrate thee `defaultFields` and *then* convert its `values()` into a `LinkedHashSet` for passing
to the Dimension.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/KeyValueStoreDimensionFactory.java
```diff
@@ -0,0 +1,89 @@
...
+        for(JsonNode node : fieldsNode) {
+            List<String> tags = new ArrayList<>();
+            if (node.has("tags")) {
+                for (final JsonNode strNode : node.get("tags")) {
+                    tags.add( strNode.textValue() );
+                }
+            }
+            dimensionFields.add(new LuthierDimensionField(
+                    EnumUtils.camelCase(node.get("name").textValue()),
+                    "Error: currently there is no description",             // TODO: Magic values!
```
archolewa: I would just use the name of the field here for now, or update the Lua stuff to include a descriptions
field for the dimension fields!

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -67,10 +72,35 @@ Dimension getDimension(String dimensionName) {
...
+    String MAGIC_luceneIndexPath = "path";
+    int MAGIC_maxResults = 10000;
+    public SearchProvider getSearchProvider(String searchProviderName) {
+        switch (searchProviderName) {
+            case "com.yahoo.bard.webservice.data.dimension.impl.NoOpSearchProvider":
+                return new NoOpSearchProvider(MAGIC_queryWeightLimit);
+            case "com.yahoo.bard.webservice.data.dimension.impl.LuceneSearchProvider":
+                return new LuceneSearchProvider(MAGIC_luceneIndexPath, MAGIC_maxResults);
+            default:
+                return new ScanSearchProvider();
```
michael-mclawhorn: When you build a search provider lua, 
each of these constructor statements should be wrapped by a SearchProviderFactory.
The map of names to search provider factories should be passed into the LuthierIndustrialPark via constructor.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -67,10 +72,35 @@ Dimension getDimension(String dimensionName) {
...
+    }
+
+    public KeyValueStore getKeyValueStore(String keyValueStoreName) throws UnsupportedOperationException {
+        switch (keyValueStoreName) {
+            // TODO: Magic values!
+            case "com.yahoo.bard.webservice.data.dimension.RedisStore":
+                throw new UnsupportedOperationException(keyValueStoreName);
+            default:
+                return new MapStore();
+        }
```
michael-mclawhorn: When you build the KeyValueStore.lua, these constructor calls should be wrapped by a KeyValueStoreFactory.
The map of named KVS Factories should be passed into the LuthierIndustrialPark via constructor.

michael-mclawhorn: also, there should be no default factory.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/KeyValueStoreDimensionFactory.java
```diff
@@ -0,0 +1,89 @@
...
+        String dimensionName = name;
+        String longName = configTable.get("longName").textValue();
+        String category = configTable.get("category").textValue();
+        String description = configTable.get("description").textValue();
+        LinkedHashSet<DimensionField> dimensionFields = fieldsBuilder(
+                configTable.get("fields")
+        );
+        KeyValueStore keyValueStore = resourceFactories.getKeyValueStore(
+                configTable.get("description").textValue()
+        );
```
michael-mclawhorn: this property should be 'domain' not 'description'
in the old spec this should have been 'keyValueStore'

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/KeyValueStoreDimensionFactory.java
```diff
@@ -0,0 +1,89 @@
...
+        SearchProvider searchProvider = resourceFactories.getSearchProvider(
+                configTable.get("searchProvider").textValue()
+        );
+        LinkedHashSet<DimensionField> defaultDimensionFields= fieldsBuilder(
+                configTable.get("defaultFields")
+        );
```
michael-mclawhorn: Per andrew's comment above you should retrieve a list of strings and copy the dimension fields created above selecting the fields whose names correspond to the strings here.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialParkSpec.groovy
```diff
@@ -0,0 +1,28 @@
...
+                .build()
+
+    }
+
+    def "An industrialPark instance is loaded up without error."() {
+        when:
+            industrialPark.load()
+        then:
+            true
+    }
```
michael-mclawhorn: Additional tests to write:
1- test that retrieving an already created dimension returns the same instance (e.g. isA()  in groovy )
2- test that building a dimension that doesn't have config results in the correct error.
3- test that building a dimension whose config refers to an invalid dependency has the correct error.

michael-mclawhorn:
Also, you have some build issues from the style checking.  Running 'mvn test' at the command line should execute the style checker locally.  I've had problems getting IDE style checking to work while also not sucking.

luthier/config/src/dimensions.lua
gab-umich: In that case, I will need to rewrite a parser in Java. Does it still make sense to take `defaultDimensionField` and build a LinkedHashSet of `DemensionField`? 
I think a list of Strings would be suffice.

luthier/config/src/utils/dimensionUtils.lua
gab-umich: Sure.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/KeyValueStoreDimensionFactory.java
gab-umich: My bad. I did not see this comment when addressing the one two above. 
Is the up-stream going to accept some kind of LuthierFactoryException or should I just throw a runtime error when a string in defaultFields is not a key we built before?

luthier/config/src/dimensions.lua
gab-umich: disregard this. Please use the comment two below to continue this conversation if need be.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/KeyValueStoreDimensionFactory.java
gab-umich: Done. Used name for now. Populating each of the item in all `fields` is a little bit more hairy (because it involves more lua change). I will do it in the next PR, and definitely close this one ASAP!

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/KeyValueStoreDimensionFactory.java
archolewa: Throw a LuthierFactoryException with a useful message, like "Attempted to use unknown field '%s' as a default field." or something.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: Thanks for the comment, these changes are refactored in another branch in my repo, and I have not fully passed the regression test to feel comfortable about a new PR. Can we ignore these for now and I will rebase and open another PR later? 
The branch is here: 
https://github.com/gab-umich/fili/commits/searchProviderTemplate

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: Please see here:
https://github.com/yahoo/fili/pull/919#discussion_r293970424

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/KeyValueStoreDimensionFactory.java
gab-umich: Please see here:
https://github.com/yahoo/fili/pull/919#discussion_r293970424

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/KeyValueStoreDimensionFactory.java
gab-umich: This is worth noting. I will need to edit that in for the next PR.

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialParkSpec.groovy
gab-umich: Some of these are addressed, again, in the new branch and eventually in the next PR, which will wrap up the dimension's implementation. I think it is not advisable to snowball this PR to something uncontrollable and we should close this without more features.

gab-umich: Great suggestions about the tests, and I will definitely include them soon.

gab-umich:
Ready to be checked again.

Issue #927, opened 2019-06-21
Fili module without moduleConfig.properties throws an unhelpful error 


gab-umich:
We get 
```
[ERROR] When new dimension's name match with the previous dimension, use a reference to the previous one(com.yahoo.bard.webservice.config.luthier.factories.KeyValueStoreDimensionFactorySpec)  Time elapsed: 0.01 s  <<< ERROR!
java.lang.NoClassDefFoundError: Could not initialize class com.yahoo.bard.webservice.data.dimension.impl.LuceneSearchProvider
	at com.yahoo.bard.webservice.config.luthier.factories.KeyValueStoreDimensionFactorySpec.setup(KeyValueStoreDimensionFactorySpec.groovy:24)

[INFO]
[INFO] Results:
[INFO]
[ERROR] Errors:
[ERROR]   KeyValueStoreDimensionFactorySpec.setup:24 » SystemConfig java.util.NoSuchElem...
[ERROR]   KeyValueStoreDimensionFactorySpec.setup:24 » NoClassDefFound Could not initial...
```
when a new module does not configure the moduleConfig.properties. 

we need to have clearer message throw when this happens.

gab-umich:
Fixed by #928. Closed.

PR #928, opened 2019-06-21
SystemConfigException fixed

gab-umich:
Fixed Issue https://github.com/yahoo/fili/issues/927#issue-459277455

new tests written and passed
logger in place

### DIFF
--- a/fili-system-config/pom.xml
+++ b/fili-system-config/pom.xml
@@ -40,6 +40,11 @@
             <artifactId>spock-core</artifactId>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>org.spockframework</groupId>
+            <artifactId>spock-core</artifactId>
+            <scope>compile</scope>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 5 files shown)


fili-system-config/pom.xml
```diff
@@ -40,6 +40,11 @@
             <artifactId>spock-core</artifactId>
             <scope>compile</scope>
         </dependency>
+        <dependency>
```
gab-umich: Not sure if this bunch is needed. 
Please double check.

fili-system-config/pom.xml
```diff
@@ -40,6 +40,11 @@
...
+            <groupId>org.spockframework</groupId>
+            <artifactId>spock-core</artifactId>
+            <scope>compile</scope>
```
will-lauer: This appears to be a mistake. At best, spock-core should be test scope

gab-umich: Addressed. Thanks for pointing it out.

gab-umich:
Just a reminder for you that I don't have write access to the yahoo/fili repo yet. 
Someone should merge this into master as long as this passes previous tests.

PR #931, opened 2019-07-01
Single Data Source physical table

gab-umich:
Encompassing StrictPhysicalTable and PermissivePhysicalTable.

- Largely refactored the Lua generation scripts
- Considerable new code regarding how we configure single data source physical tables. 
- laid groundwork for other tables
- clean up some bad comment/logic from previous PR


luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
```diff
@@ -0,0 +1,90 @@
...
+        this.logicalToPhysicalColumnNames = new LinkedHashMap<>();
+        LuthierValidationUtils.validateField(
+                configTable.get("logicalToPhysicalColumnNames"),
+                PHYSICAL_TABLE,
+                name,
+                "searchProvider"
+        );
+        JsonNode columnNameMapNode = configTable.get("logicalToPhysicalColumnNames");
+        columnNameMapNode.forEach(
+                node -> logicalToPhysicalColumnNames.put(
```
michael-mclawhorn: You could skip assigning this to a local variable

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/PermissivePhysicalTableFactory.java
```diff
@@ -0,0 +1,35 @@
...
+     *
+     * @param name  the config dictionary name (normally the apiName)
+     * @param configTable  the json tree describing this config entity
+     * @param resourceFactories  the source for locating dependent objects
+     *
+     * @return  A newly constructed config instance for the name and config provided
+     */
+    @Override
+    public ConfigPhysicalTable build(String name, ObjectNode configTable, LuthierIndustrialPark resourceFactories) {
+        super.prepare(name, configTable, resourceFactories);
```
michael-mclawhorn: Can't modify factory on build.

gab-umich: Done by using the bean approach
(setting up the Param class to hold the necessary information)

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
gab-umich: Modified.

gab-umich:
At this point, the PR is expected to be a point of reference. 
I expect to have this PR merged in at any point from now on, so I will branch off the logical table implementation from this point of the `physicalTable` branch, and then rebase when necessary.

luthier/config/src/tables.lua
```diff
@@ -76,15 +78,21 @@ local air_quality_dimensions = {
 
 M.physical = {
     wikiticker = {
+        dateTimeZone = "UTC",
+        type = "strict",
```
archolewa: Should document these two keys in the documentation above.

luthier/config/src/tables.lua
```diff
@@ -76,15 +78,21 @@ local air_quality_dimensions = {
...
         metrics = {
             "added",
             "delta",
             "deleted"
         },
         dimensions = wikipedia_dimensions,
-        granularity = "hour"
+        granularity = "hour",
+        logicalToPhysicalColumnNames = {
```
archolewa: Should document this key in the documentation above.

luthier/config/src/utils/tableUtils.lua
```diff
@@ -12,33 +12,72 @@ local misc = require 'utils./misc'
 -- Build Config
 -------------------------------------------------------------------------------
 
+--- Build a specific physical table into a json object.
```
archolewa: This is incorrect. This function doesn't actually build a JSON object, but
rather another Lua table. It massages the user provided physical table
configuration into a form that makes it easier for Fili to consume.

luthier/config/src/utils/tableUtils.lua
```diff
@@ -12,33 +12,72 @@ local misc = require 'utils./misc'
...
+--
+-- for each of the pair in logicalToPhysicalColumnNames, e.g.
```
archolewa: This is going to need some sort of context and/or transition. The
logicalToPhysicalColumnNames is only one part of the transformation, but
the documentation reads like this is the only transformation. To people
who aren't familiar with the codebase, this could be misleading. To those
who are, this reads like a non-sequitor.

You should start off with something like `Most of the transformations
are straightforward (i.e. initializing defaults), however
`logicalToPhysicalColumnNames` is converted into a rather weird
format.`

luthier/config/src/utils/tableUtils.lua
```diff
@@ -12,33 +12,72 @@ local misc = require 'utils./misc'
...
-    }
+    local physical = {}
+    local logical = {}
 
     for name, physical_table in pairs(tables.physical) do
-        local copy = misc.shallow_copy(physical_table)
-        copy.name = copy.name or name
-        copy.description = copy.description or name
-        table.insert(configuration.physical, copy)
+        physical[name] = physical_table_build(name, physical_table)
```
archolewa: We should be doing `physical[copy.name or name]` here like we do with `logical`.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -37,29 +41,36 @@
...
         Supplier<ObjectNode> dimensionConfig = new ResourceNodeSupplier("DimensionConfig.json");
         Supplier<ObjectNode> searchProviderConfig = new ResourceNodeSupplier("SearchProviderConfig.json");
-        dimensionFactoryPark = new FactoryPark<>(dimensionConfig, dimensionFactories);
-        searchProviderFactoryPark = new FactoryPark<>(searchProviderConfig, searchProviderFactories);
+        Supplier<ObjectNode> physicalTableConfig = new ResourceNodeSupplier("PhysicalTableConfig.json");
+        // Supplier<ObjectNode> logicalTableConfig = new ResourceNodeSupplier("LogicalTableConfig.json");
+        this.dimensionFactoryPark = new FactoryPark<>(dimensionConfig, dimensionFactories);
+        this.searchProviderFactoryPark = new FactoryPark<>(searchProviderConfig, searchProviderFactories);
+        this.physicalTableFactoryPark = new FactoryPark<>(physicalTableConfig, physicalTableFactories);
+        this.metadataService = new DataSourceMetadataService();
```
archolewa: We should not be hardcoding this. I'm not fully up to speed on how the
metadataService is supposed to work, but if we're expecting to use a single
service for every physical table, then we should make this a constructor
parameter on the industrial park. If we expect different physical tables to
have different metadata services, well, then we need to give our customers the
ability to configure and attach them to their physical table configuration.

@mclawhor How is the metadataService supposed to work?

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -81,9 +92,10 @@ public Dimension getDimension(String dimensionName) {
     }
 
     /**
-     * Builds a SearchProvider for a specific domain.
+     * Retrieve or build a SearchProvider for a specific domain.
      *
      * @param domain  a string that is associated with the type o
```
archolewa: Not strictly related to this PR, but it looks like the documentation here was never finished.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -98,7 +110,8 @@ public SearchProvider getSearchProvider(String domain) {
     /**
      * Bare minimum.
```
archolewa: Not strictly related to this PR, but we should not be relying on
`default` to select the `MapStore`.  We should explicitly look for the MapStore
class name, and throw an exception under `default`. Otherwise, a typo could
mean customers end up using an in-memory KeyValueStore instead of whatever
store they intended, and that could be very bad if it's a sufficiently large
dimension.

Also, I thought we were going to complete KeyValueStore before moving onto
tables. This function doesn't look particularly done, in that it still has
`TODO: Magic Values` and we haven't implemented a factory for the RedisStore
yet.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -112,10 +125,30 @@ public KeyValueStore getKeyValueStore(String keyValueStoreName) throws Unsupport
         }
     }
 
+    /**
+     * Retrieve or build a PhysicalTable.
+     *
+     * @param tableName the name for the PhysicalTable to be retrieved or built.
+     *
+     * @return the dimension instance corresponding to this name.
```
archolewa: Should be `PhysicalTable` instance, not `dimension` instance.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -112,10 +125,30 @@ public KeyValueStore getKeyValueStore(String keyValueStoreName) throws Unsupport
...
+        return physicalTableDictionary.get(tableName);
+    }
+
+    public DataSourceMetadataService getMetadataService() {
+        return metadataService;
+    }
 
     @Override
     public void load() {
         dimensionFactoryPark.fetchConfig().fieldNames().forEachRemaining(this::getDimension);
```
archolewa: I don't think we need this line anymore. Dimensions should be configured on demand as physical
table suse them. That being said, we can probably remove this in a final pass so that you don't
have to worry about tests exploding or something.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/PermissivePhysicalTableFactory.java
```diff
@@ -0,0 +1,36 @@
...
+package com.yahoo.bard.webservice.config.luthier.factories;
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.yahoo.bard.webservice.config.luthier.LuthierIndustrialPark;
+import com.yahoo.bard.webservice.data.config.LuthierPhysicalTableParam;
+import com.yahoo.bard.webservice.table.ConfigPhysicalTable;
+import com.yahoo.bard.webservice.table.PermissivePhysicalTable;
+
+/**
+ * A factory that is used by default to support Simple (non-Composite) Physical Table.
```
archolewa: We should specify here what it means for a physical table to be "permissive."

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/StrictPhysicalTableFactory.java
```diff
@@ -0,0 +1,36 @@
...
+package com.yahoo.bard.webservice.config.luthier.factories;
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.yahoo.bard.webservice.config.luthier.LuthierIndustrialPark;
+import com.yahoo.bard.webservice.data.config.LuthierPhysicalTableParam;
+import com.yahoo.bard.webservice.table.ConfigPhysicalTable;
+import com.yahoo.bard.webservice.table.StrictPhysicalTable;
+
+/**
+ * A factory that is used by default to support Simple (non-Composite) Physical Table.
```
archolewa: We should specify here what it means for a physical table to be strict.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
archolewa: After talking to @michael-mclawhorn, it looks like we're basically doing this as is already. So we don't have to worry about changing this yet.

luthier/config/src/tables.lua
gab-umich: documented. Also enriched the rest of the physical table doc

luthier/config/src/tables.lua
gab-umich: done.

luthier/config/src/utils/tableUtils.lua
gab-umich: Changed wording

luthier/config/src/utils/tableUtils.lua
gab-umich: Michael and I had the concept earlier to use a separate lua file to supply all defaulting parameter for the user to see. Due to that fact I will leave the default live in the table.lua and not mention them in tableUtils.lua (and all corresponding ones), for now. 

Also addressed your comments

luthier/config/src/utils/tableUtils.lua
gab-umich: I don't think we should. For one, the table is not doing that. 

For table, dimension, and metric, I think we had the discussion to transform the lua format from 
```
"objName": {
    "name": "objName",
    "property1": "objProperty1",
    ...
}
```
```
"objName": {
    "property1": "objProperty1",
    ...
}
```
Since the prior one does not give any benefit for the user. Since we have made the examples in `dimensions.lua` to look like the later one, there should not be this defaulting in `*Utils.lua` to hint otherwise.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: I am also not fully happy with the fact this is hard coded. 
but LIP is clearly the place that makes _most_ sense for the metadataService to live in, unless we supply user an entire different scope to configure that.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: Sometimes I wish the styler linter can semantically check whether my comments are finished...

Good catch! Fixed in this PR.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: The factory for RedisStore is scheduled to be pushed until we are finished. That being said, I agree with doing it _while not over-clogging this PR_. 

I can open a quick PR to implement KeyValueStore before opening the ones for LogicalTable.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: I only realized it and fixed it in the Logical table implementation. I can change it here.

luthier/config/src/utils/tableUtils.lua
archolewa: Consolidating all the defaults in a single file is a very good idea.

luthier/config/src/utils/tableUtils.lua
archolewa: We should update the documentation around tables then, because I think the documentation still refers to a `name` field that defaults to the key.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: Sure. 
The dimensions are not configured on-demand as for now, though (it is now being configured when the LIP is loaded). We can change that later, as I do not see that change can drastically change things.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/PermissivePhysicalTableFactory.java
gab-umich: Here or in the corresponding subclass? @archolewa

archolewa: This is the corresponding subclass.

luthier/config/src/utils/tableUtils.lua
gab-umich: Yeah, that is part of my change that will be pushed in a couple of minutes.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/PermissivePhysicalTableFactory.java
archolewa: Oh you, mean here or `PermissivePhysicalTable`. I don't care.

gab-umich: I wrote them in the corresponding subclass and hinted them here. 
But I'd like to talk with @michael-mclawhorn before pushing.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/StrictPhysicalTableFactory.java
gab-umich: same as above.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
archolewa: That's fine.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
```diff
@@ -0,0 +1,94 @@
+// Copyright 2019 Oath Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.config.luthier.factories;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
```
archolewa: Import order is incorrect. `com.*` imports are in a separate paragraph immediately after `com.yahoo.*`

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
```diff
@@ -0,0 +1,94 @@
...
+
+import org.joda.time.DateTimeZone;
+
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.Locale;
+
+/**
+ * A factory that is used by default to support Simple (non-Composite) Physical Table.
+ * Classes that extend this abstract class:
```
archolewa: Specifying this isn't necessary. If JavaDoc is generated, the JavaDoc will figure this out for you, and
if people have access to the codebase, they'll have some strategy for using their tools to find the subclasses.
Probably they'll use their IDE.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+abstract class SingleDataSourcePhysicalTableFactory implements Factory<ConfigPhysicalTable> {
+    private static final String ENTITY_TYPE = "single data source physical table";
+
+    /**
+     * Build the parameter for the subclass of SingleDataSourceParams to use.
+     *
+     * @param name  name of the LuthierTable as a String
+     * @param configTable  ObjectNode that points to the value of corresponding table entry in config file
+     * @param resourceFactories  should be the industrial park that needs to be paseed along to the builders
```
archolewa: Typo. `paseed` should be `passed`. Also, I would rephrase this as something like `@param resourceFactories  Source of the this table's dependencies (metrics, dimensions, etc)`

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
```diff
@@ -0,0 +1,94 @@
...
+        params.columns = new LinkedHashSet<>();
+
+        JsonNode dimensionsNode = configTable.get("dimensions");
+        dimensionsNode.forEach(
+                node -> params.columns.add(new DimensionColumn(resourceFactories.getDimension(node.textValue())))
+        );
+
+        JsonNode metricsNode = configTable.get("metrics");
+        metricsNode.forEach(
+                node -> params.columns.add(new MetricColumn(node.textValue()))
```
archolewa: We should be calling `resourceFactories.getMetric(node.textValue()) here as well, right? Something like:
```java
node -> {
String metricName = node.textValue();
resourceFactories.getMetric(metricName);
params.columns.add(new MetricColumn(metricName));

```
To make sure we load in the metrics.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
```diff
@@ -0,0 +1,94 @@
...
+     */
+    protected void validateFields(String name, ObjectNode configTable) {
+        LuthierValidationUtils.validateField(configTable.get("granularity"), ENTITY_TYPE, name, "granularity");
+        LuthierValidationUtils.validateField(configTable.get("dateTimeZone"), ENTITY_TYPE, name, "dateTimeZone");
+        LuthierValidationUtils.validateField(configTable.get("dimensions"), ENTITY_TYPE, name, "dimensions");
+        LuthierValidationUtils.validateField(
+                configTable.get("logicalToPhysicalColumnNames"),
+                ENTITY_TYPE,
+                name,
+                "searchProvider"
```
archolewa: Field name here is wrong. It should `"logicalToPhysicalColumnNames"` not `"searchProvider"`.

luthier/config/src/tables.lua
```diff
@@ -18,11 +18,25 @@ local M = {}
...
-
-    * name - The name of the physical table
-        Defaults to the configuration's table key
+    * The table is keyed on the name of the physical table
     * description - Brief documentation about the physical table.
         Defaults to the configuration's table key
+    * type - [Mandatory] The nature of this physical table, used to build the table internally.
+        Currently supports
+            * strict - strict single data source physical table
+            * permissive - permissive single data source physical table
```
archolewa: We should define what `strict` and `permissive` mean here as well.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
gab-umich: Makes sense. I will remove that then.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
gab-umich: modified. Together with a couple more imports.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
gab-umich: Agreed.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
gab-umich: Actually if you look slightly above that, shouldn't we keep a dictionary for DimensionColumn as well? 
That can look pretty bulky. An alternative I can think of is to extend the `DimensionColumn` class and make a singleton class to ensure different tables will query references to a same Column object when the underlying dimension matches. 
Not sure if that is really a Java-style design choice.

gab-umich: @archolewa

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
gab-umich: 😬Copy and Paste fail

luthier/config/src/tables.lua
gab-umich: added brief explanations.

archolewa:
Opened a PR with some improved documentation for permissive and strict physical table in the Lua code: https://github.com/gab-umich/fili/pull/6. I felt the old documentation was easy to misunderstand, especially if you don't understand what Fili means by "availability" (it could easily be read as "this table can't be used at all if xyz"). I've attempted to clarify that.

Once that's merged in, I'll be :+1:

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
gab-umich: reflected without caching the Column objects.

PR #936, opened 2019-07-12
Key value store using factory to build

gab-umich:
KeyValueStore is now mimicking the SearchProvider in its way of living in our code base. 

### DIFF
--- a/luthier/src/main/java/com/yahoo/bard/webservice/application/LuthierBinderFactory.java
+++ b/luthier/src/main/java/com/yahoo/bard/webservice/application/LuthierBinderFactory.java
@@ -11,6 +11,7 @@ import com.yahoo.bard.webservice.data.config.metric.MetricLoader;
 import com.yahoo.bard.webservice.data.config.metric.makers.MetricMaker;
 import com.yahoo.bard.webservice.data.config.table.TableLoader;
 import com.yahoo.bard.webservice.data.dimension.Dimension;
+import com.yahoo.bard.webservice.data.dimension.KeyValueStore;
 import com.yahoo.bard.webservice.data.dimension.SearchProvider;
 
 import java.util.Collections;
@@ -30,6 +31,7 @@ public class LuthierBinderFactory extends AbstractBinderFactory {
         LuthierIndustrialPark.Builder builder = new LuthierIndustrialPark.Builder(resourceDictionaries);
         getDimensionFactories().ifPresent(builder::withDimensionFactories);
         getSearchProviderFactories().ifPresent(builder::withSearchProviderFactories);
+        getKeyValueStoreFactories().ifPresent(builder::withKeyValueStoreFactories);
         return builder.build();
     }
 
@@ -55,6 +57,14 @@ public class LuthierBinderFactory extends AbstractBinderFactory {
     }
 
     /**
+     * Temp stuff.
+     * @return  Optional.empty()
+     */
+    protected Optional<Map<String, Factory<KeyValueStore>>> getKeyValueStoreFactories() {
+        return Optional.empty();
+    }
+
+    /**
      * Extension point to add default maker factories.
      *
      * @return  Initializing dimension factories.
@@ -72,6 +82,9 @@ public class LuthierBinderFactory extends AbstractBinderFactory {
         resourceDictionaries.getMetricMakerDictionary().putAll(LuthierResourceDictionaries.defaultMakerDictionary());
         resourceDictionaries.getSearchProviderDictionary().putAll(
                 LuthierResourceDictionaries.defaultSearchProviderDictionary()
+        );
+        resourceDictionaries.getKeyValueStoreDictionary().putAll(
+                LuthierResourceDictionaries.defaultKeyValueStoreDictionary()
         );
     }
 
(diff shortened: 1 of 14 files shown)


luthier/src/main/java/com/yahoo/bard/webservice/application/LuthierBinderFactory.java
```diff
@@ -54,6 +56,14 @@ protected ConfigurationLoader getConfigurationLoader() {
         return Optional.empty();
     }
 
+    /**
+     * Temp stuff.
```
archolewa: We should flesh this out while we're here.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/FactoryPark.java
```diff
@@ -14,14 +14,14 @@
...
-    static final String ENTITY_TYPE = "factory park";
+    private static final String ENTITY_TYPE = "factory park";
 
     private static final String FACTORY_KEY = "type";
 
     private static final String UNKNOWN_FACTORY_NAME = "factory name '%s' in config is not known to the Luthier module";
 
     // Use a supplier to support deferred loading
-    protected final Supplier<ObjectNode> configSource;
+    private final Supplier<ObjectNode> configSource;
```
archolewa: If we're going to make this private, then we should provide a protected getter so that subclasses can get their hands on the configuration.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/KeyValueStoreDimensionFactory.java
```diff
@@ -96,10 +96,8 @@ public Dimension build(String name, ObjectNode configTable, LuthierIndustrialPar
         String longName = configTable.get("longName").textValue();
         String category = configTable.get("category").textValue();
         String description = configTable.get("description").textValue();
-        KeyValueStore keyValueStore = resourceFactories.getKeyValueStore(configTable.get("keyValueStore").textValue());
-        SearchProvider searchProvider = resourceFactories.getSearchProvider(
-                configTable.get("domain").textValue()
-        );
+        KeyValueStore keyValueStore = resourceFactories.getKeyValueStore(configTable.get("domain").textValue());
+        SearchProvider searchProvider = resourceFactories.getSearchProvider(configTable.get("domain").textValue());
```
archolewa: I vaguely remember why we're doing this, but it's going to be very confusing for people. So we need to make very sure we document this. We should document this in `dimensions.lua` and on the DimensionFactory
interface as part of the contract format that @michael-mclawhorn suggested.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/MapKeyValueStoreFactory.java
```diff
@@ -0,0 +1,29 @@
...
+package com.yahoo.bard.webservice.config.luthier.factories;
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.yahoo.bard.webservice.config.luthier.Factory;
+import com.yahoo.bard.webservice.config.luthier.LuthierIndustrialPark;
+import com.yahoo.bard.webservice.data.dimension.KeyValueStore;
+import com.yahoo.bard.webservice.data.dimension.MapStore;
+
+/**
+ * A factory that is used by default to support KeyValueStore Dimensions.
```
archolewa: This documentation is incorrect. This factory doesn't build KeyValueStore Dimensions, but rather KeyValueStores.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/KeyValueStoreDimensionFactory.java
archolewa: We don't have to this yet, but is definitely something we need to mention once we've figured what we want to the contract format to look like.

archolewa:
Opened a PR against this branch with some documentation improvements: https://github.com/gab-umich/fili/pull/7

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/KeyValueStoreDimensionFactory.java
gab-umich: Agreed. We can start that as a parallel task, and now is definitely a reasonable time for us to start.

luthier/src/main/java/com/yahoo/bard/webservice/application/LuthierBinderFactory.java
gab-umich: This is marked as Temp stuff not because that it is incorrect (in fact it is exactly how we should implement in the future), but because the `concept` implementation on @michael-mclawhorn's branch will replace this with a wrapper. 

this will get changed whenever that PR is ready.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/FactoryPark.java
gab-umich: Is it really likely that users will subclass the FactoryPark? 
I thought the only LuthierBinderFactory is going to be extended.

PR #938, opened 2019-07-15
Luthier Logical Table Configuration

gab-umich:
Implements Logical Table by grouping a set of logical tables that shares everything but granularities into a `LogicalTableGroup`, then passing it through the `AbridgedLogicalTableGroupFactory`. 
(It is called abridged because the configuration can potentially be enriched (as if some of the configurable arguments are abridged). This is more of a proof of concept that _some_ meaningful LogicalTableGroup can be built.)


luthier/config/src/tables.lua
```diff
@@ -168,15 +167,26 @@ M.physical = {
 
 M.logical = {
     wikipedia = {
+        type = "abridged",
         metrics =  {"count", "added", "delta", "deleted"},
         dimensions = wikipedia_dimensions,
-        granularity = {"all", "hour", "day"},
-        physicalTables = {"wikiticker"}
+        granularities = {
+            --"all",
```
michael-mclawhorn: Why is all commented out?

luthier/config/src/tables.lua
```diff
@@ -136,16 +136,15 @@ M.physical = {
     Logical tables are defined in table M.logical. They map logical table
     names to their configuration:
 
-    * name - The name of the table
-        Defaults to the configuration's key
+    * key - The name of the logical table
```
archolewa: I see what you're trying to do here, but this is misleading. It implies there's a key named `key` on the
table that contains the logical table name. Besides, this is redundant with `They map logical table
names to their configuration:` above.

luthier/config/src/tables.lua
```diff
@@ -136,16 +136,15 @@ M.physical = {
...
     * metrics - A set of API metrics' name for this logical table, these 
         should be a subset of the metrics configured in metrics.lua.
     * dimensions - A set of dimensions for this logical table, the set of
         dimensions should be a subset of dimensions in its dependent physical
         tables.
-    * granularity - A group of available granularities of this logical table, the
+    * granularities - A group of available granularities of this logical table, the
         granularity can be "all", "hour", "day", "week", or "month".
     * physicalTables - A list of the names of the physical tables that this
         logical table depends on
```
archolewa: Missing documentation for `dateTimeZone`, and `type`. Make sure the documentation includes what the
default is, if any.

luthier/config/src/tables.lua
```diff
@@ -168,15 +167,26 @@ M.physical = {
 
 M.logical = {
     wikipedia = {
+        type = "abridged",
```
archolewa: Why `abridged`? Why not use something like `fili` since we're just building the boring old `LogicalTable` that comes
with Fili?

luthier/config/src/utils/tableUtils.lua
```diff
@@ -59,14 +59,23 @@ local physical_table_build = function(name, physical_table)
     return copy
 end
 
+local logical_table_build = function(name, logical_table)
```
archolewa: It's more idiomatic to use
```lua
local function logical_table_build(name, logical_table)
```

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -159,15 +168,56 @@ public DataSourceMetadataService getMetadataService() {
...
+     * Retrieve or build a LogicalTable.
+     *
+     * @param tableId the identifier for table name and granularity to be retrieved or built
+     *
+     * @return the logicalTable instance corresponding to this TableIdentifier.
+     */
+    public LogicalTable getLogicalTable(TableIdentifier tableId) {
+        LogicalTableDictionary logicalTableDictionary = resourceDictionaries.getLogicalDictionary();
+        if (! logicalTableDictionary.containsKey(tableId)) {
+            String tableName = tableId.getKey() + tableId.getValue().toString();
```
archolewa: How is this even working? If we have a table with name `table1` and granularities `day`, `week`, `month`,
then we're creating the following groups of logical tables: `table1day, table1week, table1month` *each* with
granularities `day, week, month`, when we should be creating a group of logical tables : `table1` with
granularities `day, week, month`.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -159,15 +168,56 @@ public DataSourceMetadataService getMetadataService() {
...
+        LogicalTableDictionary logicalTableDictionary = resourceDictionaries.getLogicalDictionary();
+        LogicalTableGroup logicalTableGroup = logicalTableGroupFactoryPark.buildEntity(tableKey, this);
+        logicalTableDictionary.putAll(logicalTableGroup);
+    }
+
     @Override
     public void load() {
         dimensionFactoryPark.fetchConfig().fieldNames().forEachRemaining(this::getDimension);
         physicalTableFactoryPark.fetchConfig().fieldNames().forEachRemaining(this::getPhysicalTable);
+        logicalTableGroupFactoryPark.fetchConfig().fieldNames().forEachRemaining(this::putLogicalTableByGroup);
```
archolewa: This should render everything else in this method redundant. We should try running things with this the only
statement in `load` and see if it works.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
```diff
@@ -0,0 +1,97 @@
...
+import com.yahoo.bard.webservice.table.PhysicalTable;
+import com.yahoo.bard.webservice.table.TableGroup;
+import com.yahoo.bard.webservice.table.TableIdentifier;
+import org.joda.time.DateTimeZone;
+
+import java.util.LinkedHashSet;
+import java.util.Locale;
+
+/**
+ * A factory that is used by default to support LogicalTableGroup.
```
archolewa: We should include some documentation here about what a LogicalTableGroup is, and why we're building a collection of
LogicalTables from one config source instead of a single LogicalTable from one config source like people might expect.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
```diff
@@ -0,0 +1,97 @@
...
+ */
+public class AbridgedLogicalTableGroupFactory implements Factory<LogicalTableGroup> {
+    private static final String ENTITY_TYPE = "abridged logical table group";
+
+    /**
+     * Build a LogicalTable instance.
```
archolewa: This is wrong. It's not actually building a LogicalTable instance, but one LogicalTable per configured granularity.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
```diff
@@ -0,0 +1,97 @@
...
+     *
+     * @param name  the config dictionary name (normally the apiName)
```
archolewa: This is weirdly phrased. I would just say `@param name  The apiName of the logical table`

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
```diff
@@ -0,0 +1,97 @@
...
+     * @param configTable  ObjectNode that points to the value of corresponding table entry in config file
```
archolewa: I would rephrase this as `@param configTable  The table's configuration`

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
```diff
@@ -0,0 +1,97 @@
...
+     * @param resourceFactories  the source for locating dependent objects
+     *
+     * @return  A newly constructed config instance for the name and config provided
+     */
+    @Override
+    public LogicalTableGroup build(String name, ObjectNode configTable, LuthierIndustrialPark resourceFactories) {
+        validateFields(name, configTable);
+
+        LogicalTableGroup tableDictionary = new LogicalTableGroup();
```
archolewa: I would call this variable `tableGroup`. Just because we're effectively using a `LogicalTableDictionary` to define
our `LogicalTableGroup`, doesn't mean we need to advertise that fact in the variable name.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
```diff
@@ -0,0 +1,97 @@
...
+                physicalTables,
+                apiMetricNames,
+                dimensions
+        );
+
+        configTable.get("granularities").forEach(
+                (JsonNode node) -> {
+                    ZonedTimeGrain granularity = DefaultTimeGrain.valueOf(
+                            node.textValue().toUpperCase(Locale.US)
+                    ).buildZonedTimeGrain(DateTimeZone.forID(configTable.get("dateTimeZone").textValue()));
```
archolewa: The formatting here can be improved abit. First, I would pull `DateTimeZone.forID(configTable.get("dateTimeZone").textValue())`
into its own variable `timeZone`. Then, I would wrap the `granularity` call on the method invocations, so:

```java
DateTimeZone timeZone = DateTimeZone.forID(configTable.get("dateTimeZone").textValue());
ZonedTimeGrain granularity = DefaultTimeGrain
.valueOf(node.textValue().toUpperCase(Locale.US))
.buildZonedTimeGrain(timeZone);
```
I generally find that it can be difficult to parse nested function invocations (i.e. `f(g(h(0, 1), 2))` . Better to
flatten things by pulling out a local variable or two. Furthermore, when I'm invoking two more methods in a single statement,
I often find it clearer to put each method invocation on its own line, rather than trying to wrap the arguments of one
method, but not the other.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * makes sure the necessary fields exist in the json config.
+     *
+     * @param name  the config dictionary name (normally the apiName)
+     * @param configTable  ObjectNode that points to the value of corresponding table entry in config file
+     */
+    private void validateFields(String name, ObjectNode configTable) {
+        LuthierValidationUtils.validateField(configTable.get("granularities"), ENTITY_TYPE, name, "granularity");
```
archolewa: The last argument should be `"granularities"` not `"granularity"`.

luthier/src/main/java/com/yahoo/bard/webservice/data/config/LogicalTableGroup.java
```diff
@@ -0,0 +1,15 @@
...
+import com.yahoo.bard.webservice.table.LogicalTable;
+import com.yahoo.bard.webservice.table.TableIdentifier;
+
+import java.util.LinkedHashMap;
+
+/**
+ * Represents the set of LogicalTables that differs by granularity.
+ * This is used to build multiple LogicalTables from the config file.
+ */
+public class LogicalTableGroup extends LinkedHashMap<TableIdentifier, LogicalTable> {
```
archolewa: I thought we were going to make `LogicalTableGroup` an interface, and have the `LogicalTableDictionary` implement `LogicalTableGroup`?

luthier/src/test/resources/MetricConfig.json
```diff
@@ -1,82 +1 @@
...
-    "dependencyMetricNames": [
-      "delta"
-    ],
-    "description": "delta",
-    "maker": "doubleSum",
-    "longName": "delta",
-    "druidMetric": "delta"
-  }
-]
\ No newline at end of file
```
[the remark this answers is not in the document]
gab-umich: I cannot configure it, but if someone can, please check if there is an option to turn off the semicolon check for json.

luthier/config/src/tables.lua
gab-umich: This is inconsistent from what I did in some of the other lua files. It might be caused by a copy-paste from an older file. 
The idea is to write: 
```markdown
    Physical tables are defined in table M.physical that map physical table's
    name (should be the same as the dataset names defined in druid's config
    file) to a dictionary of physical table configuration:
    * The table is keyed on the name of the physical table
    * description - Brief documentation about the physical table.
        Defaults to the configuration's table key
    ...
```
I will do that instead. By the way, is the above description correct? It might be outdated in the sense that we do not need to have the physical table name match with the apiName in druid store, as we now have the `logicalToPhysicalColumnNames`.

luthier/config/src/tables.lua
gab-umich: Done.

luthier/config/src/tables.lua
gab-umich: I have not implemented the casing on all, yet. I will start working on that now.

luthier/config/src/tables.lua
gab-umich: It is addressed in the most recent update in the comment of tables.lua. 
Simply put, there will be another factory to handle the build with more configurable items. 

I will make that comment-amend commit live after I get all the suggestions about documentation done.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
gab-umich: Fixed. Another copy-and-paste fail. I will be more careful about them

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
gab-umich: done.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
gab-umich: done.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
gab-umich: That is unnecessarily confusing. if you look 20 lines below that, we will have another variable typed `TableGroup`, an in fili-core, this type is used to describe a fully configured set of physical tables. (By that I mean it contains `apiMetricName` and `dimensions` as well)
We can either rename this as `logicalTableGroup`, or, maybe just leave it as is. 
Also I suggest somehow refactor that `TableGroup` type name. It is super confusing at first.

gab-umich: On a side note, the `LogicalTableGroup` is not an interface, but a concrete class at this point. 

The reason is that we wish to let `LogicalTableDictionary` to implement that eventually, but that will lead to one of the two options:

1. directly introduce `LogicalTableGroup` to fili-core/data, which is considered unstable per last talk
2. let fili-core and all other module take the luthier as a dependency, which I consider as an ugly workaround for an unnecessarily small inheritance hack.
3. (there is actually a third way I thought of:) name a concrete class `LuthierLogicalTableGroup` in luthier, and an interface `LogicalTableGroup` in luthier. let `LuthierLogicalTableGroup` extend `LogicalTableDictionary` and then implement `LogicalTableGroup`. This way we have no module level dependency and also no premature departure from master. Michael said: "no."

So for now, the `LogicalTableGroup` is just a class. We can lift it to fili-core and do the restructuring when we clear off problems in luthier.

luthier/src/main/java/com/yahoo/bard/webservice/data/config/LogicalTableGroup.java
gab-umich: See above. I did not see this comment when writing that.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
gab-umich: Yep.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
gab-umich: I wrote a doc, please feel free to give it a check.

luthier/config/src/tables.lua
archolewa: The physical table name still needs to correspond to Druid table. The `logicalToPhysicalColumnNames` only allows translating *column* names.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
archolewa: I suggest renaming it to `logicalTableGroup` then. And `TableGroup` is probably never going to be renamed. Pretty much everyone relies on it in their configuration, so it's not worth the pain of changingl.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
gab-umich: I agree. Changed in the new commit.

luthier/config/src/tables.lua
michael-mclawhorn: Make this "default" rather than abridged.

luthier/config/src/tables.lua
archolewa: Use https://github.com/gab-umich/fili/blob/logicalTable/fili-core/src/main/java/com/yahoo/bard/webservice/data/time/AllGranularity.java for the all granularity.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
```diff
@@ -0,0 +1,103 @@
...
+ * A factory that is used by default to support LogicalTableGroup.
+ *
+ * A LogicalTableGroup is essentially a map from TableIdentifier to LogicalTable.
+ * This concept exists for the reason that when we configure LogicalTable,
+ * each granularity will be treated as a separated LogicalTable internally.
+ * Thus one configuration correspond to a LogicalTableGroup, and the infrastructure
+ * will support look up for each individual LogicalTable with a TableIdentifier.
+ */
+public class AbridgedLogicalTableGroupFactory implements Factory<LogicalTableGroup> {
+    private static final String ENTITY_TYPE = "abridged logical table group";
```
michael-mclawhorn: Change to DefaultLTGF

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
```diff
@@ -0,0 +1,103 @@
...
+                dimensions
+        );
+
+        configTable.get("granularities").forEach(
+                (JsonNode node) -> {
+                    ZonedTimeGrain granularity = DefaultTimeGrain.valueOf(
+                            node.textValue().toUpperCase(Locale.US)
+                    ).buildZonedTimeGrain(DateTimeZone.forID(configTable.get("dateTimeZone").textValue()));
+                    LogicalTable logicalTable = new LogicalTable(tableName, granularity, tableGroup, metricDictionary);
+                    tableDictionary.put(new TableIdentifier(logicalTable), logicalTable);
```
michael-mclawhorn: Use the "    public LogicalTable(
            @NotNull String name,
            String category,
            String longName,
            @NotNull Granularity granularity,
            ReadablePeriod retention,
            String description,
            TableGroup tableGroup,
            MetricDictionary metricDictionary"
constructor

michael-mclawhorn: Implement retention using joda.time.Period.parse

michael-mclawhorn: Don't bother with the LogicalTableName constructor, it's just busywork here.

luthier/config/src/tables.lua
```diff
@@ -168,15 +192,26 @@ M.physical = {
...
+        },
+        physicalTables = {"wikiticker"},
+        dateTimeZone = "UTC"
     },
     air_quality = {
+        type = "abridged",
         metrics = {"averageCOPerDay", "averageNO2PerDay"},
         dimensions = air_quality_dimensions,
-        granularity = {"all", "hour", "day"},
+        granularities = {
```
michael-mclawhorn: default retention to "P1Y"

michael-mclawhorn: Build using GranularityDictionary, supplied by getGranularityDictionary from AbstractBinderFactory

luthier/config/src/utils/tableUtils.lua
gab-umich: refactored

luthier/config/src/tables.lua
gab-umich: This is now done. 
Also adapted the granularity dictionary idea.
Especially, when we build the LIP without going through the `LuthierBinderFactory > getConfigurationLoader()`, it will not load in `logicalTableGroups` due to the fact that we have no granularity dictionary. 
Instead of throwing a null-pointer error, I will instead check if the granularity dictionary exists. 
This leads to three possible cases: 

- granularityDictionary does not exist: will not load in any `logicalTable`; log a warning in using logger; proceed without exception. (If I throw exception, it will break all previous unit tests.)
- granularityDictionary exists but it does not contain the needed granularity key (String): will throw LuthierFactoryException: this means that the user clearly wants to configure the `LogicalTable`, but failed to do so by not supporting their custom time grain. From the end result, it looks harsher than the previous one. 
- granularityDictionary exists, contains needed String (lower case, including 'all'), we load in `logicalTable` as needed, proceed without exception.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
```diff
@@ -25,15 +25,21 @@
 
 /**
  * A factory that is used by default to support LogicalTableGroup.
+ *
+ * A LogicalTableGroup is essentially a map from TableIdentifier to LogicalTable.
+ * This concept exists for the reason that when we configure LogicalTable,
+ * each granularity will be treated as a separated LogicalTable internally.
+ * Thus one configuration correspond to a LogicalTableGroup, and the infrastructure
+ * will support look up for each individual LogicalTable with a TableIdentifier.
```
michael-mclawhorn: Slight rewording of this:
"One configuration string (name) defines multiple LogicalTableIdentifiers (name, granularity pair).  LogicalTableGroup allows us to collect all the distinct LogicalTables for a single name, mapping LogicalTableIdentifier to LogicalTable.

At the external configuration layer, configuration users will define a logical table group (keyed by name) with a list of granularities.  The factory will produce the collection of LogicalTables keyed by identifier (name, grain pairs), to be merged into the table dictionary."

luthier/config/src/tables.lua
```diff
@@ -136,19 +136,43 @@ M.physical = {
...
+                    * physicalTables
+                    * dateTimeZone
+                the following fields will be ignored:
+                    * category
+                    * longName
+                    * retention
+                    * description
+                    * apiFilters
+            * comprehensive - a list of rules to fully configure the logicalTable. This is not implemented yet.
+        Defaults to 'abridged'.
```
michael-mclawhorn: do we allow default factories?  I didn't think we did.

luthier/config/src/tables.lua
michael-mclawhorn: I know we have a conversation about building the default granularity dictionary in the builder.  If so we shouldn't have to handle a null granularity dictionary unless someone does it on purpose.

michael-mclawhorn: And it doesn't look done, because all is still commented out.

gab-umich: Ah I forgot to push my commits. It will go live in a few minutes.

gab-umich: And for the issue you raise, I believe we do have to handle that just to let the previously passing unit tests (which are all built directly using the `LuthierIndustrialPark.Builder` thus has a `null` as their granularity), keep passing. 
If we do not special case that, then we have to

- somehow populate the granularity, which requires _modifying the previous unit tests_ (maybe using a withGranularityDictionary(), which I also happen to have coded) OR
- make the `LuthierIndustrialPark.load()` not fail when missing the entire granularityDictionary, which is effectively what I am trying to accomplish, but just on the factory level, instead of on the LIP level.

gab-umich: I'd like to hear your suggestions, we can also examine the new commits together.

luthier/config/src/tables.lua
gab-umich: I see! 
I will be using the `* The table is keyed on the name of the logical table` instead.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: This is working because it is not used to load the logicalTable. When the system populates the logicalTables, it is using the 
`private void putLogicalTableByGroup(String tableKey) `
a couple of lines below instead. While this getter is public, and is exposed to the users, the private function is used to internally read from json and populate the config source.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
gab-umich: I see your point. I will make the name change in my new commit.

luthier/config/src/tables.lua
gab-umich: We did mention that once when we had the talk about defaulting dimension factory to KeyValueStoreDimensionFactory. 
Not sure if this really makes sense here as well (although if we do decide to make a defaulting happen, I will remove the [mandatory] tag in this doc.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: That should work in production but not in all of the unit tests. 
The reason is, in unit tests, we directly call LIP.builder.build(), which checks the granularityDictionary, sees that's `null`, and then skips the load of logicalTables, thus the recursive process will not happen. 

On the other hand, if we go through the LuthierBinderFactory to initialize LIP, we are good. 
If the test directly get a specific concept, then the recursive process start from that point, which also works as expected.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
gab-umich: addressed

luthier/config/src/tables.lua
gab-umich: done.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
gab-umich: sure

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
gab-umich: done.

luthier/config/src/tables.lua
archolewa: I think we should default the type in cases where it makes sense. We expect the vast majority of the time for people to use only one LogicalTable type, the one provided by Fili.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AbridgedLogicalTableGroupFactory.java
gab-umich: changed to the doc you suggested.

luthier/config/src/tables.lua
gab-umich: I agree. This will be more addressed and documented when we get to the luaDefaulting PR, which I hope to start after this PR.

luthier/config/src/tables.lua
```diff
@@ -136,19 +136,50 @@ M.physical = {
...
+                    * granularities
+                    * physicalTables
+                    * dateTimeZone
+                    * category
+                    * longName
+                    * retention
+                    * description
+                the following fields will be ignored:
+                    * apiFilters
+            * comprehensive - a list of rules to fully configure the logicalTable. This is not implemented yet.
```
archolewa: If this doesn't exist, I wouldn't mention it at all.

luthier/config/src/tables.lua
```diff
@@ -136,19 +136,50 @@ M.physical = {
...
+        Defaults to 'default'.
+    * metrics - A set of API metrics' name for this logical table, these
+        should be a subset of the metrics configured in metrics.lua;
+        Defaults to an empty lua table.
     * dimensions - A set of dimensions for this logical table, the set of
         dimensions should be a subset of dimensions in its dependent physical
         tables.
-    * granularity - A group of available granularities of this logical table, the
+        Defaults to an empty lua table.
+    * dateTimeZone - A case sensitive name according to joda's dateTimeZone to indicate the
```
archolewa: Typo, remove `the` at the end of the line.

luthier/config/src/tables.lua
```diff
@@ -136,19 +136,50 @@ M.physical = {
...
+        which zone this physical table belongs in. See further:
+        https://www.joda.org/joda-time/timezones.html
+        Defaults to "UTC"
+    * retention - a String in the ISO 8601 Duration format, i.e. PnYnMnDTnHnMnS
```
archolewa: In addition to the format of the data, we should also speckiy what the data means. For example `metadata
meant to tell external customers the retention policy for the underlying druid table(s). Not used
internally by Fili.`

The same applies to `dateTimeZone`.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -213,24 +271,17 @@ public ResourceDictionaries getDictionaries() {
...
+        ) {
             this.resourceDictionaries = resourceDictionaries;
             dimensionFactories = getDefaultDimensionFactories();
             keyValueStoreFactories = getDefaultKeyValueStoreFactories();
             searchProviderFactories = getDefaultSearchProviderFactories();
             physicalTableFactories = getDefaultPhysicalTableFactories();
+            logicalTableGroupFactories = getDefaultLogicalTableGroupFactories();
         }
 
-        /**
```
archolewa: Whey are we removing this constructor?

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultLogicalTableGroupFactory.java
```diff
@@ -0,0 +1,146 @@
...
+     *
+     * @return  A newly constructed config instance for the name and config provided
+     */
+    @Override
+    public LogicalTableGroup build(String name, ObjectNode configTable, LuthierIndustrialPark resourceFactories) {
+        LogicalTableGroup logicalTableGroup = new LogicalTableGroup();
+        GranularityDictionary granularityDictionary = resourceFactories.getGranularityDictionary();
+        if (granularityDictionary == null) {
+            LOG.warn(String.format(GRANULARITY_DICTIONARY_MISSING, name));
+            return logicalTableGroup;
```
archolewa: I would just explode here, with the error message what you're
currently logging as a warning. An empty table group is pretty useless, and
will lead to weird errors people won't notice until runtime, and that they'll
have to triage.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultLogicalTableGroupFactory.java
```diff
@@ -0,0 +1,146 @@
...
+        );
+        configTable.get("metrics").forEach(
+                node -> {
+                    /* go to the LIP to retrieve the metric */
+                    /* Must be done after having actual metric builders! */
+                    // apiMetricNames.add(new LuthierApiMetricName(node.textValue(), granularity));
+                    // metricDictionary.add();
+                }
+        );
+        TableGroup tableGroup = new TableGroup(
```
archolewa: This doesn't need to be wrapped.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultLogicalTableGroupFactory.java
```diff
@@ -0,0 +1,146 @@
...
+                physicalTables,
+                apiMetricNames,
+                dimensions
+        );
+
+        configTable.get("granularities").forEach(
+                 node -> {
+                    GranularityParser granularityParser = new StandardGranularityParser(
+                    );
```
archolewa: This doesn't need to be wrapped. Also, we should just lift this up to a field and build it when the factory is constructed.
No sense in building a new one over and over again.

luthier/config/src/tables.lua
gab-umich: OK. It is just a little bit strange to have only one item that we can default to in the next line.
But I will change it. I don't see the `comprehensive` which supposed to support apiFilters to be coded any time soon.

luthier/config/src/tables.lua
gab-umich: added them in.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: This builder is not used anywhere in the internal unit testing. And it is not more superior than the one above it so we don't use it in the LuthierBinderFactory. 
It should not be used by the user since the contract is that user only initialize a LIP through (some subclass of) luthier binder factory. 
There is no point to keep it in my opinion, unless we think a bean-style initialization is going to be interesting. 
I'd like to hear about it in that case @archolewa

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultLogicalTableGroupFactory.java
gab-umich: done

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultLogicalTableGroupFactory.java
gab-umich: This is not exploding since it makes previously passing unit tests unusable. (As they are initialized without a granularity dictionary, nothing will ever pass)

You concern is indeed interesting, though. I can go about and refactor all previous and future test to only use a LuthierBinderFactory to initialize the LIP, if you insist on stalling the entire process when seeing a null value.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultLogicalTableGroupFactory.java
gab-umich: done

luthier/config/src/tables.lua
archolewa: This is a case of future proofing. Keep in mind that we're also providing support for people to add their own custom types. Furthermore, we may add additional logical table types in the future. In those cases, knowing what the defaulting behavior is going to be very important.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultLogicalTableGroupFactory.java
archolewa: Creating a LuthierBinderFactory for all the tests sounds like massive overkill.

Why not just initialize their LuthierIndustrialParks with a GranularityDictionary? 

If you haven't already, you should create a utility for your tests `LuthierIndustrialPark initializeLuthierIndustrialParkBuilder` which builds a `LuthierIndustrialParkBuilder` with whatever configuration you're always going to need. 

Then, anytime we need to add additional state to the park, you can update that method rather than having to modify all your tests.

Furthermore, by returning the builder, it's easy enough for tests to attach their own custom configuration as well.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
archolewa: This is fine then.

luthier/src/main/java/com/yahoo/bard/webservice/data/config/LuthierApiMetricName.java
```diff
@@ -0,0 +1,45 @@
...
+    @Override
+    public boolean isValidFor(TimeGrain grain) {
+        return grain.satisfiedBy(timeGrain);
+    }
+
+    @Override
+    public String getApiName() {
+        return toString();
+    }
+}
```
michael-mclawhorn: We should drop this in favor of FiliApiMetricName until we need a different version.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -213,24 +266,35 @@ public ResourceDictionaries getDictionaries() {
...
          */
-        public Builder(LuthierResourceDictionaries resourceDictionaries) {
-            this.resourceDictionaries = resourceDictionaries;
-            dimensionFactories = getDefaultDimensionFactories();
-            keyValueStoreFactories = getDefaultKeyValueStoreFactories();
-            searchProviderFactories = getDefaultSearchProviderFactories();
-            physicalTableFactories = getDefaultPhysicalTableFactories();
+        public Builder(
+                LuthierResourceDictionaries resourceDictionaries
+        ) {
```
archolewa: This doesn't need to be wrapped.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -213,24 +266,35 @@ public ResourceDictionaries getDictionaries() {
...
+         * @param resourceDictionaries  a class that contains resource dictionaries including
+         * PhysicalTableDictionary, DimensionDictionary, etc.
+         * @param granularityDictionary  a named map of all supported granularities
          */
-        public Builder() {
-            this(new LuthierResourceDictionaries());
+        public Builder(
+                LuthierResourceDictionaries resourceDictionaries,
+                GranularityDictionary granularityDictionary
+        ) {
```
archolewa: This doesn't need to be wrapped (our maximum line length is 120 characters, and this is 119.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -399,6 +480,24 @@ public Builder withPhysicalTableFactories(Map<String, Factory<ConfigPhysicalTabl
             return this;
         }
 
+        public void setGranularityDictionary(GranularityDictionary granularityDictionary) {
```
archolewa: We should add a function for adding a single custom granularity to the already existing dictionary.
The vast majority of the time, if people are adding custom granularities, they want to add it to the
defaults, rather than replace the defaults completely.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
```diff
@@ -3,30 +3,37 @@
...
 import org.joda.time.DateTimeZone;
 
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
-import java.util.Locale;
 
 /**
  * A factory that is used by default to support Simple (non-Composite) Physical Table.
  */
 abstract class SingleDataSourcePhysicalTableFactory implements Factory<ConfigPhysicalTable> {
```
archolewa: We should make this public, rather than package-private. Customers may want to implement it.

gab-umich: addressed

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: Added this functionality.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: Close one! I was most likely intimidated by the IntelliJ annotation

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: sure.

luthier/src/main/java/com/yahoo/bard/webservice/data/config/LuthierApiMetricName.java
gab-umich: done.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: Got it.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
michael-mclawhorn: I'm not sure I understand the reasoning.  First, granularity dictionary should never default to null, so that failout path shouldn't happen.

Out unit test cases should be 'test build...' and 'test getFoo(FooName)' , right?

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -292,7 +356,7 @@ public Builder() {
         }
 
         /**
-         * Default physicalTable factories that currently lives in the code base.
+         * Default PhysicalTable factories that currently lives in the code base.
```
michael-mclawhorn: 'that are defined in fili-core'

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
michael-mclawhorn: Are we thinking get and modify or get and set?

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: Also changed other occurrences.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: What I added is a new function `addGranularity(String name, Granularity granularity)` that get and set a new (String, Granularity) pair. 
If key collision happens it will silently update the value (Granularity).

gab-umich: Is it unidiomatic to have both the setter and "adder" exposed to the user?

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
archolewa: Basically. In fact, you should be able to get away with just `test getFoo(FooName)` because if `FooName` doesn't exist, the system should automatically build it and all its dependencies.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
archolewa: The methods serve two different purposes. The setter overwrites the current dictionary, the adder adds to the current dictionary. So there's nothing wrong with
having both.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
```diff
@@ -3,30 +3,37 @@
...
 
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
-import java.util.Locale;
 
 /**
  * A factory that is used by default to support Simple (non-Composite) Physical Table.
  */
-abstract class SingleDataSourcePhysicalTableFactory implements Factory<ConfigPhysicalTable> {
+abstract public class SingleDataSourcePhysicalTableFactory implements Factory<ConfigPhysicalTable> {
```
archolewa: We typically put `public` before `abstract`: `public abstract class SingleDataSource...`

gab-umich: done

PR #940, opened 2019-07-17
Lua defaulting

gab-umich:
This is embodying the intermediate layer to extract away all defaulting into a separate Lua module. 

Also has a fun `verify_then_config.lua`: it is comparing the table we are about to write, against a historic snapshot stored in `resources/ref/*.json`. It will let you know what is changed, if anything. It has been handy for my task so far, because essentially in this PR it should always be `nothing is changed, proceed to overwrite`.


luthier/config/src/utils/defaultingUtils.lua
```diff
@@ -0,0 +1,71 @@
...
+M.DEFAULT_LOGICAL_CATEGORY = "GENERAL"
+M.DEFAULT_LOGICAL_RETENTION = "P1Y"
+M.DEFAULT_LOGICAL_DEPENDENT_TABLE = {}
+M.DEFAULT_LOGICAL_DATE_TIME_ZONE = "UTC"
+--- Defaulting methods:
+-- Dimensions
+function M.dimension_defaulting(dimension_name, dimension)
+    local dim_config = misc.shallow_copy(dimension)
+    dim_config.longName = dim_config.longName or dimension_name
+    dim_config.description = dim_config.description or dimension_name
```
michael-mclawhorn: this should be description or longName

PR #941, opened 2019-07-19
Upgrades jackson-databind to 2.9.9.1.

archolewa:
-- This addresses security vulnerability:
https://github.com/yahoo/fili/network/alert/pom.xml/com.fasterxml.jackson.core:jackson-databind/open.

-- Unfortunately, because this version was only released by for
the data-bind component, we had to introduce another version variable
for just jackson-databind, so there is a risk of the jackson versions
getting out of sync. The next time we upgrade Jackson, we should
remove this additional version variable, to bring all the components
back in sync.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -107,6 +107,7 @@
         <version.spring>[5.1.2,)</version.spring>
         <version.groovy>2.4.15</version.groovy>
         <version.jackson>2.9.9</version.jackson>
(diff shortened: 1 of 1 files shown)


wcekan:
See [CVE-2019-12814](https://nvd.nist.gov/vuln/detail/CVE-2019-12814)

PR #943, opened 2019-07-25
Add custom sslcontext support in fili

Chandrasekar-Rajasekar:
User must override getSSLContext method to give their own sslContext. The function should return the object of io.netty.handler.ssl.SslContext. 
`protected SslContext getSSLContext() {
        //create SSL Context here
       if(sslContext.instanceOf(io.netty.handler.ssl.SslContext.class)) 
              return sslContext;
      if(sslContext.instanceOf(javax.net.ssl.SSLContext.class))
              return new JdkSslContext(sslContext, true, ClientAuth.REQUIRE)
     .
     .
     .
     return null
  }`

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
@@ -135,6 +135,7 @@ import com.codahale.metrics.health.HealthCheck;
 import com.codahale.metrics.health.HealthCheckRegistry;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
+import io.netty.handler.ssl.SslContext;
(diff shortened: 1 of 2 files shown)


michael-mclawhorn:
definitely need a change to CHANGELOG, and you should probably create an issue to link to in the changelog ticket.

Should also add a test to config that this code gets excercised.

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -1208,6 +1209,14 @@ protected GranularityDictionary getGranularityDictionary() {
         return new HashMap<>(0);
     }
 
+    /**
+     * Get a custom configured ssl context.
+     * @return SSL context
```
michael-mclawhorn: If null, the system default SSL Context will be applied

PR #946, opened 2019-07-29
Finalizing Luthier Metric Maker and Metric

gab-umich:
long sum maker in place and build correctly. Will proceed to connect this to an actual logicalMetric.


luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -409,8 +413,35 @@ public Builder(LuthierResourceDictionaries resourceDictionaries, GranularityDict
             return logicalTableFactoryMap;
         }
 
-        public GranularityDictionary getGranularityDictionary() {
```
archolewa: Why are we getting rid of this?

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -409,8 +413,35 @@ public Builder(LuthierResourceDictionaries resourceDictionaries, GranularityDict
...
-            return granularityDictionary;
+        /**
+         * Default LogicalTable factories that are defined in fili-core.
```
archolewa: JavaDoc is wrong.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -409,8 +413,35 @@ public Builder(LuthierResourceDictionaries resourceDictionaries, GranularityDict
...
+         *
+         * @return  a LinkedHashMap of MetricMaker name to its factory
```
archolewa: The fact that it's a `LinkedHashMap` doesn't really matter (if it does, we should make that its return type). Better to just say "Map."

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/ArithmeticMakerFactory.java
```diff
@@ -21,14 +21,14 @@
  */
 public class ArithmeticMakerFactory implements Factory<MetricMaker> {
 
-    public static final String OPERATION = "operation";
+    protected static final String FUNCTION = "function";
```
archolewa: Why the change here from `operation` to `function`? I picked `operation` originally because that's what addition, subtraction, multiplication and division are called. People don't usually call them functions, even if that's what they technically are.

michael-mclawhorn: sadly, it's called function in the ArithmeticMaker.  I gave my blessing to aligning with that, even though I wish operation was what we had chosen.

archolewa: People writing config don't care what it's called in the implementation of ArithmeticMaker. With Luthier, people are probably almost never going to look at the ArithmeticMaker.

luthier/config/src/utils/defaultingUtils.lua
```diff
@@ -68,4 +71,17 @@ function M.logical_table_defaulting(logical_table_name, logical_table)
...
+    local metric_config = misc.shallow_copy(metric)
+    metric_config.longName = metric_config.longName or metric_name
+    metric_config.description = metric_config.description or metric_config.longName   -- this line must follow the previous one
+    metric_config.category = metric_config.category or M.DEFAULT_METRIC_CATEGORY
+    metric_config.dataType = metric_config.dataType or M.DEFAULT_METRIC_DATA_TYPE
+    --- the following unification is no longer necessary in my (Gabriel's) opinion.
+    --- the discrepancy can be handled by two types of Factories: Aggregation and PostAggregation
+    --metric_config.dependencyMetricNames = metric_config.druidMetric and {metric_config.druidMetric}
+    --        or metric_config.dependencies       -- if druidMetric exists, use {druidMetric}, otherwise dependencies
+    return metric_config
```
michael-mclawhorn: Actually, what really needs to happen is that the dataType should probably become optional.
The logical metric should define the default data type via the makers and the apimetricname should override that datatype if present.

My advice is to make the Factory treat this as an optional parameter and remove the defaulting altogether.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
michael-mclawhorn: yes, why?

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -409,8 +421,48 @@ public Builder(LuthierResourceDictionaries resourceDictionaries, GranularityDict
...
+            );
+            metricMakerFactoryMap.put(
+                    "com.yahoo.bard.webservice.data.config.metric.makers.LongSumMaker",
+                    longSumMakerFactory
+            );
+            metricMakerFactoryMap.put(
+                    "com.yahoo.bard.webservice.data.config.metric.makers.AggregationAverageMaker",
+                    aggregationAvgMakerFactory
+            );
+            return metricMakerFactoryMap;
```
michael-mclawhorn: as a general pattern, I would like to add className and short class name to these maker maps.  Perhaps we should make a utility method to do this for each of these that can look over the factories.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/ArithmeticMakerFactory.java
michael-mclawhorn: ok.  you've changed my mind.  call it operation until the last minute.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
```diff
@@ -86,7 +86,7 @@ LuthierPhysicalTableParams buildParams(
         metricsNode.forEach(
                 node -> {
                     String metricName = node.textValue();
-                    resourceFactories.getMetric(metricName);
+                    // resourceFactories.getMetric(metricName);
```
michael-mclawhorn: I believe we need this statement or the construction on the Table will fail.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: I was under the impression that the only communication regarding Granularity between the LIP and various Factories is one-directional. i.e. LIP gives information to Factories and Factories never have the privilege to modify the LIP's GranularityDictionary. 
This change highlights the fact that this means that the GranularityParser can hide away the implementation of GranularityDictionary inside LIP. So far, none of the Factories are calling GranularityDictionary, as that information is exposed by the GranularityParser itself. 
If user need to modify the GranularityParser, they should do it by subclassing LuthierBinderFactory thus have full control over what the GranularityDictionary is. 

Please feel free to point out the flaw in this thinking.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: corrected.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: Got it.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/ArithmeticMakerFactory.java
gab-umich: Although that is true, I believe aligning with "function" makes it simpler to track down issues from our side. Although I also find it some what awkward, people coming from Druid would most likely expect this idiom, as you can see here:
https://druid.apache.org/docs/latest/querying/post-aggregations.html#javascript-post-aggregator
The name "function" is a middle ground that I believe is kind of reasonable in this case.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
gab-umich: This is commented out because I don't have the corresponding Factories for all the metrics referenced in the physicalTable _yet_.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: This is done manually for now. I do see the value of having a helper function though. It can add clarity for sure but will need some string manipulation

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
michael-mclawhorn: during the builder building erring on the side of mutability is your friend.

gab-umich: My bad, this is not in the LIP, this is the LIP.Builder. I have put it back.

luthier/config/src/utils/defaultingUtils.lua
```diff
@@ -68,4 +71,17 @@ function M.logical_table_defaulting(logical_table_name, logical_table)
...
 end
+
+function M.metric_defaulting(metric_name, metric)
+    local metric_config = misc.shallow_copy(metric)
+    metric_config.longName = metric_config.longName or metric_name
+    metric_config.description = metric_config.description or metric_config.longName   -- this line must follow the previous one
+    metric_config.category = metric_config.category or M.DEFAULT_METRIC_CATEGORY
+    metric_config.dataType = metric_config.dataType or M.DEFAULT_METRIC_DATA_TYPE
+    --- the following unification is no longer necessary in my (Gabriel's) opinion.
+    --- the discrepancy can be handled by two types of Factories: Aggregation and PostAggregation
```
archolewa: Why is this no longer necessary? Using "dependencies" to refer to a thing in Druid is weird and confusing. We set people up to view `dependencies` as "other Fili metrics this metric depends on." So we went with `druidMetric` as an alias for it that people can use for simple metrics.

Besides, is it necessarily the case than Aggregation will never depend on other Fili metrics? I'm not convinced it is.

luthier/config/src/utils/defaultingUtils.lua
gab-umich: addressed

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/ArithmeticMakerFactory.java
archolewa: I don't see how this makes it easier for us to track down issues.

luthier/config/src/utils/defaultingUtils.lua
gab-umich: It is not necessary to unify `druidMetric` and `dependencies` into a so-called `dependencyMetricNames`. 
in the current json output, there will be `druidMetric` for AggregationMetricFactory to read, and a `dependencies` for PostAggregationMetricFactory to read, respectively.

gab-umich: For your reference, this is not changing the interface we expose to users. the `metrics.lua` is still the same.

luthier/src/test/resources/snapshots/.snapshotsHistory.json
```diff
@@ -1 +1 @@
-["2019_07_Jul_22_Mon_11:43:02/","2019_07_Jul_24_Wed_17:19:43/","2019_07_Jul_25_Thu_11:43:48/"]
\ No newline at end of file
+["2019_07_Jul_30_Tue_11:41:36/","2019_07_Jul_30_Tue_11:52:20/","2019_07_Jul_30_Tue_16:31:45/","2019_07_Jul_30_Tue_16:41:22/","2019_07_Jul_30_Tue_16:42:21/","2019_07_Jul_30_Tue_20:24:10/","2019_07_Jul_31_Wed_14:32:32/","2019_07_Jul_31_Wed_15:38:54/"]
```
archolewa: None of this should be in source control. We should add `src/test/resources/snapshots` to the .gitignore file.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/ArithmeticMakerFactory.java
gab-umich: I will be using the "operation" in place of the current "function"

luthier/src/test/resources/snapshots/.snapshotsHistory.json
gab-umich: Oh sure. 
This does clog the commit reviewable files

luthier/config/src/metrics.lua
```diff
@@ -60,8 +60,13 @@ So the metric "difference" is the (rather silly) formula:
             on directly, if any, only applies for nested metrics
 --]]
 -------------------------------------------------------------------------------
+local M = {}
 
-return {
+local aggregation_metrics = {
+    longSumCO = {
+        maker = "longSum",
```
archolewa: I'm thinking we would be better off calling this `type` rather than `maker`. It makes the
configuration more symmetrical with the rest of our config. Plus, the fact that we're using
"makers" in Fili is an implementation detail, and it isn't going to be immediately obvious to
people what a "maker" is, or why that's how we specify the type of metric this is.

luthier/config/src/utils/defaultingUtils.lua
archolewa: I don't think this is a good use of the word `type.` I think when people ask "what type of metric something is" they mean "longSum" or "complicated formula." They don't
think "metric in Fili that relies on other metrics" vs "metric in Fili that's just a simple aggregation of a Druid metric." So I think it would be better to rename `maker` to `type`,
and drop this field completely.  Requiring people to add another field whose sole purpose is to signal whether this is a simple or complex metric is ceremony. It's already unambiguous
whether a metric is simple or not: If it has the `druidMetric` field, then it's simple. If it has the `dependencies` field, then it's complex. The unification is purely an implementation
detail so that Fili doesn't have to care whether a metric is simple or not. If we want to push that decision into the Java code, then that's fine. But the decision should turn on
the existence of the `druidMetric` field.

luthier/config/src/metrics.lua
```diff
@@ -84,7 +89,10 @@ return {
     NO2M = {
         maker = "doubleSum",
         druidMetric = "NO2"
-    },
+    }
+}
+
+local post_aggregation_metrics = {
```
archolewa: This naming (`aggregation` vs `post-aggregation`) is problematic. It seems to imply that
these metrics are different because of what shows up in their Druid formula, but that's not the
case. The thing that makes these metrics different is whether or not they rely on other Fili
metrics. There is nothing stopping a customer (I don't think) from building a super complex
formula with a whole mess of aggregations and post-aggregations that relies only on Druid metrics.
That would still be an `aggregation` metric in this parlance, even though it has post aggregations
as well. In the original design, and in the comments here I believe, we went with `simple` vs.
`complex` I believe. Maybe that's not a good name either, but at least it doesn't directly
invoke the potential makeup of the Druid query.

Though I don't think having an explicit distinction with its own  field is necessary anyway. Fili doesn't
make the destinction, and the `druidMetrics` field was originally intended simply as an alias for
`dependencies` to keep things clear. I think we need to make a distinction between the two at a
system level, only at a human level.

luthier/config/src/metrics.lua
archolewa: Disregard this. I forgot that people configure makers in Luthier as well.

luthier/config/src/metrics.lua
gab-umich: This asymmetry is addressed by merging the two groups. So is the `druidMetric` vs `dependencies` are unified into a `dependencyMetricNames` list as we do before.

luthier/config/src/metrics.lua
```diff
@@ -86,11 +91,18 @@ return {
...
         dependencies = {"COM"}
     },
     averageNO2PerDay = {
-        maker = "aggregateAveragebyDay",
+        maker = "aggregateAverageDAY",
         dependencies = {"NO2M"}
     }
 }
+
+for metric_name, metric_content in pairs(metrics) do
```
archolewa: This belongs in the defaulting file.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -212,8 +205,28 @@ public GranularityParser getGranularityParser() {
...
+            LogicalMetric metric = metricFactoryPark.buildEntity(metricName, this);
+            metricDictionary.put(metricName, metric);
+        }
+        return metricDictionary.get(metricName);
+    }
+
+    /**
+     * Retrieve or build a Metric Maker.
+     *
+     * @param metricMakerName the name for the dimension to be provided.
```
archolewa: `dimension` should be `metricMakerName`.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
```diff
@@ -409,8 +423,73 @@ public Builder(LuthierResourceDictionaries resourceDictionaries, GranularityDict
...
+            metricMakerFactoryMap.put("count", countMakerFactory);
+            /* class names */
+            metricMakerFactoryMap.put("ArithmeticMaker", arithmeticMakerFactory);
+            metricMakerFactoryMap.put("LongSumMaker", longSumMakerFactory);
+            metricMakerFactoryMap.put("DoubleSumMaker", doubleSumMakerFactory);
+            metricMakerFactoryMap.put("AggregationAverageMaker", aggregationAvgMakerFactory);
+            metricMakerFactoryMap.put("CountMakerFactory", countMakerFactory);
+            /* fully qualified class names */
+            metricMakerFactoryMap.put(
+                    "com.yahoo.bard.webservice.data.config.metric.makers.ArithmeticMaker",
```
archolewa: Rather than hardcoding all of these, we should use things like `ArithmeticMaker.class.getName()`. Less prone to typos.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AggregationAverageMakerFactory.java
```diff
@@ -0,0 +1,37 @@
...
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.yahoo.bard.webservice.config.luthier.ConceptType;
+import com.yahoo.bard.webservice.config.luthier.Factory;
+import com.yahoo.bard.webservice.config.luthier.LuthierFactoryException;
+import com.yahoo.bard.webservice.config.luthier.LuthierIndustrialPark;
+import com.yahoo.bard.webservice.config.luthier.LuthierValidationUtils;
+import com.yahoo.bard.webservice.data.config.metric.makers.AggregationAverageMaker;
+import com.yahoo.bard.webservice.data.config.metric.makers.MetricMaker;
+import com.yahoo.bard.webservice.data.time.ZonelessTimeGrain;
+import com.yahoo.bard.webservice.util.GranularityParseException;
```
archolewa: Import formatting is incorrect. Everything `com.yahoo.bard` should come before `com.fasterxml`, and
`com.fasterxml` should be in a separate paragraph.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AggregationAverageMakerFactory.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Factory that supports LongSumMetricMaker.
```
archolewa: Documentation is wrong. This is an AggregationAverageMakerFactory, not LongSumMakerFactory.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AggregationAverageMakerFactory.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Override
+    public MetricMaker build(String name, ObjectNode configTable, LuthierIndustrialPark resourceFactories) {
+        LuthierValidationUtils.validateField(configTable.get(INNER_GRAIN), ConceptType.METRIC_MAKER, name, INNER_GRAIN);
+        String grainName = configTable.get(INNER_GRAIN).textValue();
+        try {
+            ZonelessTimeGrain grain = (ZonelessTimeGrain) resourceFactories.getGranularityParser()
+                    .parseGranularity(grainName);
+            return new AggregationAverageMaker(resourceFactories.getMetricDictionary(), grain);
+        } catch (GranularityParseException | ClassCastException e) {
```
archolewa: I think we should split these out into separate error messages. In the first case, we got something that
isn't a grain at all, in the second we got the wrong kind of grain. The error message we're using is very
misleading in the second case. Instead, we should say something like

```Averages cannot be computed over granularity '%s'. Averages may only be computed over instances of com.yahoo.bard.webservice.data.time.ZonelessTimeGrain.```

File: luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/CountMakerFactory.java [hunk 18]
```diff
@@ -0,0 +1,19 @@
...
+package com.yahoo.bard.webservice.config.luthier.factories;
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.yahoo.bard.webservice.config.luthier.Factory;
+import com.yahoo.bard.webservice.config.luthier.LuthierIndustrialPark;
+import com.yahoo.bard.webservice.data.config.metric.makers.CountMaker;
+import com.yahoo.bard.webservice.data.config.metric.makers.MetricMaker;
+
+/**
+ * Factory that supports LongSumMetricMaker.
```
archolewa: Should refer to the CountMaker, not LongSum.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultLogicalTableGroupFactory.java
```diff
@@ -2,16 +2,19 @@
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
 package com.yahoo.bard.webservice.config.luthier.factories;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
```
archolewa: These imports are in the wrong location. They should be after `com.yahoo` in a new paragraph.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultLogicalTableGroupFactory.java
```diff
@@ -56,78 +70,94 @@ public LogicalTableGroup build(String name, ObjectNode configTable, LuthierIndus
         if (resourceFactories.getGranularityParser() == null) {
             throw new LuthierFactoryException(String.format(GRANULARITY_DICTIONARY_MISSING, name));
         }
-        LogicalTableGroup logicalTableGroup = new LogicalTableGroup();
```
archolewa: This is a pretty big method. I wonder if we might want to break it up into a sequence of helper methods,
`buildDimensions`, `buildMetricNames`, `buildPhysicalTables`, that sort of thing. Basically, anything that
requires any kind of looping gets pulled out into a helper method.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultMetricFactory.java
```diff
@@ -0,0 +1,78 @@
+// Copyright 2019 Oath Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.config.luthier.factories;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
```
archolewa: Imports are in the wrong place.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultMetricFactory.java
```diff
@@ -0,0 +1,78 @@
...
+    private static final String DEPENDENCY_METRIC_NAMES = "dependencyMetricNames";
+
+    @Override
+    public LogicalMetric build(String name, ObjectNode configTable, LuthierIndustrialPark resourceFactories) {
+        validateFields(
+                name,
+                configTable,
+                Arrays.asList(MAKER, LONG_NAME, CATEGORY, DESCRIPTION, DEPENDENCY_METRIC_NAMES)
+        );
+        MetricMaker maker = resourceFactories.getMetricMaker(configTable.get(MAKER).textValue());
```
archolewa: One thing that we're missing in the system is some null checking around dependencies. For example,
what happens if someone types `longSumm` (note the extra `m`) instead of `longSum`? We should have
an error message "Unknown maker `longSumm`, avaiable makers are `list-of-maker-names`. But I'm not sure
we do. I think we might just null pointer somewhere in our system. We should think about tightening
that up after this PR. One option is that the Factory is expected to throw an exception if it can't find
the configuration for the given name, and the LuthierIndustrialPark catches the exception and prints an
error. That way, we have this family of errors handled in the same place, instead of scattered about
all the factories.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultMetricFactory.java
```diff
@@ -0,0 +1,78 @@
...
+                        configTable.get(DESCRIPTION).textValue(),
+                        configTable.get(DATA_TYPE).textValue()
+                );
+        List<String> dependencyMetricNames = StreamSupport.stream(
+                configTable.get(DEPENDENCY_METRIC_NAMES).spliterator(),
+                false
+        )
+                .map(JsonNode::textValue)
+                .collect(Collectors.toList());
+        return maker.make(metricInfo, dependencyMetricNames);
```
archolewa: We should catch the IllegalArgumentException that gets thrown by `make` if the
`dependencyMetricNames` can't be found, and throw a richer error message with
at least the name of the metric we're currently trying to build.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultMetricFactory.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+    /**
+     * Helper function to validate only the fields needed in the parameter build.
+     *
+     * @param name  the config dictionary name (normally the apiName)
+     * @param configTable  ObjectNode that points to the value of corresponding table entry in config file
+     * @param fieldNames  the list of field names we want to validate existence in this configTable
+     */
+    private void validateFields(String name, ObjectNode configTable, List<String> fieldNames) {
```
archolewa: If you wanted, you could have this method take a variable number of arguments: `validateFields(String name, ObjectNode configTable, String fieldNames...)`
Then you could call it with something like `validateFields(name, config, FIELD_NAME_1, FIELD_NAME_2, FIELD_NAME_3)` without having to wrap the field names
in a list.

Not a blocker.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DoubleSumMakerFactory.java
```diff
@@ -0,0 +1,19 @@
...
+package com.yahoo.bard.webservice.config.luthier.factories;
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.yahoo.bard.webservice.config.luthier.Factory;
+import com.yahoo.bard.webservice.config.luthier.LuthierIndustrialPark;
+import com.yahoo.bard.webservice.data.config.metric.makers.DoubleSumMaker;
+import com.yahoo.bard.webservice.data.config.metric.makers.MetricMaker;
+
+/**
+ * Factory that supports LongSumMetricMaker.
```
archolewa: Documentation is wrong.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/LongSumMakerFactory.java
```diff
@@ -0,0 +1,19 @@
+// Copyright 2019 Oath Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.config.luthier.factories;
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
```
archolewa: Import is wrong.

luthier/config/src/metrics.lua
gab-umich: it looks like a defaulting but I am putting it there to highlight the fact that it is a mandatory attribute the user should specify (as is for all of the `type`s in other configs), so it is more of a short hand instead of making it looks like an optional value, where defaultingUtil can help populating.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: Good catch

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialPark.java
gab-umich: Used a much more improved (?) function call to replace a lot of mindless work, while incorporating your idea.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AggregationAverageMakerFactory.java
gab-umich: fixed for many of them

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AggregationAverageMakerFactory.java
gab-umich: addressed

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/CountMakerFactory.java
gab-umich: done.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultLogicalTableGroupFactory.java
gab-umich: fixed as well

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultMetricFactory.java
gab-umich: fixed

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/LongSumMakerFactory.java
gab-umich: fixed.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DoubleSumMakerFactory.java
gab-umich: fixed

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultLogicalTableGroupFactory.java
gab-umich: done.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultMetricFactory.java
gab-umich: This method is lifted to the LuthierValidationUtils due to about 6 Factories are using it. As a result the code duplication massively reduced.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultMetricFactory.java
gab-umich: addressed. However, I think there might be other causes for an `IllegalArgumentException`, which means that I will use a more generic error message when throwing the `LuthierFactoryException`.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/DefaultMetricFactory.java
gab-umich: I can somehow envision that, but I will hold it off for now as I believe it might deserve a more thorough view when we get back together. And it is certainly not a feature we absolutely have to support in the first release. 
I'd still love to get _some_ of it done before the Summer ends, though.

luthier/config/src/utils/defaultingUtils.lua
gab-umich: transitively fixed. Now everything is defaulted to have a type `default`. Taking away this field in the end json file is not easy, as we have configured the java component to automatically look for the `FACTORY_KEY` i.e. `type` and find out which factory we need to invoke.

gab-umich:
addressed previous comments.

luthier/config/src/metrics.lua
archolewa: Why not? Do we expect people to *ever* use anything other than the default? We should document that the field exists of course, but seeing as how all the interesting customization is happening in the makers, and not in the metric, I don't see why this shouldn't be a default. For most people, this type field is going to be ceremony if they have to specify it. And we want as little ceremony as possible.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/AggregationAverageMakerFactory.java
gab-umich: done.

PR #947, opened 2019-07-30
Fixes a few typos in the metrics.lua code.

archolewa:
-- Updates some slightly wrong documentation.

-- We were using the wrong names for the average aggregation makers in
metrics.lua.

### DIFF
--- a/luthier/config/src/metricMakers.lua
+++ b/luthier/config/src/metricMakers.lua
@@ -20,8 +20,7 @@ local M = {}
     and the LongSum of metric2.
 
     Makers themselves are define in Java, as a part of your program using
-    Fili. Therefore, all references to makers are fully-qualified Java class
-    names.
+    Fili.
 
     Each maker is a table containing at least the following key:
         type: A unique string identifying the type for this maker. The full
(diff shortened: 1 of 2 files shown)


luthier/config/src/metricMakers.lua
```diff
@@ -56,8 +55,8 @@ for _, operation in ipairs({"PLUS","MINUS","MULTIPLY","DIVIDE"}) do
     }
 end
 
-for _, grain in ipairs {"HOUR", "DAY"} do
-    M["aggregateAverage" .. grain] = {
+for _, grain in ipairs {"Hour", "Day"} do
```
michael-mclawhorn: The enums are constant case 'HOUR' and 'DAY', do we want to use title case here?

archolewa: We should, because it's being used for the metric name and "aggregateAverageHOUR" is a pain in the butt metric name. 

Frankly, I think we should make it a policy of the builtin stuff at least that any time we reference enum values defined in Java, we can do so in a case insensitive manner.

PR #949, opened 2019-08-07
Luthier wiki

gab-umich:
Working luthier wiki demonstration. 


luthier/config/src/config.lua
```diff
@@ -36,16 +37,34 @@ local metricConfig = metricsUtils.build_metric_config(metrics)
 local metricMakerConfig = metricMakerUtils.build_metric_maker_config(metricMakers)
 local physicalTableConfig, logicalTableConfig = tableUtils.build_table_config(tables)
 
+--- write to testResource/tableName
+local function write_table(targetDir, tableName, configTable)
```
archolewa: I don't think we should be using "table" to refer to the the things we're writing. Too
easy to confuse with "logical table." Rather, I think we should use something like `config_entity`. So,
this function would be called `write_config_entity`. The `tableNames` variable below would
be `configEntityNames`, and so on.

luthier/config/src/config.lua
```diff
@@ -36,16 +37,34 @@ local metricConfig = metricsUtils.build_metric_config(metrics)
...
+    parser.save(targetDir .. tableName, configTable)
+    print(targetDir .. " table " .. tableName .. " Updated now.")
```
archolewa: Why not just use `targetDir .. tableName .. " Updated now."`? That way, it'll be much more
obvious to a customer where their is going.

luthier/config/src/config.lua
```diff
@@ -36,16 +37,34 @@ local metricConfig = metricsUtils.build_metric_config(metrics)
...
+end
+
+local function config(tableName, configTable)
```
archolewa: I would probably call this `write_config` instead of just `config`. I typically only use
a noun as a function name in Lua when that function is going to be returning something.

luthier/config/src/utils/defaultingUtils.lua
```diff
@@ -15,6 +15,7 @@ M.DEFAULT_DIMENSION_CATEGORY = "UNKNOWN_CATEGORY"
 M.DEFAULT_DIMENSION_IS_AGGREGATABLE = true
 M.DEFAULT_DIMENSION_FIELDS = {}
 M.DEFAULT_DIMENSION_KEY_VALUE_STORE = "memory"
+M.DEFAULT_DIMENSION_SKIP_LOADING = true
```
archolewa: I don't think we should make the system skip dimension loading by default. If people are going to
be skipping a key part of the dimension infrastructure, they should do it explicitly. Besides, if
somebody isn't using the dimension loading at all, there's nothing stopping them from defining
their own defaults.

luthier/src/main/java/com/yahoo/bard/webservice/application/LuthierWikiMain.java
```diff
@@ -0,0 +1,80 @@
+// Copyright 2019 Oath Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.application;
+
+import com.codahale.metrics.servlet.InstrumentedFilter;
```
archolewa: Your imports are all wrong here.

luthier/src/main/java/com/yahoo/bard/webservice/application/LuthierWikiMain.java
```diff
@@ -0,0 +1,80 @@
...
+import org.eclipse.jetty.servlet.ServletHolder;
+import org.glassfish.jersey.servlet.ServletContainer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.servlet.DispatcherType;
+import java.util.EnumSet;
+
+/**
+ * Launch Bard in Embedded Jetty.
```
archolewa: Should probably update this to be "Launch an instance of Fili that displays Wikipedia information, configured using
Luthier, in embedded Jetty." or something to that effect.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
```diff
@@ -110,6 +113,15 @@ LuthierPhysicalTableParams buildParams(
                 )
         );
         params.metadataService = resourceFactories.getMetadataService();
+        // registers each table we build in the metadataService
+        params.metadataService.update(
```
archolewa: It seems to me that we should be doing this in the LuthierIndustrialPark, rather than
buried in this params class. For one thing, I'm not comfortable with embedding side effects in
the construction of what appears to be a dumb bean. For another, since at the end of the day
the LuthierIndustrialPark is responsible for making sure everything is built, I think it
makes sense that it should have the responsibility to make sure that everything that should
know about our physical tables, knows about them.

luthier/config/src/config.lua
gab-umich: agreed.

luthier/config/src/config.lua
gab-umich: sure.

luthier/config/src/config.lua
gab-umich: done

luthier/src/main/java/com/yahoo/bard/webservice/application/LuthierWikiMain.java
gab-umich: done.

luthier/src/main/java/com/yahoo/bard/webservice/application/LuthierWikiMain.java
gab-umich: addressed.

luthier/config/src/utils/defaultingUtils.lua
gab-umich: done.

luthier/src/main/java/com/yahoo/bard/webservice/config/luthier/factories/SingleDataSourcePhysicalTableFactory.java
gab-umich: done.

wiki.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+mvn -pl fili-wikipedia-example exec:java -Dbard__fili_port=9998 -Dbard__druid_coord=http://localhost:8081/druid/coordinator/v1 -Dbard__druid_broker=http://localhost:8082/druid/v2
```
michael-mclawhorn: put into am luthier/scripts directory
add cd to ..
Add a comment to describe what it does

gab-umich: fixed

PR #951, opened 2019-08-12
Organize luthier

gab-umich:
Attempting to re-organize the Java and Lua code in the Luthier design module.


luthier/scripts/emptyLuthierApp.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# prepares json files
+cd $DIR/../src/main/lua/
+lua config.lua
+# this is equivalent to lua config.lua app ../../../target/classes/
+
+# install fili using mvn
+cd $DIR/../../
+mvn install -DskipTests -Dcheckstyle.skip
+# runs luthier wiki example on port 9012
```
michael-mclawhorn: Remove reference to wiki

gab-umich:
I don't think a `final char underscore = '_';` is a bad style. I will recommend ignoring this unless we have a good reason.

gab-umich:
Also, according to https://github.com/koalaman/shellcheck/wiki/SC2086, we should use `shell_script arg1 "$DIR"` instead of `shell_script arg1 $DIR`. Although the later one is valid, it might not be robust enough to handle word splitting. It should not be a problem for DIR specifically, but we should use this to encourage our user to be aware of this case.

luthier/scripts/emptyLuthierApp.sh
gab-umich: resolved

Issue #769, opened 2018-09-18
Injectable DataApiRequest Generator redesign


michael-mclawhorn:
Injecting DataApiRequest generation is one of the biggest blockers to customizing by client customers.  
DataApiRequestFactory was intended to aid in this but is only the start of a more significant change.

The proper solution is to decouple construction from binding/validating data api request values.

Create a value object DataApiRequestPojo and move the binding and validating into the factory.

PR #952, opened 2019-08-15
Pojo dari p1; Convert ApiRequestImpl to use generators

efronbs:

part 2: #954 
part 3: #955 

This PR is the first part in a series of changes focused on implementing a POJO data api request object. This step focuses on changes to the ApiRequestImpl object. It externalizes all of ApiRequest construction logic into Generator interfaces, and defers to default implementations of those interfaces.


fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/building/ResponseFormatTypeGenerator.java
```diff
@@ -0,0 +1,31 @@
...
+     */
+    ResponseFormatType generateAcceptFormat(String format) throws BadApiRequestException;
+
+    ResponseFormatTypeGenerator DEFAULT_RESPONSE_FORMAT_TYPE_GENERATOR = format -> {
+        try {
+            return format == null ?
+                    DefaultResponseFormatType.JSON :
+                    DefaultResponseFormatType.valueOf(format.toUpperCase(Locale.ENGLISH));
+        } catch (IllegalArgumentException e) {
+            throw new BadApiRequestException(ACCEPT_FORMAT_INVALID.format(format));
```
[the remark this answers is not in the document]
michael-mclawhorn: definitely pass the original exception to the new exception as a second argument to preserve stack track continuity.

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/binders/LogicalMetricGenerator.java
```diff
@@ -0,0 +1,113 @@
...
+     * @param table  The logical table for the request
+     *
+     * @throws BadApiRequestException if the requested metrics are not in the logical table
+     */
+    void validateMetrics(Set<LogicalMetric> logicalMetrics, LogicalTable table) throws BadApiRequestException;
+
+    /**
+     * Default implementation of this interface.
+     */
+    LogicalMetricGenerator DEFAULT_LOGICAL_METRIC_GENERATOR =
```
michael-mclawhorn: Why make this an anonymous instance? 
I don't disagree, necessarily, but I want to understand the thinking.

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/ApiRequestImpl.java
```diff
@@ -416,7 +314,7 @@ protected void validateMetrics(Set<LogicalMetric> logicalMetrics, LogicalTable t
             Granularity granularity,
             DateTimeFormatter dateTimeFormatter
     ) throws BadApiRequestException {
-        return generateIntervals(new DateTime(), apiIntervalQuery, granularity, dateTimeFormatter);
+        return IntervalBinders.generateIntervals(new DateTime(), apiIntervalQuery, granularity, dateTimeFormatter);
```
michael-mclawhorn: How do you feel about aligning this syntax to the generators above?

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/binders/IntervalGenerator.java
```diff
@@ -0,0 +1,9 @@
+// Copyright 2018 Oath Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.web.apirequest.binders;
+
+/**
+ * Generates intervals used by an ApiRequest implementation.
+ */
+public interface IntervalGenerator {
```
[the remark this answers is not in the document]
efronbs: I rename it in a future PR, but the intent is for the IntervalBinders class to just be a util class for help with interval generation, and have the default IntervalGenerator implemented elsewhere.

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/binders/LogicalMetricGenerator.java
efronbs: I think it's nice for the interface to provide the default implementations. I think it makes it easier to understand how the generators are used when the default implementation is directly linked through the interface, as it requires no knowledge of our coding practices (such as naming/packaging conventions) to find a good example of how the interfaces are intended to be used.

As for using an anonymous class, most of the default implementations are:
1. stateless, so the a single instance can be created here and used everywhere.
2. very small. LogicalMetricGenerator is an exception, but most of the default implementations of the generate methods are 5 lines or less. Making the default implementation an anonymous class greatly reduces clutter in most situations. While that is not true for LogicalMetricGenerator, I think the other benefits of having the anonymous class on the interface outweigh the ugliness of the long implementation being on the interface.

Additionally, using an anonymous class cuts down on the amount of classes we are making, which IMO makes it easier to navigate, and I think it makes it easier to understand how the interface is used. I think it's really easy and convenient for deciphering how the interface is supposed to be used when you can see the contract and the default implementation in one place.

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/ApiRequestImpl.java
efronbs: I think this will be clearer when IntervalBinders is renamed to IntervalUtils.

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/binders/DimensionGenerator.java
```diff
@@ -0,0 +1,115 @@
...
+                    if (!invalidDimensions.isEmpty()) {
+                        throw new BadApiRequestException(DIMENSIONS_UNDEFINED.format(invalidDimensions.toString()));
+                    }
+                    return generated;
+                }
+            }
+
+            @Override
+            public void validateRequestDimensions(
+                    Set<Dimension> requestDimensions,
```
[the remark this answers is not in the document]
michael-mclawhorn: reluctantly, I agree.  This kind of parameter reassignment is bad form, as much as I hate an extra variable being defined.

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/binders/LogicalMetricGenerator.java
michael-mclawhorn: In this case it's definitely NOT that simple.  I think we should make the generate and validate into explicit static methods, which can be tested, and then make the anonymous class delegate to them.

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/binders/ResponseFormatTypeGenerator.java
```diff
@@ -0,0 +1,40 @@
...
+import com.yahoo.bard.webservice.web.BadApiRequestException;
+import com.yahoo.bard.webservice.web.DefaultResponseFormatType;
+import com.yahoo.bard.webservice.web.ResponseFormatType;
+
+import java.util.Locale;
+
+/**
+ * Generates a ResponseFormatType object given a String representation of the desired format.
+ */
+public interface ResponseFormatTypeGenerator {
```
michael-mclawhorn: Shouldn't this one have a validate too, even if it's an empty one?

PR #953, opened 2019-08-15
Address codacy stuff

michael-mclawhorn:


luthier/src/main/lua/wikiApp/tables.lua
```diff
@@ -119,6 +119,7 @@ M.physical = {
         }
     },
     air = {
+        type = "permissive",
```
archolewa: What does this have to do with codacy?

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialParkSpec.groovy
```diff
@@ -1,6 +1,6 @@
 // Copyright 2019 Oath Inc.
 // Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
-package com.yahoo.bard.webservice.config.luthier
+package com.yahoo.bard.webservice.data.config.luthier;
```
archolewa: You fixed the package, but not the file location (should be under `webservice/data/config/luthier`).

luthier/src/test/groovy/com/yahoo/bard/webservice/config/luthier/LuthierIndustrialParkSpec.groovy
```diff
@@ -68,4 +78,71 @@ class LuthierIndustrialParkSpec extends Specification {
             secondTestDimension.is(testDimension)
             ! differentTestDimension.is(testDimension)
     }
+
+    def "When a metric name is fetched the second time, it refers to the same object as the first one"() {
```
kevinhinterlong: Typo? Looks like it was copied from the previous test
```suggestion
    def "When a metric name is fetched, it refers to the same object as the original definition"() {
```

luthier/src/main/lua/wikiApp/tables.lua
michael-mclawhorn: It increases test coverage.

PR #954, opened 2019-08-16
Pojo dari p2; DataApiRequestImpl no longer extends ApiRequestImpl.

efronbs:

part 1: https://github.com/yahoo/fili/pull/952
part 3: #955 

This PR is the second part in a series of changes to implement DataApiRequest as a POJO, that is constructed by a set of builder objects. This PR changes the existing DataApiRequestImpl to NOT extend ApiRequestImpl. This will make it easier to externalize its current behavior into Generator classes, which can then be used to build the pojo implementation.


michael-mclawhorn:
Style errors in build

PR #966, opened 2019-09-09
Pulls the Lua config in directly via LuaJ, rather than through JSON.

archolewa:
This PR isn't as big as it looks, because I'm moving the lua configuration around a lot.

-- We add a new interface, LuthierConfigNode. This is effectively a
glorified subset of the ObjectNode interface, because the original
version of Luthier read in JSON and worked with ObjectNodes. Luthier
itself now works with LuthierConfigNode. An implementation of the
interface has been added that wraps a LuaValue from LuaJ.  The
ResourceNodeSupplier has been turned into an interface
(LuthierSupplier), with an implementation that reads in and executes the
config.lua using LuaJ.

-- We remove any scripts and code that work with JSON-based
configuration, because they're no longer necessary, and their tests are
complicated by the fact that they need external JSON files (which are no
longer being auto-generated). See commit 14f7bd59 if you're interested
in what the JSON-based configuration looked like.

-- The KeyValueStoreDimensionFactory has been tweaked so that it no
longer camel cases field names. Field names should be exactly what the
customer writes in their configuration. Anything else is just asking for
frustrated customers.

-- The lua configuration has been tweaked so that now `config.lua`
returns a function that takes the app name and returns a table of
configuration. The LuaJNodeSupplier invokes the returned function,
passing in the app-name as extracted from configuration. This allows us to set the
app name using Fili's configuration framework.

-- Lua no longer performs any sort of filesystem manipulation, since it
no longer needs to generate JSON files. Similarly, we remove the
json.lua dependency since it is no longer necessary. Similarly, we
remove the verification script as no longer appropriate.

-- Imports are changed to the blatantly platform agnostic ".", rather
than Unixy "/". This plus the use of LuaJ should make Luthier fully
platform agnostic.

-- The config.lua script lives in the top level `src/main/resources`
directory. App-independent configuration lives in the
`src/main/resources/lib` directory, while all app-specific configuration
should live in `src/main/resources/<appName>`. All imports are relative
to the primary script being executed. In this case, `config.lua`.

-- Several shell scripts that are no longer necessary have been removed.
The only one that remains is `runApp.sh`, since it abstracts over the
rather fiddly mvn parameters needed to execute a local Fili instance.


michael-mclawhorn:
There's a config.lua in both resources and in resources/lua/config.lua

michael-mclawhorn:
It looks like the resources/lua/config.lua  is the one not in use.

michael-mclawhorn:
luthier/scripts/runApp.sh wikiApp doesn't seem to work anymore.  Is there an alternate way to invoke this?

archolewa:
@michael-mclawhorn That script isn't going to work anymore, because the app is now selected with the configuration parameter `bard__luthier_app`.

That being said, the unit tests were using a weird mix of test configuration and wiki configuration so I ended up consolidating them. So there's now only one app under `luthier/src/main/resources`. 

If you want a second one, you'll need to add a new folder `luthier/src/main/resources/<new_app>` and then set `bard__luthier_app=<new_app>`

PR #968, opened 2019-09-11
Lay groundwork for POJO dari impl

efronbs:


fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/rfc/Generator.java
```diff
@@ -0,0 +1,49 @@
...
+ * to generate. In your {@link AbstractBinderFactory} implementation you must make a named binding for the type of
+ * resource you wish to generate. You can only bind one generator per resource (though different resources may share
+ * a type, such as "count" and "topN"), so your implementation will replace the default generator for that resource.
+ *
+ * TODO: binding names should be written here for reference.
+ * TODO: binding order should be written here
+ *
+ * @param <T> The type of the resource this generator builds.
+ */
+public interface Generator<T> {
```
michael-mclawhorn: This interface probably deserves a slightly more specific name.  I'm not thinking of one yet, though, so this is unhelpful feedback at the moment.

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/rfc/RequestParameters.java
```diff
@@ -0,0 +1,239 @@
...
+
+import java.util.List;
+import java.util.Optional;
+
+import javax.ws.rs.core.PathSegment;
+
+/**
+ * Data object all raw pieces of data in an api request.
+ */
+public class RequestParameters {
```
michael-mclawhorn: it's generally considered more appropriate to hold the nulls internally and optional wrap on return.  I don't remember if we discussed making an exception for this class or not.

michael-mclawhorn: I suppose given the write to read ratio on this object, it actually saves cycles to wrap once.

efronbs: yeah that was the motivation. It's an immutable object that has no logic behind its internal state. Storing optionals over nulls just saves a little memory on read.

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequestValueObject.java
```diff
@@ -0,0 +1,772 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import javax.ws.rs.core.Response;
+
+/**
+ * And immutable POJO implementation of {@link DataApiRequest} contract. All data is provided through the constructor
+ * and set, with very minor or no transformations occurring on the data. Unlike the old implementation
+ * ({@link com.yahoo.bard.webservice.web.apirequest.DataApiRequestImpl}), this implementation does not build any of
+ * tis components and all components must be fully built at creation time.
```
michael-mclawhorn: spelling 'its'

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequestValueObject.java
```diff
@@ -0,0 +1,772 @@
...
+    private final Granularity granularity;
+    private final LinkedHashSet<Dimension> dimensions;
+    private final LinkedHashMap<Dimension, LinkedHashSet<DimensionField>> perDimensionFields;
+    private final LinkedHashSet<LogicalMetric> metrics;
+    private final List<Interval> intervals;
+    private final ApiFilters apiFilters;
+    private final LinkedHashMap<LogicalMetric, Set<ApiHaving>> havings;
+    private final OrderByColumn dateTimeSort;
+    private final LinkedHashSet<OrderByColumn> standardSorts;
+    private final LinkedHashSet<OrderByColumn> allSorts;
```
michael-mclawhorn: Why are we keeping all three of these?  We should either keep the separate fields or the merged one.

CHANGELOG.md
```diff
@@ -421,6 +421,10 @@ select filters)
 
 ### Added:
 
+- [Added groundwork classes for POJO `DataApiRequest` build path](https://github.com/yahoo/fili/issues/769)
+    * Generator contract, builder, and POJO object have all been added, along with some silent contract changes.
+    * Immutable implementations of `LinkedHashSet`, `LinkedHashMap`, and `ApiFilters` have been added. 
+
```
michael-mclawhorn: Need some additional CHANGELOG:

I would call out in the changed section:

Deprecated Optionals on constructors and withers
Replaced OptionalInt with Optional<Integer> on some interfaces

fili-core/src/main/java/com/yahoo/bard/webservice/config/BardFeatureFlag.java
```diff
@@ -34,7 +34,8 @@
     DRUID_DIMENSIONS_LOADER("druid_dimensions_loader_enabled"),
     CASE_SENSITIVE_KEYS("case_sensitive_keys_enabled"),
     DEFAULT_IN_FILTER("default_in_filter_enabled"),
-    REQUIRE_METRICS_QUERY("require_metrics_in_query")
+    REQUIRE_METRICS_QUERY("require_metrics_in_query"),
+    POJO_DARI_REQUIRE_ALL_STAGES_CALLED("require_all_data_request_building_stages_called")
```
michael-mclawhorn: We should add this to the default moduleConfig.properties file.

I'm inclined to default it to true, but I don't insist.

PR #969, opened 2019-09-16
Aligned code with design guidance on not constructing or holding Opti…

michael-mclawhorn:
…onals on the ApiRequest interface(s) and constructor impls.


fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/DataApiRequest.java
```diff
@@ -66,24 +66,6 @@
...
-     */
-    static LinkedHashSet<OrderByColumn> combineSorts(
-            Optional<OrderByColumn> dateTimeSort,
-            LinkedHashSet<OrderByColumn> standardSorts
-    ) {
-        LinkedHashSet<OrderByColumn> result = new LinkedHashSet<>();
-        dateTimeSort.ifPresent(result::add);
-        result.addAll(standardSorts);
-        return result;
-    }
```
efronbs: Consider moving back to DataApiRequest interface.

PR #973, opened 2019-09-17
Removed bouncer auth references.

michael-mclawhorn:


fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/AuthorizationStatus.java
```diff
@@ -0,0 +1,69 @@
+// Copyright 2016 Yahoo Inc.
```
efronbs: update to 2019 oath copyright

PR #978, opened 2019-09-19
Repackage sample apps

michael-mclawhorn:
Move the sample applications into their own subpackages.

Split Luthier into a library and a sample application.


sample-applications/luthier-sample-app/pom.xml
```diff
@@ -0,0 +1,136 @@
...
+                        <goals>
+                            <goal>java</goal>
+                        </goals>
+                        <configuration>
+                            <mainClass>com.yahoo.bard.webservice.application.LuthierMain</mainClass>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <!--
```
kevinhinterlong: Uncomment?

travis/deploy.bash
```diff
@@ -39,6 +39,8 @@ fi
 # Set environment variable for testing purposes
 export FILI_TEST_LIST=a,2,bc,234
 
+echo $(mvn -version)
```
kevinhinterlong: I don't think we need to `echo`
```suggestion
mvn -version
```

sample-applications/luthier-sample-app/pom.xml
michael-mclawhorn: This is no longer the launch mechanism.  I'm removing it.

PR #981, opened 2019-09-24
Generators p1

efronbs:


fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/generator/DefaultGranularityGenerator.java
```diff
@@ -0,0 +1,134 @@
...
+     * @return the generated granularity.
+     */
+    @Override
+    public Granularity bind(
+            DataApiRequestBuilder builder,
+            RequestParameters params,
+            BardConfigResources resources
+    ) {
+        String granularity = params.getGranularity()
+                .orElseThrow(() -> new BadApiRequestException(UNKNOWN_GRANULARITY.logFormat("null")));
```
michael-mclawhorn: I'm fairly convinced that this 'null' isn't the right messaging here.

I'm trying to trace the code, is there a code path going through here?

efronbs: I will change the message to something else. I was mostly trying to duplicate https://github.com/yahoo/fili/blob/master/fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/ApiRequestImpl.java#L237-L248 , but after looking at it again I think I just need to say granularity was missing from the query.

michael-mclawhorn: ultimately, we're going to probably treat missing granularity as meaning all granularity eventually.

Is there any code path that does down this path?

efronbs: Currently all data requests must have granularity to even hit the data servlet. Under normal circumstances this code path should never be hit.

PR #992, opened 2019-11-18
COUNT(*) support

zmz0305:
Add COUNT(\*) support.
When there is a `count` metric with `countMaker`, it will be translated into a COUNT(\*) in SQL query.

### DIFF
--- a/fili-sql/src/main/java/com/yahoo/bard/webservice/sql/presto/DruidQueryToPrestoConverter.java
+++ b/fili-sql/src/main/java/com/yahoo/bard/webservice/sql/presto/DruidQueryToPrestoConverter.java
@@ -162,17 +162,22 @@ public class DruidQueryToPrestoConverter extends DruidQueryToSqlConverter {
                     LOG.debug(msg);
                     return new RuntimeException(msg);
                 }))
-                .map(sqlAggregation -> builder.aggregateCall(
-                        sqlAggregation.getSqlAggFunction(),
-                        false,
-                        null,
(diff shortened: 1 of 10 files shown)


fili-sql/src/main/java/com/yahoo/bard/webservice/sql/presto/DruidQueryToPrestoConverter.java
```diff
@@ -162,17 +162,22 @@ protected SqlTimeConverter buildSqlTimeConverter() {
...
-                        // TODO: investigate the field name case discrepency
-                        CaseFormat.LOWER_UNDERSCORE.to(
-                                CaseFormat.LOWER_CAMEL,
-                                sqlAggregation.getSqlAggregationAsName()
-                        ),
-                        builder.field(sqlAggregation.getSqlAggregationFieldName())
-                ))
+                .map(sqlAggregation -> {
+                    if (sqlAggregation.getSqlAggFunction() == SqlStdOperatorTable.COUNT) {
+                        return builder.countStar("count");
```
kevinhinterlong: I don't think we want to hardcode the name here, I think we should be able to take it from the aggregation
```suggestion
                        return builder.countStar(sqlAggregation.getSqlAggregationAsName());
```

kevinhinterlong: Also should probably add a test around this

PR #998, opened 2019-12-20
Fix contains filter behavior

zmz0305:
Currently, if we fire a query with a contains filter, there are two cases:
1. If the filter is applied on a cached dimension, it will first look for all matching result and then fire a Druid query that looks for fields equals to those matching result.
2. If the filter is applied on a non-cached dimension, it will directly fire a Druid query that looks for fields equals to the substring in the contains filter. This will usually result in an empty result, which is not what a user would expect. 

This issue also causes the Presto connector to create a SQL query that use `=` instead of `like` for substring contains filter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Current
 -------
 
 ### Fixed:
+- [Fix contains filter behavior](https://github.com/yahoo/fili/pull/998)
+    * When there is a Contains filter applied to a non-cached dimension, it will be translated into a `SeachFilter` instead of `SelectorFilter`.
+    
 - [Fix Presto query on filtering](https://github.com/yahoo/fili/pull/995)
     * When translating from sql query to Presto query, there is no type information available for table columns. To make filtering `WHERE` clauses works in Presto, cast coulmns to varchar before comparing
 - [`MetricUnionAvailability` properly defensively copies availability map](https://github.com/yahoo/fili/pull/997)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Current
 
 ### Fixed:
 - [Fix contains filter behavior](https://github.com/yahoo/fili/pull/998)
-    * When there is a Contains filter applied to a non-cached dimension, it will be translated into a `SeachFilter` instead of `SelectorFilter`.
+    * When there is a Contains filter applied to a non-cached dimension, it will be translated into a `SearchFilter` instead of `SelectorFilter`.
     
 - [Fix Presto query on filtering](https://github.com/yahoo/fili/pull/995)
     * When translating from sql query to Presto query, there is no type information available for table columns. To make filtering `WHERE` clauses works in Presto, cast coulmns to varchar before comparing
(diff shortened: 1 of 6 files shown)


cdeszaq:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/builders/DruidOrFilterBuilder.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/Fili/fili/pullRequest?prid=4717445&bid=15751805)

cdeszaq:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/builders/DruidOrFilterBuilder.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/Fili/fili/pullRequest?prid=4717445&bid=15751805)

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ Current
 -------
 
 ### Fixed:
+- [Fix contains filter behavior](https://github.com/yahoo/fili/pull/998)
+    * When there is a Contains filter applied to a non-cached dimension, it will be translated into a `SeachFilter` instead of `SelectorFilter`.
```
kevinhinterlong:
```suggestion
    * When there is a Contains filter applied to a non-cached dimension, it will be translated into a `SearchFilter` instead of `SelectorFilter`.
```

cdeszaq:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/builders/DruidOrFilterBuilder.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/Fili/fili/pullRequest?prid=4717445&bid=15751805)

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/builders/ConjunctionDruidFilterBuilder.java
```diff
@@ -233,4 +234,16 @@ protected abstract Filter buildDimensionFilter(Dimension dimension, Set<ApiFilte
...
+
+    /**
+     * Given a dimension and its contains filter values, return a list of Contains SearchFilters.
+     * @param dimension the dimension to search for
+     * @param values the contains value
+     * @return the list of SearchFilters
+     */
+    protected List<Filter> buildContainsSearchFilters(Dimension dimension, List<String> values) {
+        return values.stream()
+                .map(value -> new SearchFilter(dimension, SearchFilter.QueryType.Contains, value))
```
colemanProjects: should the query type be insensitive contains by default?

colemanProjects: we tested this, looks like this defaults to a case insensitive search for druid, but a case sensitive search for sql. May want to standardize this to be case insensitive for both.

colemanProjects:
can we also have a few tests on the sql side?

zmz0305:
@colemanProjects @kevinhinterlong  The sql part is already tested: https://github.com/yahoo/fili/blob/master/fili-sql/src/test/groovy/com/yahoo/bard/webservice/sql/evaluator/FilterEvaluatorSpec.groovy

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/builders/ConjunctionDruidFilterBuilder.java
kevinhinterlong: > 
> 
> @colemanProjects @kevinhinterlong The sql part is already tested: https://github.com/yahoo/fili/blob/master/fili-sql/src/test/groovy/com/yahoo/bard/webservice/sql/evaluator/FilterEvaluatorSpec.groovy

@zmz0305 If we make the sql search case insensitive then we should add a test

zmz0305: @kevinhinterlong we decide to create another ticket for the SQL case insensitive search, so it won't be in this PR

PR #995, opened 2019-12-02
fix filter clause

Ruiqian1:
I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

Fili Presto support is problematic for filtering clauses. Currently all fields related to filtering is assumed to be of varchar type. All additional step to add casting upon query generation.

PR #999, opened 2019-12-31
Fix or caluse for filtering

Ruiqian1:

The solution to different types of presto tables is that we add the cast to string before comparing. In the previous PR (https://github.com/yahoo/fili/pull/995), we split the filter clause by ` AND `. In fact, ` OR ` should also be considered. Fix this issue in this PR.

git issue: https://github.com/yahoo/fili/issues/1000

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ pull request if there was one.
 
 Current
 -------
-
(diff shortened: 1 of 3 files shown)


Ruiqian1:
git issue https://github.com/yahoo/fili/issues/1000

PR #1003, opened 2020-01-24
Signal metrics

michael-mclawhorn:
Refactor LogicalMetric into a LogicalMetricImpl
Define a 'SignalMetric' interface that accepts signals
Define a SignalHandler to determine whether a signal can be accepted and to dispatch a signal to a Metric Transformer
Make AggregationAverageMaker build SignalMetrics

Tests


cdeszaq:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/LogicalMetricImpl.java  4
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/signal/DefaultSignals.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/signal/SignalMetricImpl.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/BaseSignalMetricMaker.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/signal/SignalHandler.java  5
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/signal/UnknownSignalValueException.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/signal/TimeAverageMetricTransformer.java  3
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/signal/MetricTransformer.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Fili/fili/pullRequest?prid=4884758&bid=16182917)

fili-core/src/test/groovy/com/yahoo/bard/webservice/data/metric/signal/TimeAverageTransformerSpec.groovy
```diff
@@ -0,0 +1,90 @@
...
+        MetricMaker sketchMaker = new ThetaSketchMaker(metricDictionary, 128)
+
+        LogicalMetric longSum = maker.make("foo", "bar")
+        LogicalMetric sketchUnion = sketchMaker.make("foo", "bar")
+        Map params = [(BASE_SIGNAL): value]
+
+        LogicalMetric averageMetric = timeAverageTransformer.apply(longSum, BASE_SIGNAL, params)
+        LogicalMetric sketchAverage = timeAverageTransformer.apply(sketchUnion, BASE_SIGNAL, params)
+
+        expect:
```
efronbs: Consider adding a check of time grain of result inner & outer queries.

cdeszaq:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/UnknownProtocolValueException.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/Protocol.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/LogicalMetricImpl.java  4
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/TimeAverageMetricTransformer.java  3
- fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/BaseSignalMetricMaker.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java  5
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolMetricImpl.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/MetricTransformer.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Fili/fili/pullRequest?prid=4884758&bid=16182917)

cdeszaq:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/MetricTransformer.java  2
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/Protocol.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/UnknownProtocolValueException.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/BuiltInProtocols.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/TimeAverageMetricTransformer.java  3
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java  5
- fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/BaseProtocolMetricMaker.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolMetricImpl.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/LogicalMetricImpl.java  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Fili/fili/pullRequest?prid=4884758&bid=16182917)

cdeszaq:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/UnknownProtocolValueException.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/Protocol.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/LogicalMetricImpl.java  4
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java  5
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/BuiltInProtocols.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolMetricImpl.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/TimeAverageMetricTransformer.java  3
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/MetricTransformer.java  2
- fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/BaseProtocolMetricMaker.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Fili/fili/pullRequest?prid=4884758&bid=16182917)

cdeszaq:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/BuiltInProtocols.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/TimeAverageMetricTransformer.java  3
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/MetricTransformer.java  2
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/LogicalMetricImpl.java  4
- fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/BaseProtocolMetricMaker.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/UnknownProtocolValueException.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/Protocol.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java  5
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolMetricImpl.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Fili/fili/pullRequest?prid=4884758&bid=16182917)

cdeszaq:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/TimeAverageMetricTransformer.java  3
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java  5
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/MetricTransformer.java  2
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/LogicalMetricImpl.java  4
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolMetricImpl.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/Protocol.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/UnknownProtocolValueException.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/BaseProtocolMetricMaker.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/BuiltInMetricProtocols.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Fili/fili/pullRequest?prid=4884758&bid=16182917)

cdeszaq:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/UnknownProtocolValueException.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/BuiltInMetricProtocols.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/LogicalMetricImpl.java  4
- fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/BaseProtocolMetricMaker.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/TimeAverageMetricTransformer.java  3
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/MetricTransformer.java  2
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/Protocol.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java  5
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolMetricImpl.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Fili/fili/pullRequest?prid=4884758&bid=16182917)

cdeszaq:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/LogicalMetricImpl.java  4
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java  5
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolMetricImpl.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/BaseProtocolMetricMaker.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/BuiltInMetricProtocols.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/Protocol.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/UnknownProtocolValueException.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/MetricTransformer.java  2
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/TimeAverageMetricTransformer.java  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Fili/fili/pullRequest?prid=4884758&bid=16182917)

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * An object to describe which protocols a metric can support and to invoke the apply for that protocol.
+ *
+ * In the case where a stack of metrics might be delegated to, white or black listing allows the handler to indicate
+ * whether delegation should be allowed.
+ */
+public class ProtocolSupport {
```
efronbs: I think some wires are getting crossed in this class. My understanding of the Protocol and ProtocolMetric contracts is that multiple different Protocols can share a contract but have different transformers and/or different core parameters. This contract is also intended to support a single ProtocolMetric supporting multiple different Protocols for the same contract (the different types of reagg opperations we have been discussing).

This class implicitly implies that a single metric will only support 1 Protocol per contract name. This is enforced through the instantiation of the `protocolMap` field in the constructor, as well as the accepts, withoutProtocol, withoutProtocols, and getProtocol being based solely on contract name. 

I think this class needs to be refactored to support operations based on solely protocol contract and operations based on the combination of contract + core parameter.

I also now believe that we protocol contract name and protocol core parameter name should be made into concrete classes that wrap strings, which should hopefully clarify some of the contracts here and make these types of errors less likely.

archolewa: I don't agree. It seems to me that a given metric will only have *one* ProtocolHandler, but it does support multiple Protocols. However, we don't support protocol overloading (so a single metric
can't have "currency(String denomination)" and currency(int currencyCode)" on a single metric). We don't want to support protocol overloading. Protocol overloading would be complicated to implement. Making people
use names that are unique for a given metric's namespace is fine.

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java
```diff
@@ -0,0 +1,148 @@
...
+
+    /**
+     * A substitute for Boolean to indicate Yes, No or Uknown in the case where delegates might service a signal.
+     */
+    public enum Accepts {
+        TRUE,
+        FALSE,
+        MAYBE  // If this Handler cannot accept the protocol, but doesn't make positive claims if another might.
```
archolewa: I stumbled a little here because initially I expected FALSE to mean "I
don't accept this Protocol, but a different handler might." As opposed to "You
can have my metric when you pry it from my cold, dead fingers." I worry that this
might be similarly confusing for others, and a source for bugs. One option is to
line things up with how *I* envisioned it, and do something like TRUE, FALSE, REJECT.
But I'm not convinced that helps so much as change the set of people who will be confused.
I wonder if maybe better names? Something that more emphatically states that FALSE should
be used when you want to stop processing on the metric? Something like "ACCEPT", "REJECT",
and "MAYBE"?

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java
```diff
@@ -0,0 +1,148 @@
...
+    public ProtocolSupport(
+            Collection<Protocol> protocols,
+            Collection<String> blacklist
+    ) {
+        protocolMap = protocols.stream().collect(Collectors.toMap(Protocol::getContractName, Function.identity()));
+        this.blacklist = blacklist;
+    }
+
+    /**
+     * Determine if this protocol handler or accepts this protocol.
```
archolewa: Extraneous 'or' here ("handler *or* accepts")

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java
```diff
@@ -0,0 +1,148 @@
...
+     * @param protocolName  The name of a protocol to not handle.
+     *
+     * @return A protocol handler with additional signals bound.
+     */
+    public ProtocolSupport withoutProtocol(String protocolName) {
+        return withoutProtocols(Collections.singleton(protocolName));
+    }
+
+    /**
+     * Create a modified protocol handler which doesn't accepts certain protocols.
```
archolewa: JavaDoc typo: "accepts" should be "accept".

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java
```diff
@@ -0,0 +1,148 @@
...
+     *
+     * @param protocolSupport  The protocol supports whose blacklists should not be handled.
+     *
+     * @return A protocol handler with additional protocols not supported.
+     */
+    public ProtocolSupport combineBlacklists(Collection<ProtocolSupport> protocolSupport) {
+
+        List<String> protocols =
+                protocolSupport.stream()
+                        .flatMap(support -> support.blacklist.stream())
```
archolewa: Should we be using a (protected?) getter here, rather than accessing the field directly?

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java
```diff
@@ -0,0 +1,148 @@
...
+     *
+     * @return A protocol handler with additional protocols supported.
+     */
+    public ProtocolSupport withProtocols(Collection<Protocol> addedProtocols) {
+
+        Collection<Protocol> newProtocols = Stream.concat(addedProtocols.stream(), protocolMap.values().stream())
+                .collect(Collectors.toSet());
+
+        Collection<String> newBlackList = blacklist.stream()
+                .filter(name -> !addedProtocols.contains(name))
```
archolewa: Shouldn't we be converting `addedProtocols` to a `Collection<String>` first? It looks to me like
we're trying to check that a collection of Protocols contains a String.

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/BuiltInMetricProtocols.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+public class BuiltInMetricProtocols {
+
+    // The names for default protocols
+    // Intentionally mutable
+    private static final Set<String> DEFAULT_PROTOCOL_CONTRACTS = new HashSet<>();
+
+    // The dictionary for default protocols
+    // intentionally mutable
+    private static final ProtocolDictionary PROTOCOL_DICTIONARY = new ProtocolDictionary();
```
archolewa: Why do have both of these fields? Why not just PROTOCOL_DICTIONARY, with DEFAULT_PROTOCOL_CONTRACTS
derived from PROTOCOL_DICTIONARY.keys()? Doesn't having two collections risk error, like adding
a protocol name to DEFAULT_PROTOCOL_CONTRACTS, but forgetting to add the protocol to PROTOCOL_DICTIONARY
(or vice versa) ?

archolewa: Oh wait, I think I might understand now. You don't need to (and shouldn't) always add new
protocols `DEFAULT_PROTOCOL_CONTRACTS` and `PROTOCOL_DICTIONARY`. You only add a protocol name to
`DEFAULT_PROTOCOL_CONTRACTS` if you want *every metric in your system* to support it by default. If you add
the Protocol to just `PROTOCOL_DICTIONARY`, you register it with the system, but you have to explicitly
configure metrics to handle it.

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java
efronbs: From the discussion Michael and I had it sounds like we want to support protocol contract overloading. The prime example here is with the day of typical week calculation. The current plan is to expose this as a trend metric, while still supporting standard trend metrics. day of typical week and (for example) week over week metrics will use the same contract, but will have different core parameters.

If this design is kept, then ProtocolSupport MUST be keyed by the combination of contract and core parameter. Otherwise, we need to formally change the Protocol contract.

michael-mclawhorn: ProtocolSupport is immutable on a given metric.  So on a given metric there is exactly one Protocol instance servicing a given contract.

Contract implementations /should/ be the same systemwide.  e.g. it would be weird to support two different protocol implementations with the same contract name on a given system.  Some applications may have different implementations for a given contract than the base implementation.  (e.g. reAgg supporting custom reaggregation types). 

A given protocol has exactly one 'core' parameter which is the signal used to trigger that contract.  It would be sensible, in general, to have the contract name /be/ the coreParameter name, but I can think of edge case exceptions.

Day over Typical Week is a transform triggered that essentially reaggs and trends.  The transform associated with it needs to be attached to either the reagg core parameter or the trend core parameter.  To be applied, it needs to ensure that both the reagg contract and the trend contract are available on the metric it's being applied to.

I was imagining it being implemented on an if-else in the trend protocol.  Essentially from the system point of view it would /be/ a trend, like DayOverDay, just a trend that's illegal if the reagg contract isn't available.

michael-mclawhorn: I hate making boxes around strings if there isn't some typesafety benefit.

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java
michael-mclawhorn: We're collecting new protocols into with the existing one in line 128 and we're pruning blacklisted names in the second stream.

cdeszaq:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/UnknownProtocolValueException.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolMetricImpl.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/LogicalMetricImpl.java  4
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/TimeAverageMetricTransformer.java  3
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/MetricTransformer.java  2
- fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/BaseProtocolMetricMaker.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/Protocol.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java  5
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/BuiltInMetricProtocols.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Fili/fili/pullRequest?prid=4884758&bid=16182917)

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
```diff
@@ -67,36 +74,51 @@
...
      */
     public AggregationAverageMaker(MetricDictionary metrics, ZonelessTimeGrain innerGrain) {
-        super(metrics);
+        this(metrics, innerGrain, BuiltInMetricProtocols.getStandardProtocolSupport());
+    }
+
+    /**
+     * Constructor.
+     *
+     * Any aggregation average maker will disable the protocol Aggregation.
```
archolewa: I think we intend `Reaggregation` here not `Aggregation`. Also, we may want to include a link to where the
protocol is defined in BuiltInMetricProtocols.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
```diff
@@ -107,8 +129,9 @@ private TemplateDruidQuery buildOuterQuery(
         // Build the average post aggregation
         FieldAccessorPostAggregation sumPost = new FieldAccessorPostAggregation(sum);
         PostAggregation average = new ArithmeticPostAggregation(
-                metricDictionary,
+                metricName,
                 DIVIDE,
+
```
archolewa: Extraneous blank line.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/ArithmeticMaker.java
```diff
@@ -39,18 +42,36 @@
      * @param metricDictionary  The dictionary used to resolve dependent metrics when building the LogicalMetric
      * @param function  The arithmetic operation performed by the LogicalMetrics constructed by this maker
      * @param resultSetMapperSupplier  A function that takes a metric column name and produces at build time, a result
+     * @param defaultProtocolSupport Protocols to support by default
```
archolewa: Thing this should actually go one line lower. `set mapper` just below goes with `resultSetMapperSupplier`.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/ArithmeticMaker.java
```diff
@@ -66,10 +87,18 @@ public ArithmeticMaker(MetricDictionary metricDictionary, ArithmeticPostAggregat
     }
 
     @Override
-    protected LogicalMetric makeInner(LogicalMetricInfo logicalMetricInfo, List<String> dependentMetrics) {
-        // Get the ArithmeticPostAggregation operands from the dependent metrics
+    public ResultSetMapper makeCalculation(
+            final LogicalMetricInfo logicalMetricInfo, final List<LogicalMetric> dependentMetrics
```
archolewa: Are we intentionally making these parameters final? I didn't think that was a thing we typically did. Also,
the parameters should each be on their own line if we're wrapping.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/ArithmeticMaker.java
```diff
@@ -80,13 +109,25 @@ protected LogicalMetric makeInner(LogicalMetricInfo logicalMetricInfo, List<Stri
...
-        TemplateDruidQuery query = getMergedQuery(dependentMetrics).withPostAggregations(postAggregations);
-        return new LogicalMetric(
-                query,
-                resultSetMapperSupplier.apply(logicalMetricInfo.getName()),
-                logicalMetricInfo
-        );
+    @Override
+    public ProtocolSupport makeProtocolSupport(
+            final LogicalMetricInfo logicalMetricInfo,
+            final List<LogicalMetric> dependentMetrics
```
archolewa: Another one with final parameters.

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java
```diff
@@ -0,0 +1,148 @@
...
+    /**
+     * Determine if this protocol handler or accepts this protocol.
+     *
+     * @param contractName The name of the protocol to test for.
+     *
+     * @return TRUE if this metric directly or indirectly supports this signal, FALSE if it refuses, MAYBE if it
+     * doesn't assert authority.
+     */
+    public Accepts accepts(String contractName) {
+        return protocolMap.containsKey(contractName) ? Accepts.TRUE :
```
archolewa: Shouldn't the blacklist take priority? Right now, if it's in the protocol map, and in the blacklist, we'll return Accepts.TRUE, where I would expect us to return Accepts.REJECT.

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/TimeAverageMetricTransformer.java
```diff
@@ -0,0 +1,215 @@
...
+
+/**
+ * An interface for transforming metrics into other metrics.
+ */
+public class TimeAverageMetricTransformer implements MetricTransformer {
+
+    private static final MetricDictionary EMPTY_METRIC_DICTIONARY = new MetricDictionary();
+
+    public static final String CONTRACT_NAME = "reagg";
+    private static final String CORE_PARAMETER = "reagg";
```
efronbs: This code is smelly to me. I don't have a full suggested change, and this code might be fine, but I think it should be discussed.

Protocol contracts are unique over an entire code base. You can't have multiple different Protocols with the same contract. These two fields are strongly coupling this transformer to a specific Protocol instance; the instance that takes the reagg contract with the reagg core parameter and uses this MetricTransformer as its transformer instance. However, that instance is never materialized anywhere in the code base. To use this, the client needs to materialize the specific Protocol that we are essentially defining for them.

However, the obvious issue is that the materialized Protocol might want to use a "Multiplexing" transformer to support other reagg operations in addition to this one. It is uncomfortable to just create the Protocol that this Transformer requires because it might need to be tweaked or expanded. Multiplexing on it's own raises an issue; if this contract and core parameter are shared with another Protocol, how do we determine which transformer should handle a signal? I think we need to at least define the MultiplexingTransformer, and make it enforce uniqueness of core parameter names.

That also raises another issue: It is possible to handle multiple different Protocols with the same contract in the same system but different transformers. Since you can't compare Protocols or MetricTransformers (no equals method override) it is difficult at runtime to tell if Protocols are different. They might have the same contract and core parameter, but with no way to compare transformers they could be totally different.

Depending on what we decide, it might also be worth defining an enum or collection of Protocol implementations for default Protocols.

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/MetricTransformer.java
```diff
@@ -0,0 +1,41 @@
...
+     * Transform a metric using a Protocol, and signal data.
+     *
+     * @param logicalMetric  The metric to transform.
+     * @param protocol The protocol being transformed
+     * @param parameterValues  The data associated with that signal.
+     *
+     * @return A new metric based on the signal;
+     * @throws UnknownProtocolValueException if this transformer doesn't know how to accept this signal
+     */
+    LogicalMetric apply(LogicalMetric logicalMetric, Protocol protocol, Map<String, String> parameterValues)
```
efronbs: Taking Protocol here is a little code smelly to me. Protocol is tied to a MetricTransformer instance, and it seems unlikely to me that a single MetricTransformer would be able to handle multiple different Protocol contracts. Thus, consuming the applying Protocol (which should just be the Protocol that references this transformer) feels weird. See my comment on TimeAverageMetricTransformer for more in depth thoughts on MetricTransformer.

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/TimeAverageMetricTransformer.java
```diff
@@ -0,0 +1,215 @@
...
+                protocolSupportSupplier.get()
+        );
+    }
+    @Override
+    public LogicalMetric apply(LogicalMetric logicalMetric, Protocol protocol, Map<String, String> parameterValues)
+            throws UnknownProtocolValueException {
+
+        String parameterValue = parameterValues.get(CORE_PARAMETER);
+
+        if (!makerConfigMap.containsKey(parameterValue)) {
```
efronbs: Two things:
1. We need to check if `parameterValue` is null. `Map` implementations that do not support null keys will throw a null pointer on `Map#containsKey` (https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#containsKey-java.lang.Object-)
2. I think if you are explicitly tying this Transformer to the "reagg" core parameter, the CORE_PARAMETER must be `public`. Currently it is `private`.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/ArithmeticMaker.java
michael-mclawhorn: ArithmeticMaker uses the protocolSupport as a base and then subtracts the blacklists from the dependencies before attaching it to a new metric.

michael-mclawhorn: Renaming to 'baseProtocolSupport' and clarifying in param

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/ArithmeticMaker.java
michael-mclawhorn: Nope. not intentional.

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/BuiltInMetricProtocols.java
```diff
@@ -0,0 +1,88 @@
...
+package com.yahoo.bard.webservice.data.metric.protocol;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * A collection of signal names supported.
+ */
+public class BuiltInMetricProtocols {
```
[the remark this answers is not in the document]
michael-mclawhorn: Ok, so the way I currently have it structured, the ProtocolDictionary.DEFAULT **is** the global protocol registry, and this contains the list of named protocols to attach by default. (Editable any time before actually constructing makers)

Let's see if there's a better option.

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/MetricTransformer.java
efronbs: After discussing with @michael-mclawhorn we will change Transformers to NOT close over the coreParameter, nor the contractName. The passed in Protocol will provide the core parameter to the transformer, which it can then use to pull the core parameter value out of the `parameterValues` map. Note that the Transformer WILL need to close over any auxiliary parameter names.

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/TimeAverageMetricTransformer.java
efronbs: Addressed by my comment on MetricTransformer.java

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolMetric.java
```diff
@@ -0,0 +1,45 @@
...
+     *
+     * The transformed metric is not necessarily a protocol metric.
+     *
+     * @param protocolName  The name of the protocol to apply
+     * @param parameters  A map of keys and values representing the transformation.
+     *
+     * @return A metric that has accepted this protocol transformation.
+     * @throws UnknownProtocolValueException if the protocol cannot be processed correctly
+     */
+    LogicalMetric accept(String protocolName, Map<String, String> parameters) throws UnknownProtocolValueException;
```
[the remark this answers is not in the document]
michael-mclawhorn: I'm actually going to use protocolName wherever it's not disambiguating.  There's no other name associated with protocols and I think the verbosity softens clarity rather than heightening it.

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolMetricImpl.java
```diff
@@ -0,0 +1,71 @@
...
+import com.yahoo.bard.webservice.data.metric.mappers.ResultSetMapper;
+
+import java.util.Map;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * Implement a metric that supports protocols.
+ */
+public class ProtocolMetricImpl extends LogicalMetricImpl implements ProtocolMetric {
```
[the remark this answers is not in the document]
michael-mclawhorn: adding it

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java
michael-mclawhorn: meh.  direct access of members inside a class isn't the devil.

cdeszaq:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolDictionary.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/LogicalMetricImpl.java  4
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/DefaultSystemMetricProtocols.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java  4
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/MetricTransformer.java  2
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolMetricImpl.java  5
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/Protocol.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/UnknownProtocolValueException.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/BaseProtocolMetricMaker.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/protocols/TimeAverageMetricTransformer.java  3
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/protocols/ReaggregationProtocol.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Fili/fili/pullRequest?prid=4884758&bid=16182917)

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/protocols/ReaggregationProtocol.java
```diff
@@ -0,0 +1,32 @@
...
+package com.yahoo.bard.webservice.data.metric.protocol.protocols;
+
+import com.yahoo.bard.webservice.data.metric.protocol.Protocol;
+
+import java.util.Collection;
+
+/**
+ * Define and implement a Reaggregation Protocol.
+ */
+public class ReaggregationProtocol {
```
efronbs: Consider renaming to TimeAverageReaggregationProtocol

cdeszaq:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/MetricTransformer.java  2
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolDictionary.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/UnknownProtocolValueException.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/BaseProtocolMetricMaker.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/LogicalMetricImpl.java  4
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/Protocol.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolMetricImpl.java  5
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/DefaultSystemMetricProtocols.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java  4
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/protocols/ReaggregationProtocol.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/protocols/TimeAverageMetricTransformer.java  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Fili/fili/pullRequest?prid=4884758&bid=16182917)

cdeszaq:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolMetricImpl.java  5
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/MetricTransformer.java  2
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/protocols/ReaggregationProtocol.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/LogicalMetricImpl.java  4
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/UnknownProtocolValueException.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/BaseProtocolMetricMaker.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolDictionary.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/Protocol.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/protocols/TimeAverageMetricTransformer.java  3
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java  4
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/DefaultSystemMetricProtocols.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Fili/fili/pullRequest?prid=4884758&bid=16182917)

cdeszaq:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/DefaultSystemMetricProtocols.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/protocols/ReaggregationProtocol.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java  4
- fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/BaseProtocolMetricMaker.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/UnknownProtocolValueException.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/Protocol.java  1
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/MetricTransformer.java  2
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolMetricImpl.java  5
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/LogicalMetricImpl.java  4
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/protocols/TimeAverageMetricTransformer.java  3
- fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolDictionary.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Fili/fili/pullRequest?prid=4884758&bid=16182917)

PR #1020, opened 2020-03-13
Update contract on ProtocolMetric and MetricTransformer interfaces

efronbs:
issue: https://github.com/yahoo/fili/issues/1021

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolMetric.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolMetric.java
@@ -53,10 +53,14 @@ public interface ProtocolMetric extends LogicalMetric {
     boolean accepts(String protocolName);
 
     /**
-     * Apply this protocol with these parameters to this metric and return a (typically different) metric.
+     * <p>Apply this protocol with these parameters to this metric and return a (typically different) metric. Because
+     * the result metric is inherently different than the input metric, the result must have a new (minimal) set of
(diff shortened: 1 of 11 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolMetric.java
```diff
@@ -3,40 +3,76 @@
...
 
 import com.yahoo.bard.webservice.data.metric.LogicalMetric;
+import com.yahoo.bard.webservice.data.metric.LogicalMetricInfo;
 
 import java.util.Map;
 
 /**
- * Protocol metrics are capable of handling transforming signals from the client.
+ * <p>Protocol metrics are metrics that have the ability to have transformations applied to them at query time. Protocol
+ * metrics themselves represent valid metrics even if no calculations are applied.
```
michael-mclawhorn: 'Protocol metrics should generally represent valid metrics even if no calculations are applied.'

fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/protocol/ProtocolSupport.java
```diff
@@ -52,9 +59,26 @@ public ProtocolSupport(
     public ProtocolSupport(
             Collection<Protocol> protocols,
             Set<String> blacklist
+    ) {
```
michael-mclawhorn: in the javadoc for this method add:

Name is defaulted to empty string, indicating that this protocol support will not publish metadata for the base metric.

PR #1031, opened 2020-03-26
Add missing argument to error message in RequestLog.

archolewa:
Apparently, there was a small bug in a infrequently used error path in the RequestLog, where the error message's format string expected two arguments, but we were only passing it one.


michael-mclawhorn:
I think this PR is showing the travis error that ate up my yesterday.

PR #1039, opened 2020-04-07
Making granularity available from api request

sgoplani:

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/generator/metric/ProtocolLogicalMetricGenerator.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/generator/metric/ProtocolLogicalMetricGenerator.java
@@ -76,29 +76,6 @@ public class ProtocolLogicalMetricGenerator extends DefaultLogicalMetricGenerato
     }
 
     /**
-     * Extracts the list of metrics from the url metric query string and generates a set of LogicalMetrics.
-     * <p>
-     * If the query contains undefined metrics, {@link BadApiRequestException} will be
-     * thrown.
-     *
-     * This method is meant for backwards compatibility. If you do not need to use this method for that reason please
-     * prefer using a generator instance instead.
-     *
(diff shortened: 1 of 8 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/generator/metric/ProtocolLogicalMetricGenerator.java
```diff
@@ -80,53 +75,32 @@ public ProtocolLogicalMetricGenerator(
...
-            );
-        }
-
-        if (!builder.getLogicalTableIfInitialized().isPresent()) {
-            throw new BadApiRequestException("A logical table is required for all data queries");
-        }
-
-        validateMetrics(entity, builder.getLogicalTableIfInitialized().get());
+        List<ApiMetric> apiMetrics = parseApiMetricQuery(apiMetricQuery);
+        return applyProtocols(apiMetrics, metricDictionary);
```
michael-mclawhorn, line 98: This shouldn't be here.

fili-core/src/main/java/com/yahoo/bard/webservice/web/apirequest/generator/metric/ProtocolLogicalMetricGenerator.java
```diff
@@ -135,25 +109,37 @@ public void validate(
...
      * @return set of metric objects
      */
+    @Override
     public LinkedHashSet<LogicalMetric> generateLogicalMetrics(
             String apiMetricQuery,
+            Granularity requestGranularity,
             MetricDictionary metricDictionary
     ) {
-        List<ApiMetric> apiMetrics = parseApiMetricQuery(apiMetricQuery);
+        List<ApiMetric> apiMetrics = parseApiMetricQueryWithGranularity(apiMetricQuery, requestGranularity);
```
michael-mclawhorn, line 123: We can probably drop the custom name and rely on the method signature to distinguish them.

PR #1072, opened 2020-07-24
Asynchronous callbacks that create futures have caused leaks.  This f…

michael-mclawhorn:
…ix defensively clears references to large objects in closures after they are consumed by turning them into local variables.


michael-mclawhorn:
the underlying leak is in the asynchttpclient library, and these fixes don't patch the core leak which is the http request response itself.

PR #1091, opened 2020-10-02
Make build work when JAVA_TOOL_OPTIONS is set.

jacobtolar:

This change makes the maven build work when [JAVA_TOOL_OPTIONS](https://docs.oracle.com/javase/8/docs/technotes/guides/troubleshoot/envvars002.html). 

I happened to have `JAVA_TOOL_OPTIONS=-Djava.awt.headless=true` in my environment when building Fili. With this, Java (e.g. AdoptOpenJDK 1.8) dumps a line to stderr on every jvm invocation:
```
Picked up JAVA_TOOL_OPTIONS: -Djava.awt.headless=true
```

The ant task here used to generate an Avro file [redirects stderr to `output`](https://ant.apache.org/manual/Types/redirector.html) unless you direct it somewhere else: 

> Output: Name of a file to which output should be written. If the error stream is not also redirected to a file or property, it will appear in this output.

As-is, the stderr log message about JAVA_TOOL_OPTIONS gets redirected to the top of the Avro file, which results in an obviously invalid Avro file. 

Setting `logError` redirects this message to the console:

> This attribute is used when you wish to see error output in Ant's log and you are redirecting output to a file/property. The error output will not be included in the output file/property.

which fixes things for me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ pull request if there was one.
 Current
 -------
 ### Fixed:
+- [Fix: Make build work with JAVA_TOOL_OPTIONS](https://github.com/yahoo/fili/pull/1091)
+    * Fixed maven build failure when JAVA_TOOL_OPTIONS was set by reconfiguring ant output
+
 - [Fix: Bad serialization of AllGranularity](https://github.com/yahoo/fili/issues/1093)
     * Change to rollup formatter broke serialization of all timegrain.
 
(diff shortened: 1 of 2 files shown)


michael-mclawhorn:
@jacobtolar Could you throw a quick Changelog into the changelog to summarize what you have in the PR?
That way we'll know why this change was made later?

michael-mclawhorn:
Also you'll want to rebase on master to get the owasp vulnerability  patch

michael-mclawhorn (approved):

PR #1092, opened 2020-10-02
use getName() instead of toString()

bharatmotwani:


michael-mclawhorn (approved):

michael-mclawhorn (changes requested):
Need ChangeLog and Test

michael-mclawhorn (approved):

PR #1095, opened 2020-10-07
Add logic to rename aggregation to avoid name collision

bharatmotwani:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,7 +65,7 @@ Current
 ### Added:
 - [Add logic to rename aggregation to avoid name collision](https://github.com/yahoo/fili/issues/1095)
     * Add renameIfConflicting logic for aggregations in BaseProtocolMetricMaker.
-        - Subclasses of `BaseProtocolMetricMaker` must implement `getRenamedMetricNameWithPrefix` method to have unique rename prefix for corresponding maker. 
+        - Subclasses of `BaseProtocolMetricMaker` can implement `getRenamedMetricNameWithPrefix` method to have unique rename prefix for corresponding maker. 
     
 - [Add ability in fili-sql to translate FilteredAggregation into SQL](https://github.com/yahoo/fili/pull/1083)
    * Translate a Druid query with `n` FilteredAggregation into SQL using `(n + 1)` subquery unions. 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ pull request if there was one.
 Current
 -------
 ### Fixed:
+- [Fix: Add logic to rename aggregation to avoid name collision](https://github.com/yahoo/fili/issues/1095)
+    * Add renameIfConflicting logic for aggregations in BaseProtocolMetricMaker.
+
 - [Fix: Bad serialization of AllGranularity](https://github.com/yahoo/fili/issues/1093)
     * Change to rollup formatter broke serialization of all timegrain.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,6 @@ pull request if there was one.
 Current
 -------
 ### Fixed:
-- [Fix: Add logic to rename aggregation to avoid name collision](https://github.com/yahoo/fili/issues/1095)
-    * Add renameIfConflicting logic for aggregations in BaseProtocolMetricMaker.
 
 - [Fix: Bad serialization of AllGranularity](https://github.com/yahoo/fili/issues/1093)
     * Change to rollup formatter broke serialization of all timegrain.
@@ -65,6 +63,10 @@ Current
     * Created `LegacyGenerator` as a bridge interface from the existing constructor based api request impls and the factory based value object usage.
 
 ### Added:
+- [Add logic to rename aggregation to avoid name collision](https://github.com/yahoo/fili/issues/1095)
+    * Add renameIfConflicting logic for aggregations in BaseProtocolMetricMaker.
+        - Subclasses of `BaseProtocolMetricMaker` must implement `getRenamedMetricNameWithPrefix` method to have unique rename prefix for corresponding maker. 
+    
 - [Add ability in fili-sql to translate FilteredAggregation into SQL](https://github.com/yahoo/fili/pull/1083)
    * Translate a Druid query with `n` FilteredAggregation into SQL using `(n + 1)` subquery unions. 
        - See PR description for details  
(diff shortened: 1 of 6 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/ArithmeticMaker.java
```diff
@@ -92,6 +94,17 @@ public ArithmeticMaker(MetricDictionary metricDictionary, ArithmeticPostAggregat
         );
     }
 
+    /**
+     * Abstract method to build new aggregation name.
+     *
+     * @param name The dependent metric name need rename.=
```
michael-mclawhorn, line 100: typo

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/BaseProtocolMetricMaker.java
```diff
@@ -59,12 +61,68 @@ public LogicalMetric makeInnerWithResolvedDependencies(
...
+     * If there is no name conflict the input metric is returned unchanged. Otherwise, the input metric is renamed to
+     * not conflict with the final output name.
+     * <p>
+     * This method specifically the {@link GeneratedMetricInfo} typing on the input metric if it has to be renamed. All
+     * other subclasses of {@link LogicalMetricInfo} are lost during the renaming. If maintaining more LogicalMetricInfo
+     * types is required, a withName method must «should be added to the LogicalMetricInfo interface.
+     *
+     * @param finalOutputName  The name to check for conflicts on
+     * @param dependentMetric  The dependent metric to check for a conflicting name
+     * @return  The metric without a name conflicting with finalOutputName
```
michael-mclawhorn, line 85: blank line between @param  and @return

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/BaseProtocolMetricMaker.java
```diff
@@ -59,12 +61,68 @@ public LogicalMetric makeInnerWithResolvedDependencies(
...
+        ProtocolSupport protocolSupport = makeProtocolSupport(logicalMetricInfo, renamedDependentMetrics);
         return new ProtocolMetricImpl(logicalMetricInfo, partialQuery, calculation, protocolSupport);
     }
 
+    /**
+     * Renames the provided logical metric if there is a name conflict between it and the final desired output name.
+     * If there is no name conflict the input metric is returned unchanged. Otherwise, the input metric is renamed to
+     * not conflict with the final output name.
+     * <p>
+     * This method specifically the {@link GeneratedMetricInfo} typing on the input metric if it has to be renamed. All
```
michael-mclawhorn, line 79: I think there's a grammar error here, but I'm not exactly sure what you were trying to say in this sentence.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/BaseProtocolMetricMaker.java
```diff
@@ -59,12 +61,68 @@ public LogicalMetric makeInnerWithResolvedDependencies(
...
+        }
+
+        return dependentMetric.withLogicalMetricInfo(resultInfo);
+    }
+
+    /**
+     * Abstract method to build new aggregation name.
+     * @param name The dependent metric name need rename.
+     * @return Renamed metric name with specified prefix
+     */
```
michael-mclawhorn, line 112: empty line between name and param and param and return.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/AggregationAverageMaker.java
```diff
@@ -97,6 +99,17 @@ public AggregationAverageMaker(
         this.innerGrain = innerGrain;
     }
 
+    /**
+     * Abstract method to build new renamed aggregation name.
```
michael-mclawhorn, line 103: You can used @overrides here instead of defining the javadoc.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/BaseProtocolMetricMaker.java
michael-mclawhorn: javadoc shouldn't call out that it's abstract, rather, describe the expected contract, so that javadoc by default describes the contract of implementing subclasses.

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ pull request if there was one.
 Current
 -------
 ### Fixed:
+- [Fix: Add logic to rename aggregation to avoid name collision](https://github.com/yahoo/fili/issues/1095)
+    * Add renameIfConflicting logic for aggregations in BaseProtocolMetricMaker.
```
michael-mclawhorn, line 12: This isn't a fix, this is a change.

Also, this is a breaking change, which we try to avoid, so put some bold text to tell people what they need to fix if this change breaks them (in this case, implementing the rename method)

michael-mclawhorn (changes requested):
Let's make sure to add a test where we hit the double rename problem.  You can probably feed the same metric multiple times through the transformer manually.

fili-core/src/main/java/com/yahoo/bard/webservice/data/config/metric/makers/BaseProtocolMetricMaker.java
```diff
@@ -59,12 +68,73 @@ public LogicalMetric makeInnerWithResolvedDependencies(
...
+     *
+     * @return Returns true if newName is already been used by the aggregations. otherwise false.
+     */
+    protected boolean ifConflicting(String newName, LogicalMetric dependentMetric) {
+        Set<Aggregation> aggregations = dependentMetric.getTemplateDruidQuery().getAggregations();
+        Set<PostAggregation> postAggregations = dependentMetric.getTemplateDruidQuery().getPostAggregations();
+        Set<String> allNames = Stream.concat(aggregations.stream(), postAggregations.stream())
+                .map(MetricField::getName)
+                .collect(Collectors.toSet());
+        return !allNames.add(newName);
```
michael-mclawhorn, line 135: contains would probably make more sense here.

michael-mclawhorn (commented):
Ok.

I apologize for asking for a bigger change on a followup review.  
Per the discussion I just shared, I'd prefer to move the responsibility for the generated-metric-info collision resolution up to ProtocolMetricImpl.apply rather than having it in the MetricMaker code.

My proposal is to lift up 'renameIfConflicting' and the rename to ProcolMetricImpl, rename based not on every agg and post agg but instead on the metric name, and use the protocol name as the prefix for the renaming.

michael-mclawhorn (approved):
add default implementation to base protocol maker
change changelog to reflect
unit test the rename behavior

PR #1112, opened 2020-11-13
Canonicalize Cache Key

sgoplani:


michael-mclawhorn (approved):

fili-core/src/main/java/com/yahoo/bard/webservice/util/Utils.java
```diff
@@ -251,10 +227,70 @@ public static void canonicalize(JsonNode node, ObjectMapper mapper, boolean pres
...
+
+    /**
+     * Given a JsonObjectNode, order the ObjectNode entries on the key recursively and
+     * replace context blocks with empty nodes.
+     * <p>
+     * This method is recursive.
+     *
+     * @param node            The root of the tree of json nodes.
+     * @param mapper          The object mapper that creates and empty node.
+     * @param preserveContext Boolean indicating whether context should be omitted.
```
michael-mclawhorn, line 245: The formatting from your IDEA is not the @param name(2 spaces) description of our contribution spec.  You probably need to tweak your autoformatter settings.

fili-core/src/test/groovy/com/yahoo/bard/webservice/util/UtilsSpec.groovy
```diff
@@ -83,4 +86,100 @@ class UtilsSpec extends Specification {
         expect:
         Stream.of(1, 2, 3).reduce() { a, b -> Utils.getMinValue(a, b) }.get() == 1
     }
+
+    def "Canonicalization of object with context preserved without recursion"() {
```
michael-mclawhorn, line 90: I'd be tempted to condense these into a @Unroll, where test with the where blocks having a comment field:
Similar to the https://github.com/yahoo/fili/blob/246596e04a10c0e4cde701dbd216d606e412799d/fili-core/src/test/groovy/com/yahoo/bard/webservice/async/jobs/stores/JobRowFilterSpec.groovy#L37-L73

michael-mclawhorn (changes requested):
The javadocs aren't consistent with the contribution guidelines.  Probably it's an IDE formatter setting you're using.  I'd be open to liberalizing the comment style, but not without consultation from @archolewa .

michael-mclawhorn (approved):

PR #1121, opened 2020-12-28
add presto daily table support; fix presto all time grain support; add presto topN support

Ruiqian1:
add presto daily table support:
   * provide timestamp formant and choose based on physical table time grain

fix presto all time grain support
   * When `INSTANCE` is chosen as time grain, the timestamp field should not be a group key as we want the aggression across the entire time ranges

add presto topN support
   * Allow topN druid query to be converted to presto query by using ROW_NUMBER() and adding an outer layer for the presto query


zmz0305:
Can we have an example query to explain a bit about how TopN is accomplished in the PR description for future reference.

Ruiqian1:
Sure. Here is an example of how topN works. Consider the regular query `/data/tqPublisher/hour/dealId/?dateTime=P1D/current&metrics=adOpportunities(filterDimension=None,filterType=contains,filterValues=None)&sort=adOpportunities(filterDimension=None,filterType=contains,filterValues=None)|asc&format=info`. dili-presto translates it to the following presto query

zmz0305:
The code looks good. We probably need some unit tests for the TopN case.

Ruiqian1:
Good catch. Just added unit test

fili-sql/src/main/java/com/yahoo/bard/webservice/sql/DruidQueryToSqlConverter.java
```diff
@@ -402,40 +415,99 @@ protected RelNode convertDruidQueryToRelNode(
...
+            while (metric instanceof TopNMetric) {
+                metric = ((TopNMetric) metric).getMetric();
+            }
+            if (metric == null) {
+                throw new IllegalStateException("null metric found for the topN query");
+            }
+            String metricName = (String) metric;
+            int timePartFunctions = sqlTimeConverter.timeGrainToDatePartFunctions(druidQuery.getGranularity()).size();
+            int groupBys = druidQuery.getDimensions().size() + timePartFunctions;
+            List<RexNode> timePartitions = builder.fields().subList(druidQuery.getDimensions().size(), groupBys);
```
kevinhinterlong, line 476: I think this would be cleaner if it was like 
```suggestion
            int dimensions = druidQuery.getDimensions().size();
            List<RexNode> timePartitions = builder.fields().subList(dimensions, dimension + timePartFunctions);
```

fili-sql/src/main/java/com/yahoo/bard/webservice/sql/presto/PrestoSqlBackedClient.java
```diff
@@ -138,9 +148,11 @@ private JsonNode executeAndProcessQuery(DruidAggregationQuery<?> druidQuery) {
      * </ul>
      *
      * @param sqlQuery The sql query converted from druid query.
+     * @param timestampFormat The format used to parse the timestamp
      * @return a presto compatible sql dialect.
      */
-    protected static String sqlQueryToPrestoQuery(String sqlQuery) {
+    protected static String sqlQueryToPrestoQuery(String sqlQuery, String timestampFormat) {
+        LOG.info("using fili version of PrestoSqlBackedClient");
```
kevinhinterlong, line 155: Maybe delete this? Or maybe change to not use `LOG.info`

kevinhinterlong (commented):
Small stuff, looks good

fili-sql/src/main/java/com/yahoo/bard/webservice/sql/DruidQueryToSqlConverter.java
Ruiqian1: 👍

fili-sql/src/main/java/com/yahoo/bard/webservice/sql/presto/PrestoSqlBackedClient.java
Ruiqian1: good catch. removed the logging

kevinhinterlong (approved):

zmz0305 (approved):

PR #1128, opened 2021-02-08
Fix Presto Limit

yanhan0502:
https://github.com/yahoo/fili/issues/1129
Fix nested cases. e.g. 
Before fix: 
`... FETCH NEXT 1 ROWS ONLY) AS "t4"`  ->  `LIMIT 1`
After fix:
`... FETCH NEXT 1 ROWS ONLY) AS "t4"`  ->  `LIMIT 1) AS "t4"`


yanhan0502:
SD complains about test failures however, `mvn clean test` works fine in local machine.

zmz0305:
Some tests in fili-core failed... Not sure why

Ruiqian1:
Can we also add an entry to https://github.com/yahoo/fili/blob/master/CHANGELOG.md

yanhan0502:
Looks like SD build is not stable. 
The first commit failed at `17:51:35 [INFO] Fili: Core Libraries ............................... FAILURE [01:43 min]`
But the second commit now failed at 
`09:22:02 [INFO] Fili: Luthier Config Libraries ..................... FAILURE [  2.637 s]`

Ruiqian1 (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ Current
 -------
 ### Fixed:
 
+- [Fix: Presto Limit](https://github.com/yahoo/fili/pull/1128)
+
```
michael-mclawhorn, line 13: * Create a github issue and link to it for the feature.
* Make note of the tests that you're ignoring and removing and why (flaky on the first and non-compiling on the second)

fili-sql/src/main/java/com/yahoo/bard/webservice/sql/presto/PrestoSqlBackedClient.java
```diff
@@ -246,13 +246,7 @@ private static String fetchToLimitHelper(String fixQuotePrestoQuery) {
...
                 }
             }
-            int idxOfLineReturn = nextPart.indexOf('\n');
-            if (idxOfLineReturn != -1) {
-                nextPart = "LIMIT " + rows + nextPart.substring(idxOfLineReturn);
-
-            } else {
-                nextPart = "LIMIT " + rows;
-            }
+            nextPart = nextPart.replace("FETCH NEXT " + rows + " ROWS ONLY", "LIMIT " + rows);
```
michael-mclawhorn, line 249: Is there any way to provide unit test coverage for this change?

michael-mclawhorn (changes requested):
On the documentation sides, some better CHANGELOG
On the testing side, I don't know if it's possible, but I always try to put in a confirmatory test on any fixes if at all possible.

CHANGELOG.md
yanhan0502: Sure. Created https://github.com/yahoo/fili/issues/1129

fili-sql/src/main/java/com/yahoo/bard/webservice/sql/presto/PrestoSqlBackedClient.java
yanhan0502: Added a UT for this class and included *Test.java in maven-surefire-plugin to enable the test.

michael-mclawhorn (approved):
Looks good. 
It might be an interesting experience to try rewriting your test in spock to see how much more civilized multiline strings can be.  Regardless everything is good to go in terms of contribution guidelines, so here goes.

PR #1133, opened 2021-02-15
Presto sketch

ldkge:

Add Sketch metrics support for presto.

### DIFF
--- a/fili-sql/src/main/java/com/yahoo/bard/webservice/sql/DruidQueryToSqlConverter.java
+++ b/fili-sql/src/main/java/com/yahoo/bard/webservice/sql/DruidQueryToSqlConverter.java
@@ -66,6 +66,7 @@ public class DruidQueryToSqlConverter {
     public static final int NO_OFFSET = -1;
     public static final int NO_LIMIT = -1;
     public static final String ZERO_PLACEHOLDER = "zero";
+    private final static String SKETCH_STRING = "'round(thetasketch_estimate(thetasketch_union(";
 
     /**
      * Constructs the default converter.
@@ -214,7 +215,7 @@ public class DruidQueryToSqlConverter {
         RelToSqlConverter relToSql = calciteHelper.getNewRelToSqlConverter();
         SqlPrettyWriter sqlWriter = calciteHelper.getNewSqlWriter();
 
-        return writeSql(sqlWriter, relToSql, query);
+        return formatSketchQuery(writeSql(sqlWriter, relToSql, query));
     }
 
     /**
@@ -421,6 +422,13 @@ public class DruidQueryToSqlConverter {
             queryRelNode = convertDruidQueryWithFilteredAgg(builder, druidQuery, apiToFieldMapper, sqlTable);
         } else {
             builder = builder.scan(sqlTable.getName());
+            RelBuilder finalBuilder = builder;
+            List<RexNode> sketchMetrics = druidQuery.getAggregations().stream()
+                    .filter(aggregation -> aggregation.getType().equals("thetaSketch"))
+                    .map(aggregation -> finalBuilder.alias(finalBuilder.literal(
+                            String.format("round(thetasketch_estimate(thetasketch_union(%s)))",
+                                    aggregation.getFieldName())), aggregation.getName()))
+                    .collect(Collectors.toList());
             queryRelNode = builder
                     .filter(
                             getAllWhereFilters(builder, druidQuery, apiToFieldMapper, sqlTable.getTimestampColumn())
@@ -437,6 +445,7 @@ public class DruidQueryToSqlConverter {
                     .project(
                             (Iterable) ImmutableList.builder()
                                     .addAll(builder.fields())
+                                    .addAll(sketchMetrics)
                                     .addAll(getPostAggregations(builder, druidQuery, apiToFieldMapper))
                                     .build()
                     )
@@ -839,6 +848,7 @@ public class DruidQueryToSqlConverter {
         aggToSqlType.put("longMin", SqlTypeName.BIGINT);
         return druidQuery.getAggregations()
                 .stream()
+                .filter(aggregation -> !aggregation.getType().equals("thetaSketch"))
                 .map(aggregation -> getDruidSqlAggregationConverter().apply(aggregation, apiToFieldMapper))
                 .map(optionalSqlAggregation -> optionalSqlAggregation.orElseThrow(() -> {
                     String msg = "Couldn't build sql aggregation with " + optionalSqlAggregation;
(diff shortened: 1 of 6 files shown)


fili-sql/src/main/java/com/yahoo/bard/webservice/sql/DruidQueryToSqlConverter.java
```diff
@@ -929,6 +939,55 @@ private void sqlAggregationHelper(
         return allGroupBys;
     }
 
+    /**
+     * Find all word positions in string.
+     * @param textString a string to search into.
+     * @param word a target word.
+     * @return an array of indexes of all positions word was found in the textString.
```
michael-mclawhorn, line 946: style issues @params two spaces after param name.
1 line of whitespace between params and returns/throws.

fili-sql/src/main/java/com/yahoo/bard/webservice/sql/DruidQueryToSqlConverter.java
```diff
@@ -421,6 +422,13 @@ protected RelNode convertDruidQueryToRelNode(
             queryRelNode = convertDruidQueryWithFilteredAgg(builder, druidQuery, apiToFieldMapper, sqlTable);
         } else {
             builder = builder.scan(sqlTable.getName());
+            RelBuilder finalBuilder = builder;
+            List<RexNode> sketchMetrics = druidQuery.getAggregations().stream()
+                    .filter(aggregation -> aggregation.getType().equals("thetaSketch"))
```
michael-mclawhorn, line 427: ThetaSketchAggregation.AGGREGATION_TYPE as the magic string here to support compatibility if things ever change.

fili-sql/src/main/java/com/yahoo/bard/webservice/sql/DruidQueryToSqlConverter.java
```diff
@@ -421,6 +422,13 @@ protected RelNode convertDruidQueryToRelNode(
...
+                    .map(aggregation -> finalBuilder.alias(finalBuilder.literal(
+                            String.format("round(thetasketch_estimate(thetasketch_union(%s)))",
```
michael-mclawhorn, line 429: This feels like it'd make a good constant.

fili-sql/src/main/java/com/yahoo/bard/webservice/sql/DruidQueryToSqlConverter.java
```diff
@@ -839,6 +848,7 @@ private void sqlAggregationHelper(
         aggToSqlType.put("longMin", SqlTypeName.BIGINT);
         return druidQuery.getAggregations()
                 .stream()
+                .filter(aggregation -> !aggregation.getType().equals("thetaSketch"))
```
michael-mclawhorn, line 851: ThetaSketchAggregation.AGGREGATION_TYPE as the magic string here to support compatibility if things ever change.

fili-sql/src/main/java/com/yahoo/bard/webservice/sql/DruidQueryToSqlConverter.java
```diff
@@ -929,6 +939,55 @@ private void sqlAggregationHelper(
...
+            wordLength = word.length();
+        }
+        return indexes;
+    }
+
+    /**
+     * Removes the quotes around the Sketch metric literal.
+     * @param sqlQuery the sql string built by the RelBuilder.
+     * @return the sql string without quote around the sketch metric expression.
+     */
```
michael-mclawhorn, line 969: Ditto on the style feedback.
Vertical whitepsace between description and params and between params and returns. Two spaces after the variable name on parameter comments.

fili-sql/src/main/java/com/yahoo/bard/webservice/sql/DruidQueryToSqlConverter.java
```diff
@@ -929,6 +939,55 @@ private void sqlAggregationHelper(
...
+
+        int index = 0;
+        while (index != -1) {
+            index = lowerCaseTextString.indexOf(lowerCaseWord, index + wordLength);
+            if (index != -1) {
+                indexes.add(index);
+            }
+            wordLength = word.length();
+        }
+        return indexes;
```
michael-mclawhorn, line 962: java Matcher and Pattern would be way more idiomatic to accomplish this.  why did you do this this way?

fili-sql/src/main/java/com/yahoo/bard/webservice/sql/DruidQueryToSqlConverter.java
```diff
@@ -929,6 +939,55 @@ private void sqlAggregationHelper(
...
+                beginningPos = pos + 1;
+                // Find closing quote location
+                int nextPos = beginningPos + sqlQuery.substring(beginningPos).indexOf("'");
+                // Append string up to closing quote
+                stringBuilder.append(sqlQuery, beginningPos, nextPos);
+                beginningPos = nextPos + 1;
+            }
+            stringBuilder.append(sqlQuery, beginningPos, sqlQuery.length());
+            return stringBuilder.toString();
+        } else {
```
michael-mclawhorn, line 986: This whole method feels like it would have more sensibly be evaluated with regular expressions.

michael-mclawhorn (changes requested):
It seems really odd to me to be using loops over character positions and dumping character positions into arrays of positions here.

Is there a reason to favor that stratgegy over more traditional reg-ex matching?

fili-sql/src/main/java/com/yahoo/bard/webservice/sql/DruidQueryToSqlConverter.java
```diff
@@ -929,6 +941,32 @@ private void sqlAggregationHelper(
...
     }
 
+    /**
+     * Removes the quotes around the Sketch metric literal.
+     *
+     * @param sqlQuery  the sql string built by the RelBuilder.
+     *
+     * @return the sql string without quote around the sketch metric expression.
+     */
+    private String formatSketchQuery(String sqlQuery) {
```
zmz0305, line 951: We were trying to avoid using string hacks to form queries since the string hack from intern code has created enough problems. 
I have told Ashish but I am not sure if he had synced up with you that the `RelBuilder.literal()` or some other functions in RelBuilder class could be used to create the function call sql code.

ldkge: I'm already using `RelBuilder.literal()` https://github.com/yahoo/fili/pull/1133/files#diff-564e0da92e321668ecc97c6ff955f3a03f326ad9daa6167fa11210d5503faac0R429-R433 
The intended use for literal is to create constants that's why there are quotes added that need to be removed in the end. Looking at the documentation I haven't found a better way to achieve that.

michael-mclawhorn (changes requested):
Just needs a changlog (with the link pointed to a relevant issue if you haven't created one, please)

michael-mclawhorn (approved):

PR #1139, opened 2021-02-23
Add virtual columns and first last any aggregation support

sgoplani:


fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/DruidAggregationQuery.java
```diff
@@ -91,6 +99,15 @@
...
 
+    /**
+     * Returns a copy of this query with the specified virtual columns.
+     *
+     * @param virtualColumns  the new virtual columns
+     *
+     * @return the query copy
+     */
+    Q withVirtualColumns(Collection<VirtualColumn> virtualColumns);
+
```
michael-mclawhorn, line 110: 👍

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/GroupByQuery.java
```diff
@@ -145,6 +146,81 @@ public GroupByQuery(
...
+            Having having,
+            Collection<Aggregation> aggregations,
+            Collection<PostAggregation> postAggregations,
+            Collection<Interval> intervals,
+            LimitSpec limitSpec,
+            QueryContext context,
+            boolean incrementQueryId,
+            Collection<VirtualColumn> virtualColumn
+    ) {
+        super(
```
michael-mclawhorn, line 180: Same advice as above, this becomes the primary constructor and non test invocations should migrate to it.  It's ugly, but it minimizes the changes of throwing virtual columns away under copy-with-modification which we do all over the place.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/virtualcolumns/ExpressionVirtualColumn.java
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * VirtualColumn to provide queryable column views created from a set of columns during a query.
+ */
+public class ExpressionVirtualColumn implements VirtualColumn {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ExpressionVirtualColumn.class);
+
+    private final String name;
+    private final String expression;
+    private final String outputType;
```
michael-mclawhorn, line 19: https://druid.apache.org/docs/latest/querying/virtual-columns.html

We might want to defensively make an inner enum for the types on the virtual column.

michael-mclawhorn (changes requested):
A few changes in virtual columns constructors.
A lot of small downstream changes in with-ers in response to the deprecation.
Also, looking at this code I wonder if we'd be well served to stop having classes for each of the aggregations and instead build some sort of 'simpleAggregator' type so we don't have to have so many classes.
That's definitely a nice to have and we don't have to do it now.  But we have now upwards of 15 classes that differ only in an internal type string.

If we were to do it, I'd make the new SimpleAggregation class, leave the existing explicit aggregation classes in, but deprecated, and then see how painful the refactor is to rip any explicit references to them out.

Like I said, not a thing we have to do now, but I have having lots of lines of classes with almost no behavioral difference between them.

sgoplani:
> A few changes in virtual columns constructors.
> A lot of small downstream changes in with-ers in response to the deprecation.
> Also, looking at this code I wonder if we'd be well served to stop having classes for each of the aggregations and instead build some sort of 'simpleAggregator' type so we don't have to have so many classes.
> That's definitely a nice to have and we don't have to do it now. But we have now upwards of 15 classes that differ only in an internal type string.
> 
> If we were to do it, I'd make the new SimpleAggregation class, leave the existing explicit aggregation classes in, but deprecated, and then see how painful the refactor is to rip any explicit references to them out.
> 
> Like I said, not a thing we have to do now, but I have having lots of lines of classes with almost no behavioral difference between them.

PR #1141, opened 2021-02-24
Add first last any aggs

sgoplani:


fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/virtualcolumns/VirtualColumn.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Get the type of the virtual column in the response.
+     *
+     * @return the type of the virtual column in the response
+     */
+    VirtualColumnType getType();
+
+    /**
+     * Get the expression of the virtual column in the response.
```
michael-mclawhorn, line 27: I suspect we should pull this from the interface, since the expression virtual column may be the only one that uses it.
Obviously I'm speculating since there's exactly one.  But I like your instinct to make a type to avoid refactoring interfaces later.

fili-sql/src/main/java/com/yahoo/bard/webservice/sql/SqlAggregationQuery.java
```diff
@@ -16,11 +16,11 @@
 import com.yahoo.bard.webservice.druid.model.query.DruidFactQuery;
 import com.yahoo.bard.webservice.data.time.Granularity;
 import com.yahoo.bard.webservice.druid.model.query.QueryContext;
+import com.yahoo.bard.webservice.druid.model.virtualcolumns.VirtualColumn;
 
 import org.joda.time.Interval;
 
-import java.util.Collection;
-import java.util.Optional;
+import java.util.*;
```
michael-mclawhorn, line 23: your IDE is globbing imports when it shouldn't.  There's a setting for this in IDEA assuming that's what you're using.

michael-mclawhorn (changes requested):
I'd like us to tweak the serialization of the column so that the virtual columns isn't serialized if empty.  Alternatively instead of defaulting to empty set you could default to null, which I believe will avoid serializing the virtual columns as well.

michael-mclawhorn (approved):

PR #1144, opened 2021-02-26
Allow finalizing field access to have no name

sgoplani:


fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/postaggregation/PostAggregation.java
```diff
@@ -111,7 +112,8 @@ public int hashCode() {
         SKETCH_ESTIMATE,
         SKETCH_SET_OPER,
         THETA_SKETCH_ESTIMATE,
-        THETA_SKETCH_SET_OP;
+        THETA_SKETCH_SET_OP,
+        FINALIZING_FIELD_ACCESS;
```
michael-mclawhorn, line 116: approved, but if we get a third one of these, we're going to add a type somewhere to indicate something is field-access-like

michael-mclawhorn (approved):

PR #1148, opened 2021-02-28
Sonar warnings

wcekan:
* Never ignore InterruptedException
* Use immutable collections where possible
* Use proper immutable collectors

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/KeyValueStoreDimension.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/KeyValueStoreDimension.java
@@ -475,7 +475,7 @@ public class KeyValueStoreDimension implements Dimension {
                 indexRows.put(rowIdKey, new Pair<>(dimensionRow, dimensionRowOld));
 
             } catch (IOException e) {
-                LOG.error("Exception while adding dimension row {}", e);
+                LOG.error("Exception while adding dimension row", e);
                 throw new RuntimeException(e);
             }
         }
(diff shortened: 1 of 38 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/data/dimension/impl/KeyValueStoreDimension.java
```diff
@@ -475,7 +475,7 @@ public void addAllDimensionRows(Set<DimensionRow> dimensionRows) {
                 indexRows.put(rowIdKey, new Pair<>(dimensionRow, dimensionRowOld));
 
             } catch (IOException e) {
-                LOG.error("Exception while adding dimension row {}", e);
+                LOG.error("Exception while adding dimension row", e);
```
wcekan, line 478: https://sonarcloud.io/project/issues?id=yahoo_fili&issues=AXPLWYhWIUmnqh_n8dv8&open=AXPLWYhWIUmnqh_n8dv8
> Not enough arguments

When the final argument is a Throwable it generates a stack trace and it is not used for formatting.  This maintains current functionality, but was original intent to show just the exception?
```java
LOG.error("Exception while adding dimension row {}", e.toString());
```

michael-mclawhorn: The original intent was to make sure we didn't lose the precise logging associated with the error.
Stack trace is a very nice to have.

fili-core/src/test/java/com/yahoo/bard/webservice/web/endpoints/TestFilterServlet.java
```diff
@@ -124,7 +124,7 @@ public void write(OutputStream os) throws IOException {
                 try {
                     Thread.sleep(100);
                 } catch (InterruptedException ignore) {
-                    // Ignore
+                    Thread.currentThread().interrupt();
```
michael-mclawhorn, line 127: 👍

michael-mclawhorn (approved):

michael-mclawhorn:
Could use a changelog for clarity.

PR #1157, opened 2021-03-23
Merge Virtual Columns in Template Druid Query

sgoplani:


fili-core/src/main/java/com/yahoo/bard/webservice/data/metric/TemplateDruidQuery.java
```diff
@@ -381,6 +387,50 @@ public TemplateDruidQuery merge(TemplateDruidQuery sibling) {
...
 
+    /**
+     * Given two sets of Virtual Columns, merge them into a single set, combining where possible.
+     *
+     * @param set1  First set of Virtual Columns
+     * @param set2  Second set of VirtualColumns
+     *
+     * @return the merged Virtual Columns
+     */
+    private LinkedHashSet<VirtualColumn> mergeVirtualColumns(
```
michael-mclawhorn, line 398: Needs a test on this method to verify 
1- all distinct ones are maintained.
2- duplicates that have the same meaning generate errors
3- duplicates with the same name do not generate errors.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/virtualcolumns/ExpressionVirtualColumn.java
```diff
@@ -70,4 +72,17 @@ public ExpressionVirtualColumn withExpression(String expression) {
     public ExpressionVirtualColumn withOutputType(String outputType) {
         return new ExpressionVirtualColumn(getName(), getExpression(), outputType);
     }
+
+    @Override
+    public boolean virtualColumnEquals(Object o) {
```
michael-mclawhorn, line 77: why isn't this just 'equals'?

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/virtualcolumns/ExpressionVirtualColumn.java
```diff
@@ -70,4 +72,17 @@ public ExpressionVirtualColumn withExpression(String expression) {
...
+        if (this == o) { return true; }
+        if (!(o instanceof ExpressionVirtualColumn)) { return false; }
+
+        ExpressionVirtualColumn that = (ExpressionVirtualColumn) o;
+
+        return
+                Objects.equals(expression, that.expression) &&
+                        Objects.equals(name, that.name) &&
+                        Objects.equals(outputType, that.outputType);
+    }
```
michael-mclawhorn, line 87: if we add an equals we should add a hashcode

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/virtualcolumns/VirtualColumn.java
```diff
@@ -37,4 +37,13 @@
...
     VirtualColumn withName(String name);
+
+
+    /**
+     * Compares the object with the virtual column object
+     *
+     * @param o The object being compared
+     * @return whether the objects are equal
+     */
+    boolean virtualColumnEquals(Object o);
```
michael-mclawhorn, line 48: we shouldn't need this, just use equals.

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/virtualcolumns/ExpressionVirtualColumn.java
```diff
@@ -70,4 +72,17 @@ public ExpressionVirtualColumn withExpression(String expression) {
     public ExpressionVirtualColumn withOutputType(String outputType) {
         return new ExpressionVirtualColumn(getName(), getExpression(), outputType);
     }
+
+    @Override
+    public boolean virtualColumnEquals(Object o) {
+        if (this == o) { return true; }
+        if (!(o instanceof ExpressionVirtualColumn)) { return false; }
```
michael-mclawhorn, line 79: If this is equals it will trigger automatic equals testing, so no need to test this.

michael-mclawhorn (changes requested):
I think we're better off just using 'equals' on Virtual columns.  Let me know if there's a reason why not.
Also, let's test mergeVirtualColumns.  I know we don't test merge aggregations, but we probably should.

michael-mclawhorn:
Also, needs Changelog.  Otherwise looks good.

michael-mclawhorn (approved):

PR #1196, opened 2021-08-27
Add type to meta object

sgoplani:

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/JsonAndJsonApiResponseWriter.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/JsonAndJsonApiResponseWriter.java
@@ -78,35 +78,9 @@ public abstract class JsonAndJsonApiResponseWriter implements ResponseWriter {
 
         generator.writeObjectFieldStart("meta");
 
-        if (request instanceof DataApiRequest) {
-            DataApiRequest dataApiRequest = (DataApiRequest) request;
-            Set<LogicalMetric> logicalMetricSet = dataApiRequest.getLogicalMetrics();
-            generator.writeObjectFieldStart("schema");
-            for (LogicalMetric lm : logicalMetricSet) {
-                LogicalMetricInfo lmi = lm.getLogicalMetricInfo();
-                if (lmi != null) {
-                    generator.writeObjectFieldStart(lmi.getName());
-
-                    if (lmi.getType() != null) {
-                        generator.writeStringField("type", String.format("%1$s", lmi.getType().getType()));
-                        if (lmi.getType().getSubType() != null) {
-                            generator.writeStringField("subtype",
-                                    String.format("%1$s", lmi.getType().getSubType()));
-                        }
-                    }
-
-                    if (lmi.getType().getTypeMetadata() != null && lmi.getType().getTypeMetadata().size() != 0) {
-                        for (Map.Entry<String, String> entry : lmi.getType().getTypeMetadata().entrySet()) {
-                            generator.writeObjectField(entry.getKey(), entry.getValue());
-                        }
-                    }
-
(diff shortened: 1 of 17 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/web/JsonAndJsonApiResponseWriter.java
```diff
@@ -71,6 +78,35 @@ public void writeMetaObject(
...
+                            generator.writeObjectField(entry.getKey(), entry.getValue());
+                        }
+                    }
+
+                    generator.writeEndObject();
+                }
+            }
+            generator.writeEndObject();
+        }
+
```
michael-mclawhorn, line 109: Let's wrap this new behavior in a feature flag so that we don't break anyone's existing implementation without them being able to back it out.

fili-core/src/test/groovy/com/yahoo/bard/webservice/table/availability/VolatilityTableSelectionSpec.groovy
```diff
@@ -133,6 +133,11 @@ class VolatilityTableSelectionSpec extends BaseDataServletComponentSpec {
                 }
             ],
             "meta": {
+                "schema": {
+                    "limbs": {
+                        "type":"number"
+                    }
+                },
```
michael-mclawhorn, line 140: can we enhance one of these test metrics to use the type, subtype and type metadata?
Maybe we can just change the metric for one test to minimize additional work?

fili-core/src/main/java/com/yahoo/bard/webservice/web/JsonAndJsonApiResponseWriter.java
```diff
@@ -71,6 +78,35 @@ public void writeMetaObject(
 
         generator.writeObjectFieldStart("meta");
 
+        if (request instanceof DataApiRequest) {
```
michael-mclawhorn, line 81: we should put this behavior in it's own method so it's easy to override.
We also should put tests on that method that test the various serializations (with/without type, with/without subtype and with/without typeMetadata)

fili-core/src/main/java/com/yahoo/bard/webservice/web/JsonAndJsonApiResponseWriter.java
```diff
@@ -71,6 +78,35 @@ public void writeMetaObject(
...
+                        generator.writeStringField("type", String.format("%1$s", lmi.getType().getType()));
+                        if (lmi.getType().getSubType() != null) {
+                            generator.writeStringField("subtype",
+                                    String.format("%1$s", lmi.getType().getSubType()));
+                        }
+                    }
+
+                    if (lmi.getType().getTypeMetadata() != null && lmi.getType().getTypeMetadata().size() != 0) {
+                        for (Map.Entry<String, String> entry : lmi.getType().getTypeMetadata().entrySet()) {
+                            generator.writeObjectField(entry.getKey(), entry.getValue());
```
michael-mclawhorn, line 100: this looks a little off to me.  
The values of the map are themselves strings not objects.
so either we should be writing the whole map as an object with some name or we should be using writeStringField as the lines above do.

michael-mclawhorn (changes requested):
I don't have a strong opinion on whether the type metadata should be nested inside the type block or not, but I'm pretty sure it's not going to serialize correctly as written.  There are no tests writing that metadata, however, so you should add that.

I would pull the new behavior into a method to make it easier to test and override.  I would also add a feature flag so that anyone who gets broken by this behavior can back it out.

michael-mclawhorn (approved):
looks good

michael-mclawhorn (approved):

PR #1198, opened 2021-09-08
Add Meta Flag to Properties

sgoplani:


michael-mclawhorn (approved):

michael-mclawhorn (changes requested):
I'm not sure this is a good idea.
Let's just set this true downstream.

PR #1205, opened 2021-09-29
Capability to bypass URL in Role based authentication filter

bharatmotwani:

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilter.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilter.java
@@ -17,6 +17,7 @@ import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 import javax.annotation.Priority;
 import javax.inject.Inject;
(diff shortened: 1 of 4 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/web/filters/RoleBasedAuthFilter.java
```diff
@@ -97,6 +99,19 @@ protected boolean isUserInRole (List<String> allowedUserRoles, ContainerRequestC
...
+     * Checks if an URL matches to the any of the allowed/bypassed URL in the list.
+     *
+     * @param allowedUrls List of allowed URLs
+     * @param containerRequestContext Request context contains the user role information
+     *
+     * @return boolean based on the request URL
+     */
+    protected boolean isByPassURL(List<String> allowedUrls, ContainerRequestContext containerRequestContext) {
+        String url = containerRequestContext.getUriInfo().getAbsolutePath().getPath();
+        return allowedUrls.stream().anyMatch(path -> path.startsWith(url));
```
michael-mclawhorn, line 112: Don't do starts with, use Java Pattern class to process the result.

michael-mclawhorn (changes requested):

michael-mclawhorn (approved):

PR #1247, opened 2022-03-29
Expanding and cleaning up weight check metrics.

michael-mclawhorn:
Four corner metrics establish lines scanned, returned, sketches scanned, returned.

### DIFF
--- a/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/WeightEvaluationQuery.java
+++ b/fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/WeightEvaluationQuery.java
@@ -49,14 +49,14 @@ import java.util.List;
  *        ],
  *        "aggregations": [
  *          {
- *            "name": "lines",
+ *            "name": "scannedLines",
  *            "type": "count"
  *          }
  *        ],
  *        "postAggregations": [
  *          {
  *            "type": "constant",
- *            "name": "rawSketches",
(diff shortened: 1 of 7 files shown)


fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/WeightEvaluationQuery.java
```diff
@@ -74,19 +74,23 @@
  *    "aggregations": [
  *      {
  *        "name": "resultSketches",
- *        "fieldName": "rawSketches",
+ *        "fieldName": "sketchesPerRow",
  *        "type": "longSum"
  *      },
```
will-lauer, line 79: Why is this here if its just overriden in the postAgg section?

will-lauer (approved):

fili-core/src/main/java/com/yahoo/bard/webservice/druid/model/query/WeightEvaluationQuery.java
michael-mclawhorn: The fieldname is the downward pointing reference, so resultSketches is the total of the sketchesPerRow from the inner layer.

will-lauer: ah, I missed that I was looking at fieldname rather than just `name`. makes sense now.

PR #1262, opened 2022-07-21
jdk11 with 8/9 compatibility

michael-mclawhorn:

PR  Testing a JDK bump

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1005,6 +1005,9 @@
             </plugin>
             <plugin>
                 <artifactId>maven-javadoc-plugin</artifactId>
+                <configuration>
+                    <excludePackageNames>${javadocs.package.exclude}</excludePackageNames>
+                </configuration>
                 <reportSets>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -719,7 +721,7 @@
                     <configuration>
                         <rules>
                             <requireJavaVersion>
-                                <version>[1.8,1.9]</version>
+                                <version>[1.8,12]</version>
```
will-lauer, line 724: why is this 12, not 11?

will-lauer (approved):

pom.xml
michael-mclawhorn: because I wanted all versions of 11 to be in range, I think that's a boundary marker.

will-lauer: would `[8,12)` work to make it an open interavl?

will-lauer: [8,12] usually implies that it includes 12.

PR #1269, opened 2022-09-15
Fix DataSource name mis-match problem

QubitPi:

## Why This PR Exists

Our team is setting up a Fili instance in order to report our customer's behavior data. We were referring to [fili-wikipedia-example](https://github.com/yahoo/fili/tree/master/sample-applications/fili-wikipedia-example) as setup guide and found that fili-wikipedia-example returned no data within a data-populated time interval

## Steps to produce the Issue

1. Setting up a [Druid instance locally](https://druid.apache.org/docs/latest/tutorials/index.html)
2. Setting up a fili-wikipedia-example app by following its [README](https://github.com/yahoo/fili/tree/master/sample-applications/fili-wikipedia-example)
3. Send a fili query in command line with

    `curl "http://localhost:9998/v1/data/wikipedia/day/?metrics=count&dateTime=2015-09-12/P2D" -H "Content-Type: application/json" | python -m json.tool`

The query above returns no data, which actually exists in Druid

## What This PR Do to Solve the Issue

Druid official asks people to set Datasource name to "wikipedia":

page 5

However, fili-wikipedia-example [hardcode the Datasource name to "wikiticker"](https://github.com/yahoo/fili/blob/master/sample-applications/fili-wikipedia-example/src/main/java/com/yahoo/wiki/webservice/data/config/names/WikiDruidTableName.java#L13)

So we changed `WIKITICKER` to `WIKIPEDIA` in this PR.

### DIFF
--- a/sample-applications/fili-wikipedia-example/README.md
+++ b/sample-applications/fili-wikipedia-example/README.md
@@ -18,6 +18,7 @@ This example is an entirely self contained example that provides a Fili applicat
     ```bash
     cd fili
     mvn install
+    cd sample-applications
     mvn -pl fili-wikipedia-example exec:java
     ```
 - Note that if your setup is different you can adjust it by changing the default parameters below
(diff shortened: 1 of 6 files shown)


sample-applications/fili-wikipedia-example/README.md
```diff
@@ -18,6 +18,7 @@ This example is an entirely self contained example that provides a Fili applicat
     ```bash
     cd fili
     mvn install
+    cd sample-applications
```
QubitPi, line 21: without this, the terminal gives an error at `mvn -pl fili-wikipedia-example exec:java` complaining a wrong directory

archolewa (approved):

PR #1273, opened 2022-10-10
MemCached connection refuse warning coming up even after setting up t…

hridaydubey:
…he CacheFeatureFlag as NONE

(bard__query_response_caching_strategy = NoCache)


michael-mclawhorn:
I think we determined that this required more rearchitecting to fix because of the distinct cache instances in the service.

PR #1282, opened 2022-12-09
Role query auth filter

michael-mclawhorn:


cythrawll (approved):

fili-core/src/main/java/com/yahoo/bard/webservice/servlet/RoleImpersonationFilter.java
```diff
@@ -0,0 +1,220 @@
...
+     */
+    public RoleImpersonationFilter() {
+        List<String> roleAliases = SYSTEM_CONFIG.getListProperty(ROLE_NAME_ALIAS_KEY);
+        Map<String, String> inProgressAliases = new HashMap<>();
+
+        if (!roleAliases.isEmpty()) {
+            for (String aliasMapping : roleAliases) {
+                String[] mapping = aliasMapping.split(":");
+                if (mapping.length != 2) {
+                    LOG.warn(
```
archolewa, line 66: I would just have this throw an exception and let the system crash. It happens at start up time, so it's not like this will crash anything while running, and it ensures that if there is a malformed role mapping, developers are aware of it. What's going on with role-based stuff isn't always obvious in a running system, and we don't want to be in a situation where the system is silently mishandling potentially important roles because of a dumb typo that we caught, but by and large ignored.

fili-core/src/main/java/com/yahoo/bard/webservice/servlet/RoleImpersonationFilter.java
```diff
@@ -0,0 +1,220 @@
...
+     */
+    public abstract List<String> getRolesForId(String id) throws IOException;
+
+    /**
+     * Test whether the user is authorized to impersonate using the From header.
+     *
+     * @param securityContext  The security context of the logged in user.
+     *
+     * @return true if the user has permission to impersonate other users' rights
+     * @throws IOException is authorization creates an error.
```
archolewa, line 96: `is authorization` should presumably be `if authorization`.

fili-core/src/main/java/com/yahoo/bard/webservice/servlet/DelegatedPrincipal.java
```diff
@@ -0,0 +1,18 @@
+// Copyright 2022 Yahoo Inc.
+// Licensed under the terms of the Apache license. Please see LICENSE.md file distributed with this work for terms.
+package com.yahoo.bard.webservice.servlet;
+
+import java.security.Principal;
+
+/**
+ * A subinterface to retrieve impersonator credentials from an impersonated principal.
+ */
+public interface DelegatedPrincipal extends Principal {
```
archolewa, line 10: Any particular reason we're making this an interface that we create an anonymous subclass for, instead of just making this a class that implements `Principal`? Making a class strikes me as being much clearer and just as effective for our purposes.

fili-core/src/main/java/com/yahoo/bard/webservice/servlet/RoleImpersonationFilter.java
```diff
@@ -0,0 +1,220 @@
...
+     * @param requestContext  The request context for the current request.
+     * @param fromId  The attempted impersonation.
+     *
+     * @throws IOException with an WebApplication FORBIDDEN status
+     */
+    private void throwUnauthorizedImpersonationError(
+            final ContainerRequestContext requestContext,
+            final String fromId
+    ) {
+        Response.ResponseBuilder builder = null;
```
archolewa, line 182: Why the early declaration? Why not just declare this on line 189 when we build it?

archolewa (changes requested):
Few small things (see comments above).

PR #1287, opened 2022-12-14
Unifies DataApiRequest and TablesApiRequest under ApiRequestWithTable.

archolewa:
-- It is not unusual for customers to apply security filters based on the table in the request. Sometimes, these rules (especially for rules that control whether customers can even *see* a given table) need to be applied to both the data and tables endpoint.

However, as things are currently structured, this necessitates code duplication, or handrolled checks for the tables endpoint because the RoleBasedTableValidatorRequestMapper only applies to DataApiRequests. So a new interface `ApiRequestWithTable` is introduced that extends `ApiRequest` and adds the `getTable` method. This allows us to generalize the `RoleBasedTableValidatorRequestMapper` (and possibly others if necessary) to both DataApiRequests and TablesApiRequests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Current
 
 ### Changed:
 
+- [Add new interface ApiRequestWithTable](https://github.com/yahoo/fili/pull/1287)
+   - Allows apirequests that contain a table to leverage the same request mappers for e.g. security rules
+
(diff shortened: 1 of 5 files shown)


archolewa:
We've decided to try more fully generalizing the current mapper, and potentially creating a subclass to handle tables endpoint specific concerns, rather than this approach

PR #1290, opened 2023-01-10
Quality of life improvements for test environment:

michael-mclawhorn:
Don't try to load dimensional metadata
Don't try to connect to Cache if not configured to Don't mock objects with lists

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ Current
 -------
 
 ### Fixed:
+- [Removed excessive logging and exceptions caused during test runs]()
+  - Metadata Loader during test runs shouldn't try to connect to metadata service. `NoOpDataSourceMetadataLoadTask`
+  - SegmentMetadata in tests do not parse correctly as lists, changing to a null object. `TestDruidWebService` 
+  - Noop cache shouldn't create useless connection to Memcached `StubTupleDataCache`
 
 ### Added:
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ Current
 -------
 
 ### Fixed:
+- [Removed excessive logging and exceptions caused during test runs]()
+  - Metadata Loader during test runs shouldn't try to connect to metadata service. `NoOpDataSourceMetadataLoadTask`
+  - SegmentMetadata in tests don not parse correctly as lists, changing to a null object. `TestDruidWebService`
```
will-lauer, line 14: typo: `don` vs `do`

fili-core/src/main/java/com/yahoo/bard/webservice/application/AbstractBinderFactory.java
```diff
@@ -1063,12 +1062,15 @@ protected DataCache<?> buildCache() {
      *
      * @return the instance of local signature cache
      */
-    protected DataCache<?> buildLocalSignatureCache() {
+    protected TupleDataCache<String, ?, ?> buildLocalSignatureCache() {
```
will-lauer, line 1065: Shouldn't this return a `TupleDataCache<String, Long, String>` rather than a `<String, ?, ?>`? It seems like the specific type is known in all the paths.

will-lauer (approved):

